PR #4648, opened 2015-07-23
Prefetch correct dataset for column merge datasets

### LINKED ISSUE #4645: Duplicated columns when duplicating dataset
#### Steps to reproduce:
1. Create empty dataset
2. Duplicate dataset
3. Add a column on the duplicated one
4. Merge both datasets

When you create a new empty dataset and you duplicate it, if you create a column on the duplicated one and try to merge both dataset, the columns on both datasets are the same (long gif):
dataset bug

viddo:
Fixes #4645 
cc @xavijam


Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5307/))

PR #4655, opened 2015-07-23
Fix IP addresses missing title/desc

viddo:


viddo:
verified locally since PR testing is taking too long

PR #4661, opened 2015-07-23
Remove zclip copy on publish modal

### LINKED ISSUE #4486: Remove flash placeholders (at least from full text input) at publish modal
Firefox and IE11 are already Blocking Flash, and others are to follow, so this is going to be the "default" for users... 

image

viddo:
Fixes #4486 

zclip-removed

This removes the zclip usage on the publish modal, to avoid those black boxes that overlays the inputs when flash is disabled. Unfortunately there's no way to really implement clipboard-copy from JS, so I guess this is as good as it gets for now.

### DIFF
--- a/app/assets/stylesheets/editor/publish.css.scss
+++ b/app/assets/stylesheets/editor/publish.css.scss
@@ -8,5 +8,4 @@
   margin-left: $sMargin-element;
   margin-right: $sMargin-element;
   box-sizing: border-box;
-  cursor: pointer;
 }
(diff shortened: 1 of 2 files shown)


viddo:
ping  @xavijam  @saleiva @Kartones

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5329/))

Kartones:
<3

PR #4680, opened 2015-07-24
Remove tables_count deprecated field

### LINKED ISSUE #4127: Remove tables_count field deprecated
As it seems like the field tables_count of user model has been deprecated and it's not updated anymore it should be removed with a migration and clean the code in order to remove any reference

/cc @Kartones

Kartones:
Fixes #4127

### DIFF
--- a/db/migrate/20150724121847_remove_table_count_from_users.rb
+++ b/db/migrate/20150724121847_remove_table_count_from_users.rb
@@ -1,0 +1,9 @@
+Sequel.migration do
+  up do
+    drop_column :users, :tables_count
+  end
+
+  down do
+    add_column :users, :tables_count, :integer, default: 0
+  end
+end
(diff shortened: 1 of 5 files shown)


Kartones:
@ethervoid CR please

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5353/))

db/migrate/20150724121847_remove_table_count_from_users.rb
```diff
@@ -0,0 +1,9 @@
+Sequel.migration do
```
ethervoid: As we did with `dynamic_field_enabled` field, you should put the ward to  avoid the cache of active record:

http://blog.codeship.com/rails-migrations-zero-downtime/

Kartones: but is not used/queried anywhere, will it break?

ethervoid: If no one use it should not break.

ethervoid:
:+1:

PR #4688, opened 2015-07-24
Fixing problem within new edit feature metadata

### LINKED ISSUE #4687: Edit metadata window is trimming the last character of a text when overriding
**Steps:**
* Go to an existent table (small enough so that you can see the data for all rows)
* Go to the map view and try to edit the metadata of a feature that already contains some text:

image

I wrote: " oh yeah", and the result was: 

image

The last character of the text does seem to be skipped.

**More info:**
* It doesn't seem to happen always, my first time doing the steps spotted the issue but the next times it was hard to replicate again. 
* It doesn't happen with empty boxes -  Users are removing all the content and writing again in order to workaround the issue

xavijam:
Fixes #4687.

Could you take a look @javierarce please?
cc @iriberri

### DIFF
--- a/app/assets/stylesheets/old_elements/dialog_small_edit.css.scss
+++ b/app/assets/stylesheets/old_elements/dialog_small_edit.css.scss
@@ -275,7 +275,14 @@
       p {width:17px; text-align:center; font-size:20px;}
 
       label {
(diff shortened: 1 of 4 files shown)


javierarce:
:+1: I approve

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5362/))

PR #4715, opened 2015-07-28
Fix templates escaping

viddo:
ping @CartoDB/frontend

### DIFF
--- a/lib/assets/javascripts/cartodb/old_common/header.js
+++ b/lib/assets/javascripts/cartodb/old_common/header.js
@@ -414,7 +414,7 @@ cdb.admin.Header = cdb.core.View.extend({
         },
         onDone: function(nameChanged) {
           // Check if attr saved is name to change url when
-          // visualization is table type 
+          // visualization is table type
           if (nameChanged && !this.vis.isVisualization()) {
             window.table_router.navigate(self._generateTableUrl(), {trigger: false});
             window.table_router.addToHistory();
@@ -429,7 +429,7 @@ cdb.admin.Header = cdb.core.View.extend({
         onResponse: function(data) {
           self._onSetAttributes(data);
         }
-      });  
+      });
     }
 
     dlg.appendToBody();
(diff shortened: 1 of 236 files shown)


javisantana:
:es: from my side, nice catch[s] with the javascript generated html

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5408/))

lib/assets/javascripts/cartodb/old_common/header.js
```diff
@@ -506,7 +506,7 @@ cdb.admin.Header = cdb.core.View.extend({
           } else {
             self.change_confirmation = new cdb.admin.BaseDialog({
               title: self._TEXTS.rename.title.replace('{{ vis }}', ( self.model.isVisualization() ? "visualization" : "table" )),
-              description: self._TEXTS.rename['desc_' + (self.model.isVisualization() ? "vis" : "table")],
+              descriptionSafeHtml: self._TEXTS.rename['desc_' + (self.model.isVisualization() ? "vis" : "table")],
```
javierarce: I'm curious about this change. Is `[variableName]SafeHtml` the recommended way to name the variables containing html now?

viddo: I'd think it would be preferable, to make it clear what the intention of the variable is (i.e. to be rendered w/o any escaping in the template).

javierarce:
:de: Alles klar!

lib/assets/javascripts/cartodb/old_common/header.js
javierarce: :ok_hand:

matallo:
good to know way to proceed with js templates!

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5412/))

PR #4728, opened 2015-07-28
This initializer interferes with the load of constants, reverting

ethervoid:
I put this code in the Rails initializer to enable hot reload in Resque but today I've found that interferes with the load of constants so I've to revert it :(


Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5440/))

ethervoid:
Please @Kartones check for this to merge in master

Kartones:
:+1:

PR #4742, opened 2015-07-28
Don't toggle visibility switch on adding/removing layer

### LINKED ISSUE #4093: Button for disabled layer shows as enabled when adding one
**Steps:**
* Create a map from a table
* Add a layer (empty, for example)
* Hide the first layer
image
* Add a new layer, empty again
* The disabled selector gets enabled automatically when reloading, although the layer is hidden. When you click on the hidden layer, then this button gets reloaded at it shows the correct status.
image

**Expected behaviour:**
* When adding a new layer, if a layer was hidden the button should reflect that state directly.

**Actual behaviour:**
* The button doesn't show the correct status unless the layer is selected.

image

viddo:
Fixes #4093 

This corrects a previous change d066c9d6a21a55f8056207dadc396a65dae4f3eb that introduced the bug; when a layer was added/removed it faulty changed the visibility switch state to the new layer, rather than the layer that the current layer the view represents. 

Now, the add/removal only maybe updates the visibility of the switch, but not the actual switch state (that is still handled by the event listener on the own layer model the view represents).

ping @javisantana


javisantana:
haha, a 1.5years old bug :es:

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5447/))

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5448/))

Issue #3428, opened 2015-04-29
Geocoding available_geometries "not working" with table names that start by "_"


iriberri:
The `available_geometries` function is returning `[ ]` when a table name starts by "_". This is blocking some geocoding methods as "postal code" that use the function to check the availability of the data.

Renaming a table from the dashboard to "_tablename" doesn't seem possible, the request sends the error

````
Request Method:PUT
Status Code:400 Bad Request
{"errors":["Error saving data"]}
````

but importing a table with that name is possible. (Get one [here](https://gist.githubusercontent.com/iriberri/f274ff12bfea3c93a7a2/raw/17522d6359daec359bd6fa9992afa9ba0c38a861/_untitled_table_324.csv))

Steps:

* Import [previous dataset](https://gist.githubusercontent.com/iriberri/f274ff12bfea3c93a7a2/raw/17522d6359daec359bd6fa9992afa9ba0c38a861/_untitled_table_324.csv)
* Georeference > Zipcodes 
* Choose the first selector and the second one (selecting a column in both)
* Check the available_geometries response
![image](https://cloud.githubusercontent.com/assets/1730320/7395416/08f2abc6-ee9a-11e4-9e27-fd5518ba8bd6.png)
![image](https://cloud.githubusercontent.com/assets/1730320/7395445/38e77654-ee9a-11e4-8645-31bb6b8ec0d6.png)

I'm guessing that maybe some other parts of the platform may be affected by this, but only catched this one atm.

**Comment by gfiorav** [comment]
PG gem is not supporting renaming tables to anything starting with '_'

```
rename_table('foo','_bar')
```
 fails with exception ```PG::Error: ERROR:  type "_bar" already exists```

I'm looking into the gem's source code to report the bug since standard postgres *does* allow renaming tables with '_' prefix.

iriberri:
Nice to know :)

gfiorav:
The first part of this ticket concerning table renames with '_' prefix has been solved in PR [#4752](https://github.com/CartoDB/cartodb/pull/4752)

The part about geocoding '_' prefixed tables can now be properly tested.

gfiorav:
Identified Resque error on for geo referencing tables starting with ('_'):

```** [14:11:30 2015-07-29] 6139: (Job{geocodings} | Resque::GeocoderJobs | [{"job_id"=>"xxxx"}]) failed: #<Sequel::DatabaseError: PG::Error: ERROR:  relation "public" does not exist
[...]

PR #4752, opened 2015-07-28
Fixes PG::ERROR rising when renaming or geocoding a table starting with underscore

gfiorav:
Before this fix, renaming a table to anything starting with an underscore ('_') would result in a PG::ERROR (partially described in ticket [#3428](https://github.com/CartoDB/cartodb/issues/3428)).

This error happens because the PG gem uses an intermediate table when renaming tables that happens to be the original name prefixed by an underscore, thus conflicting with the target table name and rising ```type '_original_table_name' already exists```

cc @javisantana


Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5452/))

Kartones:
Tests my friend! this are delicate things and have to come with tests, even if code is trivial.

Also, I'd use a random prefix like `aux-#{somerandom}-name`, having a fixed prefix is tempting to screw up by leaving on purpose an `aux-name` table :trollface:

and just for my info, so in the end was not Sequel but PG gem itself? interesting!

javisantana:
using "aux-" prefix is the same than using a "_" prefix, we just moved the problem to a different prefix. I'd rather fix pg gem or just not allow tables starting by "_"

Kartones:
@javisantana  that prefix is temporal, he does a two-step renaming, using an intermediate name, that's why I said adding a random prefix, to avoid collisions. Dunno if PG gem is up to date enough to fix that, but could be an option, yes

gfiorav:
@javisantana @Kartones I verified it will not collide with another table named the aux-original-name.

I have to further investigate why, though.

Kartones:
no no, you don't understand me. if I do:
- `CREATE TABLE aux-wadus` (int X);`
- and then create a table `test` and do `table.rename_to('wadus')` I'll break your fix

That's why it needs to have some random numbers in the prefix, to really minimize collisions

gfiorav:
Yes, I understood! 

It should brake the fix but it doesn't, I tested that situation yesterday.

I have to figure out why, of course!! :)

Kartones:
ahh, ok, then curious behaviour, would be interesting to know, indeed

gfiorav:
@javisantana The system does allow importing tables starting with an underscore and people actually do that (that's why ticket [#3428](https://github.com/CartoDB/cartodb/issues/3428) appeared in the first place).

For consistency, I think we should allow tables to start with an underscore. Postgres has no problem whatsoever with it so it would be weird that someone would try to create tables following a schema that prefixes underscores and couldn't.

javisantana:
yes, that's the behavior we want but if we can't fix the underscore problem is better to avoid it

gfiorav:
Ok, so the problem is avoided in the following way: a dash ('-') prefix is used for the intermediate table. The reason we can rest assured that collisions won't happen is that dash characters are not permitted in CartoDB tables, thus no intermediate table starting with a table can collide with a CartoDB table.

Unless we change our policy of not allowing dash characters in table names, we are safe.

cc @Kartones @javisantana

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5457/))

Kartones:
Wrong... remember the SQL API, or for the same purpose, the SQL query editor at cartodb:
```
CREATE TABLE kartones."-test" (x INT)
SELECT * FROM kartones."-test"
(o rows)
```
That works and I now have a table with a dash. So CartoDB does allow tables starting with a dash, they probably won't be allowed inside the editor but for sure they can live in the user's schema.

I see the same issue as having `aux` or `_`, but if you prove to me with tests that it doesn't cause issues, at @javisantana's judgement.

gfiorav:
So, basically, we can import tables with '-'? Isn't this a bit inconsistent? In the code for the editor:

https://github.com/CartoDB/cartodb/blob/master/app/models/table.rb#L1417-L1429

```
  # Gets a valid postgresql table name for a given database
  # See http://www.postgresql.org/docs/9.3/static/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
  def self.get_valid_table_name(name, options = {})
    # Initial name cleaning
    name = name.to_s.squish #.downcase
    name = 'untitled_table' if name.blank?

    # Valid names start with a letter or an underscore
    name = "table_#{name}" unless name[/^[a-z_]{1}/]

    # Subsequent characters can be letters, underscores or digits
    name = name.gsub(/[^a-z0-9]/,'_').gsub(/_{2,}/, '_')
```

Every '-' is substituted for a '_'. Maybe we should make the importer sanitize table names in the same way?

Kartones:
Careful, I am not talking about importing tables, I am talking about the [SQL API](http://docs.cartodb.com/cartodb-platform/sql-api.html), which is as important as the editor, but doesn't shares code... yoy can create tables via SQL or importing data. Every bugfix you do regarding the editor must take into account scenarios from the SQL API, and there we cannot simply forbid doing things.

I'd say let's try the TDD way of things, building first of all some tests:
- to reproduce the current renaming bug (should allow with the fix)
- to try importing first a table with the "not allowed prefix" in the name using the import API or Ruby models, then renaming (to check for collisions)
- to try creating a table via SQL (similar to [this](https://github.com/CartoDB/cartodb/blob/1167d049a742c325cd71dd2fcaf6b9f185f847cf/services/importer/lib/importer/query_batcher.rb#L77)), then renaming (to check for collisions)

You're going to need those tests anyway to push the code to the repo, plus we stop trying workarounds of this or that prefix that might cause side-effects (at least we make sure they don't).

gfiorav:
Surely we should do proper testing!

I'm just trying to understand what tests should be carried out, so thanks for the guidelines  :smile:

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5468/))

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5498/))

gfiorav:
The last commit allows to georeference tables starting with ```_```.

Further testing needs to be done to ensure nothing has broken, since this PR removes all ´´´sequel_qualified_table_name``` from use. Some queries had to be rewritten.

cc @javisantana @kartones

Kartones:
Add tests please, and then I'll gladly review the code (also if some issues arise, you might need to change again the code, etc., so better to delay reviewing)

gfiorav:
Yes, like I said, testing has to be done :)

Will nudge you again when it’s ready!

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5714/))

saleiva:
bump!

PR #4769, opened 2015-07-29
Long geocoding tasks send an email to the user

ethervoid:
This PR adds the following features:

* Now the user is going to receive an email when a geocoding finish (successfully or not) only if the geocoding took more than 5 minutes. (The same amount of time that we have in the import task)

* I've made a little refactor to extract the the success and the failure logic (as we do in the data importer). I'd be great to change the handle_failure based in exceptions but I think we should do it in phases


Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5465/))

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5469/))

saleiva:
So you are blocked by @josecruz and the design? @josecruz could you prioritize this if so?

ethervoid:
Yup, all that I need is to finish the email

matallo:
added the processed and total rows, please review @ethervoid

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5505/))

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5506/))

ethervoid:
Please @Kartones CR this :)

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5507/))

app/models/geocoding.rb
```diff
@@ -330,4 +316,53 @@ def instantiate_log
     @log
   end
 
+  def cancelled_geocoding?
```
Kartones: Uhmm, can we cancel the high-resolution geocoding (Here one)? AFAIK once you've sent the query you/we'll be charged anyway so... probably in that case we shouldn't... but not sure.

@iriberri any hint on this?

ethervoid: The check for the spent credits is in the handle_success.

iriberri: @rafatower might have added this. Non batched requests are sent one by one so I think there's a possibility we could stop it.

ethervoid: That method is mine, I've checked that cancel state is not working so I've added checks that cancel the job if the frontend change the state to `cancelled`.

@rafatower added `return false if state == 'cancelled` after the geocoding work is done and I've added another before the geocoding work is started because I thought that if the user cancel the job before the geocoder finish we could avoid that work to us and to the user.

app/models/geocoding.rb
```diff
@@ -330,4 +316,53 @@ def instantiate_log
...
+  def total_geocoded_rows(rows_geocoded_before)
+    rows_geocoded_after = table_service.owner.in_database.select.from(table_service.sequel_qualified_table_name).where('cartodb_georef_status is true and the_geom is not null').count rescue 0
+    rows_geocoded_after - rows_geocoded_before
+  end
+
+  # Used in the run! method
+  def send_report_mail(state, table_name, error_code=nil, processable_rows, number_geocoded_rows)
+    geocoding_time = @finished_at - @started_at
+    if geocoding_time >= MIN_GEOCODING_TIME_TO_NOTIFY
+      GeocoderMailer.geocoding_finished(user, state, table_name, error_code, processable_rows, number_geocoded_rows).deliver
```
Kartones: oh, why don't we enqueue like other mails? (or maybe we don't if comes from a rake, but I'd anyway do it to make each task end as early as possible).

Just personal taste

Kartones:
If the cancel process etc. is ok with high-res geocoding, good to go :+1:

app/models/geocoding.rb
ethervoid: Because this should be always inside a resque job (geocoding job). But know that I think about it its better to include inside a Resque job to avoid fails in the geocoding job if it's not possible to send the email

ethervoid: @javisantana what do you think of this

javisantana: too detailed for me to take a decision

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5511/))

app/models/geocoding.rb
javisantana: We should not allow to cancel the geocoding once the batch has sent to the external provider. Maybe this requieres some changes in frontend but it the user ask for 10k rows geocoding and then they cancel, we could be paying those geocodings

ethervoid: I can fix that removing the `return false if cancelled` after the geocoding work. Right know because is not working we are doing that way. 

IMHO I check before launch the geocoder work and cancel it but if after that the user try to cancel it he will be charged.

Is ok for all of you?

Kartones: I haven't checked the models, but this should be something like "if the geocoding is of type high-res, don't allow cancelling once sent, else allow to". So more than a matter of removing or not, I'd just make high-res one not perform that `return false if ...`. And then make sure UI also doesn't shows a cancel button, to else put a FE ticket so they disallow it for high-res one

ethervoid: Ok what I going to do is make a new issue for this case and remove from this. I don't want to hold more this PR, we can talk about how to cancel this kind of work

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5512/))

Issue #4792, opened 2015-07-30
Sanitize output of a warning message


javierarce:
![screen_shot_2015-07-30_at_12 10 00_1024](https://cloud.githubusercontent.com/assets/4933/8981033/58aeec70-36b5-11e5-973e-e09964756883.png)

javierarce:
Fixed, @matallo @javisantana

PR #4802, opened 2015-07-30
Fixes issue showing html in the flash message

javierarce:
This PR fixes a problem showing flash messages (written by us/cartodb admins) that contained HTML markup. 

Original issue:  #4792

### DIFF
--- a/app/views/shared/_flash_message.html.erb
+++ b/app/views/shared/_flash_message.html.erb
@@ -1,7 +1,7 @@
 <% if cookies[:flash] || flash[:success] || flash[:error] %>
   <div class="FlashMessage <%= 'FlashMessage--success' if (flash[:success]) %><%= 'FlashMessage--error' if flash[:error] %>">
     <div class="u-inner">
-      <p class="FlashMessage-info <%= 'success' if (flash[:success]) %><%= 'error' if flash[:error] %>"><%= cookies[:flash] || flash[:success] || flash[:error] %></p>
+      <p class="FlashMessage-info <%= 'success' if (flash[:success]) %><%= 'error' if flash[:error] %>"><%= raw cookies[:flash] || flash[:success] || flash[:error] %></p>
     </div>
   </div>
 <% end %>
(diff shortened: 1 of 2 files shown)


javierarce:
@Kartones: can you review this PR, please?

app/views/shared/_flash_message.html.erb
```diff
@@ -1,7 +1,7 @@
 <% if cookies[:flash] || flash[:success] || flash[:error] %>
   <div class="FlashMessage <%= 'FlashMessage--success' if (flash[:success]) %><%= 'FlashMessage--error' if flash[:error] %>">
     <div class="u-inner">
-      <p class="FlashMessage-info <%= 'success' if (flash[:success]) %><%= 'error' if flash[:error] %>"><%= cookies[:flash] || flash[:success] || flash[:error] %></p>
+      <p class="FlashMessage-info <%= 'success' if (flash[:success]) %><%= 'error' if flash[:error] %>"><%= raw cookies[:flash] || flash[:success] || flash[:error] %></p>
```
Kartones: that raw cookies[flash] might mean I could XSS by injecting malicious code inside?

Do you have this deployed somewhere so I can try?

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5487/))

app/views/shared/_flash_message.html.erb
javierarce: Oh, you're right. I'll give you the URL of a server in slack.

javierarce: If that doesn't work, do we have an html sanitizer we can use?

Kartones:
as you said if `.raw` already calls inside `.html_safe`... go ahead

javierarce:
Thanks!

app/views/shared/_flash_message.html.erb
viddo: Was this taken care of? You could use http://apidock.com/rails/ActionView/Helpers/SanitizeHelper/sanitize with a set of whitelisted tags, we shouldn't use raw here…

PR #4804, opened 2015-07-30
Offer the option to try lat/lon 4326 projection when not finding the .prj file

### LINKED ISSUE #4031: Offer the option to try lat/lon 4326 projection when not finding the .prj file
Great feedback from one of our users. Would increase number of imports from shp for sure.

For this particular case we should open another modal before the error one where we ask if we should add that projection automatically. This error is the 2nd most popular in shp files

cc/ @javisantana @Kartones

Kartones:
Closes #4031


Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5489/))

Kartones:
@ethervoid CR please

services/importer/lib/importer/runner.rb
```diff
@@ -148,7 +148,7 @@ def success?
 
       private
 
-      attr_reader :pg_options, :unpacker, :available_quota
+      attr_reader :pg_options, :unpacker, :available_quota, :job
```
ethervoid: Why did you add here :job?

Kartones: For the test, see below at `services/importer/spec/acceptance/shp_spec.rb`

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5493/))

ethervoid:
:+1:

Issue #3761, opened 2015-05-25
Patterns selector improvements


saleiva:
Change "Patterns" for "CartoDB Patterns Library" or something more illustrative. 
Align correctly the patterns on the thumbnails. They look blurry!

![Uploading Screen Shot 2015-05-25 at 19.00.43.png…]()

fdansv:
@saleiva Missing image!

PR #4853, opened 2015-08-03
Fixes blurry patterns + updates copy

javierarce:
This PR removes the blurriness from the patter thumbnails (caused by a difference between the default size of the thumbnails and the actual one). It also updates the title of the pattern library.

Original issue: https://github.com/CartoDB/cartodb/issues/3761

screen shot 2015-08-03 at 17 16 02

/cc: @xavijam


xavijam:
Do we have them ready for retina?

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5548/))

javierarce:
No, we don't.

xavijam:
For me there is no problem, so :jp:.

PR #4876, opened 2015-08-04
Fixed twitter details button and dialog

xavijam:
- [x] Show twitter import details button within background polling import view, in any scenario, dashboard or editor.
- [x] Let user check twitter import details when a twitter import is created within editor.
- [x] Fixed background polling success details styles.


Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5580/))

Issue #4708, opened 2015-07-27
optimize create_dev_user run time + setup sync tables on by default


Kartones:
right now calls a bunch of rakes, join functionality into one so it is way faster (but don't delete old rakes as opensource installs might use them for custom scripts). Also setup by default sync tables on

gfiorav:
Solved in PR #4879

PR #4879, opened 2015-08-05
Optimize create dev user #4708

gfiorav:
This PR fixes issue #4708 

cc @Kartones

### DIFF
--- a/lib/tasks/setup.rake
+++ b/lib/tasks/setup.rake
@@ -39,6 +39,44 @@ DESC
       puts "User #{u.username} created successfully"
     end
 
+    task :create_dev_user => 
+      ["rake:db:create", "rake:db:migrate", "cartodb:db:create_publicuser"] do
+      raise "You should provide a valid e-mail"    if ENV['EMAIL'].blank?
+      raise "You should provide a valid password"  if ENV['PASSWORD'].blank?
+      raise "You should provide a valid subdomain" if ENV['SUBDOMAIN'].blank?
+
+      # Reload User model schema to avoid errors
+      # when running this task along with db:migrate
+      User.set_dataset :users
+      
+      u = User.new
+      u.email = ENV['EMAIL']
+      u.password = ENV['PASSWORD']
+      u.password_confirmation = ENV['PASSWORD']
+      u.username = ENV['SUBDOMAIN']
+      u.database_host = ENV['DATABASE_HOST'] || ::Rails::Sequel.configuration.environment_for(Rails.env)['host']
+      u.save
+
+      raise u.errors.inspect if u.new?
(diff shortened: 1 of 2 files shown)


lib/tasks/setup.rake
```diff
@@ -39,6 +39,47 @@ DESC
...
+      ["rake:db:create", "rake:db:migrate", "cartodb:db:create_publicuser"] do
+      raise "You should provide a valid e-mail"    if ENV['EMAIL'].blank?
+      raise "You should provide a valid password"  if ENV['PASSWORD'].blank?
+      raise "You should provide a valid subdomain" if ENV['SUBDOMAIN'].blank?
+
+      # Reload User model schema to avoid errors
+      # when running this task along with db:migrate
+      User.set_dataset :users
+      
+      u = User.new
```
Kartones: For the future, convention is to use verbose names, a `vis` is not terrible instead of `visualization` because at carto is very common, there's an endpoint etc., but `user` is way better than `u`

lib/tasks/setup.rake
```diff
@@ -39,6 +39,47 @@ DESC
...
+      u.password_confirmation = ENV['PASSWORD']
+      u.username = ENV['SUBDOMAIN']
+      u.database_host = ENV['DATABASE_HOST'] || ::Rails::Sequel.configuration.environment_for(Rails.env)['host']
+      u.save
+
+      raise u.errors.inspect if u.new?
+      puts "User #{u.username} created successfully"
+      
+      # 10 Gb of quota
+      quota = 1073741824
```
Kartones: Can I make an additional small request? As I see this I noticed we should also raise local platform limits (https://github.com/CartoDB/cartodb/blob/master/lib/tasks/platform_limits.rake) for the user, maybe to the ones that `increase_limits_for_twitter_import_users` has.

Kartones:
:+1:

lib/tasks/setup.rake
gfiorav: Ok! I like it better now that you pointed it out. Will change `u` to `user`

lib/tasks/setup.rake
gfiorav: Working on this

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5589/))

gfiorav:
This new update addresses code format and suggestions made

cc @Kartones

Kartones:
good!

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5592/))

PR #4881, opened 2015-08-05
New file with countries identified by iso code and its bounding box

### LINKED ISSUE #4796: check-in countries table as part of source
thinking in on-premises here, we should have the countries dataset as part of the source code, so we can use it in data-library header https://github.com/CartoDB/cartodb/issues/4795

ethervoid:
I've added the license required by OpenStreeMap for this content

Closes #4796 

// @matallo @javisantana  @gfiorav


matallo:
the thing is: I still need the_geoms to draw the country borders in the data-library public datasets page

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5593/))

javisantana:
this is not needed, the geometries are a frontend asset, backend does not need to know anything about countries. It just search datasets within the bbox the frontend sends

PR #4902, opened 2015-08-06
update like emails with competition

matallo:
CR @CartoDB/frontend


Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5624/))

javierarce:
:+1:

PR #4904, opened 2015-08-06
Fix title if creating table from query

### LINKED ISSUE #4801: Create dataset from query shows duplication modal
When creating a dataset from query, the modal of dataset duplication appears:
image

I think that before we were using a more specific text in this case. Is this new behaviour expected?

@viddo @saleiva

viddo:
Fixes #4801 

Actually using the table.duplicate that internally duplicate table from the query, this changes the title depending on that state

@javierarce for code review


javierarce:
:+1:

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5627/))

PR #4927, opened 2015-08-06
Fixing problem checking if there is only one layer in the map

### LINKED ISSUE #4917: Right now you can delete all data layers from a map
## STR
1. Create a map with a empty layer, for example.
2. Delete the layer.
3. It allows you to delete it! :D

It shouldn't let you to delete it because if not this could happen...

screen shot 2015-08-06 at 16 23 24

So, or we so a warning dialog or we "hide" the delete button within layer panel.
cc @saleiva

xavijam:
After changes about labels on top, we have to check number of layers with function 
``` map.layers.getDataLayers() ```, if not the calculation will be always 2, counting low basemap and labels layer, that's the reason why "remove" button always appears in each layer.

Fixes #4917
REVIEWER: @viddo


viddo:
:+1: but could this be testable?

xavijam:
Yeah, let's add a couple of tests.

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5652/))

xavijam:
Fixed tests, they were failing :dancer: (_tests are our friends_)

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5659/))

PR #4928, opened 2015-08-06
Fix gmaps deprecation modal

### LINKED ISSUE #4926: Google deprecation dialog is not working
Seems like it appears but the the "ok, proceed" button is not working and the dialog doesn't disappear.

image001-1 jpg 01d0d069 d3334340

viddo:
Fixes #4926 

cc @xavijam

### DIFF
--- a/lib/assets/javascripts/cartodb/table/mapview.js
+++ b/lib/assets/javascripts/cartodb/table/mapview.js
@@ -378,6 +378,7 @@ cdb.admin.MapTab = cdb.core.View.extend({
     dialog.ok = function() {
       self.map.set('provider', 'leaflet', { silent: true });
       self.setupMap();
+      this.close && this.close();
     };
 
     dialog.cancel = function() {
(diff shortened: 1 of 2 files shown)


lib/assets/javascripts/cartodb/table/mapview.js
```diff
@@ -378,6 +378,7 @@ cdb.admin.MapTab = cdb.core.View.extend({
     dialog.ok = function() {
       self.map.set('provider', 'leaflet', { silent: true });
       self.setupMap();
+      this.close && this.close();
```
xavijam: I feel you need to use ```self``` instead of this.

viddo: No, because `this` refers to the modal view instance, self refers to the `mapTab` view.

xavijam: Ahhh, you are right my friend, then let's go.

xavijam:
:es:

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5651/))

PR #4934, opened 2015-08-07
Only expose base_url attribute for public user views

### LINKED ISSUE #4932: Don't render base_url within organization public profile

xavijam:
Fixes #4932

@CartoDB/backend


Kartones:
:+1:

PR #4970, opened 2015-08-10
Allows to export image overlays

javierarce:
This PR updates the export static map system to include the overlays in the map. 
It also moves the export logic to the final export dialog (this is needed because the method that merges the overlays with the image preview needs some seconds and made necessary to add a loading pane)

Original issue: #4933


Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5722/))

javierarce:
@viddo & @rafatower: could you review front and back-end respectively, please?

app/controllers/carto/api/image_proxy_controller.rb
```diff
@@ -0,0 +1,26 @@
+
+module Carto
+  module Api
+    class ImageProxyController < ::Api::ApplicationController
+
+      ssl_required :show
+
+      def show
+        # image url
+        url = params[:url]
```
viddo: why not name the variable `image_url` instead? that would make the comment on the line above quite unnecessary, no? :)

javierarce: I must admit that I didn't read this code, I just merged it from the export image branch, but I'll have a look.

lib/assets/javascripts/cartodb/common/dialogs/static_image/export_image_result_view.js
```diff
@@ -48,6 +41,35 @@ module.exports = BaseDialog.extend({
...
+
+    if (cdb.config.get('static_image_upload_endpoint')) {
+      callback = function(url) {
+        self._exportImage(url);
+      };
+    }
+
+    this._loadMapImage(url, function(url) {
+      callback(url);
+    });
```
viddo: Wrapping a function like this, that contains the same method signature and no additional logic, adds no value. To get right this-context you can to use `.bind`. This and the lines above can be simplified to:

```js
var callback;
if (cdb.config.get('static_image_upload_endpoint')) {
  callback = this._exportImage;
} else {
  callback = function(url) {
    this._showResult({ content: url, type: 'url' });
  };
}
this._loadMapImage(url, callback.bind(this);
```

lib/assets/javascripts/cartodb/common/dialogs/static_image/export_image_result_view.js
```diff
@@ -55,6 +77,79 @@ module.exports = BaseDialog.extend({
...
+        self._showResult({ content: content, type: "html" });
+      },
+      error: function(error) {
+        cdb.editor.ViewFactory.createDialogByTemplate('common/templates/fail', { msg: error.errors })
+        .render().appendToBody();
+      }
+    });
+  },
+
+  mergeAnnotations: function(mapImage, callback) {
```
javierarce: I need to add a _ before the name of this method.

lib/assets/javascripts/cartodb/common/dialogs/static_image/export_image_result_view.js
```diff
@@ -55,6 +77,79 @@ module.exports = BaseDialog.extend({
     this._panes.bind('tabEnabled', this.render, this);
   },
 
+  /* Load first the map image and then merge with the overlays rendered frontend side */
+  _loadMapImage: function(url, callback) {
+    var self = this;
+    var mapImage = new Image();
+    mapImage.crossOrigin = 'Anonymous';
+    mapImage.onload = function() {
+      self.mergeAnnotations(mapImage, function(url) {
```
viddo: How about `self.mergeAnnotations(mapImage, callback || function(){})` to make it a little more succinct? Or even `self.mergeAnnotations(mapImage, callback)` if possible (but looks like the method requires a callback, right?

viddo:
Apart from comments, looks :+1:

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5724/))

rafatower:
nothing to add :+1:

javierarce:
Thanks for the help!

app/controllers/carto/api/image_proxy_controller.rb
javierarce: On a closer look, I'm guessing the use of 'show' for the name is to have a simple api url like this:
```url: '/api/v1/image_proxy'```

https://github.com/CartoDB/cartodb/pull/4970/files#diff-2cfa14c28532858746536b894ca556e2R113

@javisantana: any idea?

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5738/))

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5736/))

app/controllers/carto/api/image_proxy_controller.rb
javisantana: It's just url because the service is called image_proxy but I'm not the best at naming

PR #4971, opened 2015-08-10
Close database connections before drop test DB

ethervoid:
We have some leaks in the test database that makes jekins been stuck trying to drop the test database

Trace
```
15:50:05 dropdb: database removal failed: ERROR:  database "carto_db_test" is being accessed by other users
15:50:05 DETAIL:  There are 6 other sessions using the database.
15:50:05 [sequel] Dropped database 'carto_db_test'
15:50:06 createdb: database creation failed: ERROR:  database "carto_db_test" already exists
15:50:06 [sequel] Created database 'carto_db_test'
```

### DIFF
--- a/spec/requests/admin/tables_spec.rb
+++ b/spec/requests/admin/tables_spec.rb
@@ -19,7 +19,7 @@ describe Admin::TablesController do
   end
 
   before(:each) do
-    CartoDB::NamedMapsWrapper::NamedMaps.any_instance.stubs(:get => nil, :create => true, :update => true)
+    stub_named_maps_calls
     CartoDB::Varnish.any_instance.stubs(:send_command).returns(true)
     @db = Rails::Sequel.connection
     delete_user_data $user_1
@@ -30,8 +30,9 @@ describe Admin::TablesController do
   end
 
   after(:all) do
-    $user_1.destroy
+    stub_named_maps_calls
     delete_user_data($user_1)
+    $user_1.destroy
   end
 
   describe 'GET /dashboard' do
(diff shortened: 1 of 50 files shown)


Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5725/))

ethervoid:
Please @Kartones take a look, when green and if it's ok I'd merge in master

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5739/))

spec/requests/admin/tables_spec.rb
```diff
@@ -30,8 +30,9 @@ def app
   end
 
   after(:all) do
-    $user_1.destroy
+    stub_named_maps_calls
     delete_user_data($user_1)
+    $user_1.destroy
```
Kartones: uhmm this might actually be a bug from me. No individual test battery should destroy global `$user_x`...  only clear their data

Kartones: It works now because sadly each battery recreates the users but when we really make them "global" would fail.

Kartones:
Maybe for next sweep we could use inheritance and make by default all `after(:all)` and `after(:each)` stub the named map calls? Or else monkeypatch `User.destroy` in tests so it calls the stub and then the `super`?

I know get what you mean about repetition, but +1

ethervoid:
Yes I agree with you. I thought to apply that measures but I'm a bit run of time, after the release of this 2 PR I'd take a look to improve it

PR #4980, opened 2015-08-11
Prevent polling requests from being queued up

### LINKED ISSUE #4939: geocoding checks freeze the editor
When you have some geocodings pending (lets say 3) there are checks every second for all of them so if the editor takes some time to answer the requests are queued and every action you try to do in the editor takes a lot of time because there are no slots to send the requests

Geocoding and other batch process requests should not be raised if the previous check didn't arrive
 
cc @xavijam 

screen shot 2015-08-07 at 11 05 20 am

alonsogarciapablo:
Fixes #4939.

I've basically extracted some of the polling logic from `GeocodingModel` and `ImportModel` into `GeocodingModelPoller` and `ImportModelPoller`, which are specifications of a generic `Poller`. Instead of using an interval to trigger `fetch` requests periodically, pollers now wait for the previous request to be received before triggering new ones :)

@viddo could please take a look? I'm looking forward for your feedback! Thanks!


Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5743/))

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5745/))

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5777/))

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5791/))

lib/assets/javascripts/cartodb/common/background_polling/models/geocoding_model_poller.js
```diff
@@ -0,0 +1,22 @@
...
+
+  var POLLING_INTERVAL = 2000;
+
+  var options = {
+    interval: POLLING_INTERVAL,
+    condition: function(model) {
+      return model.hasFailed() || model.hasCompleted();
+    },
+    error: function() {
+      this.model.trigger("change");
```
viddo: what's the reason for using `this.model` and not just `model`, like done in `condition` fn above?

lib/assets/javascripts/cartodb/common/background_polling/models/poller.js
```diff
@@ -0,0 +1,76 @@
...
+
+Poller.prototype._fetch = function() {
+  var self = this;
+  if (!self.polling) {
+    self.polling = true;
+    self.model.fetch({
+      success: function() {
+        self.polling = false;
+        self.numberOfRequests++;
+        if (!self.condition || (self.condition && !self.condition(self.model))) {
```
viddo: condition of what? perhaps could find a better name that fits its purpose, to make the using code clearer? 

Also, I'd try to invert the negations here (and thus the implementation functions) since it's easier to read positive checks, i.e.:

```js
if (_.isFunction(self.shouldContinuePolling) && self.shouldContinuePolling(self.model)) {
  …
```

viddo:
Apart from comments it LGTM :+1:

lib/assets/javascripts/cartodb/common/background_polling/models/poller.js
viddo: …also make the code a little more defensive by checking if the prop is actually a function

lib/assets/javascripts/cartodb/common/background_polling/models/geocoding_model_poller.js
alonsogarciapablo: Good catch! I thought about this but then forgot to revisit it. There's no specific reason.

I will also pass `model` to the `error` callback so that there's no need to know about the internals of the Poller:

```
...
error: function(model) {
   model.trigger("change");
}
...
```

PR #4981, opened 2015-08-11
Bounding box store in canonical tables at import time and bbox filtering

ethervoid:
* [x] Add bounding box to canonical visualizations in import time
* [x] Recalculate when the_geom is updated
* [x] Rake task to update a user visualizations adding or updating the bounding box field
* [x] Needed migrations for all of this
* [x] Add tests
* [x] Spec helper now don't clean postgis tables 
* [x] Endpoint to filter canonical visualizations by bbox


Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5746/))

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5749/))

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5772/))

NEWS.md
```diff
@@ -39,6 +39,8 @@ cd lib/sql; sudo make all install
 * Properly and fully disallowing multilogins, by killing other existing sessions upon login
 * Added new Platform Limit for concurrent syncs per user. Currently 2 hours TTL, 3 syncs per user max. allowed
 * Importer now tries to import shapefile zips without .prj file, by setting 4326 projection.
+* Now canonical tables store the bounding box at import time. It's also recalculated when `the_geom` change. This need to install the postgis extension in the metadata database
+* Canonical visualizations can be filtered by bounding box
```
javisantana: this is private api, I don't think we should write this in NEWS. In this case we should write soemthing like "new dataset pages blagbla..."

app/controllers/api/json/records_controller.rb
```diff
@@ -31,6 +33,7 @@ def update
       begin
         resp = @table.update_row!(params[:id], params.reject{|k,v| REJECT_PARAMS.include?(k)}.symbolize_keys)
         if resp > 0
+          BoundingBoxHelper.update_visualizations_bbox(@table)
```
javisantana: this will not have any effect, since you are calculating the bbox using stats updating a row is not going to change anything.

Calculating the real bbox for a simple row update could be killer, it's better to update from time to time

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5779/))

app/queries/carto/visualization_query_builder.rb
```diff
@@ -216,6 +222,13 @@ def build
       query = query.where("ARRAY[?]::text[] && visualizations.tags", @tags)
     end
 
+    if @bounding_box
+      bbox_coords = @bounding_box.split(',').select { |coord| true if Float(coord) rescue false }
```
javisantana: I think this sanitize should be done in the controller

app/controllers/api/json/records_controller.rb
ethervoid: I assume this is applicable to insert and delete too. I think the stats from time to time should change if `the_geom` change. If not, I should remove this lines and only recalculate in import and georeference time

app/queries/carto/visualization_query_builder.rb
ethervoid: This query builder could be called from multiples controllers, I put that here to avoid possible side effects because is a bit dangerous

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5787/))

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5790/))

Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5797/))

ethervoid:
restest this please

ethervoid:
Please @javisantana @Kartones recheck this please. Its ready to merge

app/controllers/carto/api/visualization_searcher.rb
```diff
@@ -98,6 +103,11 @@ def compose_shared(shared, only_shared, exclude_shared)
         end
       end
 
+      def validate_parameters(types, parameters)
+        if (!params.fetch(:bbox, nil).nil? && !types.include?(Carto::Visualization::TYPE_CANONICAL))
```
Kartones: Careful, we can query using multiple types... you want "at least to have canonical type" or "only have canonical type"?
Because if the 2nd, should be `&& types != [Carto::Visualization::TYPE_CANONICAL] `

app/helpers/bounding_box_helper.rb
```diff
@@ -0,0 +1,117 @@
+# encoding: utf-8
+
+module BoundingBoxHelper
+
+  DEFAULT_BOUNDS = {
```
Kartones: I'd prefer you to link [Map constants](https://github.com/CartoDB/cartodb/blob/master/app/models/map.rb#L40) to avoid repeating them

ethervoid: The idea is to refactor the Map part to avoid repeat all the bounding box logic but I didn't want to touch anything from maps yet

app/controllers/carto/api/visualization_searcher.rb
ethervoid: Good catch :)

app/helpers/bounding_box_helper.rb
```diff
@@ -0,0 +1,117 @@
...
+  end
+
+  def self.calculate_bounding_box(db,table_name)
+    get_table_bounds(db,table_name)
+  rescue Sequel::DatabaseError => exception
+    CartoDB.notify_exception(exception, { table: table_name })
+    raise BoundingBoxError.new("Can't calculate the bounding box for table #{table_name}. ERROR: #{exception}")
+  end
+
+  def self.get_table_bounds(db,table_name)
```
Kartones: uhmmm in fact this code is almost the same (if not the same) as the one at Map model, no?
What about extracting the common logic so it does not duplicate the code? (I see sometimes it uses the db, but the logic is exactly the same)

ethervoid: Yup is the same. Here I only want to add bbox to visualizations. I want to make a refactor in other PR when this becomes stable and working

app/helpers/bounding_box_helper.rb
```diff
@@ -0,0 +1,117 @@
...
+  end
+
+  def self.to_polygon(minx, miny, maxx, maxy)
+    "ST_Transform(ST_Envelope('SRID=4326;POLYGON((#{minx} #{miny}, #{minx} #{maxy}, #{maxx} #{maxy}, #{maxx} #{miny}, #{minx} #{miny}))'::geometry), 3857)"
+  end
+
+  private
+
+  def self.save_bounding_box(bounds, table_name, column_name, id)
+    update_sql = "UPDATE #{table_name} SET #{column_name} = ST_Transform(ST_Envelope('SRID=4326;POLYGON((" \
```
Kartones: You have a nice `self.to_polygon` just above, why not using it to make the query smaller for human eyes?

Kartones: Also, I prefer doing `%Q{ xxx }` insead of plain double-quoted strings for SQL stuff, it also allows variable interpolation, etc. and allows multiline by default without `\`:D

ethervoid: Yeah :+1:

Kartones:
ok then, just add a task to refactor bounding box code so we don't forget, but :+1: for the rest apart from the small comments.

If I've read correctly it won't do anything unless you're searching canonical tables AND specify a `bbox` param right? So is safe to have it live meanwhile the rake runs as nothing will query with that yet.

ethervoid:
Yep, it will only be triggered if there is a search against canonical and bbox at the same time so should be safe. I'll make the changes now

ethervoid:
Retest this please

app/queries/carto/visualization_query_builder.rb
javisantana: What I think is not right is passing the bbox as a string you need to parse. bbox is always expected in 4 floats (maybe grouped in 2 arrays), never as a string you need to parse. Create a helper if you need to parse from string but keep things as predecible as possible

ethervoid: I've made [this refactor](https://github.com/CartoDB/cartodb/commit/b9584a17f401c5ffec9cbfd45bec981b20a26a90).

javisantana:
:es:

ethervoid:
I've been thinking about all the possibilities and we have to make this deploy in two phases. The main reason is that we are going to get error when people start to import datasets and try to save the `bbox` that is not ready yet

Issue #4962, opened 2015-08-10
Deploy the new CDB_CartodbfyTable2 to production


rafatower:
Some remarks about this:
- Extension code in this PR: https://github.com/CartoDB/cartodb-postgresql/pull/78
- Name to be `CDB_CartodbfyTable` and backwards compatible. This will force a 2-step deploy since there's a dependency between editor and extension.
- No need to `move_to_schema` but probably check error scenarios
- Columns `created_at` and `updated_at` will be no longer present, this might impact editor's FE (unlikely) and BE (probably)

cc/ @pramsey

rafatower:
@Kartones a couple of questions about [cartodbfy](https://github.com/CartoDB/cartodb/blob/582dcc957bf7f997b6c3ba04862f39c628ed21a0/app/models/table.rb#L1155-L1214) method:
- Why not calling `SELECT cartodb.CDB_CartodbfyTable('#{schema_name}','#{table_name}')` directly?
- Can we get rid of the workaround for ver < 0.5.0? I guess so...

saleiva:
Soooo looking fwd to this

rafatower:
Darn, we are calling `cartodbfy` twice:
```
# 1st call
(rdb:1) puts Kernel.caller.grep(%r{/src/cartodb/})                                                                                                                                                                                             
/home/developer/src/cartodb/app/models/table.rb:1157:in `cartodbfy'
/home/developer/src/cartodb/app/models/table.rb:344:in `block in import_cleanup'
/home/developer/src/cartodb/app/models/user.rb:616:in `in_database'
/home/developer/src/cartodb/app/models/table.rb:287:in `import_cleanup'
/home/developer/src/cartodb/app/models/table.rb:374:in `before_create'
/home/developer/src/cartodb/app/models/table/user_table.rb:184:in `before_create'
/home/developer/src/cartodb/app/models/table.rb:81:in `save'
/home/developer/src/cartodb/app/models/table_registrar.rb:18:in `register'
/home/developer/src/cartodb/app/connectors/importer.rb:174:in `persist_metadata'
/home/developer/src/cartodb/app/connectors/importer.rb:64:in `register'
/home/developer/src/cartodb/app/connectors/importer.rb:49:in `block in run'
/home/developer/src/cartodb/app/connectors/importer.rb:48:in `each'
/home/developer/src/cartodb/app/connectors/importer.rb:48:in `run'
[...]

PR #4988, opened 2015-08-11
Use public function CDB_CartodbfyTable #4962

rafatower:
Use public function CDB_CartodbfyTable instead of a set of private
functions (_CDB_check_prerequisites, _CDB_drop_triggers, etc.).

Also remove old code to ensure compatibility when organizations where
added.

### DIFF
--- a/app/models/table.rb
+++ b/app/models/table.rb
@@ -1153,61 +1153,13 @@ class Table
   end
 
   def cartodbfy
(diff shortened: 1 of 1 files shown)


Cartofante:
Frontend tests were OK :+1: ([details](http://clinker.cartodb.net/jenkins/job/CartoDB-PR-testing/5763/))

Kartones:
:_) :+1:

PR #4999, opened 2015-08-12
Use pecan module instead of prior code from cartodb.js

viddo:
Since extracting pecan-specific code to an external module, use this new module as dependency instead (of cartodb.js where the code were located previously).

Some details:
- The [`_analyzeStats` method (and related code)](https://github.com/CartoDB/cartodb/compare/pecan-module?expand=1#diff-8f491e9009d9b426cec44ec6862a2446) was also extracted to [pecan module](https://github.com/CartoDB/pecan/blob/master/src/has-enough-to-guess.js), to keep "magic numbers" in one place (see https://github.com/CartoDB/pecan/issues/1 for motivation/planned work related to this)
- Changed the [guessMap](https://github.com/CartoDB/pecan/blob/master/src/guess-map.js) method signature slightly while I extracted the code, due to:
  - only ask for what's really needed
  - remove notion of Backbone in the pecan code
  - make the underscore dependency explicit and to be provided from caller (i.e. this repo), since that dependency is fixed and used from cartodb.js
- Set laxed dependency to `v1.x` to automatically get minor and patch updates on `npm install` (e.g. deploys), since they should be backward compatible. This will allow us to update pecan code w/o have to jump through all the hoops of creating code changes in this repo every time.

Verified working locally, deployed on staging ded03 if you want to have a look yourself, should work just like before.

Old code will be removed from cartodb.js after this is merged and deployed.

ping @javisantana @javierarce @fdansv 
cc @ohasselblad


viddo:
Bump! (and re-deployed on ded03 due to update w/ master)

javierarce:
:+1:  LGTM

viddo:
@javierarce apart from looking at the code, tested on staging too? You guys know these parts better than I do and I'd like to make sure we're not breaking anything when releasing this. Do we have some smoke tests for Pecan?

javierarce:
No, sorry, I'll have a look to staging now. 

I don't think we have smoke tests for Pecan. I'll add a ticket to add some tests.

javierarce:
Ok, I've tested it in staging using datasets from the Common Library and it seems to behave OK.

Issue #4996, opened 2015-08-12
Link Synchronization to Visualizations by vis.id


Kartones:
Currently synchronizations link to visualizations via the name of the canonical visualization related table. This is prone to issues and poses impediments for advanced synchronization scenarios.

Proposal is to migrate to joining by visualization.id (canonical one), so relations are direct, quicker and for external sources syncs also allows to handle scenarios where user import-syncs -> unsyncs -> import-syncs again.

This requires a small migration to update all existing synchronizations, so probably needs to be done in two steps, one adding the new field, logic and migration script, second one enforcing

Kartones:
When https://github.com/CartoDB/cartodb/issues/4775 goes live syncs will start to function by linking via this vis.id, so this ticket is done

PR #5002, opened 2015-08-12
Link Synchronization to Visualizations by vis.id

Kartones:
Relates to https://github.com/CartoDB/cartodb/issues/4996

After running manually the rake and all existing (and new) syncs have a visualization.id we can think about deprecating the visualization <- table_name -> synchronization link

### DIFF
--- a/app/models/data_import.rb
+++ b/app/models/data_import.rb
@@ -663,13 +663,18 @@ class DataImport < Sequel::Model
       synchronization.log_id  = log.id
 
       if importer.success?
+        imported_table = ::Table.get_by_table_id(self.table_id)
+        if !imported_table.nil? && imported_table.table_visualization
+          synchronization.visualization_id = imported_table.table_visualization.id
+        end
+
         synchronization.state = 'success'
         synchronization.error_code = nil
         synchronization.error_message = nil
       else
         synchronization.state = 'failure'
         synchronization.error_code = error_code
-        synchronization.error_message = get_error_text
+        synchronization.error_message = get_error_text[:title] + ' ' + get_error_text[:what_about]
       end
       log.append "importer.success? #{synchronization.state}"
       synchronization.store
@@ -788,7 +793,7 @@ class DataImport < Sequel::Model
     payload.merge!(
       file_url_hostname: URI.parse(public_url).hostname
     ) if public_url rescue nil
-    payload.merge!(error_title: get_error_text) if state == STATE_FAILURE
+    payload.merge!(error_title: get_error_text[:title]) if state == STATE_FAILURE
     payload
   end
 
--- a/app/models/table.rb
+++ b/app/models/table.rb
@@ -155,7 +155,7 @@ class Table
                 # Handy for rakes and custom ghost table registers, won't delete user table in case of error
                 :keep_user_database_table
 
-  # Getter by table uuid or table name using canonical visualizations
+  # Getter by table uuid osing canonical visualizations
   # @param table_id String
   # @param viewer_user User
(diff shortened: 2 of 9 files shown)


app/models/data_import.rb
```diff
@@ -663,13 +663,18 @@ def update_synchronization(importer)
...
+        end
+
         synchronization.state = 'success'
         synchronization.error_code = nil
         synchronization.error_message = nil
       else
         synchronization.state = 'failure'
         synchronization.error_code = error_code
-        synchronization.error_message = get_error_text
+        synchronization.error_message = get_error_text[:title] + ' ' + get_error_text[:what_about]
```
Kartones: this is a funny mini bugfix... in case of error it was exploding and not saving the error_mesage :/

app/models/table.rb
```diff
@@ -155,7 +155,7 @@ def is_raster?
                 # Handy for rakes and custom ghost table registers, won't delete user table in case of error
                 :keep_user_database_table
 
-  # Getter by table uuid or table name using canonical visualizations
```
Kartones: Tested that comment was wrong, it attacks against UUIDs always

Kartones:
@rafatower CR please :smiley:

app/models/data_import.rb
rafatower: wasn't there automatic casting to a string? anyway, ok with it of course

app/models/table.rb
rafatower: small typo

db/migrate/20150812120815_add_visualization_id_to_synchronizations.rb
```diff
@@ -0,0 +1,21 @@
...
+      add_column :visualization_id, 'uuid'
+    end
+
+    Rails::Sequel.connection.run(%Q{
+      ALTER TABLE "synchronizations"
+        ADD CONSTRAINT  visualization_id_fkey
+        FOREIGN KEY (visualization_id)
+        REFERENCES visualizations(id)
+        ON DELETE CASCADE
+    })
```
rafatower: to keep an eye when deploying. I suggest to do it first time in the morning because this can be heavy for dbm

lib/tasks/sync_tables.rake
```diff
@@ -16,4 +15,48 @@ namespace :cartodb do
...
+            end
+          else
+            printf 'S'
+          end
+      end
+      else
+        puts "\nSync id '#{record[:id]}' errored: missing synchronization entry"
+      end
+    }
+
```
rafatower: do you think there might be wrong synchronization entries? that could prevent the migration above from applying to production. worth checking locally

rafatower: the fetch_all will be a bit heavy depending on the number of rows. Don't know how many we have right now..

rafatower:
:+1: basically ok with it, my comments are mostly related to deploying it than anything.

app/models/data_import.rb
Kartones: mm not at least here, was precisely the error, storing a hash and then was breaking

lib/tasks/sync_tables.rake
Kartones: `fetch_all` fetches only ids, using pure SQL (no ruby Models) and is just around 400 items, I wouldn't worry

Kartones: And yes, there will surely be :D But in local I already messed up with them, upon running the migration it's ok as everything is left with default value (NULL), it is upon running the rake where things can break and thus I added the rescues

PR #5006, opened 2015-08-12
Log the invalidation time in postgresql

luisbosque:
This allows writing to postgresql log the time to run the invalidation against varnish.

Using log raise level with plpy is the more appropiate since LOG level is much lower in log_min_messages (http://www.postgresql.org/docs/9.3/static/runtime-config-logging.html#GUC-LOG-MIN-MESSAGES) than in client_min_messages (http://www.postgresql.org/docs/9.3/static/runtime-config-logging.html#GUC-CLIENT-MIN-MESSAGES).
With a proper logging configuration the invalidation time should not bother in the client responses but it will appear in the postgresql log.

By default when a RAISE is executed, the context of the execution is shown in the log. Getting the value before the raise and setting it after does the trick.

/cc @rochoa @javisantana @pramsey

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1939,8 +1939,17 @@ TRIGGER
         retry = #{varnish_retry}
         trigger_verbose = #{varnish_trigger_verbose}
 
-        import httplib
-        import time
+        if 'httplib' not in GD:
+          import httplib
+          GD['httplib'] = httplib
+        else:  
+          httplib = GD['httplib']
+        
+        if 'time' not in GD:
+          import time
+          GD['time'] = time
+        else:  
+          time = GD['time']
 
         start = time.time()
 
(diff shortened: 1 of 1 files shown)


lbosque:
As suggested by @rochoa this may easily moved to the function calling the invalidation method so it can be reused for any invalidation function.

The important here is validating (or not) the concept. Writing this in PL/pgSQL is basically the same lines.

javisantana:
I'd add some more information: 

- num of retries
- connection time
- errors

I understand this is temporal, right? I'd add this only for some users.

Also, doing imports is time consuming, could we import them and cache in GD?

```
if 'time' not in GD:
  import time
  GD['time'] = time
else:  
  time = GD['time']
```

lbosque:
The GD trick works fine :)

PR #5011, opened 2015-08-13
Increments bottom padding of create dialog

alonsogarciapablo:
The bottom padding didn't feel quite right and I noticed some stuff was unreachable when scrolling all the way to the bottom of the dialog's content (see first screenshot below). This PR fixes it.

Before:

screen shot 2015-08-13 at 11 41 31
screen shot 2015-08-13 at 11 47 48
screen shot 2015-08-13 at 11 47 35

After:

screen shot 2015-08-13 at 11 51 17
screen shot 2015-08-13 at 11 50 53
screen shot 2015-08-13 at 11 51 08

@viddo PTAL! Thanks!

### DIFF
--- a/app/assets/stylesheets/common/create/create_listing.css.scss
+++ b/app/assets/stylesheets/common/create/create_listing.css.scss
@@ -7,12 +7,12 @@
 .CreateDialog-listing {
   display: block;
   padding-top: 52px; // padded on top of the normal Dialog-body it seems, for content to appear below the subheader (.Filters) fold
-  padding-bottom: 70px;
+  padding-bottom: 100px;
 }
 .CreateDialog-listing.CreateDialog-listing--withFlashMessage { padding-top: 126px }
 .DatasetsPaginator {
   width: 940px;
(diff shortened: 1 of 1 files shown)


viddo:
:+1: 
FYI, we have some inconsistencies when it comes to how this is handled in the modals, in most cases we use [these `.Dialog-bodyInnerExpandedWithSubFooter` rules](https://github.com/CartoDB/cartodb/blob/master/app/assets/stylesheets/common/dialog.css.scss#L247-L250) on the body element to give the extra space to an additional footer (or footer+pre-footer). Not sure if those could be applied on the creation modal though.

alonsogarciapablo:
Thanks @viddo. I'm trying to reuse those existing rules.

viddo:
Ah, it was more of a remark for the future than a request to change the code per se. If you have it working pehaps not worth it, depends on how much changes you need to do I guess, your call. :)

alonsogarciapablo:
If it's easy I will change it, otherwise I'll just merge this.

PR #5016, opened 2015-08-13
Add new fields to metadata modal

### LINKED ISSUE #4803: Recover source and license options in dataset metadata modal
Previously, besides description and tags, the parameters source and license were available:

image

cc @viddo @saleiva 

edit after offline conversations:

Add three new fields:
- [x] source
  - free text input, markdown supported
- [x] attributions
  - free text input, markdown supported
- [x] license
  - [x] pass licenses from [server-side model](https://github.com/CartoDB/cartodb/blob/master/app/models/carto/licenses.rb) to frontend config (to be available on `cdb.config` JS obj)
  - [x] add dropdown w/ list from above

viddo:
Fixes #4803 

While I didn't have any mockups this time I simply added the new fields between desc and tags, i.e.:

Empty values/placeholders:
screen shot 2015-08-13 at 16 22 35

Licence dropdown:
screen shot 2015-08-13 at 16 57 18

@saleiva @josecruz OK? Or did you have another idae on how it should look?
@csobier and @brittanymicek for copy (see the placeholder texts above). also notice the "-" for the non-selected license perhaps we should have some text there instead?

@javierarce can you do the code review, please?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,5 @@
 The development tracker for cartodb is on github:
-http://github.com/cartodb/cartodb/
+http://github.com/cartodb/cartodb
 
 Bug fixes are best reported as pull requests over there.
 Features are best discussed on the mailing list:
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,3 @@
-
 3.10.3 (2015-mm-dd)
 ---
 * Mailchimp decorator enables category wizard and legends [#3874](https://github.com/CartoDB/cartodb/pull/3874)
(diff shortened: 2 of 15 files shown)


javierarce:
:+1: Looks good to me.

saleiva:
That icon on the dropdown should be the default one.
Also I'd not repeat 3 times the md thingym I propose we leave it just under
the description

josecruz:
What's the meaning of this right icon?

screen shot 2015-08-13 at 18 47 13

csobier:
I can suggest some text for the License field once we know what the icon indicates? Is it a tooltip? It it mandatory?  Otherwise, here is my best guess: 
"Select the type of license required for copying, distributing, or modifying this dataset"

Additionally, I would keep the field descriptions consistent- either all statements or all questions.  Typically, product UI field descriptions are statements so the Source description should say something like "Enter the source of the data"

viddo:
The icon is a mistake, I suspect the change of the icons recently (should be the default dropdown/chevron icon), fixed now:

screen shot 2015-08-14 at 15 27 49

@csobier It's a standard dropdown (see new screenshot above). It's value is optional (thus the "-" option at the top for, which I was asking if we should replace that with some text instead? Apart from that, updated the placeholders, I'll keep that guideline in mind for the next forms I'll do :+1:

saleiva:
The margin between the inputs should be the same than between the first and
the second one.

viddo:
screen shot 2015-08-14 at 17 25 48

saleiva:
:es:

viddo:
I'll go ahead and merge this, if you want any additional changes just create a new issue and assign me :)

Issue #5015, opened 2015-08-13
Hide attribution when the static map is active


javierarce:

javisantana:
why the attribution needs to be hidden?

javierarce:
Just because if I don't remove it it will appear twice (in the map and in the export area)

PR #5017, opened 2015-08-13
Remove map attribution

javierarce:
This PR removes the attribution of the map when the static export tool is active. 

Original ticket: #5015


javierarce:
Have a look, @javisantana. I don't know of a better way to remove the attribution.

javierarce:
bump

PR #5021, opened 2015-08-13
Grants publicuser when creating a new db user/role

### LINKED ISSUE #5019: Org users accessing other org users public tables
Users in the same organization should be able to query other users' public tables, right now it's not possible because users are not granted with the same permission `publicuser` has.

As an initial fix for this the best way is to `grant publicuser to :username;`.

rochoa:
This takes advantage of `publicuser` already being granted with usage permissions on user schemas.

TODO:
 - [x] rake task to grant publicuser to existent users?

This shouldn't be handled like this if we had something like https://github.com/CartoDB/cartodb-postgresql/issues/111

This will fix #5019


rochoa:
cc/ @Kartones @javisantana @zenitraM 

What do you think?

Kartones:
This also should mean  that scripts that provision a new DB machine (or the OS install's `carto_db_xxx` DB) should also setup `publicuser` before any user is created no? upon initial init I mean.

All that simplifies code, more than welcome!

rochoa:
See https://github.com/CartoDB/cartodb/blob/master/lib/tasks/setup.rake#L93-L108 + https://github.com/CartoDB/cartodb/blob/master/lib/tasks/setup.rake#L21

About db provision, check https://github.com/CartoDB/cartodb-postgresql/issues/111

Kartones:
Then marvellous :dancer:

rochoa:
Ops, I missed the last two comments :disappointed:

rochoa:
bump :exclamation:

lib/tasks/db_maintenance.rake
```diff
@@ -317,6 +317,26 @@ namespace :cartodb do
...
+        user = Carto::User.where(id: user_id).first
+        # already granted users will raise a NOTICE
+        grant_query = "GRANT publicuser to \"#{user.database_username}\""
+        conn = user.in_database(as: :cluster_admin)
+        begin
+          conn.execute(grant_query)
+        rescue => e
+          log("Failed to execute `#{grant_query}`", :grant_publicuser_to_all_users.to_s, user.database_host)
+        ensure
+          conn.close unless conn.nil?
```
Kartones: nice, we can try this, if works as it should it shouldn't leave open connections in the pooler (and we won't need to use the old model and forcibly close all user connections)

Kartones:
so... for me is a :+1:

PR #5044, opened 2015-08-17
Only use new metadata fields for dataset vis

### LINKED ISSUE #5043: We should only show new datasets fields in dataset modals
**Problem**

Now we are showing dataset only fields in edit metadata modal for maps.

**Solution**

This new fields should only appear we we are editing datasets metadata

viddo:
Fixes #5043

@javierarce can you review?
cc @ethervoid


javierarce:
:ok_hand: Looks good to me

PR #5047, opened 2015-08-17
Do not include hidden layers in the layer definition of CartoDBLayers

### LINKED ISSUE #4186: Layer hidden in visualization
## STR:

1. Create a visualization with only one layer (with geometries).
2. Go to the map and make the layer no visible.
3. It will work, but now reload the page.
4. The layer will appear in the map but if you check the layer in the panel, it appears as not visible.

layer-hidden

alonsogarciapablo:
Fixes #4186.

hidden_layers

@xavijam what do you think? Thanks!


xavijam:
:es:

PR #5048, opened 2015-08-17
Update cartodb.js due to migrated pecan module

viddo:
This updates the cdb module to latest HEAD version, since pecan code was extracted in #4999 and actual code used is used through that instead through cartodb.js

ping @javierarce @javisantana


javierarce:
:ok_hand:

PR #5057, opened 2015-08-17
4962 cleanup timestamp columns

rafatower:
This branch removes the dependency of the editor upon the `created_at` and `updated_at` columns traditionally added to user tables during cartodbfy.

Upcoming version of cartodbfy will not add them so the editor's code should be compatible with both scenarios:
- existing tables have the `created_at` and `updated_at` columns and they should not be removed because they have associated triggers.
- new tables won't have those columns but nevertheless they should be added to the dashboard.

See https://github.com/CartoDB/cartodb/issues/4962 for details

@Kartones please review (the tricky question is not what's wrong but what I might have missed).


spec/models/data_import_spec.rb
```diff
@@ -173,27 +173,6 @@
     table.records[:rows].should have(10).items
   end
 
-  it "don't touch created_at/updated_at fields if already present in the
```
rafatower: related to https://github.com/CartoDB/cartodb/issues/5074

rafatower:
retest this, please

spec/requests/api/json/columns_controller_shared_examples.rb
```diff
@@ -26,7 +26,7 @@
     it "gets the columns from a table" do    
       get_json api_v1_tables_columns_index_url(params) do |response|
         response.status.should be_success
-        (response.body - default_schema).should be_empty
+        (response.body - default_schema - [["created_at", "date"],["updated_at", "date"]]).should be_empty
```
Kartones: Maybe adding here a comment indicating that this fields are legacy and no longer required?

Kartones:
:+1: Looks simple regarding the editor, as I see mostly specs so go ahead!

spec/requests/api/json/columns_controller_shared_examples.rb
rafatower: ok

PR #5064, opened 2015-08-17
Public dataset fixes

xavijam:
- [x] One click or edit in CartoDB button appearance.
- [x] No table hover.
- [x] No table double click.
- [x] Limit maps related with public dataset to 3.
- [x] Fixed "no rows matched" with public map view problem in export view.
- [x] User breadcrumb style.
- [x] Remove public table shadow.

cc @saleiva

### DIFF
--- a/app/views/admin/visualizations/public_dataset.html.erb
+++ b/app/views/admin/visualizations/public_dataset.html.erb
@@ -257,31 +257,18 @@
           </li>
         <% end %>
         <%# at most 3 cards per line, visbility is handled by CSS depending on viewport size %>
-        <% ((3 - @total_visualizations.count + 3) % 3).times.each do %>
-          <li class="MapsList-item MapsList-item--fake">
-            <div class="MapCard">
-              <div class="MapCard-header MapCard-header--fake"></div>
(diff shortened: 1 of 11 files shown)


app/views/admin/visualizations/public_dataset.html.erb
```diff
@@ -234,7 +234,7 @@
 
     <% if @total_visualizations.count > 0 %>
       <ul class="MapsList PublicMap-mapsList">
-        <% @total_visualizations.each do |vis| %>
+        <% @total_visualizations.take(3).each do |vis| %>
```
xavijam: @javisantana ```take``` function is ok right? Just for limiting the number of maps in the "Maps created with this dataset", if not all maps will appear.

saleiva:
:es:

app/views/admin/visualizations/public_dataset.html.erb
javisantana: no idea :D

xavijam: Checked we use it in is other places:
https://github.com/CartoDB/cartodb/blob/master/lib/cartodb/stats/api_calls.rb#L130

PR #5083, opened 2015-08-18
Fix revoking access to a 3rd party service modal loading state is wrong 

### LINKED ISSUE #4749: Revoking access to a 3rd party service modal loading state is wrong
While the confirmation modal is fine, the loading one is wrong. It should only show the loading indicator and not the header.

viddo:
Fixes #4749 

Loading state now being proper like in the rest of the modals (rendered w/o header):
disconnect-loading
_here demo'ed using an already disconnected account, so it closes the modal and displays the error next to the box (that was from before)_.

@xavijam pretty please review? cc @saleiva


xavijam:
:es:

saleiva:
+1

Issue #4891, opened 2015-08-05
when like a dataset email should say dataset instead of map


matallo:
as with share maps/datasets

/cc @ethervoid

PR #5093, opened 2015-08-18
Dataset and maps have different emails now

### LINKED ISSUE #5069: Like email on dataset says "Your map got some love"
But it's a dataset

ethervoid:
We have different email for likes that comes from datasets or maps. 

At first I tried to use the same email but finally I realized that in the future is possible that we want to have different emails for it so I've split it

Closes #5069, #4891


ethervoid:
captura de pantalla 2015-08-18 17 40 23

saleiva:
Add a '!' after the bravo, and we got it!! ;)

ethervoid:
Please @Kartones check this PR :)

app/controllers/api/json/visualizations_controller.rb
```diff
@@ -371,5 +372,13 @@ def render_jsonp(obj, status = 200, options = {})
     super(obj, status, options.merge({request: request}))
   end
 
+  def send_like_email(vis, current_viewer, vis_preview_image)
+    if vis.type == Carto::Visualization::TYPE_CANONICAL
+      ::Resque.enqueue(::Resque::UserJobs::Mail::TableLiked, vis.id, current_viewer.id, vis_preview_image)
+    elsif vis.type == Carto::Visualization::TYPE_DERIVED
```
Kartones: that `elsif` instead of an `else` is on purpose, right?

app/mailers/user_mailer.rb
```diff
@@ -72,4 +72,17 @@ def map_liked(visualization, viewer_user, visualization_preview_image)
          :subject => @subject
   end
 
+  def table_liked(visualization, viewer_user, visualization_preview_image)
```
Kartones: I'd change the param name from `visualization` to `canonical_visualization` because it says "table liked" and then the first param is a vis and not a table, so to avoid confusion

vendor/assets/javascripts/cartodb.mod.odyssey.uncompressed.js
```diff
@@ -1,5 +1,5 @@
 // version: 3.15.1
-// sha: 31406160acdf8921ef9253e179b9ed2ce4ecd4a2
+// sha: a209f021e439cd0cca73ad212ff2867ae4fe4161
```
Kartones: Uhmm, this probably should have been commited ;)

Kartones:
:+1:

app/controllers/api/json/visualizations_controller.rb
ethervoid: Yes, I don't want to send email to other types

app/mailers/user_mailer.rb
ethervoid: Ok, I thought we treat everything as visualizations

vendor/assets/javascripts/cartodb.mod.odyssey.uncompressed.js
ethervoid: Ups, good catch

app/mailers/user_mailer.rb
Kartones: yes yes but is only for the shake of clarity, as both concepts appear together. I miss so much decent comments or params documentation and the like :D

PR #5100, opened 2015-08-19
Close dashboard dropdowns on opening modals

### LINKED ISSUE #5065: Creating a new map, the profile menu not closed. Create new map button is behind.
Screenshot attached

menu

viddo:
Fixes #5065

_(edit: changed approach since original PR, updated the desc to reflect this:)_

> The reason for the original 'click': 'killEvent' in this view is to prevent unselecting any selected item(s). The problem that comes with preventing that is that we, with maintaining current logic, would have to do explicit trigger for the closeDialogs event on each click callback in this view (or subviews).

> The root problem is that closeDialogs is used to unselect items too. I'd suggest to change that, to use another isolated event to control that behavior and not mix it with how dialogs are intended to behave.

So, changed to use another event to unselect items than mix it with the event used to control the open/closed behavior of dialogs.

ping @xavijam for review
cc @mariodelvalle for reporting

FYI @CartoDB/frontend

### DIFF
--- a/lib/assets/javascripts/cartodb/dashboard/filters_view.js
+++ b/lib/assets/javascripts/cartodb/dashboard/filters_view.js
@@ -38,7 +38,7 @@ module.exports = cdb.core.View.extend({
     'click .js-lock':           '_openChangeLockDialog',
     'click .js-privacy':        '_openPrivacyDialog',
     'click .js-link':           navigateThroughRouter,
-    'click':                    'killEvent'
+    'click':                    '_onClick'
   },
 
   initialize: function() {
@@ -393,6 +393,11 @@ module.exports = cdb.core.View.extend({
     } else {
       return false;
     }
+  },
+
+  _onClick: function(ev) {
+    this.killEvent(ev);
+    cdb.god.trigger('closeDialogs');
   },
 
   clean: function() {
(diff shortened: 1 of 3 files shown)


xavijam:
Sounds a good solution :+1:. And also I feel, if you only run ev.preventDefault(), it will make the trick.

viddo:
`ev.preventDefault()` instead of `killEvent` you mean? If yes, why?

xavijam:
Take into account that main views already have an event attached for this purpose:

https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb/dashboard/main_view.js#L158

Then, if you don't stop the propagation of the event this will "arrive" (luckily) to this parent view and will trigger that event. But as I said, your solution is perfect :+1:.

viddo:
Actually, trying that and further looking in the behavior here, the current code won't do :(

The reason for the original `'click': 'killEvent'` in this view is to prevent unselecting any selected item(s). The problem that comes with preventing that is that we, with maintaining current logic, would have to do explicit trigger for the `closeDialogs` event on each click callback in this view (or subviews).

The root problem is that `closeDialogs` is used to unselect items too. I'd suggest to change that, to use another isolated event to control that behavior and not mix it with how dialogs are intended to behave.

xavijam:
Create another event to listen with cdb.God?

viddo:
Yes, change [`cdb.god.bind('closeDialogs', this._unselectAll, this);` in dashboard/filters_view.js#L111](https://github.com/CartoDB/cartodb/blob/e2eafcfd091c51e5f852b120f7913699488fe2eb/lib/assets/javascripts/cartodb/dashboard/filters_view.js#L111)

xavijam:
Let's give it a try.

viddo:
ceb6607 does the trick (better method name in 7a44303)

viddo:
ceb6607 does the trick (better method name in 4d44ea1)

xavijam:
For me :+1:

Issue #4986, opened 2015-08-11
Export Dataset dialog?


csobier:
Two issues:
-  My original request was to wordsmith the Export dataset dialog, as shown below.
![image](https://cloud.githubusercontent.com/assets/13319925/9198836/de9edca4-400c-11e5-8f9f-9ccf11de660d.png)

However, this dialog has since changed???  Note that the documentation is now out of date and I must sync my images to match this window.  Please advise and be mindful before changing anything in the UI going forward.... all UI changes (image and content) should be tied to a milestone and include documentation *before* being released.

- Here is my suggested text revision. 
Please contact me to explain more details, I think I can enhance the understanding even better if you explain why they have to select the field in their query?'

CURRENT: "Export dataset
Click on the preferred format to export the dataset to.

To download any geospatial format like SHP, KML or GeoJSON don't forget to select the_geom on your query."

REVISED: "Export dataset
Select the preferred file format.

If exporting geospatial data with the SHP, KML or GeoJSON format, ensure to select the_geom on your query"

saleiva:
Again this has not changed from a week ago. Not sure what version are you refering to @csobier :(

Also, there is something weird in your proposal, 'geospacial' is wrong, right? I guess is just a mistake made of typing so fast :)

csobier:
Sorry, typo there.

Okay, am I crazy- see the  Export dataset dialog as it appears today.  See how the text is different than the snapshot above?

![image](https://cloud.githubusercontent.com/assets/13319925/9223784/2a8da8de-40cb-11e5-8f91-a9227a0a57dc.png)

saleiva:
That could be based on the query you have applied. Looks like you don't have any georef data, so it just let you export CSV. @javierarce your call here. If you think that text should be here, we leave it.

PR #5105, opened 2015-08-19
Updates export dataset copy

javierarce:
Original issue: #4986

Updated in the two export dialogs: 

Public page: 

screen shot 2015-08-19 at 13 05 31

CartoDB Editor:

screen shot 2015-08-19 at 13 06 47

Updated: the title of the captures was wrong, I changed the order


saleiva:
In the public page doesn't make sense, right? I mean, you cannot select it...

javierarce:
Well, in the particular case of my capture you can't, but in general you do and the current text is "Click on the preferred format to export the dataset to.".

saleiva:
Ok

Issue #5076, opened 2015-08-18
Consolidate duplicated models


viddo:
Right now we some models for the same imports endpoint:
- Import:
  - [`cdb.admin.Import`](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb/models/file.js) (will be located in `models/import.js` after the removal of old modals #5068)
  - [`common/background_polling/models/import_model.js`](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb/common/background_polling/models/import_model.js)
  - We should probably remove the first one, since that only has one usage (new merge modal).
- Geocoder
  - [`cdb.admin.Geocoding`](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb/models/geocodings.js)
  - [`common/background_polling/models/geocoding_model.js`](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb/common/background_polling/models/geocoding_model.js)
  - This would replace the `geocoder` param passed from [table/table.js](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb/table/table.js#L278-L279) to various child views (that need to know the `isGeocoding()` state).

Also on the subject of models, IMHO, we would be better of to keep all models that represent a concept/API endpoint in one place, e.g. `models/` (since that seems to be the commonplace). @CartoDB/frontend what do you think about moving them to that directory than have them spread out in various locations (e.g. move from `common/background_polling` to `models/`?

xavijam:
Fine by me :+1:

alonsogarciapablo:
There has been no activity on this issue for more several months. We are closing it. If you think this still needs to be addressed please open a new issue.

PR #5107, opened 2015-08-19
Possibility to duplicate a dataset from the dashboard view

### LINKED ISSUE #3423: Add Duplicate to the dashboard options when having a dataset selected
It should duplicate the dataset with the same name+_copy and appear in the list of tables.

### LINKED ISSUE #4353: Add duplicate as an option for datasets shared with you
- In the dashboard.
- In the dataset view

xavijam:
- [x] Duplicating one of your datasets (Fixes #3423).

screen shot 2015-08-19 at 13 44 43

--- 

- [x] Duplicating one of the datasets shared with you within your organization (Fixes #4353).

screen shot 2015-08-19 at 13 44 27

---

- Duplicating one of the datasets shared with you within your organization in the editor (option already available).


lib/assets/javascripts/cartodb/dashboard/filters_view.js
```diff
@@ -290,6 +292,47 @@ module.exports = cdb.core.View.extend({
...
+      cdb.god.trigger('importByUploadData', {
+        type: 'duplicate',
+        table_name: table.getUnqualifiedName() + '_copy',
+        value: tableName,
+        create_vis: false
+      });
+    }
+  },
+
+  _createDatasetFromSQL: function(e) {
```
xavijam: Import from query is not used for the moment, but ready when it is needed.

lib/assets/javascripts/cartodb/dashboard/filters_view.js
```diff
@@ -290,6 +292,47 @@ module.exports = cdb.core.View.extend({
...
+
+    var selectedDatasets = this._selectedItems();
+
+    if (selectedDatasets.length === 1) {
+      var m = selectedDatasets[0];
+      var table = m.tableMetadata();
+      var owner = table.permission && table.permission.owner.get('username');
+      var tableName = table.getUnqualifiedName();
+      var query = encodeURIComponent("SELECT * FROM " + tableName);
+      var sqlApiUrl = cdb.config.getSqlApiUrl().replace(this.user.get('username'), owner);
```
xavijam: Question: What would it be the best way to generate the SQL API url from other user? (not the current one).

xavijam:
Could you review this @viddo?
Also, I feel you can take advantage of this in your "cleaning" PR, because duplicate and sql export are available within new import model.

lib/assets/javascripts/cartodb/dashboard/filters_view.js
viddo: Why would you want to generate a SQL API URL for another user? Shouldn't they always be done from the POV of the current user?

xavijam: Example, we would like to duplicate a dataset you liked. The only way to do it is using the API SQL, and the url should be:
```
http://thatuser.cartodb.com/api/vX/sql?q=SELECT * FROM anytable
```

viddo:
The "cleaning" PR was for removing old modal code, so I don't consider this to be in that scope. 

Using this new code would be an "enhancement" (in terms of code maintenance at least, for the end-user there's no difference), similar to what is planned in #5076, consolidating models/logic to have one "proper" way of doing a particular action. Thus, if you think that's the case I'd suggest to create a separate issue instead.

xavijam:
Yep, you are right @viddo, that should be another issue. Let me create it (should I assign it to you?).

lib/assets/javascripts/cartodb/dashboard/filters_view.js
viddo: IMHO, if this code isn't used then I'd say remove it. Keep a gist for the future if you want to remember this particular code. That also postpones the discussion below until we actually need to solve that problem and know all the real requirements.

viddo:
Depends on fast you'd like this to get done, I have 20-isch issues pending for upcoming milestones. ;)

lib/assets/javascripts/cartodb/dashboard/filters_view.js
xavijam: :+1:

PR #5113, opened 2015-08-19
On search results synced data library datasets appear

### LINKED ISSUE #5110: On search results sync data library datasets appear as part of the data library
Additional details by @Kartones:
- [ ] Searching is multi-type, so the internal VQB joins only were applied if type was *exactly* remote. Need to make it work with multiple types at once
- [ ] Check tag filtering, I bet the same happens as long as searches with all types

Kartones:
Fixes #5110


Kartones:
@rafatower CR please

rafatower:
:+1:

PR #5116, opened 2015-08-19
Editor endpoints write performance analysis

### LINKED ISSUE #5085: Editor endpoints write performance analysis
For the initial phase, add StatsD logging for all editor endpoints that perform write operations in order to identify hot spots.

Task list:
- [x] Generalize `CartoDB::Importer2::ImporterStats` aggregator into `CartoDB::Stats::Aggregator` + `CartoDB::Stats::Importer` and adapt tests
- [x] Move `CartodbStats` inside `/lib/cartodb/stats/` (`UserTables` and `Authentication` aggregators)
- [x] Build `Cartodb::Stats::EditorAPIs` aggregator
- [x] Add the new stats aggregator to all editor create/update/delete endpoints
- [x] Add specific metrics for tiler calls from Ruby (possible hot spot)

Branch: `5085-new-apis-stats-and-cleanup`

Kartones:
Closes #5085


Kartones:
retest this please

Kartones:
@rafatower CR please (is just pending manual checks to confirm all is fine)

Kartones:
Wait a bit for the CR @rafatower , we've found two rogue StatsD calls plus one specific backend place were we want to also already gather stats so I'll add them

rafatower:
ok, just let me know :)

Kartones:
@rafatower Go ahead! 
I'm also going to take a look because some methods are a bit prone to errors by needing to return variables from the timing calls

lib/cartodb/stats/aggregator.rb
```diff
@@ -0,0 +1,100 @@
+require 'socket'
```
rafatower: is this file moved from somewhere? the code looks very familiar to me

Kartones: Yes, it is based on the old importer/lib/importer/importer_stats.rb class, with some minor changes but mostly only allowing to setup the prefix at initizliation time.

app/controllers/api/json/visualizations_controller.rb
```diff
@@ -24,165 +24,144 @@ class Api::Json::VisualizationsController < Api::ApplicationController
...
-      additional_fields = {
-        type:       params.fetch(:type, Visualization::Member::TYPE_DERIVED),
-        parent_id:  params.fetch(:parent_id, nil)
-      }
+        vis.privacy = vis.default_privacy(current_user)
 
-      vis = Visualization::Copier.new(
-        current_user, source, name_candidate
-      ).copy(copy_overlays, copy_layers, additional_fields)
+        # both null, make sure is the first children or automatically link to the tail of the list
```
rafatower: I would strictly separate adding stats from fixing/improving things... (but that's just me)

Kartones: This was just moving code from a +150 method to several smaller ones, no bugfixes or changes. Sorry if is hard to read but this comment was already there :)

I've only bugfixed stats-related code.

lib/cartodb/stats/authentication.rb
```diff
@@ -0,0 +1,43 @@
+require_relative 'aggregator'
+
+module CartoDB
+  module Stats
+
+    class Authentication < Aggregator
+
+      PREFIX = 'logins'
```
rafatower: We've already talked about this... I prefer `prefix` to be a method, part of the public API rather than a constant. But I'm not very opinionated on this.

lib/cartodb/stats/aggregator.rb
```diff
@@ -0,0 +1,100 @@
...
+      def decrement(key)
+        Statsd.decrement("#{fully_qualified_prefix}.#{key}")
+      end
+
+      protected
+
+      def self.read_config
+        config = Cartodb.config[:graphite]
+        config.nil? ? {} : config
+      rescue
```
rafatower: maybe it's a good time to capture specific exception (seems like it was .fetch and then KeyError and changed to array access?)

lib/cartodb/stats/aggregator.rb
```diff
@@ -0,0 +1,100 @@
...
+
+      def timing(key)
+        return_value = nil
+        @timing_stack.push(key)
+        Statsd.timing(timing_chain) do
+          begin
+            return_value = yield
+          rescue => e
+            @timing_stack.pop
+            raise e
```
rafatower: shall we fail because of stats? I'd just log the execption.

Kartones: this is done to bubble up the original error if something fails inside, careful, not if the stats themselves fail.

lib/cartodb/stats/aggregator.rb
Kartones: agreed, I'll setup the "missing config" one to just return empty hash, then if other happens log error (and yet return empty hash to not break execution)

lib/cartodb/stats/aggregator.rb
rafatower: mm true, sorry :)

rafatower:
Ok in general with the changes, testing at discretion.

Kartones:
retest this please

Kartones:
retest this please

Issue #4916, opened 2015-08-06
Review filter styles after new modals release


xavijam:
![screen shot 2015-08-06 at 16 16 27](https://cloud.githubusercontent.com/assets/132146/9113741/978b5b6c-3c56-11e5-9f04-a55d4dd648a8.png)

PR #5120, opened 2015-08-19
Fixes filter style

javierarce:
Original issue: #4916 

Original problem: 

screen shot 2015-08-19 at 18 07 00

Fixed: 
fix


javierarce:
Simple fix, @xavijam, have a look.

xavijam:
:clap:

Issue #5118, opened 2015-08-19
Text overflow in import URL input


MariaCheca:
Importing external datasets, too long urls exceed input limits:

![screen shot 2015-08-19 at 5 59 16 pm](https://cloud.githubusercontent.com/assets/2141690/9361514/0e749d2a-469c-11e5-8496-0f5c9721cf9d.png)

cc/ @xavijam

PR #5123, opened 2015-08-19
Fixes overflowing input field

javierarce:
Original issue: #5118 

Before the fix:
screen shot 2015-08-19 at 18 52 38

After the fix: 
screen shot 2015-08-19 at 18 57 00

### DIFF
--- a/app/assets/stylesheets/common/create/listing/import_selected_dataset.css.scss
+++ b/app/assets/stylesheets/common/create/listing/import_selected_dataset.css.scss
@@ -31,7 +31,10 @@ $br: 4px;
   color: #419CDB;
   text-transform: uppercase;
 }
-.DatasetSelected-itemInfo { margin-left: 20px }
+.DatasetSelected-itemInfo { 
+  width: 740px;
+  margin-left: 20px;
+}
 .DatasetSelected-itemTitle {
   display: block;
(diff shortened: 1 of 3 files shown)


javierarce:
@viddo: have a look, please.

javierarce:
Have a look to this PR, @viddo.

viddo:
The icon looks wrong, right? Also not very much in favor of these fixed widths if we can avoid them, would it be possible to let the container handle the overflow using CSS attrs? E.g.

```
  text-overflow: ellipsis;
  white-space: nowrap;
  overflow: hidden;
```

javierarce:
Sorry, I uploaded the wrong capture.

The h6 element that contains the URL has already the properties you list (.u-ellipsLongText). If I move it to the parent element we'd get this: 

screen shot 2015-08-19 at 19 00 58

If I apply the .u-ellipsLongText to both elements I get this instead… I find it a bit ugly, but maybe is better than specifying the width manually.

screen shot 2015-08-19 at 19 02 34

viddo:
Yeah, agree… if the box has a specific size then I think it's fine, but otherwise I'd let the container define the available size and use width: 100% to fill up the container (so the overflow applies correctly), but perhaps not doable in this case. If that's the case then go ahead. :+1:

Issue #5130, opened 2015-08-20
Fix overflow when you try to delete a dataset


piensaenpixel:
![image](https://cloud.githubusercontent.com/assets/1214201/9378330/e2fbfe74-4720-11e5-84bc-26aeab83d5b4.png)

PR #5133, opened 2015-08-20
Fixes text overflow in the delete dialog

javierarce:
Before:
screen shot 2015-08-20 at 10 13 16

After:
screen shot 2015-08-20 at 10 13 05

Original issue: #5130


javierarce:
Have a look, @xavijam.

xavijam:
Could you check that this is not happening in public map cards please?

javierarce:
Sure, but if it's broken there I'll fix that in another ticket.

xavijam:
:+1: :es: :coffee:

javierarce:
Public maps are OK. I'll review cards in the Editor.

saleiva:
:heart:

PR #5142, opened 2015-08-20
Increases max zoom on plain color basemaps to 32

### LINKED ISSUE #4940: when user select plain color basemap the embed does not go further than zoom level 20
- create a map and add a point
- zoom to level 24
- go to embed map

it shows level 20. If you force with zoom=24 if shows level 20 and it should be unlimited (or something like 32)

gfiorav:
Fixes #4940

@javisantana 

CR please, @xavijam


xavijam:
You can remove the comment :)

xavijam:
And then :dancer:

gfiorav:
Done, as soon as tests pass, I'll merge :+1: 

Thanks @xavijam

javisantana:
I don't think this fixes the whole problem, did you try to go to and embed and set in the url ?zoom=29 ?

gfiorav:
Could you provide more information on how you tested this?

It works on embedded for me

javisantana:
send me the embed please

Issue #4899, opened 2015-08-06
Attach the floating elements of the export image dialog to the selected area


javierarce:
Original ticket: https://github.com/CartoDB/cartodb/issues/4685

javierarce:
When the export window is very small, I've decided to move the close button to one side instead of keeping it in the same line. Not sure if it's the right solution. What do you think, @saleiva?

![anim2](https://cloud.githubusercontent.com/assets/4933/9383411/2701c720-4749-11e5-838d-1cb3f87680f9.gif)

saleiva:
Loooove, looooooove, loooove

PR #5154, opened 2015-08-20
Makes the header and footer floating around the selection

javierarce:
anim2

Original issue: #4899

### DIFF
--- a/app/assets/stylesheets/table/export_image_view.css.scss
+++ b/app/assets/stylesheets/table/export_image_view.css.scss
@@ -12,9 +12,9 @@
 .ExportImageView-Header {
   position: absolute;
   width: 100%;
-  text-align: center;
   top: -35px;
   height: 40px;
+  text-align: center;
   z-index: 105;
 }
 .ExportImageView-Header--inner {
@@ -44,9 +44,9 @@
 }
 .ExportImageView-Footer {
   position: absolute;
+  width: 100%;
   bottom: -20px;
   text-align: center;
-  width: 100%;
   z-index: 105;
 }
 .ExportImageView-Footer--inner {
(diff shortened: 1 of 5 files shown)


javierarce:
Review this, please please, @xavijam.

app/assets/stylesheets/table/export_image_view.css.scss
```diff
@@ -11,12 +11,18 @@
...
+  top: -35px;
   height: 40px;
-  top: 30px;
   z-index: 105;
 }
+.ExportImageView-Header--inner {
+  position:absolute;
+  width: 400px;
+  left: 50%;
+  margin-left: -200px;
```
xavijam: I am not a big friend of negative margins.

lib/assets/javascripts/cartodb/table/export_image_view.js
```diff
@@ -166,6 +163,12 @@ cdb.admin.ExportImageView = cdb.core.View.extend({
 
     this.$(".CanvasExport").css({ height: height, width: width });
 
+    if (width < this._MIN_WIDTH) {
```
xavijam: You can use jQuery ```.toggleClass``` function for the same behaviour.

app/assets/stylesheets/table/export_image_view.css.scss
javierarce: I need to center the element and make its size independent of the parent width. I think this is the simplest solution.

xavijam: Ah, right, i didn't pay attention to the whole thing :+1:

javierarce: Thanks!

PR #5157, opened 2015-08-20
Case insensitive tag search

### LINKED ISSUE #5108: Tag search should be case insensitive
Right now, if you search for a specific tag, ```:Cultural datasets``` and ```:cultural datasets``` don't yield the same results (some vs none).

gfiorav:
Fixes #5108


gfiorav:
Calling @pramsey to help out with this one since I hear you know a lot about PostgreSQL :)

So, I've gotten to this code:
```ruby
if @tags
      @tags.each do |t|
        t.downcase!
      end
      query = query.where("ARRAY[?]::text[] && visualizations.tags", @tags)
    end
```

In `cartodb/app/queries/carto/visualization_query_builder.rb:240`

So, `@tags` are lowercase. Now I need a way for the query to be case insensitive. I've looked around and found things like [this](http://stackoverflow.com/questions/16311289/postgresql-case-insensitive-select-on-array), but I'm worried about performance. 

Could you please suggest some efficient way of doing this?

Thanks!

pramsey:
The efficiency comes from having the column be indexed so the && operator can do magic against the index. So, the question is, can you get some way to do case-insensitive indexed calls. The answer is yes, but it will get a little bit ugly. You need a function that will upper (or lower) case a text[]. Something like
```
FUNCTION array_to_upper(text[]) RETURNS text[] 
```
Then you can create a functional index on array_to_upper, 
```
CREATE INDEX idx_test on mytable USING GIN (array_to_upper(mytags));
```
And do your query in such a way as to force use of that functional index,
```
SELECT * FROM mytable
WHERE array_to_upper(mytags) && array_to_upper(ARRAY['sometag'])
```

gfiorav:
@pramsey

Although not to orthodox, I found this *hack*. What do you think?

```ruby
if @tags
      @tags.each do |t|
        t.downcase!
      end
      query = query.where("array_to_string(visualizations.tags, ',') similar to array_to_string(ARRAY[?]::text[], ',')", @tags)
    end
```

gfiorav:
Retest this, please

pramsey:
This assumes that @tags only has one element. If it has more than one element then you're just crossing your fingers and hoping that the order of both arrays is the same. Also, it can't make use of any index, so your search better not be counting on good speed from the tag filter (maybe it isn't?). Finally, you're assuming the cases in the database column have already been coerced to lower, and I kind of assumed that was one thing you couldn't do (because then the casing the user entered would disappear, which would suck for them if they considered it significant (using HungarianNotation for example).

gfiorav:
Hi @pramsey! 

About the indexing: totally, still might not be worth it for this feature.

About single tag cases: you're absolutely right. To solve it, I'm working on something like this:

```ruby
if @tags
  @tags.each do |t|
    t.downcase!
  end
  query = query.where("array_to_string(visualizations.tags, ', ') ILIKE '%' || array_to_string(ARRAY[?]::text[], ', ') || '%'", @tags)
end
```

About the part were this code is forcing users to lowercase tags: how? This query is using `ILIKE` to compare against stored tags so it doesn't matter if they're upper or lower case.

pramsey:
`ILIKE` will be fine, I was commenting on `SIMILAR TO` which doesn't to coercion by default. I'm still not sure how you can match `ARRAY['a','b','c']` to `ARRAY['c','a']` though.

gfiorav:
Checkout the latest commit. Code finally is:
```ruby
if @tags
  @tags.each do |t|
    t.downcase!
  end
  query = query.where("array_to_string(visualizations.tags, ', ') ILIKE '%' || array_to_string(ARRAY[?]::text[], ', ') || '%'", @tags)
end
```

It works for matching with multiple tags from my testing.

pramsey:
It still seems dependent on ordering, unless there's a sort somewhere higher up in the code base forcing everything to be pushed into the same order. Even then, I would expect `ARRAY['a','b','c']` versus `ARRAY['c','a']` to fail due to the gap between `c` and `a`.

gfiorav:
If what you mean is that a user will search for multiple tags, this is not supported at the moment. 

With this fix, a user searching `:fOo` will get datasets such as:
Dataset 1 with tags `:foobar`
Dataset 2 with tags `:FoZ`, `:FOOBAR`
Dataset 3 with tags `:FoobaR`, `:FeZ`
Dataset 3 with tags `:FoR`, `:barfOO`

This was the original goal. Previously the same query would have not returned anything. After all it *is* an ILIKE %term% over a string. My testing has gone OK for any case I've thought of.

pramsey:
OK then, looks all right. There will be no index assist, but presumably the data set is small enough it's not a big deal.

gfiorav:
Yes, were losing index assistance but we should revamp tags in the future anyway :smile:

app/queries/carto/visualization_query_builder.rb
```diff
@@ -238,7 +238,10 @@ def build
     end
 
     if @tags
-      query = query.where("ARRAY[?]::text[] && visualizations.tags", @tags)
+      @tags.each do |t|
+        t.downcase!
+      end
+      query = query.where("array_to_string(visualizations.tags, ', ') ILIKE '%' || array_to_string(ARRAY[?]::text[], ', ') || '%'", @tags)
```
rafatower: Since you're downcasing tags above, the ILIKE vs LIKE operator change not make sense here, am I missing anything?

gfiorav: Tags in `visualizations.tags` are not necessarily in lowercase!

rafatower: > If what you mean is that a user will search for multiple tags, this is not supported at the moment.

this code is pretty misleading if searching by multiple tags is not supported.

gfiorav: CartoDB is *prepared* for multiple tags in the sense that @tags is an array but when parsing the search terms from the UI, it's not possible to specify more than one tag. Please see code:

```javascript
_submitSearch: function(e) {
    if (e) this.killEvent(e);
    var val = Utils.stripHTML(this.$('.js-search-input').val().trim(),'');
    var tag = val.search(':') === 0 ? val.replace(':', '') : '';
    var q = val.search(':') !== 0 ? val : '';

    this.routerModel.set({
      page: 1,
      tag: tag,
      q: q,
      shared: 'yes'
    });

    this.createModel.set('listing', 'datasets');
  }
```

https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb/common/dialogs/create/listing/navigation_view.js#L191

As @pramsey correctly pointed out, this would break in case of multiple tags since the arrays are passed onto strings so "a, b, c" != "b, a, c". 

Previous code used the `&&` inclusion operator so it didn't have this problems but, then again, it had other problems and users weren't able to input more than one tag anyway.

As soon as we decide to revamp tags, chances are that we won't be parsing arrays into string and have a model/table for tags but, until then, it's nice that users can actually use the search engine to explore tags (which is the goal of being able to searching tags).

PR #5170, opened 2015-08-21
Adds image_proxy_controller show route to routes.rb

### LINKED ISSUE #5155: Make "Export Image" export image elements too
Example uses [this map](http://cdb.io/1TZEwfZ).

Map is:
screen shot 2015-08-20 at 16 56 47

Selected "Export Image" like this:
screen shot 2015-08-20 at 16 57 09

Obtained result:
download

gfiorav:
Fixes #5155


gfiorav:
Please @juanignaciosl, could you CR? Thanks! :smiley:

juanignaciosl:
LGTM

gfiorav:
@javierarce , could you please CR the front end stuff? Thanks! :)

lib/assets/javascripts/cartodb/common/dialogs/static_image/export_image_result_view.jst.ejs
```diff
@@ -5,7 +5,7 @@
     </div>
     <p class="Dialog-headerTitle">Your image has been generated correctly</p>
     <% if (response.type === 'url') {%>
-    <p class="Dialog-headerText">It's now available in: <a href="<%- response.content %>" target="_blank" class="ExportImageResult--url"><%- response.content %></a></p>
+    <p class="Dialog-headerText">It's now available in: <a href="<%- response.content %>" target="_blank" class="ExportImageResult--url"><%- response.content.substring(0, 255) %></a></p>
```
javierarce: It's great, but I'd prefer if you truncated the string outside of the template. So we'd have `url` and `title` (for example) and title would be truncated here: https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb/common/dialogs/static_image/export_image_result_view.js#L63)

gfiorav:
@javierarce once you give me a final :+1: I'll merge

javierarce:
:ok_hand: :balloon:

lib/assets/javascripts/cartodb/common/dialogs/static_image/export_image_result_view.js
```diff
@@ -47,7 +47,7 @@ module.exports = BaseDialog.extend({
 
   generateImage: function(url) {
     var callback = function(url) {
-      this._showResult({ content: url, type: 'url' });
+      this._showResult({ content: url, type: 'url', displayedLink: url.substring(0, 255)});
```
javisantana: nice change but why don't write something useful here instead a base64 fragment?

javierarce: I was against this change when the bug appeared (since we were offering a
base64 download without a name), but because of this solution
https://github.com/CartoDB/cartodb/issues/5172#issuecomment-133374022 now
it makes sense update the name
https://github.com/CartoDB/cartodb/pull/5174/files

-- 
Javier Arce

PR #5171, opened 2015-08-21
Fix focused JS tests fail due to implicit npm dependency

### LINKED ISSUE #5147: Focused JS tests fail due to implicit npm dependency
How to reproduce: write a test case in any file that is required as a CommonJS module:

```js
fit('should fail', function() {
  expect(true).toBe(false)
});
```

This will provoke an error in the console when running the tests:

```bash
$ grunt jasmine
'/Users/jmedina/Workspace/cartodb/cartodb/.grunt/data:application/json;charset:utf-8;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpb…
```

From a first look, looks like the same issue that happened to at least [another project](https://github.com/freedomjs/freedom/issues/276). Will try to update dependencies and see if that solves it.

cc @xavijam

viddo:
Fixes #5147 

Bumping up to latest versions for outdated packages makes things works :tada: (this was needed sooner or later anyways)
…except for gulp-* modules, because at least one native binding (gulp-ttf2woff2 doesn't seem to be used in currently specified versions) requires a newer version of node (v0.12). I don't want to force an upgrade to break builds or other dependencies that requires this older versions. We can investigate and bump those in isolation.

~~There's a little caveat of bumping the versions; I documented it inline with the [code change](https://github.com/CartoDB/cartodb/commit/9b9035ad34f27817072d3ff3c9cc1176d671ea73#diff-418613da9f808dbeb01adc6244f65757)~~ edit: v3.4.2 of buffer module was released, so no temporary workaround needed anymore.

Also set all dependencies to use specific versions, to avoid dependency versions mismatches (per recommendation by @rochoa).

Verifying that things work (on my machine™):
- Testing the typical builds (both local and production)
- Running all tests locally (both headless and in browser)
- Provoke a failing test un purpose (see screenshot below)
- Browsing locally
- Review [build output](http://ci.cartodb.com/job/CartoDB-PR-testing/287/consoleFull)

screen shot 2015-08-21 at 12 26 45
_forced failed test, with proper stack trace_

@xavijam since you suffered the focused-tests issues originally
@CartoDB/frontend FYI


viddo:
Bump @xavijam can you verify that this fixes your trouble?

xavijam:
Yay! confirmed! :es:

PR #5173, opened 2015-08-21
Typo in punctuation

iriberri:
I just noticed:

image

cc @viddo


viddo:
Oh my, good catch :ok_hand:

PR #5174, opened 2015-08-21
Adds map's name as default filename for download

### LINKED ISSUE #5172: "Export Image" downloaded file has generic name
It would be nice for the downloaded image to have a sensible name like "#{table_name} - #{current_time}.png".

gfiorav:
Fixes #5172

The filename for the “Export Image” image file is now the exported map’s name.

CR @javierarce 

cc @javisantana


gfiorav:
I'm assigning this to you @javierarce since I'm out of time :smile:

The state I leave it in is: Tests passing so CR needed

Thanks!

javierarce:
Thanks, @gfiorav!

PR #5176, opened 2015-08-21
Fix background polling element in table view blocks scrolling after appearing

### LINKED ISSUE #5175: Background polling element in table view blocks scrolling after appearing
**Steps**:

* Import a table long enough so that scrolling is possible. [Sample link here](https://gist.githubusercontent.com/iriberri/61e5f94f7d39e3f15a33/raw/4536ee28c772209f5a5843a53906d3ddcb5bb802/example.csv)
* Georeference by city names, for example. Not specific options are needed.
* Once the process finishes (or meanwhile), the table is no longer scrollable. Closing the small window with the geocoding result doesn't enable scrolling neither. Open and closing the export modal window seems to help.

It's required to refresh the page to scroll the table view each time one of these polling windows appear. I think it happens also with the Pecan message if I'm not wrong.

cc @xavijam

viddo:
Fixes #5175

ping @xavijam @iriberri

### DIFF
--- a/lib/assets/javascripts/cartodb/common/dialogs/georeference/georeference_view.js
+++ b/lib/assets/javascripts/cartodb/common/dialogs/georeference/georeference_view.js
@@ -110,7 +110,7 @@ module.exports = BaseDialog.extend({
 
   _onChangeGeocodeData: function(tabModel, data) {
     cdb.god.trigger('geocodingChosen', data);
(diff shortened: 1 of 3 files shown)


xavijam:
:+1:

iriberri:
:revolving_hearts:

PR #5188, opened 2015-08-24
Go to "dataset view" from editor (being in a visualization), within the layer panel element

### LINKED ISSUE #4078: It would be extremely nice to be able to go to a dataset view from a map view
I find myself a lot of times wanting to click in a layer name in order to go to the dataset view.
That would be very useful

xavijam:
screen shot 2015-08-21 at 18 33 12

Fixes #4078.
REVIEWER: @javierarce 
Happy birthday @lbosque.


lib/assets/javascripts/cartodb/table/layer_panel_view.js
```diff
@@ -272,16 +279,18 @@
...
 
         // table name
-        this.$('.layer-info a.info .name')
+        this.$('.layer-info .info .name')
           .text(layer.get('table_name_alias') || layer.get('table_name').replace(/_/g,' '));
 
         // table name alias
-        this.$('.layer-info a.info .name_info')
-          .text('view of ' + layer.get('table_name'));
+        var layerUrl = cdb.config.prefixUrl() + '/tables/' + layer.get('table_name');
```
xavijam: I'm in doubt with this url generation. We don't have this kind of generator from a layer model using the new URLs stuff, but maybe it is worth it to do it, or not. What's your opinion @viddo @javisantana?

javisantana: why not layer.table.url() ?

xavijam: marvelous

javierarce:
:+1: LGTM. 

It bugs me a little to keep using `. name_info` and not the new preferred syntax (`.js-name` or whatever) to bind the event, but I understand we are modifying old code.

xavijam:
Yeah, that's the reason, we have to review all these old styles and ways to get/set things with js, so yes, definitely we will do :).

PR #5194, opened 2015-08-24
Fix geocoding creation

viddo:
Was accidentally removed in the removal of old modals code (see 00e36fc)

cc @iriberri @xavijam


xavijam:
:+1:

PR #5195, opened 2015-08-24
Enable edit metadata dialog when a SQL query is applied

### LINKED ISSUE #4810: Can't edit dataset metadata when custom SQL query is applied
Steps to reproduce:

1. Open a dataset on which you have admin rights
1. Link under table name should read `Edit metadata`
1. Clicking `Edit metadata` should work as expected
1. Apply a custom SQL query / view
1. Link under table name should read `View metadata`
1. Clicking that link will open a modal with greyed out fields and a message that says `You can't edit the dataset attributes.`

alonsogarciapablo:
Fixes #4810.

This PR allows users to edit the metadata of a dataset when a SQL query has been applied. I have tried to apply a query like `SELECT 1` and editing the metadata seems to work fine. I wasn't here when this constraint was introduced so I'm not sure if there's any other situation when this might not work.

@xavijam what do you think? thanks!


xavijam:
If you haven't found any problem, go ahead! :uk:

PR #5215, opened 2015-08-25
5208 cartodbfy metrics

rafatower:
Related to #5215 


app/models/data_import.rb
```diff
@@ -341,6 +341,7 @@ def self.http_timeout_for(user, assumed_kb_sec = 75*1024)
     (user.quota_in_bytes / assumed_kb_sec).round
   end
```
Kartones: uhm, just in case I'd add here a require of `lib/cartodb/stats/importer`

app/models/data_import.rb
```diff
@@ -729,24 +730,37 @@ def notify(results)
...
+                  'service_name'           => self.service_name,
+                  'data_type'              => self.data_type,
+                  'is_sync_import'         => !self.synchronization_id.nil?,
+                  'import_time'            => import_time,
+                  'file_stats'             => ::JSON.parse(self.stats),
+                  'resque_ppid'            => self.resque_ppid,
+                  'user_timeout'           => ::DataImport.http_timeout_for(current_user),
+                  'error_source'           => get_error_source,
+                  'id'                     => self.id,
+                  'total_size'             => total_size,
```
Kartones: Nice additions!

Kartones:
:+1:

app/models/data_import.rb
rafatower: as far as I tested it is loaded automatically. What's the rule there? (I personally prefer explicit requires, you know it)

Kartones: lib/xxx is tricky as doesn't works properly with autoload based on my experience (and when it does is because a related class required it before). So everytime I add something there, I do explicit requires...

rafatower: ok, adding it...

rafatower:
@ethervoid can you please review? kartones is not available right now for it.

The main change since the last review: I added a helper class pseudo-singleton (unique per process + data_import.id) called `CartodbfyTime`.

Otherwise it is really a mess because there are several DataImport instances referring to the same row in DB and also it is not easy to pass through the data_import instance to the table, where I need to get the metrics (before_ and after_ hooks in models, DataImport and Table quite coupled).

db/migrate/20150825121806_add_cartodbfy_time_to_data_imports.rb
```diff
@@ -0,0 +1,14 @@
+Sequel.migration do
+  up do
+    alter_table :data_imports do
```
ethervoid: Take care, possibly you are going to do it in two times

ethervoid:
Seems ok to me

db/migrate/20150825121806_add_cartodbfy_time_to_data_imports.rb
rafatower: why? the code won't be using that until the migration is done.

rafatower: @ethervoid I checked with Luisico. Releasing migration before code is a "good practice": decouple the release from the migration, easier rollback, avoid contention because of changes in big tables... but you're guaranteed the migrations will be applied before the new code is in place.

ethervoid: Good to know! thx @rafatower

Kartones:
:+1:

Kartones:
also, would be nice to add a test @rafatower (imports controller one at least, to avoid messing up too much with internals), to see that the value works, just with some expectations on the new singleton should be easy no?

PR #5229, opened 2015-08-25
Check for table service to not be nil

ethervoid:
It seems that sometime `table_service` could be nil so we have to check for it


ethervoid:
@rafatower could check this PR please

rafatower:
:+1:

Issue #5208, opened 2015-08-24
Add stats to measure performance of cartodbfy


rafatower:
What I think we need:
- [x] `import_time` (already present)
- [x] `total_size` (already present but within `file_stats` in json sent to metrics)
- [x] `cartodbfy_time`: to be added
- [x] `import_throughput = (total_size / import_time)` (derived, not present)
- [x] `cartodbfy_throughtput = (total_size / cartodbfy_time)` (derived)
- [x] `cartodbfy_import_ratio = (cartodbfy_time / import_time)`

After discussion with @rochoa and for simplicity, I'd add the new metrics, including the derived ones, to the json payload here:

https://github.com/CartoDB/cartodb/blob/master/app/models/data_import.rb#L736-L761

cc/ @luisbosque do you see any problem with that?

luisbosque:
No problem at all. The more stats, the better :dancer:

rafatower:
already there... preparing the dashboards. :smile:

PR #5248, opened 2015-08-26
Use counter for importer.total_size #5208

rafatower:
Use a counter instead of a gauge for total_size metrics. Gauges are only
appropriate if sending periodic data. Otherwise counters are better to
aggregate size information, especially if there are not many events.

See http://statsd.readthedocs.org/en/latest/types.html


rafatower:
Reviewed on my desk by @Kartones. I'm going ahead with this...

Issue #5222, opened 2015-08-25
Make a list of datasets and maps available in the user profile feed


javierarce:
For the page created in https://github.com/CartoDB/cartodb/issues/5220 we need an object that contains the maps and datasets of a given user (see `User Profile Feed` page in the design file) ordered by updated_at date (newest on top). 

Since we are going to have an infinite scroll pagination, we'll also need an endpoint to ask for a page.

![screen shot 2015-08-25 at 16 51 40](https://cloud.githubusercontent.com/assets/4933/9469844/76928bea-4b49-11e5-857c-9fc505f6c57e.png)

ethervoid:
You have to query using the visualizations API:

```http://user_to_query.cartodb.com/api/v1/viz?types=table,derived&page=1&per_page=16&privacy=public&order=updated_at```

javierarce:
I've made a couple of tests and the `privacy` param doesn't seem to work. 

I requested a list of tables and maps of my own account and set the privacy param to `public` and the endpoint returned me `linked` datasets. If I use `link` I get `password` tables as well.

ethervoid:
Are you using an api_key? If you don't use an api_key we could remove the `privacy` param

javierarce:
No, I didn't use the query but I still got private items.

PR #5256, opened 2015-08-26
5220 new page for user profile

### LINKED ISSUE #5220: Create new page for the user profile feed
We need a new page to place the user feed accessible under the following URLs:

```
https://team.cartodb.com/u/username 
https://username.cartodb.com
```

If the visitor is logged in, the previous URLs should have the usual behaviour (redirecting to the dashboard.)

ethervoid:
*  [x] New endpoint for the new user feed screen
*  [x] Refactor of some stuff for maps and datasets
*  [x] Added privacy flag to visualizations searcher
*  [x] Added new pages and endpoints `search` and `explore`

Closes #5220, #5222, #5285, #5283

// @javierarce


javierarce:
The frontend part looks good, but I think someone from backend should review the rest. 

Also, we should use something different for the feed URL. `/user_feed` is not pretty… maybe `/profile`, since that page will contain the user bio, twitter link, etc.? What do you think, @carlostallon?

A note about how the new path is going to work for clarity & future reference:

* If the visitors go to username.cartodb.com they'll see the profile page.
* If logged in users go to the same page, they'll be redirected to their private dashboard.
* If logged in users wants to see their public profile, they'll need to open `/profile` (or whatever URL we pick).

javierarce:
Very important: do not merge this branch until we are ready with all the changes.

javierarce:
bump @carlostallon

carlostallon:
+1 to .../profile/...

ethervoid:
We can't use `/profile` because this endpoint leads you to the user profile page

captura de pantalla 2015-08-28 10 02 28

javierarce:
Oh, bummer! Let's think about other solutions then. What about /me?

carlostallon:
what about this schema?

User public profile (feed) - cartodb.com/u/carlostallon
User public maps - cartodb.com/u/carlostallon/maps
User public datasets - cartodb.com/u/carlostallon/datasets

User dashboard - cartodb.com/u/carlostallon/dashboard

Profile user configuration -
cartodb.com/u/carlostallon/configuration/profile
Account User configuration -
cartodb.com/u/carlostallon/configuration/account
...




2015-08-28 10:20 GMT+02:00 Javier Arce <notifications@github.com>:

> Oh, bummer! Let's think about other solutions then. What about /me?
>
> —
> <https://github.com/CartoDB/cartodb/pull/5256#issuecomment-135669828>.
>



-- 
Carlos Tallón
Designer & Baker

javierarce:
Doing what you say would require to get rid of the current redirecting behaviour:

Right now if I'm the user `username` and visit `username.cartodb.com` I will be redirected to `username.cartodb.com/dashboard`.

ethervoid:
I don't dislike the `/me` idea.

carlostallon:
So, go with /me

My only concern with it is that any change for this url in the future would be impossible

javierarce:
We could remove the redirection to `/dashboard` for logged in users and always point to the user profile, but this is an important change so we should think about it.

/cc @piensaenpixel @javisantana @saleiva

saleiva:
Removing the redirection to your own dashboard is something that I don't like too much.
Maybe I'm lost here, but why do we need an extra URL? cannot we leave

cartodb.com/u/username -> feed
cartodb.com/u/username/maps or datasets -> maps or datasets page
cartodb.com/u/username/dashboard -> dashboard

?

ethervoid:
The only missing use case is when the user try to see his/her own user feed, we need an endpoint for it.

saleiva:
Ah, understood. Then I'm fine with /me

javisantana:
I think @Kartones should review this, he knows more about the redirections and so on, specially since we have organizations and subdomainless url involved

ethervoid:
Yup, I've been waiting to end this part to call him to review the backend part because some more parts could be added

Kartones:
so shall I review it now or wait for the bat-signal?

ethervoid:
@Kartones take a look to this CR when you can :)

app/controllers/admin/pages_controller.rb
```diff
@@ -382,4 +384,43 @@ def ensure_organization_correct
     end
   end
 
+  def process_dataset_render(dataset)
```
Kartones: All this code should go in a separate `PublicVoewDatasetPresenter` or something similar no?

app/controllers/admin/pages_controller.rb
```diff
@@ -13,13 +13,24 @@ class Admin::PagesController < ApplicationController
   USER_TAGS_LIMIT = 100
   PAGE_NUMBER_PLACEHOLDER = 'PAGENUMBERPLACEHOLDER'
 
-  ssl_required :common_data, :public, :datasets
+  # TODO logic as done client-side, how and where to encapsulate this better?
```
Kartones: Then client-side should read it from somewhere sent by the server, maybe setup as template variables or some similar mechanism to send data to JS...

app/controllers/admin/pages_controller.rb
```diff
@@ -82,12 +93,37 @@ def datasets
...
+    if current_user
+      index
+    else
+      user_feed
+    end
+  end
+
+  def user_feed
+    username = CartoDB.extract_subdomain(request).strip.downcase
+    @viewed_user = User.where(username: username).first
```
Kartones: I know it was already here, but with more methods becomes clear that at least setting `@viewed_user` should be a before_filter... could you setup? or at least file a ticket to do so?

app/controllers/admin/pages_controller.rb
ethervoid: I didn't want to change all at same time. This was a first refactor

app/controllers/admin/pages_controller.rb
Kartones: nah, I see is already existing so ignore for now

ethervoid: That was a TODO make in February by @viddo and I don't know what is refering to :S

app/controllers/admin/pages_controller.rb
ethervoid: yeah sure

app/controllers/admin/pages_controller.rb
```diff
@@ -82,12 +93,37 @@ def datasets
...
 
+  def public
+    if current_user
+      index
+    else
+      user_feed
+    end
+  end
+
+  def user_feed
```
Kartones: I'd add a comment here that actual template content will be requested from FE code to another endpoint, as I was searching for it :stuck_out_tongue_closed_eyes:

app/controllers/admin/pages_controller.rb
viddo: We have duplicated mapping geometry type <-> icon-name, the other place is: [lib/assets/javascripts/cartodb/models/table.js#L250-L257](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb/models/table.js#L250-L257) 
I obviously never got around to fix it, but the intention was to extract the mapping from this controller so it could also be provided to client-side through some suitable mean, e.g. `table_model.geometry_mappings`, and pass the same in the table PORO for the client-side.

config/routes.rb
```diff
@@ -23,7 +23,12 @@
   match '(/user/:user_domain)(/u/:user_domain)/sessions/create' => 'sessions#create',  as: :create_session
 
   match '(/user/:user_domain)(/u/:user_domain)/status'          => 'home#app_status'
-  match '(/user/:user_domain)(/u/:user_domain)/diagnosis'          => 'home#app_diagnosis'
+  match '(/user/:user_domain)(/u/:user_domain)/diagnosis'       => 'home#app_diagnosis'
+
+  # Explore
+  get   '/explore'         => 'explore#index',     as: :explore_index
+  get   '/search'          => 'explore#search',    as: :explore_search
+  get   '/search/:q'       => 'explore#search',    as: :explore_search
```
Kartones: Careful, naming has to be unique, rename to something like `as: explore_search_query`

Kartones: Which makes a question popup: Where are the tests for this new endpoints? :trollface: 

Same as below with the user feed, some dumb test would be really nice

Kartones:
Apart from the route repeated name, my only worry is that I see new routes and changed ones and just one test, so I'd implement at least dumb ones to see the flow is working as ok with the index stuff.

ethervoid:
Ok, I'll add dumb tests to check the routes :)

config/routes.rb
ethervoid: The are tests for user_feed (the same that the old endpoint) https://github.com/CartoDB/cartodb/pull/5256/files#diff-85a91d92b978eb1ff2e07f5335d068feR94 (I have to change the it part). I'll add test for explore endpoint

Kartones: Ok!

ethervoid:
This PR is merged in the branch `5296-explore-user-feed`

PR #5258, opened 2015-08-26
Changed on-boarding dialog when user doesn't have any map created

### LINKED ISSUE #4893: Design CTA over empty map on emtpy accounts
So we invite people to create their first map in a very easy way.

xavijam:
Now onboarding is not a dialog anymore, it will be placed over the map and green button will open create dialog.

screen shot 2015-08-26 at 18 41 56

Fixes #4893.

REVIEWER: @josecruz and @javierarce


xavijam:
cc @csobier

xavijam:
Pending: modify gif (@josecruz).

PR #5261, opened 2015-08-27
Fix API Call and Download buttons not working on public datasets

### LINKED ISSUE #5260: API Call and Download buttons not working on public datasets
When I click on the API Call or Download buttons on [this public data page](https://team.cartodb.com/u/eschbacher/tables/nyc_bikelanes/public), I only get the # added to the URL with no action happening (no modals pop up, no data downloaded, etc.)

I get the same behavior on [other public data pages](https://team.cartodb.com/u/andrew/tables/tesla_locations/public). I've tried Chrome and Safari -- same thing. @auremoser experienced the same problems.

viddo:
Fixes #5260
This event definition was causing a JS error that prevented the view from even work.

ping @javierarce

### DIFF
--- a/lib/assets/javascripts/cartodb/public_table/table_public.js
+++ b/lib/assets/javascripts/cartodb/public_table/table_public.js
@@ -7,9 +7,8 @@
   cdb.open.TablePublic = cdb.core.View.extend({
 
     events: {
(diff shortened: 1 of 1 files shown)


javierarce:
:+1:

PR #5266, opened 2015-08-27
Upgrade cdb extension to 0.9.3 #4962

rafatower:
@Kartones please review

PR #5267, opened 2015-08-27
Upgrade cdb extension to 0.9.4 #4962

rafatower:
Finally updating to 0.9.4 instead of 0.9.3. OK got in this PR: https://github.com/CartoDB/cartodb/pull/5266

The only difference is this fix: https://github.com/CartoDB/cartodb-postgresql/issues/123 that shall also fix the table tests that failed in the previous PR.


rafatower:
@Kartones quick review please

I had to change a bunch of tests, mainly because cartodb_id is now created as bigint and the order of columns is not guaranteed to be the same (actually the criteria is different, but they are equally ordered in upper layers)

Kartones:
:+1:

PR #5273, opened 2015-08-27
Adds truncate param to append method

### LINKED ISSUE #5257: GeoTIFF import error seems to be trimmed in the log
`16691887-f54a-459b-8cb9-95ec808ce911``

gfiorav:
Also applies truncate false to ogr2ogr call and exception spills.

Fixes #5257

Please CR @Kartones :smile:


Kartones:
Apart from the automated tests stabilization (that second param is going to cause some issues almost 100% sure with some Log mocks)...
1) Please add the same parameter and functionality to the activerecord model https://github.com/CartoDB/cartodb/blob/master/app/models/carto/log.rb#L22
2) This requires testing at staging of imports & syncs (first sync, which works almost as an import, and at least 1 consecutive sync which goes through "pure sync" flow), so please be careftul to check it.

Other than that... I'm ok with the change

gfiorav:
@Kartones 

Everything you requested has been tested successfully. Some changes have been made to the code regarding some ruby optional parameters formatting. Now all works well. As I'm out of time, I leave it to you to:
+ CR the changes
+ Merge

Thanks a lot! @Kartones

Kartones:
With that bugfix all scenarios working nicely and tested at staging

PR #5281, opened 2015-08-28
Fix incorrect geocoding modal for google cloud users

viddo:
- Check the (recommended) `user.google_maps_geocoder_enabled?` instead
to decide whether user has google maps geocode enabled
- Don’t show costs info (estimation, quota) for GME users, since not
relevant anyway
- Effectively enabled GME users to always be able to georeference
dataset

screen shot 2015-08-28 at 12 41 21
_modal as seen for GME users, with these changes_

@rafatower @xavijam can you code review, pretty please?
cc @iriberri


app/models/user/user_decorator.rb
```diff
@@ -60,7 +60,8 @@ def data(options = {})
           import_quota: self.import_quota,
           remove_logo: self.remove_logo?,
           sync_tables: self.sync_tables_enabled,
-          arcgis_datasource: self.arcgis_datasource_enabled?
+          arcgis_datasource: self.arcgis_datasource_enabled?,
+          google_maps_geocoder: self.google_maps_geocoder_enabled?,
```
xavijam: Wait, it is impossible to have google geocoder if the google_maps feature flag is not activated.

lib/assets/javascripts/cartodb/common/dialogs/georeference/georeference_model.js
```diff
@@ -192,7 +192,7 @@ module.exports = cdb.core.Model.extend({
   },
 
   _isGoogleMapsUser: function() {
-    return this.get('user').featureEnabled('google_maps');
+    return !!this.get('user').get('actions').google_maps_geocoder;
```
xavijam: ^^^ cc @javisantana

viddo: based this change on the info I got from @rafatower 

> The reliable way of checking if the geocoder is enabled: user.google_maps_geocoder_enabled?

xavijam: But is a google maps user function checks if the feature flag is enabled, not if he can use the google geocoder.

rafatower:
> pretty please?

XD Looks good to me though I don't master the js part :+1:

xavijam:
For me looks great, but the part about the checking of gmaps user or not, I don't agree. maybe we need @javisantana here.

rafatower:
@xavijam please check the related ticket, that were decisions taken at the time and documented.

xavijam:
If you have discussed it and you agree, no more to say.

viddo:
retest this please

app/models/user/user_decorator.rb
javisantana: Yes, that's right. 

I guess it's better to have as many switches as features but having that split is more error prone. I don't have a clear vision here, up to you

lib/assets/javascripts/cartodb/common/dialogs/georeference/georeference_model.js
javisantana: that's not good, google maps user is the one that has google client id configured (that's the minimum)

viddo: > that's not good, …

That was in reply to @xavijam or me?

Unless I misunderstand your replies, the change here is correct, since [`user.google_maps_geocoder_enabled?`](https://github.com/CartoDB/cartodb/blob/master/app/models/user.rb#L2525-L2527) does indeed check that the `client_id` is present (as well as `google_maps_private_key`). The `google_maps` feature flag could be set to true w/o the user actually using the google geocoder.

app/controllers/carto/api/user_presenter.rb
```diff
@@ -81,7 +81,8 @@ def data(options = {})
             import_quota: @user.import_quota,
             remove_logo: @user.remove_logo?,
             sync_tables: @user.sync_tables_enabled,
-            arcgis_datasource: @user.arcgis_datasource_enabled?
+            arcgis_datasource: @user.arcgis_datasource_enabled?,
+            google_maps_geocoder: @user.google_maps_geocoder_enabled?,
```
rafatower: minor detail, here and just below: I'm all for trailing commas but backend consensus is to avoid them :-/

lib/assets/javascripts/cartodb/common/dialogs/georeference/georeference_model.js
rafatower: bad naming, how about `_isGmeGeocoderEnabled`

rafatower: Also, there's a problem around these lines, regarding copies:
https://github.com/CartoDB/cartodb/blob/fix-gme-georef-modal/lib/assets/javascripts/cartodb/common/dialogs/georeference/georeference_model.js#L185-L186

viddo: >, regarding copies

not sure I follow, can you clarify what you mean with copies in this context?

viddo: :+1: for the name proposal

app/controllers/carto/api/user_presenter.rb
viddo: aight, my bad, I'll correct that

iriberri:
Hey, is this prepared to go? :-)

viddo:
Yes, but considering I'm at the conference and have spotty access/will be busy either someone else need to merge/deploy, or you'll have to wait until Thu

PR #5303, opened 2015-08-31
New search control with pin+infowindow

### LINKED ISSUE #4914: Add marker for geocoder search box result
On a public or embedded map with the Geocoder Search Box enabled, when a user types in an address the CartoDB map already zooms and centers to that location. However, we should add a map marker at that location also.

The marker does not need to persist when later geocodes are done. Could be as simple as L.marker at a minimum, or Design can think of something more advanced.

This is the behavior in ArcGIS online and Mapbox.

(moved here @saleiva @javisantana )

xavijam:
Adding a pin + infowindow with the search info, made in [CartoDB.js](https://github.com/CartoDB/cartodb.js/pull/661) and now available under the editor, but with you edit tools with help to:

- Add an annotation in that point (if user is in a visualization).
- Add a marker in that point (if data layer is made of points and a query is not applied).

---
screen shot 2015-08-31 at 14 22 40
---
screen shot 2015-08-31 at 14 23 01
---
screen shot 2015-08-31 at 14 23 18
---

Fixes #4914 
cc @iriberri 
REVIEWER: @javierarce


app/assets/stylesheets/map/map.css.scss
```diff
@@ -759,67 +762,9 @@ div.cartodb-popup.dark  {
...
     a.button {
       border-color:#676767!important;
-      &.edit_data {@include edit-bottom-states(-4, -18, #222, #222, #222, #222);}
-      &.edit_geo {@include edit-bottom-states(-35, -18, #222, #222, #222, #222);}
-      &.remove {@include edit-bottom-states(-66, -18, #222, #222, #222, #222);}
-    }
-  }
-}
-
-// FOR TOOLTIPS //
```
xavijam: Tooltip doesn't have any edit buttons on it, so removed unnecessary styles.

lib/assets/javascripts/cartodb/table/overlays/mobile.js
```diff
@@ -693,7 +693,8 @@ cdb.admin.overlays.Mobile = cdb.core.View.extend({
 
     this.search = new cdb.geo.ui.Search({
       template: this.getTemplate("table/views/search_control"),
-      model: this.options.map
+      model: this.options.map,
+      mapView: this.options.mapView
```
xavijam: Needed for new search control.

xavijam:
@josecruz, it will be available in rui servers.

Issue #4730, opened 2015-07-28
start using X_POSSIBLE_NAMES, Y_POSSIBLE_NAMES and GEOM_POSSIBLE_NAMES from ogr2ogr in the importer


javisantana:
since 2.1 those options are available to do the transfomation from lat,lng to geometry. That will improve the import of those files since that transformation does not need to be done in the database

GEOM_POSSIBLE_NAMES will also allow to remove the geojson, wkb transformations in the database

--

[From rafatower] here's my plan:
- [x] cartodb branch to use just `ogr2ogr2` binary for imports (right now we are using both `ogr2ogr` from gdal v 1.11 and `ogr2ogr2` from gdal trunk -- mind the 2, depending on the cases)
- [x] test new `ogr2ogr2` binary in staging
- [x] deploy new `ogr2ogr2` binary to production
- [x] cartodb branch to use `X_POSSIBLE_NAMES` and `Y_POSSIBLE_NAMES` ogr options
- [x] cartodb branch to use `GEOM_POSSIBLE_NAMES` ogr option https://github.com/CartoDB/cartodb/pull/5945
- [ ] get gdal fixes for ogr2ogr/ogrinfo (need another regression to make sure there are no surprises)
- [ ] deploy newer package with the fixes
- [ ] integrate the PR https://github.com/CartoDB/cartodb/pull/5931 to remove all the extra logic from `Georeferencer` ruby class except for KML

The first step is aimed at reducing complexity. Eventually we should use just one binary, from gdal official releases cc/  @Kartones @zenitraM

pramsey:
I wonder if there's some ogr2ogr options to write the webmercator column simultaneous w/ the lonlat one? If both geom columns are pre-written then the cartodbfy step reduces to just a bunch of renaming of things and checks.

javisantana:
Even told me that working with the sql option it can be possible. I tried but I got a segment fault:

```
vagrant@precise64:~$ ogr2ogr -f PostgreSQL  PG:"host=localhost port=5432 user=postgres dbname=t"  -lco DIM=2 -lco PRECISION=NO /vagrant/tm_world_borders_s_11.shp  -nln tt -nlt PROMOTE_TO_MULTI -overwrite -sql "select st_transform(GEOMETRY,4326) as the_geom, st_transform(GEOMETRY,3857) as the_geom_webmercator from tm_world_borders_s_11" -dialect sqlite
[...]

PR #5306, opened 2015-08-31
4730 just one ogr2ogr2 binary

rafatower:
Use ogr2ogr2 as default binary for imports #4730 

In order to simplify code and maintenance, use just the ogr2ogr2 binary
(mind the 2 at the end). Just keep ability to set the binary through the
config for flexibility and compatibility.

cc'ing @ethervoid as I saw you edited some of those files recently


services/importer/lib/importer/ogr2ogr.rb
```diff
@@ -14,7 +14,7 @@ class Ogr2ogr
       OSM_INDEXING_OPTION   = 'OSM_USE_CUSTOM_INDEXING=NO'
       APPEND_MODE_OPTION    = '-append'
 
-      DEFAULT_BINARY = 'which ogr2ogr'
+      DEFAULT_BINARY = 'which ogr2ogr2'
```
rafatower: this is the important change

services/importer/lib/importer/kml_splitter.rb
```diff
@@ -38,7 +38,7 @@ def source_files_for(source_file, layer_names=[])
       end
 
       def extract(extracted_file_path, source_file, layer_name)
-        `ogr2ogr -f 'KML' #{extracted_file_path} #{source_file.fullpath} "#{layer_name}"`
+        `ogr2ogr2 -f 'KML' #{extracted_file_path} #{source_file.fullpath} "#{layer_name}"`
```
rafatower: I reviewed all the usages of `ogr2ogr` and found this. I think this file is no longer used, along with the following ones:
https://github.com/CartoDB/cartodb/blob/master/services/importer/lib/importer/unp.rb
https://github.com/CartoDB/cartodb/blob/master/services/importer/spec/unit/unp_spec.rb

can I get rid of them?

Kartones: UNP is used for uncompressing, are you really sure it is not used?

kml_splitter might not be frequently used, but my knowledge of KML is minimal, it might be for a very specific scenario...

Kartones:
:+1: for the change, but I'd be careful with removing UNP, as I said I think it is indeed used

services/importer/lib/importer/ogr2ogr.rb
ethervoid: What about this https://github.com/CartoDB/cartodb/blob/master/config/app_config.yml.sample#L3 ?

services/importer/lib/importer/kml_splitter.rb
rafatower: Mmm you're right, `Unp` is indeed in use.

services/importer/lib/importer/ogr2ogr.rb
rafatower: you're right, that also needs to be changed to `ogr2ogr2` thanks!

PR #5310, opened 2015-08-31
Use table_names from user instead of get the visualizations

ethervoid:
Due connection problems because we use too many connections to retrieve the table names from the users databases, we have to stop update this field by now.

The solution is to use the `table_name` from the layer options to set this value

### DIFF
--- a/lib/tasks/explore_api.rake
+++ b/lib/tasks/explore_api.rake
@@ -255,14 +255,15 @@ namespace :cartodb do
     end
 
     def get_visualization_tables(visualization)
-      table_names = visualization.related_tables.map{ |table| table.name }
+      # We are using layers instead of related tables because with related tables we are connecting
+      # to the users databases and we are reaching the connections limit
+      table_names = visualization.layers(:carto_and_torque).map{ |layer| layer.options['table_name'] }
       %Q{{#{table_names.join(",")}}}
     end
 
     def update_mapviews_and_likes_query(visualization)
       v = visualization
-      # Only derived visualizations could add or remove tables (layers)
-      update_tables = %Q{, visualization_table_names = '#{get_visualization_tables(v)}'} if v.type == CartoDB::Visualization::Member::TYPE_DERIVED
+      update_tables = %Q{, visualization_table_names = '#{get_visualization_tables(v)}'}
       # Synchronization method from Visualization::Relator uses empty Hash when there is no sync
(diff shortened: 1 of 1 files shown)


ethervoid:
@Kartones please review this PR (it's a bit urgent)

lib/tasks/explore_api.rake
```diff
@@ -255,14 +255,15 @@ namespace :cartodb do
...
 
     def get_visualization_tables(visualization)
-      table_names = visualization.related_tables.map{ |table| table.name }
+      table_names = visualization.layers(:carto_and_torque).map{ |layer| layer.options['table_name'] }
       %Q{{#{table_names.join(",")}}}
     end
 
     def update_mapviews_and_likes_query(visualization)
       v = visualization
       # Only derived visualizations could add or remove tables (layers)
```
Kartones: remove this comment then :P

lib/tasks/explore_api.rake
```diff
@@ -255,14 +255,15 @@ namespace :cartodb do
...
-      update_tables = %Q{, visualization_table_names = '#{get_visualization_tables(v)}'} if v.type == CartoDB::Visualization::Member::TYPE_DERIVED
+      # TODO after the first time all the user_tables are updated, we have to update only the derived ones
+      update_tables = %Q{, visualization_table_names = '#{get_visualization_tables(v)}'} #if v.type == CartoDB::Visualization::Member::TYPE_DERIVED
```
Kartones: don't comment, delete :D

lib/tasks/explore_api.rake
```diff
@@ -255,14 +255,15 @@ namespace :cartodb do
     end
 
     def get_visualization_tables(visualization)
-      table_names = visualization.related_tables.map{ |table| table.name }
+      table_names = visualization.layers(:carto_and_torque).map{ |layer| layer.options['table_name'] }
```
Kartones: please document here why you're going for a non-straightforward way of getting the tables, as looks suspicious

Kartones:
Apart from the comments, ok, but again: How this would affect any opensource? Why not making a new rake labeled "xxx_updates" to leave people in the previous state able to do the migration?

If this only affects our SaaS fine then

ethervoid:
Changes done. 

This right now AFAIK is affecting only our SaaS

Kartones:
:+1:

PR #5311, opened 2015-08-31
Simplify groups transitions/states handling

viddo:
This is a pre-step refactoring for the add-groups-users use-case.

While I started to work on this feature I realized that the current rendering was becoming too complicated, especially to handle the different pre-steps (creating, saving, deleting, prefetching etc.), so pushed that logic "up" (e.g. the `group.fetch()` call).

Both router and the main view should be both easier to understand and change now (e.g. can actually be unit-tested), as well as handling the preloading-state at the top-level instead of complicating individual child views (and get rid of that smelly switch-case in the main view).

Also: 
- [x] Header menu + back button to navigate between different views/"levels" in the view hierarchy
- [x] Changed the URL path to the group to be more proper "REST" (i.e. `/groups/:id/edit`)
- [x] Moved navigate-through-router up to main view
- [x] Using new layout container for groups (+add tab)

@alonsogarciapablo Can you review please?
cc @juanignaciosl FYI

### DIFF
--- a/app/views/admin/shared/_org_subheader.html.erb
+++ b/app/views/admin/shared/_org_subheader.html.erb
@@ -22,6 +22,13 @@
           <% if (!Cartodb.config[:cartodb_com_hosted].present? && (!current_user.organization.present? || current_user.organization_owner?))  %>
             <li class="Filters-typeItem"><a href="<%= current_user.plan_url(request.protocol) %>" class="Filters-typeLink">Billing</a></li>
           <% end %>
+          <% if current_user.has_feature_flag?('groups') %>
+            <li class="Filters-typeItem">
+              <a href="<%= CartoDB.url(self, 'organization_groups') %>" class="Filters-typeLink <%= 'is-selected' if "#{params[:controller]}_#{params[:action]}" == "admin/organizations_groups" %>">
+                Groups
+              </a>
+            </li>
+          <% end %>
         </ul>
       </div>
     </div>
--- a/lib/assets/javascripts/cartodb/organization_groups/create_group_view.js
+++ b/lib/assets/javascripts/cartodb/organization_groups/create_group_view.js
@@ -52,8 +52,8 @@ module.exports = cdb.core.View.extend({
   },
 
   _redirectToGroupsIndex: function() {
-    // Redirect back to list
-    this.options.router.navigate(this.options.router.rootUrl, { trigger: true });
+    // Redirect back to show group
+    this.options.router.navigate(this.options.router.rootUrl.urlToPath(this.options.groups.last().id), { trigger: true });
   },
 
   _showErrors: function() {
--- a/lib/assets/javascripts/cartodb/organization_groups/create_group_view.js
+++ b/lib/assets/javascripts/cartodb/organization_groups/create_group_view.js
@@ -1,3 +1,4 @@
+var _ = require('underscore');
 var cdb = require('cartodb.js');
 var randomQuote = require('../common/view_helpers/random_quote');
 
@@ -12,7 +13,10 @@ module.exports = cdb.core.View.extend({
   },
 
   initialize: function() {
-    if (!this.options.groups) throw new Error('groups is required');
+    _.each(['group', 'onCreated'], function(name) {
+      if (!this.options[name]) throw new Error(name + ' is required');
+    }, this);
+
     this.model = new cdb.core.Model();
     this._initBinds();
   },
@@ -41,25 +45,18 @@ module.exports = cdb.core.View.extend({
     var name = this._name();
     if (name) {
       this.model.set('isLoading', true);
-      this.options.groups.create({
+      this.options.group.save({
         display_name: name
       }, {
         wait: true,
-        success: this._redirectToGroupsIndex.bind(this),
+        success: this.options.onCreated,
         error: this._showErrors.bind(this)
       });
     }
   },
 
-  _redirectToGroupsIndex: function() {
-    // Redirect back to show group
-    this.options.router.navigate(this.options.router.rootUrl.urlToPath(this.options.groups.last().id), { trigger: true });
-  },
-
   _showErrors: function() {
-    this.model.set({
-      isLoading: false
-    });
+    this.model.set('isLoading', false);
   },
 
   _onChangeName: function() {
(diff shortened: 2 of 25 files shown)


alonsogarciapablo:
Taking a look!

app/views/admin/shared/_org_subheader.html.erb
```diff
@@ -22,8 +22,15 @@
           <% if (!Cartodb.config[:cartodb_com_hosted].present? && (!current_user.organization.present? || current_user.organization_owner?))  %>
             <li class="Filters-typeItem"><a href="<%= current_user.plan_url(request.protocol) %>" class="Filters-typeLink">Billing</a></li>
           <% end %>
+          <% if current_user.has_feature_flag?('groups') %>
```
alonsogarciapablo: Will this feature be available for every kind of user or just users that belong to an org?

alonsogarciapablo: I didn't realized this file is called `_org_subheader` so I guess it will only be rendered for users in an org. I asked before I saw some checks like `current_user.organization.present?` in this same file.

lib/assets/test/spec/cartodb/organization_groups/edit_group_view.spec.js
```diff
@@ -181,12 +112,7 @@ describe('organization_groups/edit_group_view', function() {
     });
   });
 
-  it("should not have leaks", function() {
```
alonsogarciapablo: Shouldn't we leave this here?

lib/assets/test/spec/cartodb/organization_groups/group_header_view.spec.js
```diff
@@ -0,0 +1,94 @@
...
+      expect(this.view.$('h3').text()).toEqual('my group');
+    });
+
+    it('should render the menu', function() {
+      expect(this.innerHTML()).toContain('Users');
+      expect(this.innerHTML()).toContain('Settings');
+    });
+
+    it('should only render users count if has any', function() {
+      expect(this.innerHTML()).not.toContain('0 Users');
```
alonsogarciapablo: If the number of users == 1, it should say "1 User" (singular). Not sure if you're handling that case.

lib/assets/javascripts/cartodb/organization_groups/create_group_view.js
```diff
@@ -12,7 +13,10 @@ module.exports = cdb.core.View.extend({
   },
 
   initialize: function() {
-    if (!this.options.groups) throw new Error('groups is required');
+    _.each(['group', 'onCreated'], function(name) {
```
alonsogarciapablo: I realised we are doing this kind of check every time. We could perhaps think about creating a helper method to take care of this... Something like `ensureOptionsArePresent('group', 'onCreated')`... Just an idea...

lib/assets/javascripts/cartodb/organization_groups/router.js
```diff
@@ -42,23 +53,75 @@ module.exports = Router.extend({
   },
 
   routeToGroupsIndex: function() {
```
alonsogarciapablo: I like how these guys read now :+1:

lib/assets/javascripts/cartodb/organization_groups/groups_main_view.js
```diff
@@ -15,7 +12,7 @@ module.exports = cdb.core.View.extend({
   },
 
   initialize: function() {
-    _.each(['user', 'groups', 'router'], function(name) {
+    _.each(['router'], function(name) {
```
alonsogarciapablo: It looks like we're only using the router to access some methods on the model (`this.options.router.model`). Why not pass the model directly to make Demeter happy?

alonsogarciapablo:
Left some minor comments. Code looks great :clap: :clap: :clap:

app/views/admin/shared/_org_subheader.html.erb
viddo: Only for organization owner (there's a separate UI for organization users, TBD)

lib/assets/test/spec/cartodb/organization_groups/edit_group_view.spec.js
viddo: Yes, it was just moved to the top of the file :)

lib/assets/test/spec/cartodb/organization_groups/group_header_view.spec.js
viddo: Ouch, good catch :+1:

lib/assets/javascripts/cartodb/organization_groups/create_group_view.js
viddo: Yes, I've been thinking about that too. Have not yet had the time to figure out where would be suitable though (especially since `cdb.core.View` is located in the cartodb.js lib), not sure we should push it there or extend the View in the cartodb-editor context. What do you think?

lib/assets/javascripts/cartodb/organization_groups/groups_main_view.js
viddo: Router is used implicitly for the Backbone navigation (see [`navigateThroughRouter`](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb/common/view_helpers/navigate_through_router.js)

lib/assets/test/spec/cartodb/organization_groups/edit_group_view.spec.js
alonsogarciapablo: :+1:

lib/assets/javascripts/cartodb/organization_groups/groups_main_view.js
alonsogarciapablo: :+1:

lib/assets/javascripts/cartodb/organization_groups/create_group_view.js
alonsogarciapablo: Not sure either. I guess it could be useful in cartodb.js too, so probably makes sense to add it there.

viddo: :+1: I'll look into this later

Issue #5009, opened 2015-08-12
Seems like importer is using the user queries timeout


rafatower:
See import id `41c250a6-3d4a-49d6-a963-8130e144560b`
```
2015-08-12 16:33:55 UTC: Finished importing data from /tmp/imports/20150812-19901-nasszk/twitter_alfabet_alphabet_abcxyz_google.csv
2015-08-12 16:33:56 UTC: Proceeding to register
2015-08-12 16:33:56 UTC: Before renaming from importer_6cab8642410111e5a8420e6e1df11cbf to twitter_alfabet_alphabet_abcxyz_google
2015-08-12 16:33:56 UTC: Before moving schema 'twitter_alfabet_alphabet_abcxyz_google' from cdb_importer to piensaenpixel
2015-08-12 16:33:56 UTC: Before persisting metadata 'twitter_alfabet_alphabet_abcxyz_google' data_import_id: 41c250a6-3d4a-49d6-a963-8130e144560b
2015-08-12 16:39:34 UTC: Exception: PG::Error: ERROR:  canceling statement due to statement timeout
CONTEXT:  SQL statement "UPDATE twitter_alfabet_alphabet_abcxyz_google SET the_geom_webmercator = cartodb.CDB_TransformToWebmercator(the_geom) "
PL/pgSQL function _cdb_populate_the_g
```

The stacktrace beautified:
```
app/connectors/importer.rb:71:in `rescue in register'
app/connectors/importer.rb:57:in `register'
app/connectors/importer.rb:49:in `block in run'
app/connectors/importer.rb:48:in `each'
app/connectors/importer.rb:48:in `run'
app/models/data_import.rb:617:in `new_importer'
app/models/data_import.rb:350:in `dispatch'
app/models/data_import.rb:161:in `run_import!'
lib/resque/importer_jobs.rb:9:in `block in perform'
lib/resque/base_job.rb:21:in `call'
lib/resque/base_job.rb:21:in `run_action'
lib/resque/importer_jobs.rb:9:in `perform'
```

The user `piensaenpixel` had the "Timeout for user queries" set to `300000` which should be in ms. That roughly corresponds to 5 min before crashing with db statement timeout error.

rafatower:
Note: I raised the timeout as a temporary workaround.

zenitraM:
ogr2ogr connects with the proper user and does not override the statement_timeout so this is expected

ethervoid:
[...]

PR #5322, opened 2015-09-01
Add db timeout for ogr2ogr #5009

rafatower:
independent from user's statement timeout

@Kartones please review. Tested in local setting 1s timeout for my user, could import public dataset "populated places".

The 1 hour timeout is arbitrary, but taken into account saleiva has 50 minutes and most imports work as expected, 1 hour seems reasonable.


Kartones:
:+1:

ethervoid:
:clap: :clap: :clap:

PR #5329, opened 2015-09-01
Remove quotes and take only the table name

ethervoid:
Some layers' `table_name` field have the following format `"rilos-katia".alcohol_export_v2` and this could be parse with the actual method.

Please @Kartones take a look to this PR :)

### DIFF
--- a/lib/tasks/explore_api.rake
+++ b/lib/tasks/explore_api.rake
@@ -36,7 +36,6 @@ namespace :cartodb do
                 visualization_description,
                 visualization_type,
                 visualization_synced,
-                visualization_table_names,
                 visualization_tags,
                 visualization_created_at,
                 visualization_updated_at,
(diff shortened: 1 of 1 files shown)


Kartones:
Please try this with organization users who have visualizations with tables from other users of the same org.

Eg.: User B has a vis which has a layer with User A shared table. This will cause similar "conflictive names" and it should work...

ethervoid:
@Kartones take a look to the change. I think this way is going to check for all the cases. ie.

`"user_username": "ethervoid",`
`"visualization_table_names": [`
`"fl_insurance_sample",`
`"\"ethervoid-10\".alchohol_export_v2"`
],`

lib/tasks/explore_api.rake
```diff
@@ -254,16 +253,35 @@ namespace :cartodb do
...
 
-    def update_mapviews_and_likes_query(visualization)
+    def extract_table_name(layer, explore_visualization)
+      table_username = layer.options['user_name']
+      table_name = layer.options['table_name'].gsub(/"/,'\"').split('.')
+      #Here we check for:
+      # a) return as comes, but escaping quotes, if the table_name have a . this means an old format username.tablename
+      # b) return table_name only if there is only the tablename and the visualization owner and the layer username is the same
+      # c) return username.table_name if there is only the tablename and the visualization owner and the layer username are not the same
+      #Always escape the username because to avoid problems with non-alfanumeric characters like '-'
```
Kartones: "alfanumeric" -> "alphanumeric" :trollface:

Kartones:
Looks better, for the future, as soon as a rake becomes more complex than a mere query, I'd build cartodb class/helper/utility/whatever, both to keep rakes small (a defect from the past) and to be able to add tests to non-trivial parts (like yours)

Kartones:
:+1:

ethervoid:
Yup I've thought about that and I come to the same conclusion, this rake is too complex for just a rake so I have to move logic to a helper or make a service for it

PR #5340, opened 2015-09-01
Privacy toggler within create dataset dialog

### LINKED ISSUE #3111: Design CTA + Check for creating private tables on import
and inform users that tables are public by default.

xavijam:
Basically it will let you to change the privacy of the import you want to start. But, if you don't have private_tables enabled under your user, it will redirect to your upgrade page when it is clicked.
For open source users that link will be disabled and it will show a "standard" text.
One important thing, by default, if user can create private datasets, the privacy value will be "PRIVATE", if the user can create private tables, of course, the value will be "PUBLIC".

Example:

privacy_toggler

---

We should review the texts of the tooltips cc @csobier @javierarce.

---
Possibility to change the privacy:
screen shot 2015-09-01 at 17 36 26
---
Upgrade:
screen shot 2015-09-01 at 18 55 07
---

Fixes #3111 
REVIEWER: @alonsogarciapablo 
cc @javierarce @saleiva


xavijam:
Deployed in staging ruis

saleiva:
@javier  could you take a quick look to this, specially to the texts?

alonsogarciapablo:
Code looks ~~good~~ great to me! I've done some QA locally and works as expected. So :+1: 

@javierarce what do you think?

javierarce:
:+1: code looks good to me. I'm going to review the texts before merging.

lib/assets/javascripts/cartodb/common/dialogs/create/footer/privacy_toggler_template.jst.ejs
```diff
@@ -0,0 +1,16 @@
+<% if (canUpgrade) { %>
+  <a href="<%- upgradeUrl %>" class="PrivacyToggler js-toggler PrivacyToggler--<%- privacy %> is-disabled" data-title="You cannot change the privacy of your new datasets. Click here to upgrade your account.">
+    <i class="iconFont iconFont-<%- icon %>"></i>
+  </a>
+<% } else {%>
+  <button type="button" class="PrivacyToggler js-toggler PrivacyToggler--<%- privacy %> <%- isDisabled ? 'is-disabled' : '' %>"
+    data-title="
+      <% if (!isDisabled) { %>
+        Your new dataset will be <%- privacy.toLowerCase() %>. Click to change it to <%- nextPrivacy.toLowerCase() %>
```
javierarce: Let's add a new line here, so "Click" appears in the next line.

lib/assets/javascripts/cartodb/common/dialogs/create/footer/privacy_toggler_template.jst.ejs
```diff
@@ -0,0 +1,16 @@
...
+      <% } else { %>
+        Your user can change the privacy of the new datasets
```
javierarce: "You can change the privacy of the new datasets"… sounds better, right?

alonsogarciapablo: Actually I think it should read: "You **cannot** change the privacy of new datasets.". I think this is the message that should be displayed when the users can't upgrade and can't change the privacy of datasets... @javierarce: good catch!

javierarce: Oh, that's true!

lib/assets/javascripts/cartodb/common/dialogs/create/footer/privacy_toggler_template.jst.ejs
javierarce: screen shot 2015-09-04 at 15 56 13

screen shot 2015-09-04 at 15 56 07

javierarce:
You're totally right, I'll make the change.

xavijam:
:clap: :clap:

PR #5352, opened 2015-09-02
[NO MERGE] 2202 cast date to number

rafatower:

### DIFF
--- a/app/models/table/column_typecaster.rb
+++ b/app/models/table/column_typecaster.rb
@@ -173,7 +173,7 @@ module CartoDB
     end #number_to_boolean
 
     def date_to_number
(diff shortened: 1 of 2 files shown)


Kartones:
Uhm, but the original issue is that somehow we didn't had properly installed the extension? because you're basically extracting the logic from the SQL to Ruby but is exactly the same, so I don't know why we don't keep using the extension function...

Kartones:
I mean, the function exists... https://github.com/CartoDB/cartodb-postgresql/blob/master/scripts-available/CDB_DateToNumber.sql

rafatower:
Okay, I've look for that and did not found it. The problem are types. A `timestamp with time zone` is different from a `timestamp`

rafatower:
Fixed in the cdb extension.

PR #5371, opened 2015-09-03
Move connection creation inside begin block as it may fail

rochoa:


rochoa:
cc @zenitraM

zenitraM:
:hamster:

PR #5374, opened 2015-09-03
Reject empty custom attributions

alonsogarciapablo:
Custom attributions are "" by default and we're joining them in the viz.json. That was causing some attributions of layers with multiple sublayers show things like: ", ". This prevents that issue.

@Kartones can you please review this? Thanks!


Kartones:
:+1:

Issue #5373, opened 2015-09-03
Misalignment of the visualization title under Firefox


javierarce:
![screen shot 2015-09-03 at 18 00 52](https://cloud.githubusercontent.com/assets/4933/9663808/9a50733e-5266-11e5-8735-23ba5ee5b3f7.png)

PR #5391, opened 2015-09-04
Fix layer name missalignment in Firefox

javierarce:
This PR fixes a problem with the position of the layer name under Firefox.

Original ticket: #5373 

Can you review this @alonsogarciapablo, please?


alonsogarciapablo:
:+1:

It also fixes this issue in Chrome:
screen shot 2015-09-04 at 16 00 16

Great catch!

Issue #5368, opened 2015-09-03
Choropleth wizard ramp shows just two colors per ramp


javierarce:
How to reproduce it: 

1. Go to the choropleth wizard.
2. Select a ramp in the ramp selector.

![screen shot 2015-09-03 at 16 53 01](https://cloud.githubusercontent.com/assets/4933/9661689/1eef00ac-525c-11e5-8b68-4a0f933d8f39.png)

This happens in Chrome but not in Firefox or Safari.

javierarce:
After some help from the team, we discovered it's related to the latest version of Chrome (45.0.2454.85)

javierarce:
This browser bug could be related: https://code.google.com/p/chromium/issues/detail?id=527343&q=table&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Cr%20Status%20Owner%20Summary%20OS%20Modified

xavijam:
Fixed!

PR #5402, opened 2015-09-07
Fix color ramp bucket rendering

javierarce:
Temporary fix for the color ramp bug:

screen shot 2015-09-03 at 16 53 01

I noticed that adding a width to the `td.color` element the rendering of the cells gets fixed.

Original issue: #5368.


javierarce:
@matallo: I added a temporary solution. Could you review this, please?

app/assets/stylesheets/table/table_panel/layer-views-panels/wizard_panel.css.scss
```diff
@@ -324,3 +322,9 @@
       }
     }
   }
+  
+  /* Temporary hack for a Chrome v45 bug */
+  /* https://github.com/CartoDB/cartodb/issues/5368 */
+  .select2-result-label td.color { 
+     width: 10%;
```
matallo: I tried with 14% and it work with any range between 1-14%, no idea why

matallo:
:+1:

PR #5413, opened 2015-09-07
Added job_role and industry fields to user

ethervoid:
We need this two fields to integrate them in the explore api.

Please @Kartones check this two so you have check the other related PRs. Thank you


Kartones:
+1

PR #5418, opened 2015-09-07
3501 notify when reach max retries

### LINKED ISSUE #3501: Suggestion: Send an email to user when sync has reached max_retries
Feedback from @d4weed

When a sync table fails to be synced and it gets to "max retries", it's never synced again. We could warn the users through an email that this is happening so that they're aware that their data will be no longer updated.

cc @saleiva

ethervoid:
Now we send an email to the owner of the dataset when reaches the max number of retries while trying to synchronize.
 
Closes #3501

### DIFF
--- a/app/models/synchronization/member.rb
+++ b/app/models/synchronization/member.rb
@@ -228,7 +228,7 @@ module CartoDB
         notify
 
       rescue => exception
-        Rollbar.report_exception(exception)
+        CartoDB.notify_exception(exception)
         log.append exception.message, truncate = false
         log.append exception.backtrace.join('\n'), truncate = false
 
@@ -385,7 +385,13 @@ module CartoDB
           self.run_at         = Time.now + interval
         end
       rescue => e
-        Rollbar.report_exception(e)
+        CartoDB.notify_exception(e,
+          {
+            error_code: error_code,
+            error_message: error_message,
+            retried_times: self.retried_times
+          }
+        )
       end
 
       # Tries to run automatic geocoding if present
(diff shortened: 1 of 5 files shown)


ethervoid:
Please @rafatower check this PR :)

app/models/synchronization/member.rb
```diff
@@ -368,9 +368,12 @@ def set_failure_state_from(importer)
           default_message = CartoDB::IMPORTER_ERROR_CODES.fetch(self.error_code, {})
           self.error_message = default_message.fetch(:title, '')
         end
-        self.retried_times  = self.retried_times + 1
+        self.retried_times  += 1 unless self.retried_times >= MAX_RETRIES
```
rafatower: instead of the `unless`, shouldn't be the increment right before `::Resque.enqueue`?

app/models/synchronization/member.rb
```diff
@@ -380,12 +383,21 @@ def set_general_failure_state_from(exception, error_code = 99999, error_message
         self.state          = STATE_FAILURE
         self.error_code     = error_code
         self.error_message  = error_message
-        self.retried_times  = self.retried_times + 1
+        self.retried_times  += 1 unless self.retried_times >= MAX_RETRIES
```
rafatower: same as above here

rafatower:
aside from that minor comment, ok

app/models/synchronization/member.rb
ethervoid: Inside the if not the else i assume. If left this way because is how its done and I don't know if there is some implication. If its ok I'll put inside the `if self.retried_times < MAX_RETRIES`

rafatower: yes, in the `if` part.

PR #5428, opened 2015-09-08
Fix problem with limits dialog when user has more layers than allowed

### LINKED ISSUE #5411: Maximum number of layers CTA only showing when you're in the limit
If you have exceeded your layers limit, you can keep on adding layers.

**Steps to reproduce:**
- In a map, start adding layers. When you're going to hit the limit, open the same URL in a new tab and add a layer in both tabs at the same time. Now you exceed your layers quota.
- You can keep adding layers, no CTA.
- Get back to your map and delete your extra layers to have exactly the number of allowed layers you can have.
- Try to add a new layer: the CTA appears

A related question: is there any call we can do to check the real status of the map regarding layers? The tab trick is an easy one to avoid the restriction.

cc @saleiva  @xavijam

xavijam:
Related with #5411 (frontend)

REVIEWER: @viddo 
cc @iriberri


lib/assets/javascripts/cartodb/table/actions_menu.js
```diff
@@ -40,14 +37,9 @@
       this.vis = this.options.vis;
       this.master_vis = this.options.master_vis;
       this.map = this.vis.map;
+      this.user = this.options.user;
       this.globalError = this.options.globalError;
 
-      // Set max layers if user has this parameter
-      var max_user_layers = this.options.user.get('max_layers');
```
xavijam: Kill me for doing this :(

viddo:
:+1:

PR #5429, opened 2015-09-08
Removing Bitly credentials from JS, setting them using application config

xavijam:
No more credentials set within url_shortener class, they will be set from the cdb.config model and those values will be provided through the frontend_config from the backend.

@luisbosque @javisantana


xavijam:
Sounds good?

lib/assets/javascripts/cartodb/common/url_shortener.js
```diff
@@ -1,29 +1,33 @@
...
   var cachedUrl = this.localStorage.search(originalUrl);
+  var self = this;
+
   if (cachedUrl) {
     return callbacks.success(cachedUrl);
   }
 
-  var self = this;
+  if (!this._LOGIN || !this._KEY) {
+    return callbacks.error(originalUrl);
```
javisantana: why error?

app/helpers/application_helper.rb
```diff
@@ -80,6 +80,8 @@ def frontend_config
       oauth_gdrive:               Cartodb.config[:oauth]['gdrive']['client_id'],
       oauth_instagram:            Cartodb.config[:oauth]['instagram']['app_key'],
       oauth_mailchimp:            Cartodb.config[:oauth]['mailchimp']['app_key'],
+      bitly_login:                Cartodb.config[:bitly]['login'],
```
javisantana: could we make this optional, if we do this in this way everyone will need to add that configuration

xavijam: You mean:

```rb
...
bitly_login: Cartodb.config[:bitly].present ? Cartodb.config[:bitly]["login"] : "",
bitly_key: Cartodb.config[:bitly].present ? Cartodb.config[:bitly]["key"] : "",
...
```

lib/assets/javascripts/cartodb/common/url_shortener.js
xavijam: Choosing between success or error, it sounds me as an error, but yeah, success could fit as well.

luisbosque:
Yep, bitly should be used only if the login/key exists and it's not empty. Otherwise the real CartoDB map url should be returned instead of the bitly url.

app/helpers/application_helper.rb
javisantana: yes

xavijam:
Until this https://github.com/CartoDB/cartodb-platform/issues/1055 issue is not ready, we can't deploy this pull request.

xavijam:
Apart from that, everything ok @luisbosque @javisantana ?

luisbosque:
:+1:

PR #5431, opened 2015-09-08
Add or remove group users

viddo:
Part-milestone for #4883

edit: This implements the last must-have tasks for the group admin (organization owner):
  - [x] add users to a group
  - [x] browse add-org-users list through pagination
  - [x] search add-org-users
  - [x] apply visual styles
  - [x] remove users from a group

Noteworthy: 
- The UI proposal had two different states depending on the group having users or not, which made things very complicated so opted for a simpler version to start with, showing the add-users modal if entering a group with no members.
- Refactored the share-permissions view to reuse the "organization users" part that contains the search bar since the logic/view here is exactly the same (although allow to inject how the list-view should be created since that varies for the context).

screen shot 2015-09-08 at 19 29 00
screen shot 2015-09-08 at 19 29 09

Also had to re-organize the files (move to organization/ dir), since there were a bunch of setup stuff that was needed for this entry point too (e.g. header dropdowns etc.)

@juanignaciosl can you review?


app/views/admin/shared/_org_subheader.html.erb
```diff
@@ -29,6 +26,9 @@
               </a>
             </li>
           <% end %>
+          <% if (!Cartodb.config[:cartodb_com_hosted].present? && (!current_user.organization.present? || current_user.organization_owner?))  %>
```
juanignaciosl: We should extract this to the controller or a helper.

juanignaciosl:
I've taken a quick look to this, but it's mostly frontend code, do you want me to test it?

app/views/admin/shared/_org_subheader.html.erb
viddo: Sure, where is the best place to put this kind of logic in your opinion?

viddo:
Yes it's mostly frontend. I think it's better to wait with testing until I'm done with the rest of the must-have's on the issue (in a day or two)

app/views/admin/shared/_org_subheader.html.erb
juanignaciosl: I think OrganizationsController covers all the uses of this subheader, so a begin_filter inside it might be the best fix. And, if it's needed elsewhere, you can always extract it to a mixin.

viddo: Fix in 117121b

viddo:
While waiting for the code-review, I've added the missing styles (minor view changes), so things looks more OK. There are obviously details missing (the things I consider "nice-to-have" for the moment), they can be filled in later.

@CartoDB/copywriters FYI this PR is not final, but heads up that groups UI is coming, so if you have any early feedback on the stuff it's appreciated, subscribe to #4883 if you want updates

screen shot 2015-09-09 at 15 09 05
screen shot 2015-09-09 at 15 09 11
screen shot 2015-09-09 at 15 09 16
screen shot 2015-09-09 at 15 10 04
screen shot 2015-09-09 at 15 10 13

app/assets/stylesheets/organization/organization_users_list.css.scss
```diff
@@ -19,13 +22,31 @@
...
 }
 .OrganizationList-user {
+  box-sizing: border-box;
+  position: relative;
   width: 100%;
 
-  &:not(:first-child) { border-top: 1px solid $cStructure-softLine; }
+  &:not(:last-child) { border-bottom: 1px solid $cStructure-softLine; }
+}
+// TODO: this is duplicated from .DatasetsList-item, can we consolidate these styles somehow?
```
viddo: @CartoDB/frontend This is unfortunate, since duplication leads to more maintenance issues down the line. Ideally we should find a way to make components sufficiently standalone so they can be reused in various contexts. 

Refactoring this would probably be a day or two of work (which I don't have right now), and the risk of introducing new visual bugs, so will not do it now. But I kindly ask you to simply keep this in mind before introducing new components in the future.

MariaCheca: We could start a new stylesheet for this elements with many CSS properties used in multiple views. A good reference to start would be the visual styleguide (buttons, paragraphs, border lines, etc).

alonsogarciapablo:
Massive PR! I didn't spotted anything crazy. So :clap: :clap: :clap:

PR #5439, opened 2015-09-09
Remove duplicates from explore api table names

ethervoid:
This PR removes duplicates in the table names. This duplicates are caused because the layers could be using the same table N times


ethervoid:

Kartones:
:+1:

PR #5448, opened 2015-09-09
If there is not map we should return nil for zoom

### LINKED ISSUE #5447: If the visualization don't have map avoid get the zoom attribute
In the [`vis_items`](https://github.com/CartoDB/cartodb/blob/master/app/controllers/admin/pages_controller.rb#L268) we should check if the map exists before we try to get the zoom element

ethervoid:
Closes #5447 

Please @Kartones  check this PR


Kartones:
:+1:

PR #5449, opened 2015-09-09
Calculate zoom on newly imported maps

### LINKED ISSUE #5307: Calculate zoom on newly imported maps
Map model default options set the bounding box, center and zoom as in https://github.com/CartoDB/cartodb/blob/18565865a8f1ba53aed56ab925d4799b7397931c/app/models/map.rb#L47-L54

I guess on import bounding box is calculated and set in map model but center and zoom keep their default values.

Named map send for the view option both, the zoom+center and the bounding box as in https://github.com/CartoDB/cartodb/blob/18565865a8f1ba53aed56ab925d4799b7397931c/services/named-maps-api-wrapper/lib/named-maps-wrapper/named_map.rb#L222-L242

But as in view option [zoom+center takes precedence over bounding box](https://github.com/CartoDB/Windshaft-cartodb/blob/2.12.0/docs/Map-API.md#arguments), named maps end having the default zoom=3, lat=30, lon=0 default options.

So all maps end having the same view after importing, making all canonical visualizations to have a, most of the time, incorrect view.

## STR:
1. Download https://common-data.cartodb.com/tables/districtes_barcelona/public as geojson/shp or other format.
1. Upload the dataset
1. Go to the canonical visualization static image preview

### Current result:
Image is using center+zoom with default zoom=3, lat=30, lon=0 although the bounding box is OK for the dataset.

### Expected result:
Image using bounding box.

## Examples:

I uploaded the dataset twice:
 1. https://team.cartodb.com/u/rochoa/viz/cb8e0456-4fe2-11e5-a2d6-0e5e07bb5d8a/public_map
 1. https://team.cartodb.com/u/rochoa/viz/e3343652-4fe7-11e5-a1a9-0e49835281d6/public_map

But then I modified the named map of the first one by just removing center and zoom from view. Have a look at the image preview difference by doing that:

after removing zoom + center

vs

default zoom + center

I guess the best thing is to not send the center+zoom if it's the default one and there is a bounding box available.

Thanks a lot @javierarce for your help and patience debugging this issue.

p.s.: I think this explains https://github.com/CartoDB/cartodb/issues/2974

Kartones:
Fixes #5307

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,7 @@
 * Log.append now allows to disable truncating (by default active)
 * Detection of lat/long columns now is done in `ogr2ogr2` rather than rails code [#5349](https://github.com/CartoDB/cartodb/pull/5349). In order to get this feature working (and some related tests), execute this to get the ogr2ogr2 package updated: `sudo apt-get update; sudo apt-get upgrade`
 * Removed Mixpanel tracking code [#5410](https://github.com/CartoDB/cartodb/pull/5410)
+* Newly imported datasets now properly calculate the map bounds and zoom and store them
 
 3.10.3 (2015-08-13)
 ---
--- a/app/models/map.rb
+++ b/app/models/map.rb
@@ -53,6 +53,8 @@ class Map < Sequel::Model
     center:          [30, 0]
   }
 
+  MAXIMUM_ZOOM = 18;
+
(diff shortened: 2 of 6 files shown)


Kartones:
@ethervoid CR please

ethervoid:
Good for me

PR #5451, opened 2015-09-09
If you try to import from data-library and fails, disappear from the list

### LINKED ISSUE #5132: If you try to import from data-library and fails, disappear from the list
**Problem**

I've trying to import a dataset from data-library without connection and failed. After that I don't have the datasets imported and has disappear from the data-library list

**Solution**

If there is an import error we should not add to `external_data_import` so we could re-import the dataset

Kartones:
Fixes #5132

### DIFF
--- a/app/models/data_import.rb
+++ b/app/models/data_import.rb
@@ -639,7 +639,7 @@ class DataImport < Sequel::Model
       self.table_id   = importer.table.id if importer.success? && importer.table
 
       if importer.success?
-        self.visualization_id = importer.data_import.visualization_id if importer.data_import.create_visualization
+        update_visualization_id(importer)
         update_synchronization(importer)
       end
 
--- a/app/models/data_import.rb
+++ b/app/models/data_import.rb
@@ -653,6 +653,12 @@ class DataImport < Sequel::Model
     end
   end
 
+  def update_visualization_id(importer)
+    if importer.data_import.create_visualization
+      self.visualization_id = importer.data_import.visualization_id
+    end
+  end
+
   def update_synchronization(importer)
     if synchronization_id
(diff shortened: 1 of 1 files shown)


Kartones:
@ethervoid CR please

app/models/data_import.rb
```diff
@@ -638,11 +638,11 @@ def store_results(importer=nil, runner=nil, datasource_provider=nil, manual_fiel
       self.table_name = importer.table.name if importer.success? && importer.table
       self.table_id   = importer.table.id if importer.success? && importer.table
 
-      if importer.success? && importer.data_import.create_visualization
-        self.visualization_id = importer.data_import.visualization_id
+      if importer.success?
+        self.visualization_id = importer.data_import.visualization_id if importer.data_import.create_visualization
```
ethervoid: I'd put the if condition above with &&

ethervoid:
Ok despite a minor comment

ethervoid:
retest this please

app/models/data_import.rb
Kartones: Careful, check the diff at the left, setting the vis. id is only to be done if the second condition is true, but updating the sync is to be done always that importer is a success

ethervoid: Hum you're right! . Anyway I don't like the inline ifs, maybe `importer.data_import.visualization_id if importer.data_import.create_visualization` should be extract away as a method. There are too much logic implicit that could lead to error

Kartones: I don't think is too complex or bad to have two ifs but ok, I'll extract to a method because if you already got confused others might too :+1:

Issue #5468, opened 2015-09-10
Resque failures not logged in rollbar


rafatower:
Whenever there's an exception, it is captured by `resque-metrics` and the job is logged in metrics as failure, but the exception is not logged in rollbar unless done explicitly i.e: `CartoDB.notify_exception(e)`

alonsogarciapablo:
There has been no activity on this issue for more several months. We are closing it. If you think this still needs to be addressed please open a new issue.

PR #5469, opened 2015-09-10
Log exceptions in LinkGhostTables #5464

rafatower:

I tried to find a quick solution to the more generic issue, bc that happens in all resque jobs, but couldn't. There's a ticket to deal with that: https://github.com/CartoDB/cartodb/issues/5468


Kartones:
:+1:

PR #5475, opened 2015-09-10
staging public pages are not indexed anymore

### LINKED ISSUE #4906: staging public pages are being indexed
https://www.google.es/search?q=cartodb-staging.com&oq=cartodb-staging.com&aqs=chrome..69i58j69i60l3j69i57.2974j0j4&sourceid=chrome&es_sm=91&ie=UTF-8

matallo:
- [x] Public pages
- [x] Login
- [x] Signup
- [ ] Error pages (404, 500, HTML5, ...)

close #4906

CR @xavijam

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -7,6 +7,9 @@
     <%= stylesheet_link_tag 'cartodb.css', 'common.css', :media => 'all' %>
     <%= yield :css -%>
     <%= csrf_meta_tag -%>
+    <% if Rails.env.staging? %>
+      <meta name="robots" content="noindex, nofollow">
+    <% end %>
   </head>
   <body>
     <% if ( ( controller_name != "visualizations" && controller_name != "tables" )  || action_name != "show" ) %>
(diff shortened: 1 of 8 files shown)


matallo:
added the rest of layouts (better safe than sorry) and the env :sweat_smile: 
 @xavijam

xavijam:
Looks ok :+1:

matallo:
I'm takling the error pages independently (as they need to be changed)

btw, @CartoDB/systems shouldn't we approach this with a robots.txt better? These will work anyway as we can use the `ENV` in the layout views

viddo:
Considering _we_ (cartodb) don't want to index staging servers, why not add this at the nginx level instead of enforcing it here, inline in the code? Something like https://gist.github.com/sergejmueller/0e508f3a4d92fd7ec110 using the x-robots-tag Bonus: this way we don't have to remember adding it to new layout files.

xavijam:
Uff... totally agree @viddo, @luisbosque could you take care of it?

PR #5484, opened 2015-09-11
5438 set srs for guessed csv files

rafatower:
Want to run the regression asap, I'll write a test before the review.


rafatower:
@Kartones please review

What this does: forces a source and target SRS when using csv guessing

Why? Because converting from lat/long columns with SRID = 0 (unknown) is the root cause of the problem. Probably worth a a fix in gdal trunk. Cartodbfy is not the place to fix because it does not have the context of "hey, this comes from degrees in a plain text CSV file".

I tried to write a test but cannot get a reasonably small csv file to reproduce the issue, the original has 128K rows. I keep trying but this is kind of urgent as it breaks syncs and some imports.

Kartones:
:+1:

PR #5491, opened 2015-09-11
show organization link for org. owner in public dashboard dropdown

### LINKED ISSUE #4400: https://org_name.cartodb.com/maps avatar dropdown doesn't show Your organization if being the admin

matallo:
close #4400

the problem with these two dropdowns is that they are created differently, one model (dashboard) comes from the `window.user_data` which is exposed in the rails view as this:
```
var user_data = <%= safe_js_object current_user.data.to_json %>;
```

the other model (public dashboard) comes from the authenticated user as this:
```
var user = new cdb.admin.User(authenticatedUser.attributes);
```

these two models carry different attributes thus the specific `_isOrgAdmin`

CR @CartoDB/frontend @viddo


lib/assets/javascripts/cartodb/public_common/user_settings/dropdown_view.js
```diff
@@ -21,6 +21,14 @@ module.exports = cdb.admin.DropdownMenu.extend({
     cdb.god.bind('closeDialogs', this.hide, this);
   },
 
+  _isOrgAdmin: function(user) {
+    if (user.organization) {
+      return user.organization.owner.email === user.email
```
matallo: Thanks @houndci

lib/assets/javascripts/cartodb/public_common/user_settings/dropdown_view.js
```diff
@@ -28,7 +36,8 @@ module.exports = cdb.admin.DropdownMenu.extend({
     this.$el.html(this.template_base({
       name: user.get('name') || user.get('username'),
       email: user.get('email'),
-      isOrgAdmin: user.isOrgAdmin(),
+      isOrgAdmin: this._isOrgAdmin(user),
+      organizationUrl: userUrl.urlToPath('organization').get('base_url'),
```
viddo: `groups.options.user.viewUrl().urlToPath('organization')` should be sufficient, `.toString()` on that object should be called implicitly when rendered in the template

lib/assets/javascripts/cartodb/public_common/user_settings/dropdown_view.js
```diff
@@ -21,6 +21,14 @@ module.exports = cdb.admin.DropdownMenu.extend({
     cdb.god.bind('closeDialogs', this.hide, this);
   },
 
+  _isOrgAdmin: function(user) {
+    if (user.organization) {
+      return user.organization.owner.email === user.email;
+    }
+
+    return false;
+  },
```
viddo: The view should not need to know these internals of the user model. From what you point out in the PR description it rather seems like the issue is that the data is not matching what the user model expect, so why not fix that instead?

matallo: because of this https://github.com/CartoDB/cartodb/issues?q=is%3Aopen+milestone%3ATorrelodones+assignee%3Amatallo+is%3Aissue :(

also, the quota progress isn't appearing, it's a toally different dropdown/model. I'd be delighted to work on that, but right now I'm very time constrained, let's discuss this with @CartoDB/frontend

viddo: From what I can tell that this item is not appearing is not a hotfix/urgent, so better to solve it in a way that is easier to maintain long-term, no need to stress this.

csobier:
Adding myself as a watcher, I reported to Maria that these options appear differently and was wondering about consistency.  I'll need to know if/when these options are changed in a future milestone so that I can update the doc accordingly.
image

matallo:
alright! in the so I modified the `api/user_controller` to return the needed info for the `isOrgAdmin` to work plus replicated some methods from the model (`upgrade_url`) https://github.com/CartoDB/cartodb/blob/4400-org_public_dropdown/app/controllers/carto/api/users_controller.rb#L64-L92

it would be great for some @CartoDB/backend to review this

Kartones:
Being a url that usually points to Central looks ok

matallo:
k, going through the hound madness and merging :disappointed:

app/controllers/carto/api/users_controller.rb
```diff
@@ -69,9 +69,28 @@ def render_auth_users_data(user, referrer, subdomain, referrer_organization_user
           email: user.email,
           organization: Carto::Api::OrganizationPresenter.new(user.organization).to_poro,
           base_url: user.public_url,
+          id: user.id,
```
javisantana: these are private attributes and should be exposed only in auth endpoints. Not sure this is the case but please, make sure

Kartones: uhm, good catch, we usually only expose `username`, and also I don't see any use of the `id`

matallo:
screen shot 2015-09-15 at 15 40 32

app/controllers/carto/api/users_controller.rb
matallo: uhm, I need it for `user.isOrgAdmin()` in the front

saleiva:
bump!

matallo:
I'll try to make it fit while work on other feature

matallo:
the bug is solved but I cannot invest time in fixing it properly, as per conversation with @xavijam I'm closing the PR but leaving the branch open

Issue #5052, opened 2015-08-17
Design a proposal for a step by step onboarding in the loading state (or after)


saleiva:
cc/ @mariodelvalle

PR #5506, opened 2015-09-14
Allowing drag&drop a file to a map so it's added as a layer

### LINKED ISSUE #4955: allow to drag a file to a map so it's added as a layer
cc @saleiva

xavijam:
And, of course, if the user can't add more layers to the map:

screen shot 2015-09-14 at 17 19 24

Pretty similar to one we have when user reaches his/her quota limits.

REVIEWER: @alonsogarciapablo.
Fixes #4955.


javisantana:
that upgrade is also show in onpremise? maybe in that case we should not say that, just a you reached maximum number of layers

xavijam:
ok, well spotted.

xavijam:
@javisantana, it checks if there is an upgrade_url, an in onpremise I feel there is no upgrade url, so it won't show the CTA dialog.

lib/assets/javascripts/cartodb/common/dialogs/map/add_layer_view.js
```diff
@@ -41,7 +41,7 @@ module.exports = BaseDialog.extend({
   },
 
   _initBinds: function() {
-    this.model.bind('addLayerDone', this.hide, this);
+    this.model.bind('addLayerDone', this.close, this);
```
xavijam: ```hide``` function doesn't trigger 'closeDialog' event and it provokes undesired behaviours :(. For example: mamufas dialog don't know if it has to be enabled again...

lib/cartodb/import_error_codes.rb
```diff
@@ -226,6 +226,11 @@ module CartoDB
       what_about: "The columns you have chosen don't have the same column type in both tables. Please change the types so the columns will have the same type and try again.",
       source: ERROR_SOURCE_USER
     },
+    8005 => {
```
xavijam: @CartoDB/backend adding in advance a possible import error, when user has reached his/her layers limit per map.

lib/assets/javascripts/cartodb/table/table.js
```diff
@@ -313,6 +320,14 @@ $(function() {
 
     _initBinds: function() {
       cdb.god.bind('geocodingChosen', this._onGeocodingChosen, this);
+      cdb.god.bind('dialogOpened', function() {
```
xavijam: Mamufas can't be enabled when there is another dialog opened.

lib/assets/javascripts/cartodb/common/dialogs/delete_layer/delete_layer_view.js
```diff
@@ -54,11 +65,12 @@ module.exports = BaseDialog.extend({
     this.model.destroy({
       wait: true,
       success: function() {
-        self.clean();
+        self.close();
```
xavijam: ```clean``` function doesn't trigger dialogClosed event and it provokes undesired behaviours :(. For example: mamufas dialog don't know if it has to be enabled again...

xavijam:
@alonsogarciapablo could you take a look please ;)

lib/assets/javascripts/cartodb/editor/background_polling_view.js
```diff
@@ -0,0 +1,48 @@
...
+      return false;
+    }
+
+    // Check if user can't add more layers to the map
+    // (erroring upload before sending the request)
+    var dataLayers = this.vis.map.layers.getDataLayers();
+    var maxLayers = this.user.getMaxLayers();
+
+    if (dataLayers.length >= maxLayers) {
+      _.extend(
```
xavijam: Not sure about how to "trigger" this error, should it be upload process? import process?

lib/assets/test/spec/cartodb/common/dialogs/delete_column/delete_layer_view.spec.js
```diff
@@ -1,62 +0,0 @@
-var DeleteColumnView = require('../../../../../../javascripts/cartodb/common/dialogs/delete_column/delete_column_view');
```
xavijam: Duplicated with delete_column_view.

lib/assets/javascripts/cartodb/common/background_polling/models/imports_model.js
```diff
@@ -75,20 +75,21 @@ module.exports = cdb.core.Model.extend({
   },
 
   _checkStatus: function() {
+    if (!this.upl.isValid()) {
+      // Upload is invalid so we need
```
alonsogarciapablo: I know you just moved this code, but I find this comment a bit confusing... Not sure what "we need checking init attributes" means...

xavijam: When you create a new import, it could come with the upload attributes already set. When you initialize an upload it checks if it is valid or not from the beginning. So that's the reason to check from the beginning if it is ok.

lib/assets/javascripts/cartodb/common/views/layers_error_details_upgrade_template.jst.ejs
```diff
@@ -0,0 +1,28 @@
+<div class="Dialog-header ErrorDetails-header">
```
alonsogarciapablo: Why aren't we re-using [the existing template](https://github.com/CartoDB/cartodb/blob/ac5b93b343539b575ff1d53dfa8c33b0efcb93fa/lib/assets/javascripts/cartodb/common/dialogs/limits_reach/limits_reached.jst.ejs)?

lib/assets/javascripts/cartodb/common/background_polling/models/imports_model.js
alonsogarciapablo: I was talking about the comment. It doesn't clarify anything to me.

lib/assets/javascripts/cartodb/common/views/layers_error_details_upgrade_template.jst.ejs
xavijam: Becuase I wanted to follow the same behaviour we have when a user tries to import a dataset and it fails because it can't fit in his/her account quota. Also, not sure if that template/dialog is a good representation of what happened when the user tried to add a new layer... I don't know, what do you think @josecruz ?

lib/assets/javascripts/cartodb/editor/background_polling_view.js
alonsogarciapablo: I have the feeling that this logic shouldn't be here. @xavijam what do you think about moving it [here](https://github.com/CartoDB/cartodb/blob/8a9c38746557a2f6e759519415c5e42d98556992/lib/assets/javascripts/cartodb/common/background_polling/background_polling_model.js#L80)? We could remove this class and this limitation will be enforced everywhere...

lib/assets/test/spec/cartodb/common/dialogs/delete_layer/delete_layer_view.spec.js
```diff
@@ -37,12 +37,16 @@ describe('common/dialogs/delete_layer/delete_layer_view', function() {
...
       });
 
-      it('should clean this view', function() {
-        expect(this.view.clean).toHaveBeenCalled();
+      it('should close this view', function(done) {
+        var self = this;
+        setTimeout(function() {
+          expect(self.view.close).toHaveBeenCalled();
+          done();
+        }, 100);
```
alonsogarciapablo: Some tests work just fine with `setTimeout(function(){.. }, 0)`. We could try here and save some milliseconds...

lib/assets/javascripts/cartodb/common/background_polling/models/imports_model.js
xavijam: Shall i remove it?

lib/cartodb/import_error_codes.rb
xavijam: do you agree then?

xavijam:
@salevia, I feel I have to ask you some things that @alonsogarciapablo has asked to me:

1. If user is in viewing a dataset (only one layer), mamufas should be disabled right?
---
2. When the "upload" is stopped, the button to show the details should open the "limits plan dialog":

6da50190-3bd9-11e5-8707-0f608e91f702

 or the error + upgrade one:

3b85dd48-5b09-11e5-8c4f-6bd1fc510343
---
3. Do we want to expose this functionality somewhere?

josecruz:
I agree with @xavijam 's design instead of the layer limit modal that we launch when clicking on add layer button.

- Do we want to expose this functionality somewhere?
I have this ticket pending #5052 where we can show that kind of features.

xavijam:
@alonsogarciapablo moved the logic to the background polling model (related to editor, not the general one).

lib/cartodb/import_error_codes.rb
xavijam: ping @kartones @juanignaciosl, last call :D.

alonsogarciapablo:
:clap:

PR #5522, opened 2015-09-15
Disable the row count and table size update because is really inefficient

ethervoid:


ethervoid:
please @juanignaciosl check this PR

Issue #5524, opened 2015-09-15
Add "application/vnd.ms-excel;charset=UTF-8" as a valid downloader content type


iriberri:

javisantana:
where?

iriberri:
In the URL downloader: https://github.com/CartoDB/cartodb/edit/master/services/importer/lib/importer/downloader.rb

@Kartones told me to add it directly, so I'm right now adding it to the XLS format array, if that's okay.

iriberri:
For more info:

If the content-type doesn't match completely (which doesn't in the case of charsets), the format guessing is based in the name of the resource according to the URL. In the specific scenario that failed, the URL is not explanative enough to make it work and the format ".do" is being guessed from it.

PR #5525, opened 2015-09-15
Allows charset parameter in content type for utf-8 cases

iriberri:
and trims empty spaces to compare the Content-Type header, as according to [RFC 2068](http://www.w3.org/Protocols/rfc2068/rfc2068.txt) it could present spaces. 

Sample URL which is suffering this issue: http://geoportalgasolineras.es/descargarPreciosTodasEstaciones.do?tipoBusqueda=0

Fixes  https://github.com/CartoDB/cartodb/issues/5524
cc @Kartones :-)

### DIFF
--- a/services/importer/lib/importer/downloader.rb
+++ b/services/importer/lib/importer/downloader.rb
@@ -45,15 +45,15 @@ module CartoDB
 
       CONTENT_TYPES_MAPPING = [
         {
-          content_types: ['text/plain', 'text/plain;charset=utf-8'],
+          content_types: ['text/plain'],
           extensions: ['txt']
         },
         {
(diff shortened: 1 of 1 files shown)


Kartones:
:+1:

ethervoid:
We should strip [here](https://github.com/CartoDB/cartodb/blob/master/services/importer/lib/importer/downloader.rb#L283) all the unnecessary charsets or properties from the content-type to avoid this kind of situations.

iriberri:
Thanks @ethervoid, without any doubt a much more clean solution. Let me work on this as soon as mi Vagrant is alive again.

PR #5566, opened 2015-09-17
Using the wrong vis model for create-vis-first dialog

### LINKED ISSUE #5563: Layer disappears from map on Visualize click if layer was added via "OK, Create Map" button
#### Expected:
"Publish" button should appear after you've clicked "OK, Create Map" when trying to add a second layer to a dataset.

#### Happening:
"Visualize" button appears instead of "Publish". If you hit "Visualize" you're prompted to create a map again. If you do, the layer you just added disappears.

To replicate:
1. From within a dataset click the plus button
2. On "A map is required to add layers" page click "OK, Create Map" button
3. From next "Add a new layer" page select a dataset and click "Add Layer" button 
[both layers are in map but "Visualize" button is visible instead of "Publish" button
4. Click "Visualize"
5. Click "OK, Create Map" button again b/c you've been taken to "A map is required to add layers" page
[Layer you just added has disappeared but now "Publish" button is visible]

https://github.com/CartoDB/cartodb/issues/5562

xavijam:
Being in a dataset, if you try to add a new layer, the "create a map first" dialog will appear. The model of the dialog is turned into a derived visualization, but unfortunately this model is the wrong one.

This is another example of the problems we have with ```master_vis``` and ```vis```. 
I'm really considering to remove odyssey code preventing more bugs in the future.
In any case, what would you do to prevent this error happening again?
cc @CartoDB/frontend 

REVIEWER: @viddo 
Fixes #5563

### DIFF
--- a/lib/assets/javascripts/cartodb/table/actions_menu.js
+++ b/lib/assets/javascripts/cartodb/table/actions_menu.js
@@ -233,7 +233,7 @@
         var createVisFirstView = new cdb.editor.CreateVisFirstView({
           clean_on_hide: true,
           enter_to_confirm: true,
-          model: this.vis,
+          model: this.master_vis,
           router: window.table_router,
           title: 'A map is required to add layers',
           explanation: 'A map is a shareable mix of layers, styles and queries. You can view all your maps in your dashboard.',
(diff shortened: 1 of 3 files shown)


viddo:
>  what would you do to prevent this error happening again?

Change looks legit, but is there some way to verify that we're using the "correct" vis? E.g. test assertion?

+1 to remove the Odyssey integration if it's not planned to release it finally.

xavijam:
Yeah, I was asking for something like that.

PR #5573, opened 2015-09-17
Adds X-Frame-Options for embedded maps

gfiorav:
Fixes [#99](https://github.com/CartoDB/cartodb-platform/issues/99)

Please CR @Kartones :)

cc @zenitraM @javisantana

### DIFF
--- a/app/controllers/admin/admin_controller.rb
+++ b/app/controllers/admin/admin_controller.rb
@@ -1,0 +1,8 @@
+class Admin::AdminController < ApplicationController
+  before_filter :x_frame_options_deny
+
+  protected
+  def x_frame_options_deny
+    response.headers['X-Frame-Options'] = 'DENY'
+  end
+end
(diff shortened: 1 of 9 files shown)


Kartones:
:+1:

gfiorav:
@Kartones like we spoke, I extended securization to all `/admin`. Please give a final :+1: so I can deploy! Thanks! :smiley:

xavijam:
Probably arriving late to this party, but, what about redirecting those public_maps embeded pages to the embed_map page?

gfiorav:
@xavijam AFAIK this can only be detected in the client side. Maybe we could create another ticket for the frontend to implement the redirection on public maps upon detection of being embedded? That would save all the grey boxes

xavijam:
Ok :+1:, in my opinion we should do that before merging this thing, shouldn't we?

Kartones:
:+1: to the code. 

And yep, this is for client-side, especifically for the browsers, we cannot control serverside if a page is inside a frame :disappointed:

xavijam:
Go ahead, no more to say.

PR #5588, opened 2015-09-18
Adds rake tasks to delete users by username or email

### LINKED ISSUE #5587: Create rake task to drop a user
cc @javisantana

gfiorav:
Fixes #5587

Now available:

```
bundle exec rake user:deletion:by_username[‘manolo']
bundle exec rake user:deletion:by_email['manolo@cartodb.com']
```

Please CR @rafatower :)

cc @javisantana

### DIFF
--- a/lib/tasks/user.rake
+++ b/lib/tasks/user.rake
@@ -5,7 +5,9 @@ namespace :user do
       raise 'Please specify the username of the user to be deleted' if args[:username].blank?
       
       user = User.find(username: args[:username])
-      raise "The username '#{args[:username]}' does not correspond to any user"
+      raise "The username '#{args[:username]}' does not correspond to any user" if user.nil?
+
+      raise 'Deletion aborted due to bad confirmation' if !deletion_confirmed?(user)
 
       user.destroy
     end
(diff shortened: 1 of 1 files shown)


ethervoid:
As @miguelarias said in a email, we should take care of things like if the user has tables, visualizations or this kind of stuff.

@CartoDB/backend

gfiorav:
@ethervoid I took the `UserController`'s code for deletion as a reference:

```ruby
def delete
    deletion_password_confirmation = params[:deletion_password_confirmation]
    if !@user.validate_old_password(deletion_password_confirmation)
      raise PASSWORD_DOES_NOT_MATCH_MESSAGE
    end

    @user.delete_in_central
    @user.destroy
    cdb_logout

    if Cartodb::Central.sync_data_with_cartodb_central?
      redirect_to "http://www.cartodb.com"
    else
      render(file: "public/404.html", layout: false, status: 404)
    end
  rescue CartoDB::CentralCommunicationFailure => e
    Rollbar.report_message('Central error deleting user at CartoDB', 'error', { user: @user.inspect, error: e.inspect })
    flash.now[:error] = "Error deleting user: #{e.user_message}"
    render 'account'
  rescue => e
    CartoDB.notify_exception(e, { user: @user.inspect }) unless e.message == PASSWORD_DOES_NOT_MATCH_MESSAGE
    flash.now[:error] = "Error deleting user: #{e.message}"
    render 'account'
 end
```

I thought that the `user.destroy` code would be prepared to delete dependent on the user. Can you provide insight to what other actions are carried out when a user deletes an account? 

Your help would be greatly appreciated :)

rafatower:
:+1: 

@ethervoid & @gfiorav logical deletion and this rake  are separated issues. This rake is intended to be executed by someone who knows what he's doing and the logical deletion or any other approach will be taken as a project/feature in its own. See https://github.com/CartoDB/cartodb/issues/5578

Kartones:
- User deletion should take care of everything related with the user, else it's a bug and should be fixed at `User.delete`, indeed
- As there is no soft delete for now, we should do just a "dry run" option in the rake to see what will be deleted before deleting (and also output something when deleting the user)
- I'd made the rake explode if there is no username specified ;)

gfiorav:
@rafatower @ethervoid 

That settles it I guess. Merging. 

Thanks for the comments and CR :smiley:

ethervoid:
I think we should have warnings or confirmations in the rake prior to the deletion. We should warn the task executor about that the user is a paid user, or have vis and tables, etc and expect confirmation. My two cents here

gfiorav:
@ethervoid you have a point certainly! Since @javisantana asked for this rake task maybe he could intercede on wether to leave it as is or check for dependencies and ask for confirmation.

gfiorav:
@Kartones You're right, the rake task isn't to failsafe. Now it's corrected, thanks :)

gfiorav:
Now there is a dry run as per @Kartones and @ethervoid suggestion:

```
You are about to delete the following user:
        > username: manolo
        > email: manolo@cartodb.com

Alongside ‘manolo', the following will be deleted:
        > 6 maps
        > 5 datasets

Type in the user's username (manolo) to confirm:
```

Since the code has changed a lot, a re-CR would be optimal @rafatower :smiley: 
Sorry for the double trouble!

rafatower:
Looks much better :+1:

PR #5590, opened 2015-09-18
Removes &hellip; and truncates the text with CSS

javierarce:
This fix makes rid of the `&hellip;` HTML code and truncates the text using CSS. Users applying a filter to the New Zealand hill Taumatawhakatangihangakoauauotamateaturipukakapikimaungahoronukupokaiwhenuakitanatahu won't see the ugly symbol anymore.

screen shot 2015-09-18 at 12 17 09

cc: @xavijam


xavijam:
Do you think it is a good moment to add filters_panel to scss-lint exclude list?

javierarce:
I'm adding the CSS file to the ignore list.

xavijam:
:+1:

lib/assets/javascripts/cartodb/table/menu_modules/filters/selection.js
```diff
@@ -272,7 +271,7 @@
         status: status,
         column_type: this.model.get("column_type"),
         legend: this.model.escape('column'),
-        short_legend: this._cleanString(this.model.escape('column'), 15),
+        short_legend: this._cleanString(this.model.escape('column')),
```
xavijam: Coud ```cdb.Utils.stripHTML``` function work here?

javierarce: On a second look I think it's safe to remove the cleanString altogether. I think we were using cleanString not to get rid of the HTML (because it should be already sanitized) but to do the truncation.

xavijam: Much better!

xavijam:
:us:

xavijam:
Could we merge this one?

nhibchung:
Hello CartoDB team, I have the &hellip problem for Layer Selector when viewing my maps on a phone. Is there anyway you can fix this too?

PR #5610, opened 2015-09-21
Add groups items on sharing modal

### LINKED ISSUE #4851: Groups item on the sharing modal
@viddo edit: some planning

An user can:
- [x] see groups (that the user is member of) in sharing lists for both dataset and maps
  - [x] group name, 
  - [x] default avatar (since groups don't have avatars)
  - [x] show member count
    - [x] requires #5529 to return group users on grantables API too
- [x] search groups (by name) in sharing lists
- [x] grant access to whole organization, which should be reflected for all users
- [x] grant access to a group, which should be reflected on all users which are members of that group
- [x] not change inherited access, but see which ones are available (UI details in #5521)
  - [x] see which org/groups that the access is inherited from
  - [x] requires #5556 to return users groups on grantables' API 

Technicalities:
- use new `grantables` API endpoint to render a mixed list of users/groups, ordered by name alphabetically and ascending, see #5160
- use existing ACL structure/API endpoint for modifying access (now supports `group` type)

These use-cases have also been identified to be updated due to introducing groups, but since they're not blocking this issue or breaks they can postponed, to be solved in isolation:
- #5533 editor header, indicator for shared entities to account for groups
- #5534 change-privacy modal, shared-with info to account for groups
- #5575 delete vis (dataset/map), affected entities to take into account groups
- #5576 warn user if revoking access of a group that's using the visualization

viddo:
Fixes #4851 - see the issue for scope and details
screen shot 2015-09-18 at 19 34 06

Some noteworthy details regarding changes:
- Extracted fetch params organization to shared model that is "attached" on the collection (since it affects both `url` and `parse` methods), see the `PagedSearchModel`
- Produce models bundle through browserify instead. The models are still attached to the global `cdb.admin` namespace, but this allows using npm packages in models. Bonus: models now supports source maps and are visible in the dev console :tada: 
- Users that don't have groups yet will still use the new models/views/grantables API

@xavijam can you review?
cc @alonsogarciapablo since he reviewed previous PRs
cc @juanignaciosl


lib/assets/javascripts/cartodb/common/dialogs/change_privacy/share/grantables_view.js
```diff
@@ -0,0 +1,91 @@
...
+        return new UserDetailsView(_.defaults({
+          isUsingVis: _.any(usersUsingVis, function(u) { return u.id === grantableEntity.get('id'); })
+        }, defaultOpts))
+        break;
+      case 'group':
+        return new GroupDetailsView(_.defaults({
+        }, defaultOpts))
+        break;
+      default:
+        // cdb.log.error('No details view for grantable model of type ' + type);
```
xavijam: Let's remove this comment please ;).

xavijam: There is no org option here?

lib/assets/javascripts/cartodb/common/dialogs/change_privacy/share/permission_toggler_view.js
```diff
@@ -0,0 +1,82 @@
...
+
+  events: {
+    'mouseover .js-toggler.is-disabled': '_onHoverDisabledToggler',
+    'mouseout .js-toggler': '_closeTooltip',
+    'mouseleave .js-toggler': '_closeTooltip',
+    'change .js-input': '_onChangeInput'
+  },
+
+  initialize: function() {
+    _.each(['model', 'permission', 'hasAccess', 'canChangeAccess', 'toggleAccess', 'label'], function(name) {
```
xavijam: oh!

lib/assets/javascripts/cartodb/common/dialogs/change_privacy/share/grantables_view.js
viddo: This method is only called for grantables items, the organization option is not returned through that API, but added explicitly, see `_renderOrganizationPermissionView` in this same file.

lib/assets/javascripts/cartodb/common/dialogs/change_privacy/share/share_view.js
```diff
@@ -34,96 +32,66 @@ module.exports = BaseDialog.extend({
...
+      quote: randomQuote()
+    });
+    loadingView.appendToBody();
+
     var permission = this.options.vis.permission;
     permission.overwriteAcl(this.model.get('permission'));
     permission.save()
-      .done(function() {
-        self._openChangePrivacy();
+      .always(function() {
```
xavijam: why don't you like 'done' premise?

viddo: They are not the same; `done` is only called for a resolved promise, while `always` is also called when it's rejected.

lib/assets/javascripts/cartodb/common/views/paged_search/paged_search_view.js
```diff
@@ -179,7 +180,7 @@ module.exports = cdb.core.View.extend({
 
   _submitSearch: function(e) {
     var search = this.$('.js-search-input').val().trim();
-    this.organizationUsers
+    this.collection
```
xavijam: Why have you changed the name? Do you feel it is easier to manage it?IMO, we should start defining what are the models and collections in the description to have an idea what they are, what do you think?

viddo: `organizationUsers` is no longer accurate, since `collection` now can be either a `organizationUsers` or `grantables` colllection, so I figured `collection` was more accurate and follows Backbone's conventions. 
Regarding documentation/comments, I changed the class desc to _View to render a a searchable/pageable collection._ to reflect this, how would you like to phrase it to make it more clear?

viddo: Clarified in 54467bc

xavijam:
Lovely feature :clap: :clap:  :es:
@josecruz, i feel we need a new icon for groups, it is the same as organization and it is a bit confusing.

josecruz:
@xavijam I don't find any other icon that illustrates groups :(

PR #5613, opened 2015-09-21
Improved too many rows error message in the importer

### LINKED ISSUE #5502: Improve error message when importing over the row limit
Right now it says there are too many rows, and that's it. I'd suggest to complete a bit more the error (in the modal and the email as well)

cc/ @ethervoid  can you paste the actual test to @csobier so we find a better one?

ethervoid:
Closes #5502 


Kartones:
:+1:

Issue #5612, opened 2015-09-21
Arcgis import issue


rafatower:
S/B: 6671042

The import is failing but it is unclear the root cause. Here's a trimmed down trace:
```
...
true: Running in append mode
true: ogr2ogr output:          Warning 6: driver GeoJSON does not support open option X_POSSIBLE_NAMES
Warning 6: driver GeoJSON does not support open option Y_POSSIBLE_NAMES

true: ogr2ogr exit code:       0
true: Errored importing data from /tmp/imports/20150917-7490-1ux674k/somefile.json:
===LOG END===
true: ----------------------------------------------------
false: [backtrace]
true: ----------------------------------------------------
2015-09-17 18:45:39 UTC: Proceeding to register
2015-09-17 18:45:39 UTC: After importer run
2015-09-17 18:45:39 UTC: After dispatch
2015-09-17 18:45:39 UTC: ERROR!
===LOG END===
```

and here the relevant backtrace:
```
services/importer/lib/importer/loader.rb:319:in `check_for_import_errors'
services/importer/lib/importer/loader.rb:262:in `run_ogr2ogr'
services/importer/lib/importer/loader.rb:90:in `streamed_run_continue'
services/importer/lib/importer/runner.rb:214:in `streamed_loader_run'
services/importer/lib/importer/runner.rb:175:in `import'
services/importer/lib/importer/runner.rb:325:in `execute_import'
services/importer/lib/importer/runner.rb:256:in `block (3 levels) in single_resource_import'
services/importer/lib/importer/runner.rb:249:in `each'
services/importer/lib/importer/runner.rb:249:in `each_with_index'
services/importer/lib/importer/runner.rb:249:in `block (2 levels) in single_resource_import'
lib/cartodb/stats/aggregator.rb:45:in `block in timing'
lib/cartodb/stats/aggregator.rb:43:in `timing'
services/importer/lib/importer/runner.rb:248:in `block in single_resource_import'
lib/cartodb/stats/aggregator.rb:45:in `block in timing'
lib/cartodb/stats/aggregator.rb:43:in `timing'
services/importer/lib/importer/runner.rb:226:in `single_resource_import'
services/importer/lib/importer/runner.rb:95:in `run_import'
services/importer/lib/importer/runner.rb:88:in `block in run'
lib/cartodb/stats/aggregator.rb:45:in `block in timing'
lib/cartodb/stats/aggregator.rb:43:in `timing'
[...]

PR #5614, opened 2015-09-21
Remove [X|Y]_POSSIBLE_NAMES from params #5612

rafatower:
That option should be there just for CSV files with guessing enabled.
Although it just produces a warning with ret value 0 with other gdal
drivers other than csv_drv, remove it for housekeeping.

If you look below, the option was already included in the `guessing_option`.

@Kartones please review

cc/ @gfiorav


Kartones:
:+1:

rafatower:
I modified a couple of test related bits. Feel free to review again.

I need to give a pass through staging before deploying

rafatower:
and tested in staging, merging to master...

PR #5622, opened 2015-09-21
Fix negative geocoding quota in georeference modal

### LINKED ISSUE #5223: Negative geocoding quota in georeference modal
When `monthly_use` >  `quota`, which could happen when users can exceed the hard limit, a message like the following one showing negative credits left appears:

image

In case  `monthly_use` >  `quota`, I'd say that the best approach would be to show "0 credits left".

Re: https://github.com/CartoDB/cartodb/blob/d5a34998654eb01978935846405088c508a60f5c/lib/assets/javascripts/cartodb/common/dialogs/georeference/user_geocoding_model.js

viddo:
fixes #5223 
screen shot 2015-09-21 at 16 25 47

@xavijam can you review?
cc @iriberri


iriberri:
Awesome!

xavijam:
:+1:

PR #5625, opened 2015-09-21
Use right protocol with id previews 

javierarce:
This PR fixes the static map generation to use the right protocol depending on the page where the image is loaded.

### DIFF
--- a/lib/assets/javascripts/cartodb/common/views/mapcard_preview.js
+++ b/lib/assets/javascripts/cartodb/common/views/mapcard_preview.js
@@ -41,7 +41,7 @@ module.exports = cdb.core.View.extend({
 
   _loadFromVisId: function() {
 
-    var protocol = this._isHTTPS ? 'https': 'http';
+    var protocol = this._isHTTPS() ? 'https': 'http';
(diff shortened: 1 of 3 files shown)


lib/assets/javascripts/cartodb/common/views/mapcard_preview.js
```diff
@@ -41,7 +41,8 @@ module.exports = cdb.core.View.extend({
 
   _loadFromVisId: function() {
 
-    var url = 'http://' + this.options.username + '.' + this.options.account_host + '/api/v1/map/static/named/' + this._generateImageTemplate() + '/' + this.width + '/' + this.height + '.png';
+    var protocol = this._isHTTPS() ? 'https': 'http';
+    var url = protocol + '://' + this.options.username + '.' + this.options.account_host + '/api/v1/map/static/named/' + this._generateImageTemplate() + '/' + this.width + '/' + this.height + '.png';
```
xavijam: Couldn't we use '//whatever_in_url'?

javierarce: Yes, but since we were already using that method for the previews with vizjson I thought it would make more sense.

xavijam: whatever you decide is :+1:.

xavijam:
:us:

PR #5626, opened 2015-09-21
Store invalidation client with unique name

luisbosque:
When the invalidation function stores the client in a python global var,
store it under a unique name based on the invalidation service host and
port

@rochoa please review


javisantana:
do we have different service ip/port?

luisbosque:
If you change the IP or port of the service, in order to get the changes instantly the new client must not be cached.

PR #5627, opened 2015-09-21
Add the missing trace of ogr2ogr falures #5612

rafatower:
@Kartones mini-review

This is needed to confirm the error I can reproduce locally is the same that the one hitting production. Otherwise the output of ogr2ogr is lost.


rafatower:
@ethervoid can you please review?

ethervoid:
Seems good to me :+1:

Kartones:
:+1:

PR #5628, opened 2015-09-21
Exclude shared visualizations from the user feed

ethervoid:
We want to exclude the shared maps and datasets from the user feed


javierarce:
:cool:

javierarce:
@ethervoid: have a look, please.

ethervoid:
done @javierarce

PR #5632, opened 2015-09-21
Allows to specify a host in the AuthenticatedUser

javierarce:


javierarce:
@xavijam: can you review this, please?

lib/assets/javascripts/cartodb/public/authenticated_user.js
```diff
@@ -12,7 +12,11 @@ cdb.open.AuthenticatedUser = cdb.core.Model.extend({
   },
 
   url: function() {
-    return "//" + window.location.host  + "/api/v1/get_authenticated_users"
-  }
+    var host = this.get('host') ? this.get('host') : this._getCurrentHost();
+    return "//" + host + "/api/v1/get_authenticated_users";
```
xavijam: I feel this won't work with users under organization, because if you are in:

https://team.cartodb.com/u/xavijam/maps
window.location.host will be team.cartodb.com...  and you need /u/xavijam...

xavijam: Forget my last comment, but not sure if we should use common-data as host, what do you think @javisantana ?

javisantana: I guess you did this because of the redirection. My main concern is for onpremise but this is really tied to cartodb.com ...

javierarce: Yes, the custom option is meant to be used in the explore page. The normal behaviour is the same we already have.

javierarce:
Is it ok to merge, @xavijam?

xavijam:
Yeah, go ahead!

PR #5688, opened 2015-09-23
Fix dependency bundles due to new groups models

viddo:
Reverts the models-bundle-through-browserify in a0ffc723edde48476ca69bedcecea38fce18e480, since models are not motivated to be browserified anymore. 

Also add new files that were missing in dependency bundles that broke public pages (cc @juanignaciosl for spotting this one).

@xavijam can you please review? Added the missing dependencies for user and organization models that have special bundles on public pages. 
Considering the amount of duplicated files on the public bundles there I'd recommend we DRY these ones up to have a common bundle, it doesn't make sense to have a separate bundle per page if the files are basically the same.

### DIFF
--- a/lib/assets/javascripts/cartodb/models/assets.js
+++ b/lib/assets/javascripts/cartodb/models/assets.js
@@ -1,13 +1,11 @@
-if (typeof module !== 'undefined') {
-  module.exports = {};
-}
+
 
   /*
    *  User asset Model
    */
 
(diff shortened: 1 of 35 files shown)


xavijam:
Agree with you @viddo. Separate bundles per each page is nice, but if we have several repeated bundles is a synonym that something is not ok, yep.

xavijam:
In any case :+1:, and great to have found the issue on time.

PR #5699, opened 2015-09-23
Workaround for append mode PK issues #5612

rafatower:

This is needed because of differences in behavior between ogr gdal 1.11.0 and ogr v2.0+. An explanation of why this used to work is in the ticket, from this [comment](https://github.com/CartoDB/cartodb/issues/5612#issuecomment-142562236) on.


Kartones:
:+1:

PR #5700, opened 2015-09-23
Multiple additions to notify trending

ethervoid:
* Reapply the notify trending deploy because of the failed revert made 2 days ago
* Added simulation message to now what mode are we using

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -87,6 +87,7 @@ class UserMailer < ActionMailer::Base
 
   def trending_map(visualization, mapviews, visualization_preview_image)
     @user = visualization.user
+    @mapviews = mapviews
     @map_name = visualization.name
     @preview_image = visualization_preview_image
     @subject = "Recent activity on one of your maps!"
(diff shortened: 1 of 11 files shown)


ethervoid:

app/mailers/user_mailer.rb
```diff
@@ -85,4 +85,18 @@ def table_liked(canonical_visualization, viewer_user, visualization_preview_imag
          :subject => @subject
   end
 
+  def trending_map(visualization, mapviews, visualization_preview_image)
+    @user = visualization.user
+    @bcc = "subscriptions@cartodb.com"
```
Kartones: This must be a config param read from `Cartodb.config`, remember opensource & custom installs...

Kartones: I'd also implement either a sampling rate or a semaphore (with the PlatformLimits should be easy) to not spam and only send 1 of each X maps, but that's pure personal taste, for starting is ok as it is.

Kartones:
Apart from the hardcoded email, everything else is fine, so change that and you have my :+1:

ethervoid:
Removed that part in order to deploy this asap because we need it working

PR #5702, opened 2015-09-23
Fix bug trying to change users permission

### LINKED ISSUE #5676: Bug trying to change users permission
### STR:
- Having a user within an organization and belonging to a group.
- Open sharing options from a dataset.
- If you try to change the permissions over a group, it works weird.
   - Disabling WRITE permission, group members still have write permissions.
   - And if we try to remove READ permission, it disables both READ and WRITE checks.

groups-bug

Found by @urbanphes

viddo:
Don’t allow revoking access if it’s inherited, effectively only allow to modify the item’s access if it’s adds a more privileged access on top of the inherited one.

fixes #5676 

screen shot 2015-09-23 at 13 10 23
screen shot 2015-09-23 at 13 10 34
screen shot 2015-09-23 at 13 11 04

@xavijam can you review the code?
cc @urbanphes @juanignaciosl

### DIFF
--- a/lib/assets/javascripts/cartodb/models/permissions.js
+++ b/lib/assets/javascripts/cartodb/models/permissions.js
@@ -69,16 +69,24 @@ cdb.admin.Permission = cdb.core.Model.extend({
   },
 
   hasWriteAccess: function(model) {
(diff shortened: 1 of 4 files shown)


xavijam:
:+1:

PR #5709, opened 2015-09-23
Adds proper message for password to short

### LINKED ISSUE #2878: Password too short not raising error
At [User model](https://github.com/CartoDB/cartodb/blob/master/app/models/user.rb#L457), is not raising an error if password is too small.

It should directly raise a ValidationError or similar (as checking at validate method is not possible as password is already hashed).

gfiorav:
Fixes #2878 
Fixes [#33](https://github.com/CartoDB/cartodb-bloomberg/issues/33)

Since you were the last one to modify password validation and author of #2878, please CR @Kartones Thanks! :smiley:


app/models/user.rb
```diff
@@ -577,7 +577,10 @@ def self.make_token
   end
 
   def password=(value)
-    return if !value.nil? && value.length < MIN_PASSWORD_LENGTH
+    if !value.nil? && value.length < MIN_PASSWORD_LENGTH
```
Kartones: Spec for this capturing the expected error, please (`user.errors.should ...`).

Also test that organization user signup (/signup) works (it should, so this means manually, no need for spec)

Kartones:
Once there's a test, :+1:

gfiorav:
@Kartones Tested organization login.

Could you CR the test and give a final :+1: ?

Thanks!

spec/requests/signup_controller_spec.rb
```diff
@@ -75,6 +75,18 @@
...
     end
 
+    it 'triggers validation error is password is too short' do
+      user = User.new
+
+      user.username = 'testusername'
+      user.email = 'manolo@escobar.es'
+      user.password = 'short'
+      user.password_confirmation = user.password
+      user.database_host = ::Rails::Sequel.configuration.environment_for(Rails.env)['host']
```
Kartones: is this line needed?

Kartones:
:+1:

gfiorav:
@Kartones removed unnecessary line, will merge as soon as CI test pass

PR #5712, opened 2015-09-23
data library index page

matallo:
- [x] load cards (using Vis collection)
- [x] implement more
- [x] implement filters
 - [x] use Leaflet with hovers for the header country navigation
- [x] add first tests
- [x] implement preview
- [x] implement search
- [x] finish tests


matallo:
started the data library index page, inspired by the new explore, I came across the first questions

screen shot 2015-09-23 at 16 18 58

I wrote the filter links as they are in the dashboard, but had to add a new class for the styles, they live outside the list and item in the explore template, though

we should unify this, worth crating an issue?

app/assets/stylesheets/common/filters.css.scss
```diff
@@ -220,6 +227,12 @@ a.Filters-cleanSearch { line-height:32px }
     text-decoration: none;
   }
 }
+.Filters-typeLink.Filters-typeLink--grey {
+  color: #979ea1;
+  &:hover {
+    color: #979ea1;
```
matallo: ಠ_ಠ why?

@CartoDB/frontend I feel more comfortable writing hex color in lowercase, is this open for discussion?

fdansv: how dare you

app/assets/stylesheets/public_data_dashboard/public_data_dashboard.css.scss
```diff
@@ -0,0 +1,31 @@
...
+  margin-top: 65px;
+}
+.DataLibrary-columns * {
+  -moz-box-sizing: border-box;
+  -webkit-box-sizing: border-box;
+  box-sizing: border-box;
+}
+.DataLibrary-text {
+  color: #333;
+  font-weight: 200;
```
matallo: from the shorthand I prefer:
- font-weight
- font-size
- line-height

@CartoDB/frontend

matallo:
summoning @CartoDB/frontend here, per @xavijam request as I don't want to get off track from the decisions originally discussed with @javisantana, also I applied some of the implementations that have already been done for the explore, thinking of the filtering, ordering and load more (to be implemented).

the `/data-library` resides in datasets view of common-data user, and we show a different layout/view with a feature_flag

https://github.com/CartoDB/cartodb/pull/5712/files#diff-07eeed666445b4ccdc97b529e78e6998R239

thinking about it, and despite this was decided with @javisantana, maybe it's not the best solution (I'm pretty sure it isn't) as this leaves us in a position where data-library is enabled for all users or none, instead of progressively open the feature to users. Thus, we could have a new route as it's been done with explore (correct me if I'm wrong @javierarce) and make proper use of feature flag.

app/assets/stylesheets/public_data_dashboard/icons.css.scss
```diff
@@ -0,0 +1,25 @@
+// Icons
+// --------------------------------------------------
+
+.Icon {
+  display: inline-block;
+  width: 60px;
+  height: 60px;
+  border: 2px solid rgba(#9b9b9b, .4);
```
matallo: I prefer not to use the zero for < 1 decimals, also open for change but worth discussing

matallo:
hey @xavijam I've already updated this with the new controller

lib/assets/javascripts/cartodb/data_library/view.js
```diff
@@ -0,0 +1,157 @@
...
+    this.loader = ViewFactory.createByTemplate(this.loaderTemplatePath, {
+      title: this._LOADER_TITLE,
+      quote: randomQuote()
+    });
+
+    this.$el.append(this.loader.render().$el);
+  },
+
+  _fetchItems: function(reset) {
+    debugger;
```
xavijam: hello debugger

xavijam:
I feel we should remove likes from those cards because there is no user behind it, and you don't need to like it in order to have them in your "likes" section.  @saleiva @carlostallon

xavijam:
IMHO:

- Loading template shouldn't be duplicated, only for adding a new class in the parent div. I'd try to use the same we have in common and make it customizable.
- Review (and probably remove) likes functionality, there are several changes from explore behaviour.
- You don't need this, because there is no big avatar as we have in user profile:
```js
$('.UserAvatar-img').wrap($('<a>',{
   href: userDashboardUrl
}));
```
- Be careful with the cards link, we already have an [issue](https://github.com/CartoDB/cartodb/issues/5761) in the new user profile, so please, take it into account.

lib/assets/javascripts/cartodb/data_library/view.js
```diff
@@ -0,0 +1,157 @@
...
+
+  _initModels: function() {
+    this.model = new cdb.core.Model({
+      type: 'table',
+      page: 0,
+      order_by: 'likes'
+    });
+
+    this.collection = new Visualizations();
+    this.collection.bind('reset', this._renderVisualizations, this);
```
xavijam: You have to add this.collection as a related model in order to be "unbinded" when the view is cleared.

```js
this.add_related_model(this.collection);
```

lib/assets/javascripts/cartodb/data_library/view.js
```diff
@@ -0,0 +1,157 @@
...
+  _onChangeShowLoader: function() {
+    if (this.model.get('show_loader')) {
+      this.loader.show();
+    } else {
+      this.loader.hide();
+    }
+  },
+
+  _onChangeVisCount: function() {
+    if (this.model.get('vis_count') >= this.collection.total_entries) {
```
xavijam: I'd use a ternary instead a if with else.

lib/assets/javascripts/cartodb/data_library/view.js
```diff
@@ -0,0 +1,157 @@
...
+      this.model.set({ page: 0, show_more: false, show_loader: true, show_mast: false });
+      this.$('.js-items').empty();
+    }
+
+    var type = this.model.get('type');
+    var orderBy = this.model.get('order_by');
+    var page = this.model.get('page');
+    var limit = this._LIMIT;
+
+    this.collection.fetch({ type: type, orderBy: orderBy, page: page, limit: limit, reset: reset });
```
xavijam:
```js
this.collection.fetch({
  property: value,
  property: value,
  ...
});
```
For easy reading?

lib/assets/javascripts/cartodb/data_library/view.js
```diff
@@ -0,0 +1,157 @@
...
+        datasetSize: this._getDatasetSize(item),
+        geomType: this._getGeometryType(item.get('geom_types')),
+        account_host: 'cartodb.com'
+      });
+
+      this.$('.js-items').append(el);
+
+      var $item = this.$('.js-items').find('[data-vis-id="' + item.get('id') + '"]');
+
+      if (item.get('type') === 'derived') {
```
xavijam: Why do you need this?

lib/assets/javascripts/cartodb/data_library/view.js
viddo: Why not?
```js
this.model.set({ 
  show_more: this.model.get('vis_count') < this.collection.total_entries
});
```

The ternary operator makes more sense when one wants a return value that is not a boolean, e.g. `var aString = isSomething ? 'this' : 'orThat'`

xavijam: Yeah, agree, ternary or that option you have just explained ;).

matallo:
+1 to remove likes in the data-library

matallo:
> Let's avoid to use 'view.js' names because it is more difficult to find desired files when there are several 'view.js', so better to use data_library_view.js and data_library_template.js.

Agree! What about the `entry.js`?

xavijam:
```Entry.js``` for me is ok, because there could be 5-7 files, no more. But imagine the number of views...

carlostallon:
the "like" maybe are not necessary for a data library, so go for it

xavijam:
ok, thanks @carlostallon

viddo:
> @matallo: Agree! What about the entry.js?
> @xavijam: Entry.js for me is ok, because there could be 5-7 files, no more. But imagine the number of views...

IMHO it would help to use the same name as the bundle it represents, and place it in `lib/assets/javascripts/cartodb/`. There are already a few of those there, we could do that for new entry points from now on at least.

matallo:
> IMHO it would help to use the same name as the bundle it represents, and place it in lib/assets/javascripts/cartodb/

*___* music to my ears

xavijam:
Up to you!

jesusvazquez:
skip ci

jesusvazquez:
retest this please

matallo:
retest this, please

PR #5739, opened 2015-09-24
Fix Explore page

javierarce:
This PR enables CORS to make the requests to the Explore page.

Changes in frontend:

- Adds CORS to Backbone's sync method.
- Removed the hardcoded endpoints.
- Updated the URLs for the Explore endpoints.


javierarce:
@xavijam: can you review the frontend part?

xavijam:
Checking!

xavijam:
BTW, checking the code of the explore with @matallo, we discover the use of ```cdb.Backbone.Model```, what is that?

lib/assets/javascripts/cartodb/app.js
```diff
@@ -51,6 +51,23 @@ var i18n = {
     return Backbone.originalSync(method, model, options);
   };
 
+  Backbone.currentSync = Backbone.sync;
+  Backbone.withCORS = function(method, model, options) {
```
xavijam: :+1:

lib/assets/javascripts/cartodb/explore/entry.js
```diff
@@ -67,7 +68,9 @@ $(function() {
 
     this.exploreModel.fetch({ success: onSuccess });
 
-    var authenticatedUser = new cdb.open.AuthenticatedUser({ host: 'common-data.cartodb.com'});
+    var authenticatedUser = new cdb.open.AuthenticatedUser({ host: cdb.config.get('explore_user') + '.' + cdb.config.get('account_host')});
+
+    authenticatedUser.sync = Backbone.withCORS;
```
xavijam: Not sure about this, what do you think @viddo?

javierarce:
It's a mistake. It should be `cdb.core.Model`. I'll fix it.

xavijam:
Thanks man!

lib/assets/javascripts/cartodb/explore/entry.js
viddo: From what I can see the only thing that changes compared to the default sync is the options argument, which is passed from the `fetch` method, so I'm not convinced either.

If this model _always_ should use CORS it should be added internally in the class definition of `cdb.open.AuthenticatedUser` i.e. override sync internally and avoid the monkey-patch here. Or encapsulate the CORS options somewhere so they can be used per use-case, such as `authenticatedUser.fetch(Backbone.withCORS({ other: 'options', … })); // adds the CORS settings to the given object`

xavijam: What if...:
```js
Backbone.ModelWithCors = {...}
```

viddo: Can you elaborate a little more what you have in mind, please?

xavijam: After a second though: 

```js
Backbone.Model = Backbone.Model.extend({
   sync: function(method, model, options) {
      if (options.withCORS) {
         ...
      }
   }
});
var model = new cdb.open.Authentication();
model.fetch({
   withCORS: true // by default it would be false?
});
```

xavijam: As if it is only a sync change, we could activate when we need it, just in sync operations, I don't know, ^^^ this thing sounds good to me.

viddo: That would work, but I'm a bit reluctant about adding more customizations inside Backbone, especially for things that aren't common/default behavior.

javierarce: Thanks! I'll refactor this in another ticket

ethervoid:
Please @Kartones check this PR (specially the CORS part)

app/controllers/application_controller.rb
```diff
@@ -82,13 +83,35 @@ def set_asset_debugging
       (Cartodb.config[:debug_assets].nil? ? true : Cartodb.config[:debug_assets]) if Rails.env.development?
   end
 
+  def cors_preflight_check
+    if request.method == :options && check_cors_headers_for_whitelisted_referer
+      response.headers['Access-Control-Allow-Origin'] = request.headers['origin']
+      response.headers['Access-Control-Allow-Methods'] = 'GET, POST, DELETE'
+      response.headers['Access-Control-Allow-Headers'] = 'Content-Type'
+      response.headers['Access-Control-Max-Age'] = '172800'
```
rafatower: 48 hours?

rafatower:
:+1:

PR #5740, opened 2015-09-24
4730 ogr geom possible names

rafatower:


rafatower:
@javisantana, there's something I'm having trouble to understand. In this ticket I'm moving some functionality from rails to ogr2ogr. By running tests I found a difference in behavior that concerns me:

Basically when importing a csv with lat/long I'm obtaining a table with `ST_MultiPoint` geometries instead of `ST_Point`. For this particular case it seems reasonable to use `ST_Point` instead, as every row will contain one and just one point, no need for multi-geometry.

But ogr2ogr is doing so because of the `PROMOTE_TO_MULTI` option (that option has been there since 58324e3e). But then in the "geometrify" process in rails, it seems as it hated multipoint, since 86464b56 (and geometry collections in general):

https://github.com/CartoDB/cartodb/blob/master/services/importer/lib/importer/georeferencer.rb#L82-L95

https://github.com/CartoDB/cartodb/blob/master/services/importer/lib/importer/georeferencer.rb#L318-L332

I guess the `PROMOTE_TO_MULTI` is needed for SHP and other sources containing multi-polygons and stuff like that.

Is there any trouble in leaving it as `ST_MultiPoint`? maybe for rendering? cc/ @rochoa 
I think that is the safest move.

rafatower:
I'll answer myself:
- I don't see any major issue in using multi-points, so I'll go for it.
- The ideal situation would require to have more information prior to running ogr2ogr, so that we do not have to "fix things" afterwards and keep the process fast. This would require an updated version of `ogrinfo` that we currently do not have. So I'd push harder on a solution for gdal versions.

rochoa:
Sorry, this slipped through the cracks. Just confirm there is no issue using multipoint for rendering.

rafatower:
I created this issue to deal with ogr2ogr crashes: https://trac.osgeo.org/gdal/ticket/6137

rafatower:
please do not delete that branch yet.

rochoa:
Update: I missed the torque use case, torque renderer does not support multipoint. Sorry I didn't realise about it.

PR #5743, opened 2015-09-24
Adds cast in case column was not typed as string

### LINKED ISSUE #5595: the_geom guessed as double instead of Geometry
Here's the input file: https://gist.githubusercontent.com/rafatower/f2b6c2918a66749d5898/raw/5a345f34cb57c41f506af54a335f9389dbfc69f0/untitled_table_3%2520(2).csv

And here's the failure:
```
true: Errored importing data from /tmp/imports/20150918-26933-1dz6hjn/untitled_table_3%25202.csv:
false: Sequel::DatabaseError: PG::Error: ERROR:  invalid input syntax for type double precision: ""
LINE 4:           AND the_geom != ''
                                  ^

```
It happens around the georeferencer.rb and column.rb classes.

gfiorav:
Fixes #5595

Really, it only attempts to make the importer more robust in case of column mistyping.

I tried different things but this seemed the easiest way:
```
SELECT #{column_name} FROM "#{schema}"."#{table_name}"
  WHERE #{column_name} IS NOT NULL
  AND #{column_name}::text != ''
  LIMIT 1
```
+ **In case the column's type is a string**: casting `::text` doesn’t hurt
+ **In case it's a numerical type**: the only check we want is for NULL (since a number can’t be empty) so casting does nothing.

Could you provide some insight on this, @rafatower ? Thanks!

*Update*: Just realised this might cause problems if there is a null (since `NULL::text` raises syntax error). What do you think about that?


rafatower:
Works on my machine (tm): :smile: 
```sql
select NULL::text as c;
 c 
---
 
(1 row)
```

:+1:

rafatower:
If this solves the import crash for the moment, I'm all for it despite the overlap

gfiorav:
@rafatower Ok, I'll merge it then!

rafatower:
Not without a test

spec/connectors/importer_spec.rb
```diff
@@ -200,7 +200,20 @@
 
     data_import.success.should_not eq true
   end
-end
 
+  it 'should be able to handle wrong type guessing' do
```
rafatower: The bug you're fixing is in `Column` class, that crashes unless fed with a text type. So this should go into https://github.com/CartoDB/cartodb/blob/master/services/importer/spec/unit/column_spec.rb

As things stand now, I can call `Column#records_with_data` from anywhere in the code and the fact that this is related to imports is circumstantial. 

ok this time, but keep in mind for the upcoming occasions

rafatower: rephrase it to `'should be able to handle the_geom with numeric types'`

rafatower: (forget about the deleted comment)

rafatower:
:+1:

gfiorav:
Ok, merging as soon as CI tests are good :+1:

PR #5750, opened 2015-09-24
Fix likes initialization in user feed and explore page

javierarce:
This PR fixes the faulty initialization of the likes in both the explore page and the user feed.

### DIFF
--- a/lib/assets/javascripts/cartodb/user_feed/view.js
+++ b/lib/assets/javascripts/cartodb/user_feed/view.js
@@ -21,7 +21,7 @@ module.exports = cdb.core.View.extend({
   },
 
   initialize: function() {
-    _.bindAll(this, '_initLike', '_onFetchError', '_onWindowResize', '_renderStaticMaps');
+    _.bindAll(this, '_initLike', '_fetchLike', '_onFetchError', '_onWindowResize', '_renderStaticMaps');
 
     this.maps = [];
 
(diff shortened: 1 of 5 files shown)


javierarce:
Please have a look, @xavijam

lib/assets/javascripts/cartodb/user_feed/view.js
```diff
@@ -263,23 +251,24 @@ module.exports = cdb.core.View.extend({
...
-      likeModel.url = url;
-      likeModel.sync = Backbone.withCORS;
-      likeModel.fetch();
-    }
+      this._fetchLike(likeModel, url);
+    } else {
+      var self = this;
+      this.options.authenticatedUser.bind('change', function() {
+        self._fetchLike(likeModel, url);
+      });
```
xavijam: If you add ```this``` at the end, you won't need the ```self```.

lib/assets/test/spec/cartodb/common/views/feed/feed.spec.js
```diff
@@ -74,10 +74,6 @@ describe('common/views/feed/view', function() {
     expect(this.view.$('.js-more').hasClass('is-hidden')).toBeTruthy();
   });
 
-  it('should init the likes', function() {
```
xavijam: We should add a new test checking if the authenticated user arrives, likes model should have a new url set.

javierarce: Ok

xavijam:
Thanks :+1:

PR #5754, opened 2015-09-25
Make a shared vis URL to have the correct format

viddo:
This takes the current user into account when generating the vis URL (similar to how it works for a dataset URL already), so the URL format is the expected `/user/<current-user-username>/vis/<owner-username>.<vis-id>` in the case of a shared map.

@xavijam can you review the code please

cc @Kartones @juanignaciosl for offline discussion


viddo:
deployed on ded03 if you want to review

Kartones:
Re-deployed and tested and both public and private shared map urls work fine now, so good to go from backend side

xavijam:
:+1:

PR #5756, opened 2015-09-25
Daily report for the trending maps

### LINKED ISSUE #5706: Add BCC support for the notify trending mail

ethervoid:
Closes #5706

### DIFF
--- a/app/mailers/reporter_mailer.rb
+++ b/app/mailers/reporter_mailer.rb
@@ -1,5 +1,5 @@
 class ReporterMailer < ActionMailer::Base
-  default from: "cartodb.com <support@cartodb.com>"
+  default from: Cartodb.get_config(:mailer, 'from')
   layout 'mail'
 
   def trending_maps_report(mail_to, trending_visualizations)
--- a/app/mailers/reporter_mailer.rb
+++ b/app/mailers/reporter_mailer.rb
@@ -1,0 +1,11 @@
+class ReporterMailer < ActionMailer::Base
+  default from: "cartodb.com <support@cartodb.com>"
+  layout 'mail'
+
+  def trending_maps_report(mail_to, trending_visualizations)
+    @subject = "Daily report for trending maps"
+    @trending_visualizations = trending_visualizations
+
+    mail to: mail_to, subject: @subject
+  end
+end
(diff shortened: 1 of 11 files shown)


app/mailers/reporter_mailer.rb
```diff
@@ -0,0 +1,11 @@
+class ReporterMailer < ActionMailer::Base
+  default from: "cartodb.com <support@cartodb.com>"
```
Kartones: Again.. this should be config based :/ I now takes time to do config changes but we shouldn't skip them... 
Even if OSS installs won't trigger the call

lib/cartodb/trending_maps.rb
```diff
@@ -18,7 +18,8 @@ def get_trending_maps(simulation=false)
         yesterday_mapviews = stats_manager.get_api_calls_from_redis(username, {from: date, to: date, stat_tag: visualization_id})
         total_mapviews = stats_manager.get_total_api_calls_from_redis(username, visualization_id)
         if is_trending_map?(yesterday_mapviews[date_key], total_mapviews)
-          trending_maps[visualization_id] = { user: username, mapviews: total_mapviews }
+          visualization = Carto::Visualization.find(visualization_id)
+          trending_maps[visualization_id] = { user: username, user_mail: visualization.user.email, user_public_url: visualization.user.public_url, mapviews: total_mapviews, visualization_name: visualization.name }
```
Kartones: multiline please, here Hound is correct, really hard to read in the diff :P

Kartones:
Apart from the hardcoded email & one-liner big hash, +1

app/mailers/reporter_mailer.rb
ethervoid: Ok but that is this way in every mailer we have. The other mail I'm going to put in the config. I'll change all this too in a new property

lib/cartodb/trending_maps.rb
ethervoid: Ok

app/mailers/reporter_mailer.rb
Kartones: uhm, ok, but would be nice to have it via config anyway... but then as you see best

ethervoid: Yup I'm changing and this way I learn how to add new configuration elements

ethervoid:
Done the default from change for all the mailers and promoted the configuration to production. I'll deploy it when the groups thing is done.

PR #5773, opened 2015-09-28
Fixes maps an datasets URLs in the user feed page

javierarce:
This PR fixes the malformed URLs for maps and datasets that appear in the user feed. Prepending the `base_url` part adds the `/u/username` part that was missing in the links from organization users.

### DIFF
--- a/lib/assets/javascripts/cartodb/explore/view.js
+++ b/lib/assets/javascripts/cartodb/explore/view.js
@@ -170,8 +170,7 @@ module.exports = Feed.extend({
         date: this.model.get('order_by') === 'updated_at' ? item.get('updated_at') : item.get('created_at'),
         datasetSize: this._getDatasetSize(item),
         geomType: this._getGeometryType(item.get('geom_types')),
-        account_host: cdb.config.get('account_host'),
-        base_url: cdb.config.get('base_url')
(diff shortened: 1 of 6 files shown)


javierarce:
Can you have a look, @xavijam?

javierarce:
Can you have a look, @viddo?

viddo:
With this approach the first part of the URL would always be based on a configuration setting (current user, right?), AFAIK subdomains and first part of path (/u/…) may differ depending on the vis _owner_. Does this actually work as expected for all URLs? 

Is there a reason for not using `vis.viewUrl().public()` instead, which handles this? `vis.viewUrl(currentUser).public()` might be even better since it takes the current user access into account too.

javierarce:
In this case the `base_url` variable refers to the vis owner, and all the maps and datasets showed in the feed belongs to him/her.

The reason we aren't using `vis.viewUrl` is because in this case the `vis` elements of the collection is just the default backbone model obtained querying the `/api/v1/viz` endpoint.

viddo:
Aha, thx for clarifying! In that case :+1: 
cc @xavijam let's keep this (:point_up:) in mind for the future, I'm not very fond of having logic duplicated but fixing it would require quite a change I think.

PR #5780, opened 2015-09-28
Get proper website url for user feed and maps/datasets public pages

### LINKED ISSUE #5775: Link to url doesnt work
For example here - https://catalanas2015-porcentual.cartodb.com

xavijam:
#### Problem:
People is used to entering the url w/o http, and the url link is not working in the new user feed.
We solved in the maps/datasets page so I've decided to create a helper method, what do you think @kartones? Is it the right place (biggest doubt)?

Fixes #5775
REVIEWER:  @Kartones 
cc @piensaenpixel, thanks for spotting it.


Kartones:
:+1: fine, as that's very specific presentation logic only for this, perfect to have it here

PR #5781, opened 2015-09-28
Add search and pagination to groups view

viddo:
Implement the browse feature missing on groups view, i.e. pagination+search filtering.
Part milestone for #4883 

Groups view:
screen shot 2015-09-28 at 16 27 08
screen shot 2015-09-28 at 16 27 24

Explore page:
screen shot 2015-09-25 at 17 51 12
screen shot 2015-09-25 at 17 51 15

Technical noteworthy changes:
- Applied `PagedSearchView` to groups_index
- Refactored `.Filters` in various ways and for different purposes:
  - Extracted the explore styles, since although the rules are similar the visual style is different, should make it easier to maintain them both.
  - Removed redudant markup+styles that either didn't have any styles or could be simplified using a single component, `Filters-order/-orderList/-list` => `Filters-group`
  - Moved filters search state handling to JS instead (following existing convention), makes the CSS less complex and focused on styles only
  - Bonus: Make filters work consistently in create-modal too
    - search stays focused after a search submit
    - ESC resets search-state instead of closing modal (2nd ESC closes modal, as expected, ofc)

Note that with these changes and the `PagedSearchView` we could/should DRY the organization users view too, and thus remove the redundant [`.OrganizationSearch`](https://github.com/CartoDB/cartodb/blob/master/app/assets/stylesheets/organization/organization_user_search.css.scss) and view code that's already solved, although I'd like to that in an isolated PR.

@xavijam can you review?
cc @juanignaciosl


viddo:
cc @javierarce @MariaCheca FYI re: the changes to the `.Filters`, feedback appreciated :)

viddo:
deployed on ded03 for testing

app/assets/stylesheets/explore/context-menu.css.scss
```diff
@@ -0,0 +1,96 @@
+@import '../variables/mixins';
```
xavijam: Great, better to have this one separated cc @javierarce

lib/assets/javascripts/cartodb/common/dialogs/create/listing/navigation_view.js
```diff
@@ -99,14 +105,40 @@ module.exports = cdb.core.View.extend({
...
+      this._focusSearchInput();
+    }
+  },
+
+  _$searchInput: function() {
+    return this.$('.js-search-input')
+  },
+
+  _focusSearchInput: function() {
+    this._$searchInput().select();
```
xavijam:
```js
this._$searchInput()
  .select()
  .focus();
```

viddo: If I recall correctly the chaining calls didn't work when I implemented this, why I split it up on two lines. :/ I'll double-check

lib/assets/javascripts/cartodb/common/dialogs/map/image_picker/navigation_template.jst.ejs
```diff
@@ -1,52 +1,48 @@
 <span class="Filters-separator"></span>
 <div class="Filters-row">
```
xavijam: Great, I checked this element before getting into this change.

xavijam:
:es:

lib/assets/javascripts/cartodb/common/dialogs/create/listing/navigation_view.js
xavijam: Yeah, I didn't try it, so probably it will not work :S, but it should be!

app/assets/stylesheets/explore/context-menu.css.scss
javierarce: :+1:

PR #5795, opened 2015-09-29
Adds REGEX to better find filenames within a url

### LINKED ISSUE #5704: Australian bureau of statistic files cannot be imported by URL
Example: 

http://www.abs.gov.au/ausstats/subscriber.nsf/log?openagent&1259030001_MB_2011_ASGC_NSW_csv.zip&1259.0.30.001&Data%20Cubes&2787F2FFF3F6E607CA2578CC001268ED&0&July%202011&05.10.2011&Latest

This comes from this big library: 

http://www.abs.gov.au/AUSSTATS/abs@.nsf/DetailsPage/1259.0.30.001July%202011?OpenDocument

CartoDB is downloading a "log" file and throwing a `Unable to open datasource `/tmp/imports/20150923-19668-azep6v/log' with the following drivers.` error.

gfiorav:
Fixes #5704

The idea here is to improve the way we look for filenames inside a URL (when the filename is not specified in the HTTP headers). 

The REGEX is generated programatically so if in the future we support more filetypes, this code will still be up to date.

If a filename can’t be matched, the code continues as it did before when no name could be found. This, however, is bound to not have happened very often (since the previous code would almost certainly return something, even an empty string), and maybe we should revise that situation. Should we open another issue for that?

Still have to add a test, but I wan’t to know if you think there might be a better approach.

What do you think @rafatower ?


services/importer/lib/importer/downloader.rb
```diff
@@ -24,6 +24,9 @@ class Downloader
       DEFAULT_HTTP_REQUEST_TIMEOUT = 600
       URL_ESCAPED_CHARACTERS = 'áéíóúÁÉÍÓÚñÑçÇàèìòùÀÈÌÒÙ'
 
+      EXTS = CartoDB::Importer2::Unp::SUPPORTED_FORMATS.concat(CartoDB::Importer2::Unp::COMPRESSED_EXTENSIONS).join('|')
```
rafatower: Use meaningful names

rafatower:
Add a test case before code review, please

gfiorav:
I think this is ready for a CR @rafatower 

Thanks! :smile:

services/importer/spec/unit/downloader_spec.rb
```diff
@@ -6,8 +6,9 @@
 describe Downloader do
   before do
     @file_url =
-      "https://developer.mozilla.org/samples/video/chroma-key/foo.png" 
-    @file_filepath  = path_to('foo.png')
+      "http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/physical/ne_110m_lakes.zip"
+    @hard_url = "https://manolo.escobar.es/param&myfilenameparam&zip_file.csv.zip&otherinfo"
```
rafatower: what is `@hard_url`?

services/importer/spec/unit/downloader_spec.rb
```diff
@@ -225,14 +226,20 @@
     it 'gets the file name from the URL if no Content-Disposition header' do
       headers = {}
       downloader = Downloader.new(@file_url)
-      downloader.send(:name_from, headers, @file_url).should eq 'foo.png'
+      downloader.send(:name_from, headers, @file_url).should eq 'ne_110m_lakes.zip'
+    end
+
+    it 'gets the file name from the URL if no Content-Disposition header and custom params schema is used' do
+      headers = {}
+      downloader = Downloader.new(@hard_url)
```
rafatower: if `@hard_url` is only used here, then I'd use a local variable.

services/importer/spec/unit/downloader_spec.rb
gfiorav: `@hard_url` is just an example of a url where the filename is obvious (as in right after the las `/`) but within custom delimiters.

services/importer/spec/unit/downloader_spec.rb
gfiorav: Yes, I agree. Besides, it's not a real url so it can't be used for anything else than testing file name finding.

rafatower:
:+1: remember to test different import/sync scenarios that could be affected by this on staging.

gfiorav:
I've tried importing/syncing files from URLs with the filename embedded in various ways and it looks good. Will merge. Thanks!

PR #5798, opened 2015-09-29
Display annotations and use CDN in highlighted explore map

### LINKED ISSUE #5789: Highlighted map is not showing annotations
Checking current highlighted map in explore page it appears as:

screen shot 2015-09-29 at 00 12 44

But in the [public map](https://kjethoe.cartodb.com/viz/7f431526-65ce-11e5-b6b2-0e49835281d6/public_map):

screen shot 2015-09-29 at 00 13 53

xavijam:
We should use CDN tiles instead of application tiles, and showing annotations if the map has any of those.

Fixes #5789.

REVIEWER: @javierarce

### DIFF
--- a/lib/assets/javascripts/cartodb/explore/entry.js
+++ b/lib/assets/javascripts/cartodb/explore/entry.js
@@ -60,14 +60,13 @@ $(function() {
         createVis: {
           url: visURL,
           opts: {
-            no_cdn: true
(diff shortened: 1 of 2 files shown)


xavijam:
Sounds good @javierarce?

javierarce:
Like music in my hears :ear: :notes:

PR #5804, opened 2015-09-30
Adds the possibility to export an image by specifying the size

javierarce:
This PR adds the possibility to export an image by specifying the size. 

#### Export view

screen shot 2015-09-30 at 12 30 33

#### Dropdown 
screen shot 2015-09-30 at 12 31 31

#### Text for review:

> Export custom image size
The image will be centered using the map's current center and won't export your widgets.
Please, be sure you have the pop-up blocker deactivated for this site.


javierarce:
Disregarding the hound warnings (woking on them), this PR is ready for review, @xavijam.
@josecruz: can you have a look to the attached screenshots? Since we had to get rid of the .gif option I had to slightly modify the layout.

javierarce:
I defined these three styles inside of the `export_image_view.scss` and I'm not completely sure they should be there:
```
.Form-rowData.Form-rowData--export 
.Form-rowLabel.Form-rowLabel--export
.RadioButton.RadioButton--export```

**Comment by xavijam** [comment]
staging?

**Comment by javierarce** [comment]
@xavijam, @josecruz: ded06

**Comment by xavijam** [comment]
Format element within the dialog looks weird, shouldn't it be align to the left?
screen shot 2015-09-30 at 14 50 05

**Comment by xavijam** [comment]
Without checking the code, opening the format dropdown works strange, don't you feel it slow? And there is no animation.

**Comment by javierarce** [comment]
Regarding the aligment, that's why I asked @josecruz. The original design with 3 formats is not aligned to the left.

How did the bug happen?

**Comment by xavijam** [comment]
Select jpg and open the dropdown again.

**Comment by javierarce** [comment]
It's animated for me :/

**Comment by javierarce** [comment]
Oh, I see the bug. Fixing it.

File: lib/assets/javascripts/cartodb/common/dialogs/static_image/advanced_export_view.js [hunk 1]
```diff
@@ -0,0 +1,67 @@
+var cdb = require('cartodb.js');
+var BaseDialog = require('../../views/base_dialog/view');
+var ViewFactory = require('../../view_factory');
+var randomQuote = require('../../view_helpers/random_quote');
+var moment = require('moment');
+
+module.exports = BaseDialog.extend({
+
+  defaults: {
```
xavijam: defaults is more for Backbone.Model. If you don't want to use options, I'd try to use it as a Static variable.

lib/assets/javascripts/cartodb/common/dialogs/static_image/advanced_export_view.js
```diff
@@ -0,0 +1,67 @@
...
+    this._initModel();
+    this._initBinds();
+  },
+
+  render_content: function() {
+    return this.getTemplate('common/dialogs/static_image/advanced_export_view')(this.options);
+  },
+
+  _initModel: function() {
+    this.model = new cdb.core.Model(this.options);
```
xavijam: I'd create the model in the initialize

lib/assets/javascripts/cartodb/common/dialogs/static_image/advanced_export_view.js
```diff
@@ -0,0 +1,67 @@
...
+  events: BaseDialog.extendEvents({
+    'click .js-format': '_onClickFormat',
+    'focus .js-textInput': '_onFocus',
+    'blur .js-textInput': '_onBlur'
+  }),
+
+  initialize: function() {
+    this.elder('initialize');
+
+    this.options = _.extend(this.defaults, this.options);
```
xavijam: I've checked that format is being set in the model, should we only use the model?

xavijam: And what about to create a proper model only for image export purposes and initialize it before the creation of the view?

lib/assets/javascripts/cartodb/table/export_image_view.js
```diff
@@ -246,6 +260,88 @@ cdb.admin.ExportImageView = cdb.core.View.extend({
     this._cancel();
   },
 
+  _openFormatDropdow: function(e) {
+    var self = this;
```
xavijam: I feel this self is not longer needed

xavijam:
Surprisingly, in Firefox and in Safari the dropdown is opened very smooth, but in my chrome is showed with a jump. What about IE10 btw?

javierarce:
In IE is smooth too.

-- 
Javier Arce

javierarce:
Maybe this alignment is better, @josecruz? 

screen shot 2015-09-30 at 18 07 38

josecruz:
Sorry for the delay! LGTM. The only thing I would separate the format labels by 20px and also would lock constraints (just like the non-modal one)

javierarce:
Cool. I improved the alignment and added error indicators.

screen shot 2015-09-30 at 18 39 34

PR #5814, opened 2015-09-30
Removing possibility to render map previews using cdb.Image

### LINKED ISSUE #4670: migrate previews to preview api
Remove all the cartodb.js instanciation thing and use the static url for the visualization:

https://github.com/CartoDB/cartodb/blob/master/app/views/admin/visualizations/embed_map.html.erb#L5

xavijam:
We used to have two possibilities, render the map previews with:

- Vizjson: it was using the cdb.Image we introduced in CartoDB.js (old method).
- Vizid: using the Static API we have in the editor (newish method).

Due to the fact that Preview API is faster that the old method, we have removed that one and we only offer de new one.

Places changed:

- Dashboard maps list.
- Delete items dialog.
- Public organization dashboard.
- Public user feed.
- Explore page.
- Public user maps.
- Public user datasets.
- Public map.
- Public dataset.
- Pecan?: no, it uses ```cdb.Image...```

Fixes #4670.
REVIEWER: @javierarce

cc @matallo


xavijam:
Available on RUIs in a while.

lib/assets/javascripts/cartodb/explore/entry.js
```diff
@@ -13,12 +13,9 @@ var ExploreModel = require('./model');
  */
 $(function() {
   cdb.init(function() {
-
     cdb.templates.namespace = 'cartodb/';
+    cdb.config.set(window.config);
```
xavijam: exposing all the configuration in the cdb.config model, where we have ```maps_api_template```, for example.

lib/assets/javascripts/cartodb/explore/view.js
```diff
@@ -217,7 +217,7 @@ module.exports = Feed.extend({
         username: username,
         visId: visId,
         className: className,
-        account_host: cdb.config.get('account_host')
+        mapsApiHost: cdb.config.getMapsApiHost()
```
xavijam: It should use the same approach we take in the whole editor, despite of the fact that explore page is a bit special. But if we need to check that works in local, we should "create" the same environment with a common-data user in development.

javierarce:
### Checked
- [ ] Dashboard maps list.
- [ ] Delete items dialog.
- [ ] Public organization dashboard.
- [ ] Public user feed.
- [ ] Explore page.
- [ ] Public user maps.
- [ ] Public user datasets.
- [ ] Public map.
- [ ] Public dataset.
- [ ] Pecan

xavijam:
Fixed problem when a map is shared in the organization and the map preview was not working due to it was not taking the map owner.

javierarce:
:+1: code and manual tests :are :ok:

xavijam:
Waiting the green light to deploy :dancer:, thanks @javierarce.

xavijam:
Merged with master due to package conflict, on frontend green, :rocket:.

PR #5820, opened 2015-10-01
Visualization export/restore v1.0

### LINKED ISSUE #5710: Visualization export/restore v1.0

Kartones:
Closes #5710


Kartones:
@ethervoid CR please (this one is fun)

NEWS.md
```diff
@@ -15,6 +15,8 @@
   4. Trigger existing org owner role assignment: `RAILS_ENV=development bundle exec rake cartodb:db:assign_org_owner_role`.
   5. Increase your database pool size to 50 (10 x # threads, see next line) at config/database.yml. Sample development configuration: config/database.yml.sample
   6. From now on you must run the server in multithread mode: `bundle exec thin start --threaded -p 3000 --threadpool-size 5`.
+* New visualization backups feature. Upon viz deletion a special vizjson will be stored in a new DB table. Bakcups live for Carto::VisualizationsExportService::DAYS_TO_KEEP_BACKUP days and can be recovered with `cartodb:vizs:import_user_visualization` rake by visualization id. Needs new feature flag `visualizations_backup`. Check https://github.com/CartoDB/cartodb/issues/5710 for additional details
```
ethervoid: s/Bakcups/Backups

app/models/table.rb
```diff
@@ -502,38 +503,18 @@ def default_baselayer_for_user(user=nil)
   end
 
   def create_default_map_and_layers
+    base_layer = CartoDB::Factories::LayerFactory.get_default_base_layer(owner)
 
-    # Adds the default baselayer
-    baselayer = default_baselayer_for_user
-    provider = ::Map.provider_for_baselayer(baselayer)
-    m = ::Map.create(::Map::DEFAULT_OPTIONS.merge(table_id: self.id, user_id: self.user_id, provider: provider))
+    m = CartoDB::Factories::MapFactory.get_map(base_layer, user_id, id)
```
ethervoid: I'll be great remote consonant like `m`

Kartones: true, was like that before but no excuse as I was refucktoring, so fixing it ;)

app/models/visualization/member.rb
```diff
@@ -249,8 +249,17 @@ def fetch
...
 
-      def delete(from_table_deletion=false)
-          # Named map must be deleted before the map, or we lose the reference to it
+      def delete(from_table_deletion = false)
+        if user.has_feature_flag?(Carto::VisualizationsExportService::FEATURE_FLAG_NAME)
+          begin
+            Carto::VisualizationsExportService.new.export(id)
+          rescue => exception
+            # Don't break deletion flow
+            CartoDB.notify_exception(exception, user: user, visualization_id: id)
```
ethervoid: Notification exception only allow request and user. Change it for notification_error

app/models/visualization/vizjson.rb
```diff
@@ -22,6 +21,31 @@ def initialize(visualization, options={}, configuration={}, logger=nil)
         logger.info(map.inspect) if logger
       end
 
+      # TODO: move to exporter class
```
ethervoid: TODO's are evil

app/models/visualization/vizjson.rb
```diff
@@ -22,6 +21,31 @@ def initialize(visualization, options={}, configuration={}, logger=nil)
...
+      def to_export_poro(version = 1)
+        # Redcarpet markdown renderer adds "garbage" that would otherwise get reimported
+        description = visualization.description_html_safe.nil? ? "" :
```
ethervoid: To much logic in one line, extract a method for example

app/models/visualization/vizjson.rb
```diff
@@ -91,6 +115,24 @@ def bounds_from(map)
...
+        data_layers = visualization.layers(:cartodb).map do |layer|
+          CartoDB::Layer::Presenter.new(layer, options, configuration).to_vizjson_v2
+        end
+        layers_data.push(data_layers)
+
+        layers_data.push(other_layers_for(visualization))
+
+        layers_data += non_basemap_base_layers_for(visualization)
+
+        layers_data.compact.flatten
```
ethervoid: Beware, here should be compact!

app/services/carto/visualizations_export_service.rb
```diff
@@ -0,0 +1,243 @@
...
+    DAYS_TO_KEEP_BACKUP = 15
+
+    SERVICE_VERSION = 1
+
+    def purge_old
+      items = retrieve_old_backups
+      items.each do |item|
+        remove_backup(item)
+      end
+      items.length
```
ethervoid: Check for nil? here if needed

app/models/visualization/vizjson.rb
Kartones: To be sincere I don't know what `.flatten` is doing here, as everything should be already one-dimensional arrays, but left when copying from `def all_layers_for`. But I'd better remove first the nulls (compact) & then flatten, not the other way around...

app/services/carto/visualizations_export_service.rb
Kartones: Checked, `.pluck` returns always an array, either empty or with items, so items is always an array too

app/services/carto/visualizations_export_service.rb
```diff
@@ -0,0 +1,243 @@
...
+                                            .to_json
+      backup_entry = Carto::VisualizationBackup.new(
+        username: visualization.user.username,
+        visualization: visualization.id,
+        export_vizjson: data
+      )
+      backup_entry.save
+
+      true
+    rescue => exception
```
ethervoid: Much more clear to have rescue VisualizationsExportServiceError (and reraise) and then rescue => exception and raise a new Visualiza...

app/services/carto/visualizations_export_service.rb
```diff
@@ -0,0 +1,243 @@
...
+    private
+
+    # Mainly intended for testing
+    def export_version
+      SERVICE_VERSION
+    end
+
+    def retrieve_old_backups
+      max_date = Date.today - DAYS_TO_KEEP_BACKUP
+      Carto::VisualizationBackup.where("created_at < ?", max_date).pluck(:visualization)
```
ethervoid: Less than or less or equal than

app/services/carto/visualizations_export_service.rb
```diff
@@ -0,0 +1,243 @@
...
+        true
+      else
+        false
+      end
+    end
+
+    def restore_backup(visualization_id, skip_version_check)
+      # TODO: support partial restores
+      visualization = Carto::Visualization.where(id: visualization_id).first
+      if visualization
```
ethervoid: raise .... if visualization.nil?

app/services/carto/visualizations_export_service.rb
Kartones: You're right, changed to to multiple catches, one specific and one generic

app/services/carto/visualizations_export_service.rb
```diff
@@ -0,0 +1,243 @@
...
+        raise VisualizationsExportServiceError.new("Visualization with id #{visualization_id} already exists!")
+      end
+
+      dump_data = get_restore_data(visualization_id, skip_version_check)
+
+      user = ::User.where(id: dump_data["owner"]["id"]).first
```
ethervoid: What if dump_data["owner"]["id"] is nil?

app/services/carto/visualizations_export_service.rb
Kartones: haha ok, no need to be really precise but will change :D

app/services/carto/visualizations_export_service.rb
```diff
@@ -0,0 +1,243 @@
...
+
+      true
+    end
+
+    def set_map_data(map, exported_data)
+      map.recalculate_bounds!
+
+      map.scrollwheel = exported_data["scrollwheel"] if exported_data["scrollwheel"]
+      map.legends = exported_data["legends"] if exported_data["legends"]
+      if exported_data["bounds"]
```
ethervoid: I'll check for the length here just in case

app/services/carto/visualizations_export_service.rb
```diff
@@ -0,0 +1,243 @@
...
+        map.view_bounds_sw = exported_data["bounds"][0].to_s
+        map.view_bounds_ne = exported_data["bounds"][1].to_s
+      end
+      map.center = exported_data["center"] if exported_data["center"]
+      map.zoom = exported_data["zoom"] if exported_data["zoom"]
+      map.provider = exported_data["map_provider"] if exported_data["map_provider"]
+
+      map.save.reload
```
ethervoid: I'd put two calls instead one

app/services/carto/visualizations_export_service.rb
Kartones: Line too long then ;) so changed to if ... end

app/services/carto/visualizations_export_service.rb
Kartones: Then I let the world explode because that's critical :) (as other fields). I recover from, ignore or recreate non-critical ones, others let the fireworks happen

app/services/carto/visualizations_export_service.rb
```diff
@@ -0,0 +1,243 @@
...
+    def layer_kind_from_type(exported_layer_type)
+      if exported_layer_type == 'CartoDB'
+        'carto'
+      else
+        exported_layer_type.downcase
+      end
+    end
+
+    def create_base_layer(user, exported_data)
+      layer_data = exported_data["layers"].select { |layer| ::Layer::BASE_LAYER_KINDS.include?(layer["type"]) }.first
```
ethervoid: What if this has 2 or more layers and you get only the first?

Kartones: Basemap is always layer 0. If there are more (which will almost surely be), we only care about the first one, but I'll add a comment (this is juglar knowledge, not written anywhere but transmited by word of mouth)

app/services/carto/visualizations_export_service.rb
```diff
@@ -0,0 +1,243 @@
...
+    end
+
+    def create_map(user, base_layer)
+      map = CartoDB::Factories::MapFactory.get_map(base_layer, user.id)
+      map.add_layer(base_layer)
+      map
+    end
+
+    def add_data_layers(map, exported_data)
+      exported_data["layers"].select { |layer|
```
ethervoid: Don't nest two .select and inside a .each. This should be more readable use the do |layer| blabla :)

app/services/carto/visualizations_export_service.rb
ethervoid: Are you sure the order is granted because the select could mess with the array order (I've suffered it IIRC)

Kartones: Yep, the `.layers(:TYPE)` guarantees an intrinsic ORDER BY internally (when coding AR version of the model I suffered it). I've also done a lot of real visualization tests (because the editor and "scripted" layers creation differ a bit) and is fine

ethervoid:
Despite the comments, seems :+1:

PR #5823, opened 2015-10-01
Captures random_name return 5822

### LINKED ISSUE #5822: URL importer download: TypeError: can't convert nil into String
https://github.com/CartoDB/cartodb/blob/master/services/importer/lib/importer/downloader.rb#L387

GDELT API Syncs are failing due to this error, their URL has the shape:

`http://api.domain.org/api/v1/gkg_geojson?QUERY=Some%20Name&TIMESPAN=1440&OUTPUTFIELDS=url,name,sharingimage`

gfiorav:
Fixes #5822

Because of previous `name_in(url)` method always returning something, the following code was not ready for `random_name`'s return to be used. 

My bad for not testing a no-url-name no-http-header-name import, and not discovering this earlier.

It should now work.

Please, CR @rafatower or @ethervoid

cc/ @iriberri

### DIFF
--- a/services/importer/lib/importer/downloader.rb
+++ b/services/importer/lib/importer/downloader.rb
@@ -277,11 +277,11 @@ module CartoDB
 
       def name_from(headers, url, custom=nil)
         name =  custom || name_from_http(headers) || name_in(url)
+
         if name == nil || name == ''
-          random_name
-        else
-          name
+          name = random_name
         end
+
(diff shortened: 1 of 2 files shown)


rafatower:
:+1:

gfiorav:
Thanks @rafatower, will deploy as soon as CI tests pass

PR #5834, opened 2015-10-01
Better import error page

### LINKED ISSUE #5778: Error import page lacks details
A bit of a common theme in some of my reports...

screenshot 2015-09-28 21 47 45

It would be much better for the user to include some diagnostic and contextual information. 

- what URL was attempted?
- what HTTP code was returned?
- any status message?

(In this case, I think probably the source server has some kind of internal failure, so don't worry about trying to solve this one for me...)

gfiorav:
Fixes #5778

We can work on this PR, since the relevant info has been piped through to the front end.

screen shot 2015-10-01 at 19 36 25

cc @saleiva @rafatower @stevage @javisantana


javisantana:
too much info, for the user is enough to know the url and the error code (and the text) returned by it

gfiorav:
@javisantana 

+ The error text is already displayed
+ The HTTP code is shown when relevant (i.e, 404) but I'll work on piping it from the downloader
+ The URL will be shown

I'll remove the other info, it was just so we could decide what's important and what not.

iriberri:
I think your list in your last comment is enough, @gfiorav. :)

saleiva:
I thought you were assigning this to me. Happy to jump in!

gfiorav:
@saleiva I'll push the necessary info to the backbone model. You take it from there! :smiley: 

Still figuring out the least messy way of getting that HTTP response code to the front end.

app/controllers/carto/api/data_import_presenter.rb
```diff
@@ -62,6 +62,16 @@ def public_values
...
         values.merge!(success: @data_import.success) if @data_import.final_state?
+        values.merge!(original_url: @data_import.original_url)
+        values.merge!(state: @data_import.state)
+        values.merge!(content_guessing: @data_import.content_guessing)
+        values.merge!(type_guessing: @data_import.type_guessing)
+        values.merge!(server: @data_import.server)
+        values.merge!(service_name: @data_import.service_name)
+        values.merge!(host: @data_import.host)
+        values.merge!(upload_host: @data_import.upload_host)
+        values.merge!(resque_ppid: @data_import.resque_ppid)
```
rafatower: all sensitive information (internal network, job pid and so) to be removed from here

gfiorav:
Ok, so I think this might be more like it. What if we add a dictionary so we can show more info on the code? i.e: "404 - Not found"

screen shot 2015-10-02 at 15 12 31

gfiorav:
Now we have an error code message as well. I think this should be enough. 

screen shot 2015-10-02 at 16 29 43

saleiva:
IMO, that "403 - CartoDB is not autho..." should be in the point 1. (I'd remove the "HTTP Response" title)
then, point 2 should have the "Want to know more?"
Point '+' should have the "Persisting error" content

gfiorav:
@saleiva Agree, I changed the order.

The only thing thats bugging me is that "Follow these steps" doesn't make much sense right?

What if we change the error code to match the HTTP error code (support doesn't need anything else that the identifier anyway) and change the subtitle to "Don't panic, here's some useful information"?

Something like this:
screen shot 2015-10-02 at 17 14 54

gfiorav:
Retest this please

gfiorav:
@saleiva 

Like we agreed, this seems to be a good balance between human readability and functionality.

screen shot 2015-10-02 at 18 28 31


Ready for CR @rafatower

gfiorav:
Final design:

screen shot 2015-10-05 at 11 23 45

I'm currently having some trouble with Jasmine, as I can't quite figure out how the test that is breaking ever passed before.

It's complaining about the view not including the text `We could not create table from your query` but I can't see where that string is produced in master.

Looking to figure that out and close this soon.

stevage:
Looks great :) My one last quibble would be the "Download error (1001)" - that error code doesn't really mean much to anyone. Much more meaningful would be "Download error (403)" or failing that, "Download error". The only time the 1001 would be relevant is when contacting support, and you already have that covered by including the GUID.

Just another $0.02 from me... :)

gfiorav:
@stevage thanks for the feedback. Since we have documented our internal error codes, we would like to keep them. You can find said documentation [here](http://docs.cartodb.com/cartodb-editor.html#error-codes-and-solutions).

Hope it makes sense for you! :smiley:

gfiorav:
Fixed Jasmine issues. Need a CR @rafatower so we can close this!

I'll do some staging tests in the mean time.

Thanks!

jesusvazquez:
retest this please

gfiorav:
Retest this please

gfiorav:
Retest this please

INFO: We are having some CI problems, sorry for the repeated messages

gfiorav:
retest this please

gfiorav:
Thanks for helping @xavijam !

Could I get a CR so I can close this @rafatower ?

Thanks :smile:

iriberri:
Just a super small detail. Remember that for any change that implies a change in the texts shown to the user, a revision from a native person is required.

gfiorav:
Oh! @iriberri could you ping such a person here?

iriberri:
@csobier, could you please check if the sentences written [here](https://github.com/CartoDB/cartodb/pull/5834/files) sound fine? :)

csobier:
Looks fine to me :+1:

app/controllers/carto/api/data_import_presenter.rb
```diff
@@ -5,6 +5,47 @@ module Carto
...
+        "404" => "This usually means no file was found at the specified URL.",
+        "405" => "This usually means CartoDB could not negotiate the download with the file's provider. The " \
+                  "provider is probably using a non standard method to serve this file.",
+        "407" => "This usually means CartoDB is not authorized to retrieve this file as proxy authentication " \
+                  "is required. If you can use the apropiate proxy, download the file manually and upload " \
+                  "it from your computer.",
+        "408" => "This usually means a timeout request was produced. You may want to try again.",
+        "409" => "This usually means a confilict in the request was produced. You might want to try again.",
+        "410" => "This usually means the file specified is now longer available at this location.",
+        "411" => "This usually means CartoDB did not perform the request properly. 'Content Length' header is missing.",
```
rafatower: I'm not for adding such interpretations on top of HTTP status codes

rafatower: I'd replace this by some standard status message that should be located somewhere within rails

db/migrate/20151002105834_add_http_response_code_to_data_imports.rb
```diff
@@ -0,0 +1,13 @@
+Sequel.migration do
+  up do
+    alter_table :data_imports do
+      add_column :http_response_code, :text
```
rafatower: why is it a text?

app/controllers/carto/api/data_import_presenter.rb
gfiorav: If you mean the "friendly speech", the idea is to provide more aid to a non-expert user and to sound more friendly.

Anyhow, maybe this is more for the design reviewer to decide? For me, it's a design issue although I might be wrong.

Since the DR was done by @saleiva: Should we go for more sterile messages or use these ones instead?

rafatower: (I'm open to discussion here, but at first sight I'm afraid of causing more confusing to the user rather than solving things on one hand and maintaining the list on the other hand)

db/migrate/20151002105834_add_http_response_code_to_data_imports.rb
gfiorav: As opposed to an `integer`? 

If so: because the header value is text and if it were a number it would never be operated on (as in sum, subtraction, etc). We avoid casting and problems derived from it. HTTP codes are never treated like numbers AFAIK.

rafatower:
:+1: fair enough. Let's go ahead with this and we can refine texts if needed

gfiorav:
Ok, thanks @rafatower @saleiva !

PR #5839, opened 2015-10-02
Sends the right dimensions to the advanced export dialog 

javierarce:
Fixes a problem that prevented sending the right dimensions to the advanced dialog (the problem was that this.options.width and this.options.height was being overwritten in the extend)


javierarce:
cc @xavijam

xavijam:
:+1:

PR #5857, opened 2015-10-05
Filter legacy remote visualizations

### LINKED ISSUE #5097: When an external_source has been used, can't be deleted
If a user has used an external_source to import a dataset, a row in `external_data_imports` is created. When we want to update the remote visualizations and external sources, if there are rows that isn't going to appear anymore we remove them but it they are linked by a row in `external_data_imports` its impossible to remove.

The best solution is to remove them logically

ethervoid:
Closes #5097


ethervoid:
retest this please

ethervoid:
retest this please

ethervoid:

Kartones:
+1

PR #5868, opened 2015-10-06
Add support for background image layer at named maps

### LINKED ISSUE #4130: Add support for background image layer at named maps
After #4129, we're missing one scenario for basemap layers. If type is `background` and `options -> image` is not empty (and thus, `options -> color` **is** empty), currently Named Maps at RoR won't render that layer.

After Windshaft implements the bg image support, add it to named maps so the layer is sent properly.

/cc @rochoa @javisantana 

Branch: `4130-static-map-bgs`

Kartones:
Closes #4130


Kartones:
@rafatower Really simple CR.

When added support for plain basemap layers, Windshaft didn't had a property for image urls/background urls. Now that it has, I just added detection & support for it (as is just a different param to send instead of `color` one)

rafatower:
:+1:

PR #5875, opened 2015-10-06
Added click tracking for some emails

ethervoid:
We want to add some click trackers for like and trending mails send from the platform.

We need to add this properties to the config but its backwards compatible with the configuration


ethervoid:

Kartones:
:+1:

ethervoid:
retest this please

PR #5876, opened 2015-10-06
Do not run cartodb postgresql extension tests

rochoa:
Extension is already running tests in its own repository

cc @jesusvazquez


rochoa:
Ok, closing it. We will make sense of this if we come back to actively use travis.

PR #5884, opened 2015-10-06
Export image using bbox instead of center

javierarce:
This PR changes the current advanced export method from center to bbox.


javierarce:
cc @xavijam

xavijam:
:+1:

PR #5903, opened 2015-10-07
Fixes spec

javierarce:
Fixed by stubbing the `pixelToLatLon` method.

### DIFF
--- a/lib/assets/test/spec/cartodb/table/map/export_image_view.spec.js
+++ b/lib/assets/test/spec/cartodb/table/map/export_image_view.spec.js
@@ -267,8 +267,7 @@ describe("ExportImage", function() {
         map: this.map
       });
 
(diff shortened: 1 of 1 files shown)


javierarce:
cc: @viddo

lib/assets/test/spec/cartodb/table/map/export_image_view.spec.js
```diff
@@ -267,6 +267,9 @@ describe("ExportImage", function() {
         map: this.map
       });
 
+      this.pixelToLatLon = sinon.stub(this.mapView, "pixelToLatLon");
+      this.pixelToLatLon.returns({ lat: 20, lng: 30 });
```
viddo: why use sinon instead of a default Jasmine spy? I.e. `spyOn(this.mapView, 'pixelToLatLon').and.returnValue({ lat: 20, lng: 30 })`

viddo:
:+1:

PR #5925, opened 2015-10-13
Configurable uploads folder

Kartones:
Makes configurable both the public uploads folder (althoug h see app_config.yml.sample as there are some limitations) and the UNP temporal folder.

Both default to current hardcoded constant values if config is not present

### DIFF
--- a/app/helpers/file_upload.rb
+++ b/app/helpers/file_upload.rb
@@ -1,10 +1,22 @@
 # coding: utf-8
 
-module  FileUploadHelper
+module CartoDB
+  class FileUpload
 
-    UPLOADS_PATH  = 'public/uploads'
+    DEFAULT_UPLOADS_PATH = 'public/uploads'
     FILE_ENCODING = 'utf-8'
-    MAX_SYNC_UPLOAD_S3_FILE_SIZE = 52428800   # bytes
+    MAX_SYNC_UPLOAD_S3_FILE_SIZE = 52428800 # bytes
+
+    def initialize(uploads_path = nil)
+      @uploads_path = uploads_path || DEFAULT_UPLOADS_PATH
+      unless @uploads_path[0] == "/"
+        @uploads_path = Rails.root.join(@uploads_path)
+      end
+    end
+
+    def get_uploads_path
+      @uploads_path
+    end
 
     def upload_file_to_storage(params, request, s3_config = nil, timestamp = Time.now)
       results = {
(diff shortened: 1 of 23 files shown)


Kartones:
retest this please

Kartones:
Tested at staging and works fine so, @rafatower CR please

rafatower:
hound makes this harder if we don't listen to it :/

Kartones:
@rafatower try now, sorry but else this task would have taken 2 days properly indenting all the files affected...

app/helpers/file_upload.rb
```diff
@@ -1,10 +1,22 @@
...
-    UPLOADS_PATH  = 'public/uploads'
+    DEFAULT_UPLOADS_PATH = 'public/uploads'
     FILE_ENCODING = 'utf-8'
-    MAX_SYNC_UPLOAD_S3_FILE_SIZE = 52428800   # bytes
+    MAX_SYNC_UPLOAD_S3_FILE_SIZE = 52428800 # bytes
+
+    def initialize(uploads_path = nil)
+      @uploads_path = uploads_path || DEFAULT_UPLOADS_PATH
+      unless @uploads_path[0] == "/"
+        @uploads_path = Rails.root.join(@uploads_path)
```
rafatower: this can be a problem if a deploy happens in the middle.

services/importer/lib/importer/loader.rb
```diff
@@ -115,7 +115,7 @@ def normalize
           .inject(source_file.fullpath) { |filepath, normalizer_klass|
 
             @importer_stats.timing(normalizer_klass.to_s.split('::').last) do
-              normalizer = normalizer_klass.new(filepath, job)
+              normalizer = normalizer_klass.new(filepath, job, options[:importer_config])
```
rafatower: have you tested with the 2 normalizers? (IIRC there were: shp, csv)

rafatower:
I made a couple of comments: one is a legacy issue (and I don't know how to get around it), the other is just a question. Other than that, ok

services/importer/lib/importer/loader.rb
Kartones: Yep, tested with CSV, KML, added support for the new param to SHP one too. Also adapted the XLS/XLSX ones

app/helpers/file_upload.rb
Kartones: Yes but this also was happening before, the url was hardcoded anyway. so if the Rails root changed it would anyway fail with a file/folder not found error...

The good thing is that now we support config-based absolute folders, so if we make a separate, HTTP-accessible path, the uploads folder becomes independant of Rails. /cc @luisbosque @zenitraM

PR #5928, opened 2015-10-13
Update CartoCSS docs reference

viddo:
I noticed that the docs ref was updated for cartodb.js, in CartoDB/cartodb.js#709, I figure we should do the same here?

@javisantana @csobier since you two were involved in the other PR


javisantana:
:es:

PR #5929, opened 2015-10-13
Add missing AR models and relationships

zenitraM:
This adds:
- Shallow AR models for (I think) all of the non-AR existing models. They have no logic at all
- The missing `belongs_to`, `has_many` relationships between the models (so model traversal using reflection from the user-mover works)

It also changes the `has_and_belongs_to_many` relationships to safer, less-magic `has_many :through`, which are the same but the relationship object actually exists and you can put logic on it if you wanted (ref: http://blog.flatironschool.com/why-you-dont-need-has-and-belongs-to-many/ )


zenitraM:
cc @Kartones

Makefile
```diff
@@ -205,7 +205,7 @@ endif
 	# TODO skip this if db already exists ?
 	# Clean DB connections before drop test DB
 	psql -U postgres -c "select pg_terminate_backend(pid) from pg_stat_activity where datname='carto_db_test'"
-	MOCHA_OPTIONS=skip_integration RAILS_ENV=test bundle exec rake cartodb:test:prepare
+	MOCHA_OPTIONS=skip_integration RAILS_ENV=test bundle exec rake cartodb:test:prepare --trace
```
Kartones: is this intended?

zenitraM: That outputs the trace only when test:prepare fails, I used it to debug my troubles with the autoload. Not really related, but it doesn't hurt either.

Kartones: If only outputs in case of error, +1 for the change as I want as detailed error outputs as possible!

app/models/carto/automatic_geocoding.rb
```diff
@@ -0,0 +1,11 @@
+# encoding: utf-8
+
+require 'active_record'
+
+module Carto
+  class AutomaticGeocoding < ActiveRecord::Base
```
Kartones: It is missing a `has_many :geocodings` no? ([check here](https://github.com/CartoDB/cartodb/blob/master/app/models/automatic_geocoding.rb#L7)) (also please include either the default order or a TODO so we don't forget to migrate that sorting when we add logic)

app/models/carto/client_application.rb
```diff
@@ -0,0 +1,12 @@
+# encoding: utf-8
+
+require 'active_record'
+
+module Carto
+  class ClientApplication < ActiveRecord::Base
+
+    belongs_to :user, class_name: Carto::User
```
Kartones: Missing `has many :access_tokens`? ([sequel has it](https://github.com/CartoDB/cartodb/blob/master/app/models/client_application.rb#L7))

app/models/carto/data_import.rb
```diff
@@ -23,6 +23,9 @@ def instance_method_already_implemented?(method_name)
     STATE_STUCK     = 'stuck'
 
     belongs_to :user
+    belongs_to :log
+    has_many :external_data_imports, inverse_of: :data_import
```
Kartones: no need to specify here the `class_name`s? (maybe just to be sure)

app/models/carto/client_application.rb
zenitraM: There's no `access_tokens` table, it seems that the access_tokens model just extends `oauth_token` with this plugin: http://sequel.jeremyevans.net/rdoc-plugins/classes/Sequel/Plugins/SingleTableInheritance.html

It seems that it picks one class or other based on `type` attribute, right now there's no difference between them.
Should I add a TODO to add this logic once this model has real logic on it?

app/models/carto/external_source.rb
```diff
@@ -7,4 +7,6 @@ class Carto::ExternalSource < ActiveRecord::Base
   # Seconds
   REFRESH_INTERVAL = 30*24*60*60
 
+  has_many :external_data_imports
```
Kartones: External sources also have a mandatory visualization_id (not enforced by Ruby except validating is not nil, but should be constrainted)

app/models/carto/client_application.rb
zenitraM: It seems there are two kind of tokens, `AccessToken` and `RequestToken`, both inherit from `OauthToken` which seems like the abstract model for both.

ClientApplication is related only to `AccessToken` but User is related to `OauthToken`; I'll add a filtered relation for both of them just in case

app/models/carto/automatic_geocoding.rb
zenitraM: Oops, missed that one, added!

Kartones:
Nice work! I think I don't have anything else to add, so when build goes green :+1:

PR #5931, opened 2015-10-13
4730 ogr geom possible names take2

rafatower:


rafatower:
please do not delete that branch yet.

PR #5933, opened 2015-10-13
5880 multilayer map after multilayer import

### LINKED ISSUE #5880: Multilayer formats should create a multilayer map at the end of the importation
Multilayer formats or compressed files (zip, kml, etc) should create a multilayer map once the import ends.

For example: If we have a KML with 5 different layers we should and a user wants to create a map with that file we should create a multilayer map at the end of the import process.

* [ ] Create multilayered map from multilayer file
* [ ] Create multilayered map from multilayer datasource source

//cc @rafatower

### LINKED ISSUE #5947: Multilayer import warning
Related PR: https://github.com/CartoDB/cartodb/pull/5933

create - file importing warning

@saleiva u like it?
/cc @gfiorav

gfiorav:
Fixes #5880 
Fixes #5947

Advanced map creation.

Adds ability to create a multilayered map in one go from a compressed archive containing multiple datasets.

I’m PRing this to get some CI feedback. Still have some work to do.

- [x] Multilayered map naming
- [x] Handle `max_layers` exceeding 
- [x] Add warnings for 'some layers were not added'
- [x] Tests for multilayered imports
- [x] Smoketests 
- [x] Design warnings

@rafatower @ethervoid


gfiorav:
Warnings are now functional. @josecruz will design the warnings page properly and then someone will DR.

Some screenshots of the current proof of concept:
screen shot 2015-10-14 at 17 45 44
screen shot 2015-10-14 at 17 45 49

app/connectors/importer.rb
```diff
@@ -19,6 +19,8 @@ class Importer
       # @param quota_checker CartoDB::QuotaChecker
       # @param database
       # @param data_import_id String UUID
+      # @param imported_table_ids Array UUID
+      # @param rejected_layers Array String|nil
```
rafatower: good to document, but this two are not part of the initialize params

app/connectors/importer.rb
```diff
@@ -32,7 +34,9 @@ def initialize(runner, table_registrar, quota_checker, database, data_import_id,
         @destination_schema     = destination_schema
         @support_tables_helper  = CartoDB::Visualization::SupportTables.new(database,
                                                                             {public_user_roles: public_user_roles})
-        @data_import            = nil
+        @data_import            = data_import
+        @imported_table_ids = []
+        @rejected_layers = nil
```
rafatower: why one is initialized to empty array and the other to nil? wouldn't it be better to treat both alike?

gfiorav: Marios code checks for emptiness, I went for nullity.

It should be unified, I agree.

app/connectors/importer.rb
```diff
@@ -32,7 +34,9 @@ def initialize(runner, table_registrar, quota_checker, database, data_import_id,
         @destination_schema     = destination_schema
         @support_tables_helper  = CartoDB::Visualization::SupportTables.new(database,
                                                                             {public_user_roles: public_user_roles})
-        @data_import            = nil
+        @data_import            = data_import
```
rafatower: This is a bit confusing (and I think maybe what confused you with this code): when you write `data_import`, are you referring to the method below (I guess so) or the `attr_reader`? I'd remove the `attr_reader :data_import` defined above and stick to using the method defined below in order to always use clean lazy initialization, unless you see a problem with that approach.

gfiorav: In this case I'm using the method to initialise `@data_import`. It basically looks for the `DataImport` with id `@data_import_id`

app/connectors/importer.rb
```diff
@@ -177,14 +205,15 @@ def rename_the_geom_index_if_exists(current_name, new_name)
 
       def persist_metadata(result, name, data_import_id)
         table_registrar.register(name, data_import_id)
-        self.table = table_registrar.table
+        @table = table_registrar.table
+        @imported_table_ids << @table.table_visualization.id
```
rafatower: I'd use `@imported_table_visualization_ids`, and document somewhere (close to the `attr_reader`) its meaning: an Array of canonical visualization id's. Sounds correct?

app/connectors/importer.rb
rafatower: `rejected_layers` seem to be a String  containing a comma separated list of layers, rather than an Array. And it is latter used like `rejected_layers.split(',')`.

app/connectors/importer.rb
gfiorav: Yeah, it's more explicit. I like it.

app/connectors/importer.rb
rafatower: (Or maybe not, was this done to avoid storing arrays in postgres in `DataImport` model? anyway, the interface exposed to the outside word can be more uniform. Thoughts?)

gfiorav: `@rejected_layers` lives in `Importer` as an Array, but is stored in `@data_import` as a string to ease it's database storage in the `data_imports` table. It's only a string when stored or read. It's messy, but since it has to pass through sequel it's the easiest way. In other parts of the code, others used JSON stringing instead of comma separated values to achieve the same. Since it's just a list I went for simplicity.

app/models/visualization/derived_creator.rb
```diff
@@ -0,0 +1,54 @@
+# encoding: utf-8
+
+module CartoDB
+  module Visualization
+    class DerivedCreator
+      DEFAULT_MAP_NAME = 'Untitled Map'
+
+      def initialize(user, tables = [])
+        @rejected_layers = nil
```
rafatower: If you initialize this as an empty array, you won't have to deal with "is it an array or nil?" dilemma that obscures the code. My 2 cents

db/migrate/20151014085055_add_rejected_layers_to_data_imports.rb
```diff
@@ -0,0 +1,13 @@
+Sequel.migration do
+  up do
+    alter_table :data_imports do
+      add_column :rejected_layers, :text
```
rafatower: @gfiorav please remember to cherry-pick and release this migration in advance.

spec/requests/api/json/visualizations_controller_spec.rb
```diff
@@ -76,7 +76,7 @@
     it "when a map is liked should send an email to the owner" do
       user_owner = create_user
       table = new_table({user_id: user_owner.id, privacy: ::UserTable::PRIVACY_PUBLIC}).save.reload
-      vis = table.create_derived_visualization
+      vis, rejected_layers = CartoDB::Visualization::DerivedCreator.new(user_owner, [table]).create
```
rafatower: you can check as part of this test that `rejected_layers` is empty/nil as part of the test

db/migrate/20151014085055_add_rejected_layers_to_data_imports.rb
gfiorav: noted

app/models/data_import.rb
```diff
@@ -635,6 +636,7 @@ def store_results(importer=nil, runner=nil, datasource_provider=nil, manual_fiel
     else
       self.results    = importer.results
       self.error_code = importer.error_code
+      self.rejected_layers = importer.rejected_layers.join(',') if !importer.rejected_layers.nil?
```
rafatower: You should isolate in this class the fact that the internal representation of `rejected_layers` in the DB is a String whereas it is normally used as an Array outside of it.

spec/requests/api/json/visualizations_controller_spec.rb
gfiorav: I was going to to `vis, nil` but I like this more :smile:

app/connectors/importer.rb
rafatower: you won't have that problem if you don't use `@data_import` but  `data_import` instead except in the public method to lazy-initialize it below.

gfiorav:
@jose 

will work on UI on PR #5953. This will be released without UI.

gfiorav:
Pre-deploying migration in #5954

gfiorav:
The migration file was missing at the begging of this CI.

Retest this, please

jesusvazquez:
retest this please

lib/assets/javascripts/cartodb/common/views/warnings_details_view.js
```diff
@@ -0,0 +1,35 @@
...
+
+  render: function() {
+    var warnings = this.warnings;
+    var user = this.user;
+
+    var template = cdb.templates.getTemplate(this._TEMPLATES['partial_import']);
+
+    this.$el.html(
+      template({
+        user: user,
```
xavijam: I'd try to use only needed attributes or variables and not pass the whole models.

vendor/assets/javascripts/cartodb.mod.odyssey.uncompressed.js
```diff
@@ -1,5 +1,5 @@
-// version: 3.15.2
```
xavijam: Don't commit odyssey file, revert it ;).

vendor/assets/javascripts/cartodb.uncompressed.js
```diff
@@ -1,6 +1,6 @@
-// cartodb.js version: 3.15.7
```
xavijam: Pointing to last develop commit in CartoDB.js?

lib/assets/javascripts/cartodb/common/views/warnings_details_view.js
xavijam: BTW, do you need user Model in the template?

xavijam:
Fine by me, looks :+1:, just take care of those things I've just commented in your changes.

vendor/assets/javascripts/cartodb.uncompressed.js
gfiorav: If that is the result of running `grunt cdb` then yes

lib/assets/javascripts/cartodb/common/views/warnings_details_view.js
gfiorav: No, you're right. No need for user :)

vendor/assets/javascripts/cartodb.uncompressed.js
gfiorav: Ah, ok I see what you mean. I think that 3.15.8 is already taken, i'll checkout and recommit.

PR #5936, opened 2015-10-14
[Metrics] 5870 segment events

jmartinter:
New code for sending the following events to segment.com:
- created dataset
- deleted dataset
- created map
- deleted map

Each new event call is sent to resque queue and processed in an independent process.
Before each call to resque, the existence of a segment api key is checked. If unavailable, the call is just ignored.

A new gem is required: analytics-ruby

cc'ed @xavijam


app/controllers/api/json/tables_controller.rb
```diff
@@ -37,6 +39,9 @@ def create
           @table.valid? && @table.save
         end
 
+        custom_properties = {'privacy' => @table.table_visualization.privacy, 'type' => @table.table_visualization.type,  'vis_id' => @table.table_visualization.id, 'origin' => 'blank'}
```
Kartones: Better put this lines inside the `if save_status` at line 45, so in case of error you don't get a false positive created dataset call

app/controllers/api/json/visualizations_controller.rb
```diff
@@ -17,7 +18,7 @@ class Api::Json::VisualizationsController < Api::ApplicationController
 
   ssl_allowed  :notify_watching, :list_watching, :add_like, :remove_like
   ssl_required :create, :update, :destroy, :set_next_id unless Rails.env.development? || Rails.env.test?
-  skip_before_filter :api_authorization_required, only: [:add_like, :remove_like]
```
Kartones: Why did you removed this line? because this will probably make fail on live the likes feature... (likes can be "called" from unauthenticated scenarios and we don't want that to cause a 401 http error)

app/controllers/api/json/tables_controller.rb
Kartones: Why sending a `blank` origin here if at other calls below you simply omit the parameter? (this is more out of curiosity)

app/models/data_import.rb
```diff
@@ -879,4 +884,22 @@ def set_datasource_audit_to_failed(datasource)
     end
   end
 
+  def track_new_datasets(results)
```
Kartones: uhmm a few things to note here:
- We cannot have an import failing due to stats tracking, so you should add a `rescue` block around this code (and send an error using `CartoDB.notify_exception`) so in any case the import goes forward
- This can be triggered in multiple scenarios. e.g. without a `current_user` object, so you should check for that and don't do anything if missing
- Initially this will only get called upon real data imports (or first sync of a synchronization, then goes a different flow), so is fine to have a stat, but probably Common Data imports and/or syncs will trigger false positives (when importing common data, not when creating a table from a common data dataset, there is fine IMO)

config/app_config.yml.sample
```diff
@@ -92,6 +92,8 @@ defaults: &defaults
         wizard: ''
         filter: ''
         query: ''
+  segment:
```
Kartones: You'll need to talk with Systems team to deploy the staging & live config changes

lib/cartodb/segment_tracker.rb
```diff
@@ -0,0 +1,41 @@
...
+    def track_event(user_id, event, properties)
+      return unless enabled?
+
+      begin
+        @analytics.track(
+          user_id: user_id,
+          event: event,
+          properties: properties
+        )
+      rescue Exception => e
```
Kartones: Catching exception is not a good practice, use instead `rescue => e`

app/controllers/api/json/tables_controller.rb
```diff
@@ -1,4 +1,6 @@
 # coding: UTF-8
+
+#require 'cartodb/event_tracker'
```
Kartones: Add it as a `require_relative` but do it explicitly please (autoloading sometimes gets messy)

app/controllers/api/json/visualizations_controller.rb
```diff
@@ -1,5 +1,6 @@
 # encoding: utf-8
 require 'json'
+require 'cartodb/event_tracker'
```
Kartones: We tend to do `require_relative` for our requires and leave require for gems (to avoid autoload issues)

Kartones:
Apart from needing a master merge, I've expressed some some concerns, some improvements and I'd add a test, but don't know your time pressure, if needed could be a separate task.

Also you should test this at staging before releasing, I can help you deploying the branch but you need to talk with Systems for the config stuff.

app/controllers/api/json/tables_controller.rb
jmartinter: Every 'Created dataset' has an 'origin' parameter to identify how it was generated. Possible values are:
- blank (if created from scratch)
- copy (if created by duplicating)
- import
- common-data (if created from data library)

'Created map' doesn't have this parameter

jmartinter:
Code corrected with comments by @Kartones

jmartinter:
Need to reset resque to apply changes in staging/production (same issue as https://github.com/CartoDB/cartodb-central/pull/851)
CC/ @santisaez

Kartones:
Code looks fine now :+1:

jmartinter:
Needs new resque worker launch (assigned to new event queue) 
If not created, events are just ignored
https://github.com/CartoDB/cartodb-platform/issues/1230

PR #5945, opened 2015-10-15
4730 ogr geom possible names take3

rafatower:


rafatower:

This is not half what it was in the original https://github.com/CartoDB/cartodb/pull/5931, but it is tested in staging and should evolve as we get things fixed in gdal.

services/importer/lib/importer/ogr2ogr_params_helper.rb
```diff
@@ -0,0 +1,72 @@
...
+      def guessing_args
+        [
+          '-oo AUTODETECT_TYPE=YES',
+          quoted_fields_as_string_option,
+          x_y_possible_names_option,
+          '-s_srs EPSG:4326 -t_srs EPSG:4326',
+          '-skipfailure',
+          geom_possible_names_option,
+          keep_geom_columns_option
+        ].join(' ')
```
Kartones: +10 to this way of building a big command line string instead of an interpolated, doublequoted huge string

services/importer/lib/importer/ogrinfo.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      def executed?
+        @executed
+      end
+
+      def command
+        "#{DEFAULT_BINARY} #{arguments.join(' ')} #{@input_file_path} #{@layer}"
+      end
+
+      def arguments
```
Kartones: why not joining arguments inside the method as you do above with `guessing_args`? Just for consistency, either both returning a string array or the concatenated string, IMHO

services/importer/spec/unit/ogrinfo_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  describe '#geometry_type' do
+    it 'parses the geometry type as returned by ogrinfo' do
+      ogrinfo = ogrinfo_factory('TM_WORLD_BORDERS_SIMPL-0.3.shp', 'TM_WORLD_BORDERS_SIMPL-0.3')
+      ogrinfo.geometry_type.should == 'Polygon'
+    end
+    it 'raises an exception if the file could not be opened' do
+      ogrinfo = ogrinfo_factory('ThisFileDoesNotExist')
+      expect {ogrinfo.geometry_type}.to raise_error(CartoDB::Importer2::OgrInfoError)
+    end
+    it "returns 'None' with a CSV with the_geom (with current version)" do
```
Kartones: ok, I guess this are some of the errors you talk about. A pity but nice to have a test so when they get fixed it will complain.

Kartones:
:+1: my comments were either only comments or dumb suggestions about code style so... go ahead

services/importer/spec/unit/ogrinfo_spec.rb
rafatower: To give a bit more of background:

Actually that's ok with ogrinfo from gdal 1.11 (used in that code). With a more recent ogrinfo it could analyze with the same guessing options, so figure out a `the_geom` can potentially contain wkt, geojson and then report as such and build the arguments to ogr2ogr in such a way there's no need for extra massaging in the `Georeferencer`.

But when used that way I found ogr crashed badly under some circumstances, which prevented me from fully deploying the whole thing.

If you look carefully this branch, you would notice the params helper is pretty dummy.

But I left the scaffolding code plus the tests with the hope of merging the other branch. They kind of serve as "documentation" to anyone touching any of the zillion ogr params :smile:

PR #5950, opened 2015-10-16
Fixed organization user maps counter

### LINKED ISSUE #5786: Organization users maps counter doesn't show right number
Might be related to https://github.com/CartoDB/cartodb/issues/5503
When the organization owner checks the basic data of the rest of members, the map counter shows a wrong name.

screen shot 2015-09-28 at 22 24 44

MariaCheca:
Fixes #5786 
The organization user maps counter now shows the number of **public** maps the user has.

screen shot 2015-10-16 at 14 20 32

CR @Kartones


xavijam:
Backend review -> @kartones or @juanignaciosl please.
Also, not sure if we only want to show public stats, it is likely we want everything, public and private stats.

Kartones:
This is probably wrong...

- maps_count used to be the total maps, both public and private. To display at the user dashboard, etc. (and I guess at the organization management page). But here the term maps means "ruby model maps", not "editor maps". Editor maps == visualizations of type derived. Model maps == Visualizations of **any** type (because all of them have a map and each map is unique per vis). [This deleted lines](https://github.com/CartoDB/cartodb/pull/5950/files#diff-c6081688a012bc9178036ba7f2153227L157).
- public_visualizations_count mean... what the name implies: editor maps aka visualizations of type derived whose privacy is public. This count is meant for the user public profile page (and I guess any public listing, like his/her feed, or some public user listing)

So, the old counter showed all "maps and datasets" regardless of the privacy, the new one shows only "public Maps". If this is correct, go ahead, but you should think where this counter is going to be displayed (and I guess it is at an internal group management page so... smeels wrong to not count private Maps too).

So depending on the intent, this is a :-1: or a :+1: 


PS: And if is right, adding a comment to the PR like "this count was wrong due to blablabla" helps a lot to have context for the review... 

PS2: I'd add also a test to demonstrate the desired behaviour of "User decorator should only count public visualizations blablabla"

jesusvazquez:
retest this please

app/queries/carto/visualization_query_builder.rb
```diff
@@ -20,10 +20,18 @@ def self.user_public_visualizations(user)
...
+  def self.user_all_visualizations(user)
+    self.user_all(user).with_type(Carto::Visualization::TYPE_DERIVED)
+  end
+
   def self.user_public(user)
     new.with_user_id(user ? user.id : nil).with_privacy(Carto::Visualization::PRIVACY_PUBLIC)
   end
 
+  def self.user_all(user)
+    new.with_user_id(user ? user.id : nil)
```
Kartones: this `user_all` is not a bit redundant? Check `user_public` just above, directly uses `with_user_id`, so I'd remove it

Kartones:
Apart from the redundant internal method, :+1:

app/queries/carto/visualization_query_builder.rb
MariaCheca: For this case I wanted to retrieve all kind of visualizations, not only the ones with public privacy (which `user_public` returns).

Kartones: I mean doing this:
```
def self.user_all_visualizations(user)
  new.with_user_id(user ? user.id : nil).with_type(Carto::Visualization::TYPE_DERIVED)
end
```

So you don't need `def self.user_all(user)` at all

MariaCheca:
retest this please

MariaCheca:
retest this please

MariaCheca:
retest this please

MariaCheca:
:sparkles:

PR #5968, opened 2015-10-20
Silence duplicate viz export messages

Kartones:
Upon AR patched reconnections, the transactionability replays the full viz deletion process, which includes exporting its backup. This generates noise in the error logging system, when it is actually not any critical error because the backup was commited before the error (or wouldn't be stored at the DB).

No need for new automated tests as the return value was just ignored, but I tested manually using the Rake.

@ethervoid CR please


ethervoid:
:shipit:

PR #5979, opened 2015-10-21
Provide an option to invalidate embedded maps in redis after a deploy

Kartones:
Closes https://github.com/CartoDB/cartodb-platform/issues/866


Kartones:
Tested at staging (outputing embed redis cache hits/misses:
```
REDIS CACHE HIT: 
vis:ab191e54-606d-11e5-99c8-040146752f01:embed:https:4:3.22.3:4126147946
REDIS CACHE MISS: 
vis:ab191e54-606d-11e5-99c8-040146752f01:embed:https:4:3.22.3:4179801954
REDIS CACHE HIT: 
vis:ab191e54-606d-11e5-99c8-040146752f01:embed:https:4:3.22.3:4179801954

```

Kartones:
@rafatower CR please

Kartones:
Branch is actually ok: http://ci.cartodb.com/job/CartoDB-PR-testing/1542/console

Kartones:
retest this please

Kartones:
Also if you want, @rochoa & @alonsogarciapablo take a look to confirm this will solve the issues (I've tried and skipping Varnish it works as expected)

app/helpers/embed_redis_cache.rb
```diff
@@ -21,40 +22,79 @@ def get(visualization_id, https_request)
...
-    protocol = https_request ? 'https' : 'http' 
-    "visualization:#{visualization_id}:embed:#{protocol}:#{VERSION}"
+  def key(visualization_id, https_request = false)
+    protocol = https_request ? 'https' : 'http'
+    [
+      "vis",
+      visualization_id,
+      "embed",
+      protocol,
+      VERSION,
```
alonsogarciapablo: Why are we keeping the `VERSION`?

Kartones: I like it as a safeguard just in case we screw up something and need to bump up to a new version to unpollute cache, or we change some logic that returns a fragment of the key (that could return false positives/negatives), scenarios like those.

Call it paranoia, but once or twice in the past I've had to make use of it with caching.

app/helpers/embed_redis_cache.rb
```diff
@@ -21,40 +22,79 @@ def get(visualization_id, https_request)
...
+    @@key_fragment_fe_version ||= read_frontend_version
+  end
+
+  def embed_template_hash
+    # INFO: New deploys restart the server, invalidating this value + we want to hit disk as less as possible
+    @@key_fragment_embed_template_hash ||= calculate_embed_template_hash
+  end
+
+  def read_frontend_version
+    JSON::parse(File.read(Rails.root.join("package.json")))["version"]
```
alonsogarciapablo: What would happen if, for any reason, `package.json` or `app/views/admin/visualizations/embed_map.html.erb` no longer exist?

app/helpers/embed_redis_cache.rb
alonsogarciapablo: :+1:

app/helpers/embed_redis_cache.rb
Kartones: error will bubble up to get/set/invalidate, where there is a generic catch that returns `nil` (as if there was no cache). Plus logging the error, of course.

app/helpers/embed_redis_cache.rb
```diff
@@ -21,40 +22,79 @@ def get(visualization_id, https_request)
   rescue Redis::BaseError => exception
```
alonsogarciapablo: I guess we can get rid of this `rescue` block now.

app/helpers/embed_redis_cache.rb
alonsogarciapablo: :+1:

app/helpers/embed_redis_cache.rb
rafatower: +1

app/helpers/embed_redis_cache.rb
```diff
@@ -21,40 +22,79 @@ def get(visualization_id, https_request)
...
-                               })
+    serialized = JSON.generate(headers: response_headers,
+                               body: response_body
+                              )
     redis.setex(key(visualization_id, https_request), 24.hours.to_i, serialized)
-  rescue Redis::BaseError => exception
+  rescue => exception
+    # Captures:
+    # - Redis::BaseError if redis is down
+    # - IO errors due to deploys changing physical path (see read_frontend_version / calculate_embed_template_hash)
```
rafatower: are we gonna face the `Rails.root` issue here as well? I guess we'll have to wait and see.

that scares me a little :/

rafatower:
:+1: 
Just added a paranoid comment. I think it would be very, very improbable to hit the `Rails.root` issue

app/helpers/embed_redis_cache.rb
Kartones: Indeed, either a merge brought it back or I forgot to remove that one, but redundant :)

app/helpers/embed_redis_cache.rb
Kartones: Yes... that's why in case of error it will return nil as if hadn't cache (so it doesn't fails). The `Rails.root` issue is quite global, from Loggers, JS templates, file uploads... Maybe we should file a ticket to attack that, as it is out of this scope :/

But I share your fear, that's why I though about the best approach to handle that scenario and I think is this (just ignoring cache without erroring)

jesusvazquez:
retest this please

jesusvazquez:
retest this please

PR #5983, opened 2015-10-21
SAML Integration

vishalya:
Please integrate the code for the SAML based organization. This is based on the inputs from our previous discussion on. This implementation is generic.
Please refer to https://github.com/bloomberg/cartodb/wiki/SAML-integration for the requirement/design & the details.


vishalya:
@javisantana : hundci is not very happy with my PR. is there any other tool I pass my code and get it fixed?
But please take a look at the wiki and the diff and please send me your questions or comments.

javisantana:
@vishalya 
- @jesusvazquez can take a look what happened with the CI
- @Kartones @juanignaciosl please CR

jesusvazquez:
Yes tests went ok in the parallel environment but they didn't in the serial one. This is a known issue and we're trying to fix it today.

app/controllers/saml_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+ 
+    if user == nil
+
+      @organization = BBOrganization.new
+
+      @signup_errors[:saml_error] = ["Single sing on athentication failed."]
+
+      logger.error @signup_errors[:saml_error].first
+
+      render 'shared/signup_issue'
```
Kartones: it is better to reuse the `if user == nil` to do an `if ... render signup_issue else redirect_to` as the last method from an action should be a render or have the form `render 'shared/signup_issue' and return`

jesusvazquez:
retest this please

app/controllers/saml_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+    
+    redirect_to CartoDB.url(self, 'dashboard', {trailing_slash: true}, user)
+ end
+
+ def load_organization(saml_idp)
+
+    @organization = Carto::Organization.where(saml_idp_name: saml_idp).first 
+
+    logger.info "--- The organization is #{@organization.name}"
```
Kartones: I'd add `unless @organization.nil?` because else this can fail. (same below with @org_admin, is missing null check)

app/views/admin/organization_users/_form.html.erb
```diff
@@ -21,6 +21,24 @@
     </div>
   </div>
 
+
```
Kartones: @xavijam could you or somebody from FE check this is all in order, aligned with main CSS style, etc.?

app/controllers/saml_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    scope = user_info.cartodb_username
+
+    logger.info "uuid after calling Saml " + scope
+
+    user = authenticate!(:saml_header, {}, :scope => scope) unless authenticated?(scope)
+ 
+    if user == nil
+
+      @organization = BBOrganization.new
```
Kartones: This `BBOrganization` model is new? You should use standard `::Organization` / `Carto::Organization`, and if need to add custom fields maybe create a children organization class with the custom behaviour, but we cannot have in the code calls to a model that doesn't exists in the codebase.

app/controllers/saml_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      @signup_errors[:saml_error] = ["Single sing on athentication failed."]
+
+      logger.error @signup_errors[:saml_error].first
+
+      render 'shared/signup_issue'
+
+    end
+    
+    redirect_to CartoDB.url(self, 'dashboard', {trailing_slash: true}, user)
```
Kartones: Another reason to place inside the if-else, if user is nil we'd be rendering the signup issue, then trying to rendirect to the dashboard of `user=nil`

db/migrate/20151014154530_create_saml_identity_providers.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    create_table :saml_users do
+      String :saml_name_id, primary_key:true
+      foreign_key:idp_name, :saml_identity_providers, :type=>'text'
+      String :cartodb_username, null:false
+    end
+
+    alter_table :organizations do
+      add_foreign_key :saml_idp_name, :saml_identity_providers, :type=>'text'
```
Kartones: Would have to check to confirm, but shouldn't this also have a `null:true` so non-SAML organizations work?

app/models/saml_identity_provider.rb
```diff
@@ -0,0 +1,4 @@
+class SamlIdentityProvider <  Sequel::Model
+  include CartoDB::MiniSequel
```
Kartones: No need for this (main `User` model has it because some PG array stuff is not fully supported outside of Sequel)

app/models/saml_user.rb
```diff
@@ -0,0 +1,5 @@
+class SamlUser < Sequel::Model
+  include CartoDB::MiniSequel
```
Kartones: No need for this (main `User` model has it because some PG array stuff is not fully supported outside of Sequel)

app/views/admin/organization_users/_form.html.erb
xavijam: Checking changes in my local.

config/initializers/warden_saml.rb
```diff
@@ -0,0 +1,68 @@
+require 'yaml'
+
+Warden::Strategies.add(:saml_header) do
+
+  def valid?
+    true
```
Kartones: shouldn't you check presence of `params[:saml_idp]` to decide if this is valid? ([example](https://github.com/CartoDB/cartodb/blob/master/config/initializers/warden.rb#L130))

config/initializers/warden_saml.rb
```diff
@@ -0,0 +1,68 @@
...
+  def authenticate!
+    user_info = SamlAuthenticator.get_user_info(params)
+
+    user = SamlAuthenticator.check_user(user_info)
+    
+    user ? success!(user) : fail!
+  end
+end 
+ 
+class SamlAuthenticator
```
Kartones: Ruby good practices say 1 class per file, exceptions being custom Errors/Exceptions, which are usually located after the class to avoid excessive requires.

config/initializers/warden_saml.rb
```diff
@@ -0,0 +1,68 @@
...
+    saml_idp = SamlIdentityProvider.where(idp_name: idp_name).first
+    
+    settings.idp_cert = saml_idp.idp_cert
+    
+    settings
+  end
+
+
+  def self.check_user(user_data)
+    existing_user = User.where("username = '#{user_data.cartodb_username}'").first
```
Kartones: you can also use the syntax: `::User.where(username: user_data.cartodb_username).first`

But please, always namespace `User` (as `::User`) everywhere you use it as it can collide with `Carto::User` ActiveRecord model if not properly specified as being "the Sequel model"

config/initializers/warden_saml.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def self.get_saml_settings(idp_name)
+    settings = OneLogin::RubySaml::Settings.new
+
+    settings.soft = true
+    
+    saml_idp = SamlIdentityProvider.where(idp_name: idp_name).first
+    
+    settings.idp_cert = saml_idp.idp_cert
```
Kartones: what if `saml_idp` returns `nil`? I feel in general the code is too optimistic regarding not checking for nuls/non-existance of stuff.

config/routes.rb
```diff
@@ -20,7 +22,7 @@
   get   '(/user/:user_domain)(/u/:user_domain)/resend_validation_mail/:user_id' => 'account_tokens#resend',     as: :resend_validation_mail
   get   '(/user/:user_domain)(/u/:user_domain)/account_token_authentication_error' => 'sessions#account_token_authentication_error',     as: :account_token_authentication_error
 
-  get   '/login'           => 'sessions#new',     as: :login
+  get   '(/user/:user_domain)(/u/:user_domain)/login'           => 'sessions#new',     as: :login
```
Kartones: Why did you added this? Do you have a specific need? because being a login path has no user yet, etc.

config/routes.rb
```diff
@@ -346,6 +348,10 @@
     get '(/user/:user_domain)(/u/:user_domain)/api/v1/viz/:visualization_id/overlays'     => 'overlays#index',    as: :api_v1_visualizations_overlays_index,  constraints: { visualization_id: /[^\/]+/ }
     get '(/user/:user_domain)(/u/:user_domain)/api/v1/viz/:visualization_id/overlays/:id' => 'overlays#show',     as: :api_v1_visualizations_overlays_show,   constraints: { visualization_id: /[^\/]+/ }
 
+    # Synchronizations
+    get    '(/user/:user_domain)(/u/:user_domain)/api/v1_1/synchronizations/:id'          => 'synchronizations#show',     as: :api_v1_1_synchronizations_show
```
Kartones: Same here, this smells like a bad merge, so if you haven touched this routes please re-remove them

Kartones:
Apart from my comments around the code, something as sensible as a new warden authentication strategy should have some tests, both mocking the saml auth path and when there are missing components like no saml config or no requested user, to make sure the code doesn't throws errors.

app/views/admin/organization_users/_form.html.erb
xavijam: Looks great!

jesusvazquez:
Merge current master release into this branch to solve the CI issues.

db/migrate/20151014154530_create_saml_identity_providers.rb
vishalya: I think, that's the default. I am able to add a regular organization without any SAML IDP.
I can make it explicity 'null:true'.

Kartones: If works with a regular org, no need to make it explicit :)

config/routes.rb
vishalya: @vmanohar1 : This one came from Vinayak.
Vinayak : could you please comment?

vishalya:
@Kartones : thanks for the prompt review and appreciate the comments! 

I am planning to add the following tests: 
(1) A script to add a saml based org and it's admin user. This will be run before we start the APP.
(2) A curl based test with a fake IDP and mocked up SAML with failure case and successful case. 
This will need the APP running.
Let me know - in which directories they can reside.

Kartones:
- For tests that simulate requests, check https://github.com/CartoDB/cartodb/tree/master/spec/requests (and there is where they should live at).
- For unit/integration tests, we use https://github.com/CartoDB/cartodb/tree/master/spec/models, https://github.com/CartoDB/cartodb/tree/master/spec/lib, etc
- If you wish to add a script, we put them at https://github.com/CartoDB/cartodb/tree/master/script (if is regarding SAML would be nice to have something like `scripts/saml/xxxx.sh`)

app/controllers/saml_controller.rb
vishalya: BBOrganization : thought I had removed it, it's replaced by the load_organization   I need to remove some dead code, here. Sorry about the confusion.

config/initializers/warden_saml.rb
vishalya: yes, I think that makes sense. I will add it.

config/initializers/warden_saml.rb
vishalya: I am all about - good/better n best practices, will take care of it.

vishalya:
@jesusvazquez : should I rebase to pass the tests? I also see that you have requested for the 'retest', how can I do it?

jesusvazquez:
@vishalya yes you should rebase in order to avoid the CI errors. You can't retest because you're not a member of CartoDB organization. This is a security measure. 

If you need to retest, ping us and we will run the tests.

config/routes.rb
vmanohar1: This is needed for our domainless urls.. Otherwise when we click from login button, since the route does not exist, it will throw exceptions..

config/initializers/warden_saml.rb
vishalya: Added the check.

config/initializers/warden_saml.rb
vishalya: Using the ::User and new syntax.

config/routes.rb
vishalya: Yes, it looks like a bad merge, removing these lines.

app/controllers/saml_controller.rb
vishalya: Added the check!

app/controllers/saml_controller.rb
vishalya: I don't like the nested if then else, but definitely I see the issue of not being returned from the ```user == nil ```

app/models/saml_identity_provider.rb
vishalya: removed.

app/models/saml_user.rb
vishalya: removed.

config/initializers/warden_saml.rb
```diff
@@ -0,0 +1,68 @@
...
+   
+     if response.is_valid?
+       puts "The response is valid"
+   
+       user_data = SamlUser.where(saml_name_id: response.nameid).first
+       return user_data
+   
+     else
+   
+       puts "Response Invalid. Errors: #{response.errors}"
```
vishalya: done.

config/initializers/warden_saml.rb
```diff
@@ -0,0 +1,68 @@
...
+ 
+class SamlAuthenticator
+  def self.get_user_info(params)
+   
+     response = OneLogin::RubySaml::Response.new(params[:SAMLResponse])
+   
+     response.settings = get_saml_settings(params[:saml_idp])
+   
+     if response.is_valid?
+       puts "The response is valid"
```
vishalya: It's not really needed. removed.

config/routes.rb
vishalya: yes, bad merge. fixed it.

config/routes.rb
Kartones: uhm, but the login routes should always be `root_domain/login`, eg. `cartodb.com/login`, not subdomain-based, e.g. `kartones.cartodb.com/login`.

For subdomainless urls should allow to navigate to urls like `localhost.lan:3000/login`, it did in the past and else there is a regression bug that we should research.

Also, by not having the user subdomain specified, you get as a bonus "for free" subdomain based logins, so `kartones.cartodb.com/login` works because it matches to `/login` but the MVC engine doesn't stores `kartones` as `user_domain` as it would otherwise (and undesired for login page).

Kartones:
@vishalya have you pushed the changes? because the pull request code is still the old one despite your comments.

vishalya:
Not yet, this was just re-based version to see if the CI errors are fixed.
I will push the sometime today.

vishalya:
- Review comments are implemented (tests and hound fixes are TBD)
- Rebased with the latest master
- Please check the routes.rb, I see some routes being removed. 
- How to take care of the hound issues? do you guys use any special editors or scripts to fix these or it's manual labor?

Kartones:
About Hound issues, what we do is manual fixes, but the best approach is to install locally [rubocop](https://github.com/bbatsov/rubocop) code analyzer and work fixing all complaints before commiting/pushing the code. 
For example I use as my development IDE Sublime Text and I can plugin the style checks inside the editor using [SublimeLinter](http://sublimelinter.readthedocs.org/en/latest/) and [SublimeLinter-rubocop](https://github.com/SublimeLinter/SublimeLinter-rubocop).

Rubocop allows some `-a` flag that self-heals some issues but we haven't tried it with our codebase.

app/controllers/saml_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+ layout 'frontend'
+
+ # acs method
+ def initialize
+   @signup_source = 'SAML Sign-On'
+   @signup_errors = {}
+ end
+
+ def acs
+    logger.info "inside smal acs"
```
Kartones: I guess its a typo and you want to log `"inside saml acs"`

app/controllers/saml_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+    
+    redirect_to CartoDB.url(self, 'dashboard', {trailing_slash: true}, user)
+ end
+
+ def load_organization(saml_idp)
+
+    @organization = Carto::Organization.where(saml_idp_name: saml_idp).first 
+
+    @org_admin = Carto::User.where(id:@organization.owner_id).first unless @organization.nil?
```
Kartones: Does your intended saml auth flow contemplate having saml users without organization? Because if not, I'd make this code raise an exception if there is no `@organization` or `@org_admin` (and do a `rescue` in the `acs` to do the `render 'shared/signup_issue' and return`)

config/initializers/saml_autheticator.rb
```diff
@@ -0,0 +1,50 @@
...
+       return nil
+    end
+
+    settings.idp_cert = saml_idp.idp_cert
+
+    settings
+  end
+
+
+  def self.check_user(user_data)
```
Kartones: You should check user_data is not nil at least, because [this code flow](https://github.com/CartoDB/cartodb/pull/5983/files#diff-166a4c1747b7473bb3cf3dc5542b7416R13) can cause a call to `check_user` to be made with nil if `get_user_info`has invalid or missing data.

config/routes.rb
```diff
@@ -9,8 +9,10 @@
 # Can be also done at controller source files by using -> skip_before_filter :ensure_org_url_if_org_user
 
 CartoDB::Application.routes.draw do
+
   # Double use: for user public dashboard AND org dashboard
-  get   '/[(user/:user_domain)(u/:user_domain)]'                 => 'admin/pages#public', as: :root
+  post '/(user/:user_domain)(u/:user_domain)/:saml_idp/acs' => 'saml#acs'
+  get   '/(user/:user_domain)(u/:user_domain)'                 => 'admin/pages#public', as: :root
```
Kartones: Careful, Add your route above the comment on line 12/13, and leave original code as we bugfixed an issue and if you see on the deleted line the route optional params now come inside `[]` brackets.

config/routes.rb
```diff
@@ -525,14 +526,9 @@
   scope :module => 'superadmin', :format => :json do
     get '/superadmin/get_databases_info' => 'platform#databases_info'
     get '/superadmin/stats/total_users' => 'platform#total_users'
-    get '/superadmin/stats/total_pay_users' => 'platform#total_pay_users'
```
Kartones: this looks like a bad merge, please add them again (when in doubt, please check cartodb master as reference): https://github.com/CartoDB/cartodb/blob/master/config/routes.rb#L528

app/controllers/saml_controller.rb
vishalya: agreed, I think the signup_issue needs organization in order to correctly render itself. Is there any other generic error page which I can use?

Kartones: I think you can use that same one and just place different error messages inside, check the source code of the template as I recall something like that being available

vishalya: I had to add small check in sign_issue template for organization != nil & I can reuse the same template.

config/initializers/saml_autheticator.rb
vishalya: agreed! done.

vishalya:
Apart from the hound and tests, it should have all the other comments implemented.

vishalya:
Sorry, I got pulled into some other work, will resume work on the following 
(1) Hound errors
(2) Adding the Tests

javisantana:
we added 3rd party auth though headers (which is a more general approach)

PR #5986, opened 2015-10-22
5981 missing event

jmartinter:
Fixes:
 - it generates a 'Delete map' events in case a deleted dataset has dependent visualizations.

Improves:
 - check if provided user is valid before queuing a new event
 - add exception message to Rollbar call properties

@Kartones


Kartones:
:+1:

Issue #5932, opened 2015-10-13
Remove dead files


Kartones:
create_test_user script
drop_all.sql

Branch: `dead-code-cleanup`

Kartones:
create_test_user script
drop_all.sql

PR #5987, opened 2015-10-22
Remove dead files/scripts

Kartones:
Closes https://github.com/CartoDB/cartodb/issues/5932


Kartones:
@ethervoid CR please (checked with systems that this scripts are no longer used)

ethervoid:
:shipit:

PR #5998, opened 2015-10-23
Removing user mover spec from Makefile

jesusvazquez:
Trying to fix ci issues.


jesusvazquez:
@Kartones I think we can merge this now.

Kartones:
ok :(

PR #6000, opened 2015-10-23
Unicorn workers restart

Kartones:
Relates to https://github.com/CartoDB/cartodb-platform/issues/1214

As there are really few options for the combo Ruby 1.9.3 + Unicorn other than upgrading to Ruby 2.0 or higher (where the GC is better and there are a lot of memory profilers), using the unicorn-worker-killer avoids memory high memory usage (and possible memory leaks). 

It is a patch for the problem, not an answer, but considering the soon-to-be-made upgrade to Ruby 2.2 this is a quick way towards that.

It's been tested at staging and works fine, gracefully killing and restarting the workers without interfering with requests and without any loss of them.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,6 +37,7 @@ gem 'net-ldap',                '0.11'
 
 group :production, :staging do
   gem 'unicorn',               '4.8.2'
+  gem 'unicorn-worker-killer'
   gem 'raindrops',             '0.12.0'
 end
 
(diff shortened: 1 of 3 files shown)


Kartones:
@luisbosque @zenitraM As @santisaez is on vacations, if you think this short-term approach is fine, we can safely release this branch, as by default the unicorn gem is not used (unless you comment the `config.ru` lines)

luisbosque:
+1

zenitraM:
:+1:

PR #6008, opened 2015-10-26
GPX multilayer

### LINKED ISSUE #5974: Multi-layer import for GPX files
Implement multi-layer import for GPX files, pretty much the same as for other data sources such as zip and KML files.

As the code is now, this would need another splitter (see `KmlSplitter` and co.).

ethervoid:
Closes #5974 

GPX Splitter to import multi layer GPX files to create multilayer maps with it. 

If you import one file with "one layer" its going to create a two layer map or two datasets. For example if you have a file with a `<trk>` block, **ogr2ogr** create two layers with data: **track** and **track_points**. `track` have the the `multilinestring` for the points and `track_points` all the points gathered by the GPS

### DIFF
--- a/services/importer/lib/importer/gpx_splitter.rb
+++ b/services/importer/lib/importer/gpx_splitter.rb
@@ -1,0 +1,78 @@
+# encoding: utf-8
+require 'open3'
+require_relative './source_file'
+require_relative './unp'
+require_relative './exceptions'
+
+module CartoDB
+  module Importer2
+    class GpxSplitter
+      MAX_LAYERS = 50
+      GPX_LAYERS = ['waypoints','routes','tracks','route_points','track_points']
+      ITEM_COUNT_REGEX = 'Feature Count:\s'
+      def self.support?(source_file)
+        source_file.extension == '.gpx'
+      end
+
+      def initialize(source_file, temporary_directory)
+        @source_file          = source_file
+        @temporary_directory  = temporary_directory
+      end
+
+      def run
+        n_layers = layers_in(source_file).length
+        return self if n_layers <= 1
+        raise CartoDB::Importer2::TooManyLayersError.new("File has too many layers (#{n_layers}). Maximum number of layers: #{MAX_LAYERS}") if n_layers > MAX_LAYERS
+        @source_files = source_files_for(source_file, layers_in(source_file))
+        self
+      end
+
+      def source_files
+        return [source_file] unless multiple_layers?(source_file)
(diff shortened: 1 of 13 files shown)


ethervoid:
Please @rafatower check this PR :)

services/importer/lib/importer/gpx_splitter.rb
```diff
@@ -0,0 +1,86 @@
...
+require_relative './exceptions'
+
+module CartoDB
+  module Importer2
+    class GpxSplitter
+      MAX_LAYERS = 50
+      GPX_LAYERS = ['waypoints', 'routes', 'tracks', 'route_points', 'track_points']
+      ITEM_COUNT_REGEX = 'Feature Count:\s'
+      OGRINFO_BINARY = 'ogrinfo'
+      DEFAULT_OGR2OGR_BINARY = 'ogr2ogr2'
```
rafatower: couldn't we pick this from Ogr2ogr class?

rafatower: nevermind

services/importer/spec/fixtures/multiple_layer.gpx
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<gpx xmlns="http://www.topografix.com/GPX/1/1" version="1.1">
+    <metadata>
+        <name>Multi layer Test file</name>
```
rafatower: just FYI: there are other gpx test files lurking in the repo

rafatower:
Looks good to me.

Just one quick question: in the description of this PR you say that it will pick "track and track_points", but nothings prevent it from importing other layers or with other names, right? (though I think those are standard names).

ethervoid:
Yes, the layer names are standard AFAIK. You can add more elements using the extensions but I don't think this should be treated as new layers. [schema](http://www.topografix.com/GPX/1/1/gpx.xsd)

When you import a GPX with only <trk> (for example) date, OGR is going to create a layer for tracks (the line defining the track) and track_points (with all the points)

rafatower:
Mmm, the `multilinestring` is maybe what required the `promote_to_multi` dreaded param? mental note to myself for testing when/if we remove it

ethervoid:
Nope (for what I know), in the [GDAL doc](http://www.gdal.org/drv_gpx.html) don't say nothing about the need of `promote_to_multi` to import GPX

PR #6016, opened 2015-10-26
Restrict asset upload

Kartones:
Closes https://github.com/CartoDB/cartodb-management/issues/4289


Kartones:
@rafatower CR please

app/models/asset.rb
```diff
@@ -6,7 +6,9 @@ class Asset < Sequel::Model
 
   KIND_ORG_AVATAR = 'orgavatar'
 
-  PUBLIC_ATTRIBUTES = %W{ id public_url user_id kind }
+  PUBLIC_ATTRIBUTES = %w{ id public_url user_id kind }
+
+  VALID_EXTENSIONS = %w{ jpg jpeg gif png svg }
```
rafatower: who can confirm this is the complete list of assets valid types?

Kartones: @xavijam , spoke with him but just in case summoning him

rafatower:
:+1:

app/models/asset.rb
xavijam: IMHO, those are the only types we should allow to upload.

Kartones: Yep, this are the ones on the UI literal listing allowed images. BMP and TIFF are too big in general for markers or BGs, and we don't support here compressed files so... I also think it is more than enough

PR #6023, opened 2015-10-27
Hardened s3 import rake

Kartones:
Relates to RB 8582

### DIFF
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -1,6 +1,5 @@
 # encoding: utf-8
 require 'fileutils'
-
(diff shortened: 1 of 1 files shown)


Kartones:
@ethervoid CR please

Real source of error is unknown as the error was [just outputting the data import item id](https://github.com/CartoDB/cartodb/pull/6023/files#diff-9028707827a598e003d6ad79182e831dR38), so I've made the rescue blocks resilient to wrong? data_import_item contents, plus sending a `.inspect` so we learn what it comes inside.

ethervoid:
Oki :+1:

PR #6026, opened 2015-10-27
Added country code filter for visualizations

ethervoid:
As a fix for the bounding box problem with the dateline @matallo suggested tu use ISO country codes to "tag" the visualizations so we can filter the tables based on that.

Please @rafatower check this PR


rafatower:
:+1:

javisantana:
not agree with this fix, this is really tied to some concrete use case and does not fix the problem at all

PR #6027, opened 2015-10-27
Map not created after importing dataset from synced URL dataset

### LINKED ISSUE #5806: Map not created after importing dataset from synced URL dataset
This problem is reproduced:

1. From maps view in dashboard click "New map"
2. Click "Connect dataset"
3. Import from URL (for example: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/physical/ne_110m_coastline.zip)
4. Choose to sync every hour

The dataset is imported but the map is not created.

cc @xavijam

Kartones:
Closes #5806


Kartones:
@xavijam CR please

xavijam:
:+1:, thanks mate!

PR #6034, opened 2015-10-27
Propagates UserTable privacy to Table 6021

### LINKED ISSUE #6021: When importing through IMPORT API viz privacy is not set as dataset privacy is set on upload
According to the referenced commit:

https://github.com/CartoDB/cartodb/pull/5022

"the privacy of the viz is the same as whatever the privacy param was set to."

I can't see that behavior on the latest version. Setting a dataset's privacy does work but when create_vis is set to "true" on the import api the resulting visualization is always set as private regardless of the privacy setting specified through the IMPORT API call.

gfiorav:
Fixes #6021

There was a typo where the privacy was changed for `UserTable` but not saved.

To Do:
- [x] CR ( @ethervoid )
- [x] Modify test to check for this
- [x] Staging tests

cc/ @cesarandreslopez

### DIFF
--- a/app/models/table.rb
+++ b/app/models/table.rb
@@ -422,6 +422,8 @@ class Table
           raise "Error: User '#{self.owner.username}' doesn't have private tables enabled"
         end
         @user_table.privacy = @data_import.privacy
+        privacy = @data_import.privacy
+        save
       end
 
       decorator = CartoDB::Datasources::Decorators::Factory.decorator_for(@data_import.service_name)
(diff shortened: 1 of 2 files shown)


gfiorav:
Ready for CR @ethervoid :godmode:

app/models/table.rb
```diff
@@ -424,6 +423,8 @@ def after_create
         @user_table.privacy = @data_import.privacy
       end
 
+      save
```
ethervoid: Here we are saving but this is the after_create, not the after_save (even there you can't do this). Maybe you want to do `@user_table.save` ?

ethervoid:
Check the comment before deploy

app/models/table.rb
gfiorav: You're right!

PR #6048, opened 2015-10-28
Adds get_unused_port method to helper and applies it in serve_file 5909

### LINKED ISSUE #5909: Errno::EADDRINUSE: (Test parallelization)
**spec/requests/api/imports_spec.rb**
```
Failures:

  1) Imports API raises an error if the user attempts to import tables when being over quota
     Failure/Error: serve_file(Rails.root.join('spec/support/data/ESP_adm.zip')) do |url|
     Errno::EADDRINUSE:
       Address already in use - bind(2)
     # ./spec/support/helpers.rb:30:in `new'
     # ./spec/support/helpers.rb:30:in `serve_file'
     # ./spec/requests/api/imports_spec.rb:154:in `block (2 levels) in <top (required)>'

  2) Imports API raises an error if the user attempts to import tables when being over disk quota
     Failure/Error: serve_file(Rails.root.join('spec/support/data/ESP_adm.zip')) do |url|
     Errno::EADDRINUSE:
       Address already in use - bind(2)
     # ./spec/support/helpers.rb:30:in `new'
     # ./spec/support/helpers.rb:30:in `serve_file'
     # ./spec/requests/api/imports_spec.rb:167:in `block (2 levels) in <top (required)>'

Finished in 1 minute 28.03 seconds
13 examples, 2 failures, 1 pending
```

gfiorav:
Fixes #5909

Now `serve_file` method in helper uses new `get_unused_port` method to find an unused port instead of using a fix port.

This prevents `EADDRINUSE` from rising when two tests using `serve_file` are executed in parallel.

Please CR @rafatower 

cc/ @jesusvazquez

### DIFF
--- a/spec/support/helpers.rb
+++ b/spec/support/helpers.rb
@@ -26,11 +26,13 @@ module HelperMethods
   end
 
   def serve_file(file_path, options = {})
+    port = get_unused_port
+
     require 'webrick'
     server = WEBrick::HTTPServer.new(
       :AccessLog       => [],
       :Logger          => WEBrick::Log::new("/dev/null", 7), #comment this line if weird things happen
-      :Port            => 9779,
+      :Port            => port,
       :DocumentRoot    => File.dirname(file_path),
       :RequestCallback => Proc.new() { |req, res|
         options[:headers].each { |k, v| res[k] = v } if options[:headers].present?
@@ -45,7 +47,7 @@ module HelperMethods
     a = Thread.new { server.start }
 
     begin
-      yield "http://localhost:9779/#{File.basename(file_path)}" if block_given?
+      yield "http://localhost:#{port}/#{File.basename(file_path)}" if block_given?
     rescue => e
       raise e
     ensure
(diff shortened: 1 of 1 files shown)


spec/support/helpers.rb
```diff
@@ -56,6 +58,16 @@ def serve_file(file_path, options = {})
     end
   end
 
+  def get_unused_port
+    used_ports_command = `netstat -tln | tail -n +3 | awk '{ print $4 }' | cut -f2 -d ':'`
+    used_ports = used_ports_command.split("\n").map(&:to_i)
+
+    (10000..65535).each do |port|
```
rafatower: scanning that many ports looks too much

how about choosing a random one in that range and giving up after 10 retries?

spec/support/helpers.rb
```diff
@@ -45,7 +47,7 @@ def serve_file(file_path, options = {})
     a = Thread.new { server.start }
 
     begin
-      yield "http://localhost:9779/#{File.basename(file_path)}" if block_given?
+      yield "http://localhost:#{port}/#{File.basename(file_path)}" if block_given?
```
rafatower: good catch

spec/support/helpers.rb
gfiorav: The idea was to be deterministic. In reality, no more than 10-15 ports are used an this will 99.99% be done on the first iteration. Also 65535 is the maximum a 16 bit number can have so it's not magic :)

spec/support/helpers.rb
gfiorav: :smile:

spec/support/helpers.rb
rafatower: the whole thing is a bit tricky, but it is ok for just testing.

gfiorav: Agreed. Only one system call though per method invocation, so no major drama if a lot of ports are used (which will probably never happen).

Issue #6052, opened 2015-10-29
Add primary key to Likes table


zenitraM:
ActiveRecord needs a PK even on that table, else UserMover fails if user has Likes

cc @Kartones

PR #6054, opened 2015-10-29
Add primary key to Likes table

Kartones:
Closes https://github.com/CartoDB/cartodb/issues/6052

Respects the unique constraint of actor-subject, but adding the new ID


Kartones:
@zenitraM CR please (tested already up & down at local, with a table containing likes)

zenitraM:
:+1:

PR #6063, opened 2015-10-30
[Metrics] 5871 trending map event

jmartinter:
Adds:
 - Sends new 'Scored trending map' event to segment.io from the rake task that notifies users by email

@ethervoid


Kartones:
+1, but as you mentioned @ethervoid assigning him too

ethervoid:
:+1:

PR #6069, opened 2015-11-02
5074 fix data import specs

### LINKED ISSUE #5074: Fix or delete ./spec/models/data_import_spec.rb
Currently not being executed in `Makefile`.

Quick fix for some tests:
```patch
diff --git a/spec/models/data_import_spec.rb b/spec/models/data_import_spec.rb
index 1c2ac79..9624e76 100644
--- a/spec/models/data_import_spec.rb
+++ b/spec/models/data_import_spec.rb
@@ -2,14 +2,15 @@
 require_relative '../spec_helper'
 
 describe DataImport do
+
   before(:each) do
     User.all.each(&:destroy)
+    stub_named_maps_calls
     @user = create_user(:username => 'test', :email => "client@example.com", :password => "clientex")
     @table = create_table :user_id => @user.id
   end
 
   after(:all) do
-    stub_named_maps_calls
     @user.destroy
   end
```

but there are other tests broken. cc/ @juanignaciosl

Kartones:
Fixes #5074

### DIFF
--- a/spec/models/data_import_spec.rb
+++ b/spec/models/data_import_spec.rb
@@ -42,20 +42,6 @@ describe DataImport do
     data_import.error_code.should == 8004
   end
 
-  it 'should allow to append data to an existing table' do
-    pending "not yet implemented"
-    fixture = '/../db/fake_data/column_string_to_boolean.csv'
-    expect do
-      DataImport.create(
-        user_id: @user.id,
-        table_id: @table.id,
-        data_source: fixture,
-        updated_at: Time.now,
-        append: true
-      ).run_import!
-    end.to change { @table.reload.records[:total_rows] }.by(11)
-  end
-
   it 'raises a meaningful error if over storage quota' do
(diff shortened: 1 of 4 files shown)


Kartones:
retest this please

Kartones:
@rafatower CR please

spec/models/data_import_spec.rb
```diff
@@ -41,29 +42,15 @@
     data_import.error_code.should == 8004
   end
 
-  it 'should allow to append data to an existing table' do
```
rafatower: Is this gone forever? :((

Kartones: uhmm the test was pending, and right now we don't use the "append mode" right? Code is still around DataImport model but if we're not using nor testing it... at least test out I'd say.

But if you want to keep it I can revert that

rafatower:
:+1:

PR #6105, opened 2015-11-10
Fixed error on basemaps URLs

### LINKED ISSUE #6059: Duplication of attributions
Open a map and change the base map from positron to stamen or for that matter any other base map.
The attributions keep on appending.
Only the last base maps attributions should be available on the map.

acanimal:
reviewer @xavijam 

close #6059 

Invalid space within the basemap urls.


xavijam:
Yay! :) :+1:

PR #6119, opened 2015-11-12
data library: finished pending tasks

matallo:
- made dataset urls relative to `cartodb.com`
- replaced font to proxima nova

CR @xavijam


xavijam:
I would try to differentiate between environments in order to make data-library work in development, staging and production. Example: pseudocode.
```
  if (development) {
    base_url = ":username." + whatever
  } else if (staging) {
    base_url = "cartodb.com"
  } else {
    base_url = "cartodb.com"
  }
```

app/controllers/data_library_controller.rb
```diff
@@ -9,16 +9,10 @@ class DataLibraryController < ApplicationController
   def index
     render_404 and return if @viewed_user.nil? || !@viewed_user.has_feature_flag?("data_library")
 
-    respond_to do |format|
-      format.html { render 'index' }
-    end
-  end
+    @base_url = "#{request.protocol}#{CartoDB.account_host}"
```
xavijam: This could work in production and in staging but, what about development?
Also, we are setting "/dataset" within the url and that won't work in your development stack (i guess).

matallo: so, how can we solve this in order to make it relative to "cartodb.com"? it's the best solution I found, after your previous comments

xavijam: what about adding in the controller:
```
if development
 @base_url = ":username.:account_host/table/“
else if staging
 @base_url = "cartodb-staging.com/datatet”
...
```

matallo:
check now @xavijam

xavijam:
:+1: Much better buddy

PR #6129, opened 2015-11-13
Disable asset spec from parallel execution

jesusvazquez:


jesusvazquez:
@gfiorav As soon as this is green we can merge it to master. The reason can be found [here](https://github.com/CartoDB/cartodb/issues/5906)

gfiorav:
:+1:

PR #6141, opened 2015-11-17
Enabled invalidation duration log in python

luisbosque:
Enable invalidation log inside the invalidation service plpython trigger. The logging is done through syslog. It's a more natural way to do it if postgresql is already configured to log through syslog.

Note that the format is just a string representation of a python dict. Not sure how expensive can be converting a dict to json to string every time a invalidation occurs.

I've taken the chance to write some more info that may help to debug in case of problems like the number of retries and error string.

/cc @rochoa @javisantana @rafatower @zenitraM

### DIFF
--- a/app/models/user/db_service.rb
+++ b/app/models/user/db_service.rb
@@ -1270,7 +1270,9 @@ module CartoDB
                 current_time = time.strftime("%Y-%m-%d %H:%M:%S %Z", time.localtime())
                 session_user = plpy.execute("SELECT session_user", 1)[0]["session_user"]
                 invalidation_result = {'timestamp': current_time, 'duration': round(invalidation_duration, 8), 'termination_state': termination_state, 'retries': retries, 'error': error, 'database': '#{@user.database_name}', 'table_name': table_name, 'dbuser': session_user}
-                syslog.syslog(syslog.LOG_INFO, "invalidation: %s" % invalidation_result)
+                
+                if trigger_verbose:
+                  syslog.syslog(syslog.LOG_INFO, "invalidation: %s" % invalidation_result)
 
             $$
             LANGUAGE 'plpythonu' VOLATILE;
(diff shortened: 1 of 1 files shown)


zenitraM:
Would it make sense to make this configurable? so by default syslog logging is disabled

luisbosque:
Agree. That's a nice addition

app/models/user/db_service.rb
```diff
@@ -1241,7 +1262,16 @@ def create_function_invalidate_varnish_invalidation_service
                       if critical:
                         plpy.error('Invalidation Service error: ' +  str(err))
                       break
+                    retries = retries + 1
                     retry -= 1 # try reconnecting
+                  
+                end = time.time()
+                invalidation_duration = (end - start)
+                current_time = time.strftime("%Y-%m-%d %H:%M:%S %Z", time.localtime())
```
rochoa: What about using `time.gmtime()`?

luisbosque: I'm using localtime instead because it seemed to me more consistent with
other logs that just rely on the system local time.
Another way would be capturing the timezone used by the running postgresql
cluster and using it.

On Tue, Nov 17, 2015 at 2:38 PM, Raul Ochoa <notifications@github.com>
wrote:

> In app/models/user/db_service.rb
> <https://github.com/CartoDB/cartodb/pull/6141#discussion_r45059894>:
>
> >                      retry -= 1 # try reconnecting
> > +
> > +                end = time.time()
> > +                invalidation_duration = (end - start)
> > +                current_time = time.strftime("%Y-%m-%d %H:%M:%S %Z", time.localtime())
>
> What about using time.gmtime()?
>
> —
> <https://github.com/CartoDB/cartodb/pull/6141/files#r45059894>.
>



-- 
Luis Bosque
@luisico

www.cartodb.com
Map, analyze and build applications with your data

luisbosque: What do you think about this @rochoa?

rochoa: I though we always logged everything in UTC time, but if ES later can get the `%Z` I'm fine with this.

PR #6151, opened 2015-11-18
Escaping quotes in available_geometries query

iriberri:
## Description:

This issue was reported at our Data Services repo: https://github.com/CartoDB/data-services/issues/161.
The problem happens explicitly in the  `postal_code` geocoding process when the country input information is obtained from a column (a free text is correctly sanitized by the current code).

When one of the countries in that column contains a `' ` character, which happens for Cote d'Ivore, the following line crashed:

https://github.com/CartoDB/cartodb/blob/master/app/controllers/carto/api/geocodings_controller.rb#L71

as the code was generating an array with the countries in the column without escaping the values inside. This was generating something like:

````
SELECT (admin0_available_services(Array['Spain', 'Cote d'Ivore'])).*
`````

Which provokes a clear syntax error in Postgres.

## Proposed solution

The code gets the distinct values for the countries in a separate function and generated the contents of the SQL Array here:
````
 list = input.map{ |v| "'#{ v }'" }.join(",")
````

From: https://github.com/CartoDB/cartodb/blob/master/app/controllers/carto/api/geocodings_controller.rb#L69

This new code escapes single quotes inside the mapping substituting them by two single quotes, which is the way in Postgres to escape single quotes:

````
list = input.map{ |v| "'#{ v.to_s.gsub("'", "''") }'" }.join(",")
```
## Tests

I created a new test that generates a table whose content is syntactically incorrect for the SQL API and runs the available_geometries request. It expects to get a 200 response, which basically would mean that there's no syntax error in the query.

The test is running, but here's the behaviour of the test with the old code (notice the syntax error):

![image](https://cloud.githubusercontent.com/assets/1730320/11245899/800c9e36-8e15-11e5-9f02-6f3ce672860d.png)


cc @rafatower @jgoizueta 

Fixes https://github.com/CartoDB/data-services/issues/161

### DIFF
--- a/app/controllers/carto/api/geocodings_controller.rb
+++ b/app/controllers/carto/api/geocodings_controller.rb
@@ -66,7 +66,7 @@ module Carto
         return head(400) if input.nil? && params[:table_name].present?
         render(json: []) and return if input.nil? || input.empty?
 
-        list = input.map{ |v| "'#{ v.to_s.gsub("'", %q('')) }'" }.join(",")
+        list = input.map { |v| "'#{v.to_s.gsub("'", "''")}'" }.join(",")
 
         services = CartoDB::SQLApi.new({ username: 'geocoding', timeout: GEOCODING_SQLAPI_CALLS_TIMEOUT}).fetch("SELECT (admin0_available_services(Array[#{list}])).*")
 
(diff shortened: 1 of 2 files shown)


iriberri:
Okay, it's failing because the tests don't know about geocoding and it only runs in instances where the pg available_services query actually exists.

I'm going to set the response of the SQL API such as: 
```
      CartoDB::SQLApi.any_instance.expects(:fetch).with("SELECT (admin0_available_services(Array['myfreetext'])).*").returns([ ....])
```
As in the line I can control that the input is actually escaped.

jesusvazquez:
retest this please

iriberri:
@rafatower could you review, please?

rafatower:
:+1: good job!

PR #6152, opened 2015-11-18
Csv normalization 6116

### LINKED ISSUE #6116: CSV normalization doesn't support in-cell linebreaks
A CSV row may be split in multiple lines when quoted fields contain lines breaks. This is correctly supported by `ogr2ogr` (and, I assume, by Ruby's `CSV` module).

Handling such rows is checked by the specs ('imports records with cell line breaks') but the test is perfomed with a CSV file that doesn't require normalization, so this problem hasn't been detected so far.

CSV files are *normalized* when either:

* The encoding is different from UTF-8, ISO-8859-1/2
* The delimiter is not the comma
* It contains a single column (single-column rows are omitted during normalization)

In these cases Ruby CSV normalization is performed changing the  encoding to UTF-8 and the separator to `','` (and likely slowing noticeably the process along the way).

This Ruby normalization process reads each *line* and parses it with `CSV.parse_line`, silently ignoring anything that causes errors, which causes multiline rows to be dropped (an possibly other evil things).

jgoizueta:
This introduces support for line breaks inside cells in CSV normalization and also makes CSV normalization more consistent with imports without normalization:

* Row separators are detected and handle by the Ruby CSV module, which results in more consistency
  with direct ogr2ogr handling (e.g. `\r\n` separators where incorrectly handle by the normalization)
* Reverse line feeds in Windows-1252 encoding where previously removed for some reason
  (see f54a953345a17be45336f9065257c1eb629db34e). Now we do as ogr2ogr and just preserve them.

Fixes #6116

Some regression tests are added to specify the new desired behaviour.

Malformed CSV lines are being silently ignored as was done before. This is ok for some cases in which such lines are blank (e.g. some trailing `\n\r\n` found in some files). It would be nice to inform the user of non-blank dropped lines, so a counter has been introduced in the code to be used in some future feature.


jgoizueta:
Please @rafatower @ethervoid review this

ethervoid:
Did you make load tests with huge files? I tried to make something like this change in the past and the performance loss was significant.

services/importer/lib/importer/csv_normalizer.rb
```diff
@@ -125,13 +125,25 @@ def self.supported?(extension)
...
+
+        nonblank_ignored_lines = 0
+
+        CSV.open(filepath, "rb:#{encoding}", col_sep: @delimiter) do |input|
+          loop do
+            begin
+              row = input.shift
+              break unless row
+            rescue CSV::MalformedCSVError
+              nonblank_ignored_lines += 1 unless row.to_s.strip.empty?
```
ethervoid: What do we do with ignored lines that aren't empty ones?

services/importer/lib/importer/csv_normalizer.rb
```diff
@@ -125,13 +125,25 @@ def self.supported?(extension)
...
       def normalize(temporary_filepath)
-        temporary_csv = ::CSV.open(temporary_filepath, 'w', col_sep: OUTPUT_DELIMITER)
-        File.open(filepath, 'rb', external_encoding: encoding)
-        .each_line(line_delimiter) { |line|
-          row = parsed_line(line)
-          next unless row
-          temporary_csv << multiple_column(row)
-        }
+
+        temporary_csv = CSV.open(temporary_filepath, 'w', col_sep: OUTPUT_DELIMITER, encoding: 'UTF-8')
```
ethervoid: Is there any possibility to get an error trying to force the encoding to UTF-8. Below we have a fallback to Win-1252.

Other thing is that we should use constants instead of pain strings

rafatower:
There's some test failing:
```
09:58:43 Failures:
09:58:43 
09:58:43   1) gz and tgz regression tests returns error if csv is invalid with supported gzip file
09:58:43      Failure/Error: runner.results.first.error_code.should eq 1002
09:58:43        
09:58:43        expected: 1002
09:58:43             got: 2002
09:58:43        
09:58:43        (compared using ==)
09:58:43      # ./services/importer/spec/acceptance/gz_tgz_spec.rb:77:in `block (2 levels) in <top (required)>'
09:58:43 
09:58:43 Finished in 6 minutes 21.9 seconds
09:58:43 299 examples, 1 failure, 3 pending
09:58:43 
09:58:43 Failed examples:
09:58:43 
09:58:43 rspec ./services/importer/spec/acceptance/gz_tgz_spec.rb:66 # gz and tgz regression tests returns error if csv is invalid with supported gzip file
```

rafatower:
In general looks good to me.

I'd keep an eye on the stats once released.

But I do believe performance should improve, for the case where ruby was in charge of normalizing and putting everything into a huge line because of the wrong splitting with ill-detected line separators (most likely hitting mem and swapping in some cases).

jgoizueta:
Regarding the failing test: it tries to load a JPEG image as it was a CSV file. Is used to file with 1002 (bad file type) and now fails with 2002 (bad encoding).

The former error message may seem more accurate to the user, but this is what's happening:

Before we had the normalization going over the JPEG without error and ignoring all the "lines" because they are malformed, then ogr2ogr chokes on this empty file

Now we have the CSC normalization failing because the binary JPEG data is not valid in the detected encoding (or any other encoding for that matter).

So, the situation now is more reasonable, but the resulting error message may not be as clear to the user.

Since the error message was more accurate only by chance I think we should change the test and validate 2002 as the correct result with the new implementation. If we want to report cases such as this to the user as invalid file type instead of invalid encoding we'd need to add some kind of file type detection based on contents.

jgoizueta:
retest this, please

jgoizueta:
retest this, please

PR #6186, opened 2015-11-26
Rake to install geocoder client extension

ethervoid:
Added rake and model logic to install the geocoder extension to the passed user or organization:

- [x] Add `cdb_geocoder_client` schema to the search path
- [x] Install `plproxy` and `cdb_geocoder_extensions`
- [x] Add required geocoder server config

@rafatower pls check this out

### DIFF
--- a/app/models/user/db_service.rb
+++ b/app/models/user/db_service.rb
@@ -12,6 +12,7 @@ module CartoDB
       # Also default schema for new users
       SCHEMA_PUBLIC = 'public'
       SCHEMA_CARTODB = 'cartodb'
+      SCHEMA_CDB_GEOCODER = 'cdb_geocoder_client'
       SCHEMA_IMPORTER = 'cdb_importer'
       SCHEMA_GEOCODING = 'cdb'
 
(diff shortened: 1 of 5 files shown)


app/models/user/db_service.rb
```diff
@@ -392,6 +393,48 @@ def configure_extension_org_metadata_api_endpoint
...
+            db.run(geocoder_server_config_sql)
+          end
+        end
+
+        if !organization.nil?
+          org_users = organization.users
+          org_users.each do |u|
+            # TODO This part is a copy of the build_search_path method to avoid put the geocoder schema path
+            # until we open this to all the users
+            search_path = "\"#{u.database_schema}\", #{SCHEMA_CARTODB}, #{SCHEMA_CDB_GEOCODER}, #{SCHEMA_PUBLIC}"
```
rafatower: I just noticed something: probably this will fail if the `cdb_geocoder_client` extension is not installed. Worth testing it @ethervoid

ethervoid: I've check this before merge and don't rail. You could put in the `search_path` a schema that doesn't exists

rafatower:
@luisbosque roughly equivalent to other rake tasks to be applied on all users

app/models/user/db_service.rb
rafatower: never mind: right above is the extension creation. I'd maybe force a particular version to avoid trouble. Even if hardcoded.

ethervoid: Ok, I'll add it in other PR

PR #6204, opened 2015-12-02
Geocoder api user creation

ethervoid:
Added rake and model logic to install the geocoder extension to the passed user or organization:

- [x] Install and configure geocoder extension to a user when is created
- [x] Rakes to install the geocoder api extension to all the users and / or organization
- [x] Add the `entity_config` configuration (v2 but no harm is done to add here now)
- [x] Add the `_cdb_username()` function to get the username based on the schema (org users) or the config (non-org users)
- [x] Change the config part to fulfill the new config requeriments


ethervoid:

rafatower:
looks good! :+1:

PR #6207, opened 2015-12-03
Mobile panning issue

xavijam:
Just adding the fix applied into CartoDB.js (https://github.com/CartoDB/cartodb.js/pull/903).


xavijam:
cc @alonsogarciapablo who reviewed the change in cartodb.js :palm_tree:

PR #6213, opened 2015-12-04
Set include_all param to true in respond_to method

luisbosque:
Since ruby 2 the behavior of respond_to? method has changed. Now, it
only checks, by default, the public methods, but not private and
protected ones. In order to enable check of all methods, a second
parameter "include_all" must be enabled. More info in:
http://tenderlovemaking.com/2012/09/07/protected-methods-and-ruby-2-0.html

Please @rafatower and @juanignaciosl review


rafatower:
:+1:

PR #6231, opened 2015-12-09
Add password max length check

gfiorav:
CR please, @ethervoid

### DIFF
--- a/app/models/carto/user.rb
+++ b/app/models/carto/user.rb
@@ -10,7 +10,9 @@ class Carto::User < ActiveRecord::Base
   extend Forwardable
 
   MIN_PASSWORD_LENGTH = 6
+  MAX_PASSWORD_LENGTH = 64
   GEOCODING_BLOCK_SIZE = 1000
+
   # INFO: select filter is done for security and performance reasons. Add new columns if needed.
   DEFAULT_SELECT = "users.email, users.username, users.admin, users.organization_id, users.id, users.avatar_url," +
                    "users.api_key, users.database_schema, users.database_name, users.name, users.location," +
@@ -64,7 +66,7 @@ class Carto::User < ActiveRecord::Base
   end
 
   def password=(value)
-    return if !value.nil? && value.length < MIN_PASSWORD_LENGTH
+    return if !value.nil? && value.length < MIN_PASSWORD_LENGTH && value.length >= MAX_PASSWORD_LENGTH
 
     @password = value
     self.salt = new_record? ? service.class.make_token : ::User.filter(:id => self.id).select(:salt).first.salt
--- a/app/models/carto/user.rb
+++ b/app/models/carto/user.rb
@@ -66,7 +66,9 @@ class Carto::User < ActiveRecord::Base
   end
 
   def password=(value)
-    return if !value.nil? && value.length < MIN_PASSWORD_LENGTH && value.length >= MAX_PASSWORD_LENGTH
+    return if value.nil?
+    return if value.length < MIN_PASSWORD_LENGTH
+    return if value.length >= MAX_PASSWORD_LENGTH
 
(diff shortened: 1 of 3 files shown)


app/models/carto/user.rb
```diff
@@ -64,7 +66,9 @@ def name_or_username
   end
 
   def password=(value)
-    return if !value.nil? && value.length < MIN_PASSWORD_LENGTH
+    return if value.nil?
```
ethervoid: This is not true. You have to merge all the condition if not, this is going to be an OR clause

spec/models/user_spec.rb
```diff
@@ -407,6 +407,23 @@ def get_user_by_id(user_id)
     user.destroy
   end
 
+  it "should validate password presence and length" do
```
ethervoid: Same check in [the carto model too](https://github.com/CartoDB/cartodb/blob/bb68-add-password-max-length/spec/models/carto/user_spec.rb)

app/models/user.rb
```diff
@@ -372,9 +373,16 @@ def validate_password_change
     if @new_password != @new_password_confirmation
       errors.add(:new_password, "New password and confirm password are not the same")
     end
-    errors.add(:new_password, "Missing new password") if @new_password.nil?
-    if !@new_password.nil? && @new_password.length < MIN_PASSWORD_LENGTH
-      errors.add(:new_password, "New password is too short (6 chars min)")
+
```
ethervoid: I should refactor this and the `def password=` logic to one method due is the same logic

ethervoid:
Check the things I've pointed before merge it

app/models/carto/user.rb
gfiorav: yeah, tests are failing, I'll bump you when they actually pass, sorry

Thanks for pointing this out

spec/models/user_spec.rb
gfiorav: Will do

luisbosque:
retest this please

gfiorav:
See if you like it now, @ethervoid !

Thanks!

app/models/carto/user.rb
```diff
@@ -65,6 +67,7 @@ def name_or_username
 
   def password=(value)
     return if !value.nil? && value.length < MIN_PASSWORD_LENGTH
+    return if !value.nil? && value.length >= MAX_PASSWORD_LENGTH
```
ethervoid: Still wrong. If the value.length is greater than MIN_PASSWORD_LENGTH is going to return and don't check the MAX_PASSWORD_LENGTH

app/models/user.rb
```diff
@@ -458,13 +473,10 @@ def self.make_token
   end
 
   def password=(value)
-    if !value.nil? && value.length < MIN_PASSWORD_LENGTH
-      errors.add(:password, "must be at least #{MIN_PASSWORD_LENGTH} characters long")
-      return
-    end
+    return if !value.nil? && !valid_password?(:password, value, value)
```
ethervoid: This should be if !value.nil? && valid_password? isn't it?

app/models/carto/user.rb
gfiorav: They will never happen at the same time though. A password cannot be shorter than the min and longer than the max at the same time. Will refactor with an OR if you insist but there's not much of a point IMHO.

app/models/user.rb
gfiorav: No, I want to return without doing anything if the password is invalid. To check the password is invalid the password cannot be nil.

app/models/carto/user.rb
ethervoid: Ok I see your point. Let it be

ethervoid:
:+1:

PR #6235, opened 2015-12-10
Add more info when the import fail

### LINKED ISSUE #5693: [User feedback] Add more information into failed import emails
We just received a suggestion in which the user was asking for the email to contain:

- Name of the file :heavy_check_mark: 
- Name of the user: which I personally think it's a good idea in terms of user communication (and in case someone works in more than 1 account...)
- Format of the file: In this case, the error was due to an invalid support format which we didn't detected (a error HTML file), but wondering if in other cases we could detect it and add it to the email to make it more useful and for it to feel less "automatic".

Low priority, but nice idea for user care.

@ethervoid 
@rafatower

ethervoid:
Closes #5693 
Add more info to the email received by the user when an import fail

### DIFF
--- a/app/mailers/import_mailer.rb
+++ b/app/mailers/import_mailer.rb
@@ -1,13 +1,17 @@
+require 'json'
+
 class ImportMailer < ActionMailer::Base
   default from: Cartodb.get_config(:mailer, 'from')
   layout 'mail'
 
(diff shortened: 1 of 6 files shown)


ethervoid:
retest this please

ethervoid:
Please  @rafatower check this

rafatower:
:+1:

PR #6242, opened 2015-12-11
5463 register used geocoder in geocoder logs

### LINKED ISSUE #5463: register in geocoding log what geocoder is being used
We support many geocoders (here and gmaps so far), we should keep record of what is being used in a geocoding process.

cc @rafatower

ethervoid:
Closes #5463 

**Deploy first the migration**

Added geoder type info in:

- Geocoding logs
- Geocoding table with the field `geocoder_type`

### DIFF
--- a/spec/requests/carto/api/geocodings_controller_spec.rb
+++ b/spec/requests/carto/api/geocodings_controller_spec.rb
@@ -243,14 +243,25 @@ describe 'legacy behaviour tests' do
     end
 
     it 'returns started geocodings but not finished' do
-      geocoding1 = FactoryGirl.create(:geocoding, user: @user1, kind: 'high-resolution', created_at: Time.now, processed_rows: 1, state: 'started')
-      FactoryGirl.create(:geocoding, user: @user1, kind: 'high-resolution', created_at: Time.now, processed_rows: 1, state: 'finished')
+      geocoding1 = FactoryGirl.create(:geocoding, user: @user1, kind: 'high-resolution', created_at: Time.now,
+                                      processed_rows: 1, state: 'started')
+      FactoryGirl.create(:geocoding, user: @user1, kind: 'high-resolution', created_at: Time.now,
+                         processed_rows: 1, state: 'finished')
 
       get api_v1_geocodings_index_url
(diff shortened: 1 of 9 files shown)


db/migrate/20151211103229_add_geocoding_type_to_geocodings.rb
```diff
@@ -0,0 +1,13 @@
+Sequel.migration do
+  up do
+    alter_table(:geocodings) do
+      add_column :geocoder_type, :text
```
rafatower: This should go in a separate branch. Let's talk with systems about migrations (don't see an obvious risk but just in case)

spec/requests/carto/api/geocodings_controller_spec.rb
```diff
@@ -243,14 +243,25 @@ def remove_dates(geocoding_hash)
     end
 
     it 'returns started geocodings but not finished' do
-      geocoding1 = FactoryGirl.create(:geocoding, user: @user1, kind: 'high-resolution', created_at: Time.now, processed_rows: 1, state: 'started')
-      FactoryGirl.create(:geocoding, user: @user1, kind: 'high-resolution', created_at: Time.now, processed_rows: 1, state: 'finished')
+      geocoding1 = FactoryGirl.create(:geocoding, user: @user1, kind: 'high-resolution', created_at: Time.now.current,
```
rafatower: In order to fix the tests, I think it suffices to remove the `.current`, here and below.

db/migrate/20151211103229_add_geocoding_type_to_geocodings.rb
ethervoid: I usually leave the all in the "original" branch and when I'm going to deploy I make a cherry-pick of that commit (only the migration) and deploy it

ethervoid:
@rafatower Test passing, is this ok to deploy? (check migration)

rafatower:
LGTM

I think you conflicted with yourself :dancer:

ethervoid:
Yes, with the test change in the migration. No problem, working as expected

PR #6254, opened 2015-12-14
Add geocoder type column to geocodings table

ethervoid:
**Deploy this PR first, then deploy #6242**

Add new `geocoder_type` column to the `geocodings` table to store the type of used geocoder.


rafatower:
:+1:

PR #6259, opened 2015-12-14
Added default error in case there is no error at all in the log

### LINKED ISSUE #6136: Empty error code in "maximum attempts sync" email
I just received a "There was some problem while syncying your dataset" email in which the error code is empty:

image

ethervoid:
Closes #6136 

Added default error in case the synchronization log don't have any error code inside.


ethervoid:
@rafatower check this out

rafatower:
:+1:

PR #6261, opened 2015-12-14
Generate logs for created import in after_create callback

iriberri:
This code forces the generation of a log when the import has been created (whose corresponding status is `pending`).


iriberri:
Already tested in staging/local by importing a bunch of different files. :+1:

PR #6262, opened 2015-12-14
Added geocoder type to the geocodings log

ethervoid:
We need to add the geocoder type to the geocodings metrics in order to use it in Kibana


ethervoid:
@rafatower check this out

rafatower:
:+1

PR #6264, opened 2015-12-14
Activated geocoder api extension install for new users

ethervoid:
Last part to enable geocoder api install to all the users where are created


ethervoid:
@rafatower check this out

rafatower:
after-the-fact-okay

Issue #6263, opened 2015-12-14
Detect errors caused by exceeing max number of columns from sequel_pg


iriberri:
[Here](https://github.com/CartoDB/cartodb/blob/86c42a7306a8198d61ceb1fe6b7c3497e4ce3ce4/services/importer/lib/importer/ogr2ogr.rb#L96) we are detecting the OGR2OGR warning for more than 1600 columns, but sequel_pg is throwing the error when the number of columns is 256 due to the default restriction `SPG_MAX_FIELDS = 256`.

Therefore, the files with more than 255 columns but less than 1600 are reported with error code = 9999 (unknown) and the user doesn't get any useful feedback.

Besides recognizing the error itself, this also involves treating: 

* [ ] https://github.com/CartoDB/cartodb/issues/6251
* [ ] https://github.com/CartoDB/cartodb/issues/3975

iriberri:
At first sight, this error seems to come from geocoding guessing.

iriberri:
Okay, so: Let's generate a new exception from this issue that will be matched with the current TooManyColumns error code. We will update this message to say that max is 250 columns.

Issue #3975, opened 2015-06-08
Discrepancy in number of max tables in postgres


iriberri:
I was checking this file https://github.com/CartoDB/cartodb/blob/dc3cbba77d26c0acbcf5d70476199df76aaa5d55/services/importer/lib/importer/loader.rb#L240 and found things like:

```
raise TooManyColumnsError.new(job.logger) if ogr2ogr.command_output =~ /tables can have at most 1600 columns/
```

This explained error (and others) isn't being raised appropriately because Postgres [has a lower limit](http://www.postgresql.org/about/) which seems to depend on column type:

`Maximum Columns per Table    250 - 1600 depending on column types``

Also, the error is being raised by Postgres and not by ogr2ogr, so as we're using `ogr2ogr.command_output` we're not seeing some errors that are being caused after the ogr2ogr processing part.

@Kartones suggested that this may be caused by us using an old Postgres driver 

@pramsey any tricks? :-)

cc @javisantana

=====

PS:
Another example is for example the `DuplicatedColumnError`: when we detect columns with repeated names we rename them, but the following use case fails:
  * User has long column names, which are detected as different
  * We trim the column names 
  * In some cases, the beginning of the names are equal, so we're creating in our end columns with the same name. This import fails but the DuplicatedColumnError is not show.

pramsey:
I think it really might be 1600…

https://github.com/postgres/postgres/blob/master/src/include/access/htup_details.h#L47

iriberri:
Actually forgot to copy something important: the import log from a test:

2015-06-08 18:56:09 UTC: Disabling autovacuum for "cdb_importer"."importer_0221ef4c0e1011e5bebf0e8dde98a187"
2015-06-08 18:56:09 UTC: Errored importing data from /tmp/imports/20150608-15131-13gw6hr/carla.csv:
2015-06-08 18:56:09 UTC: RangeError: more than 256 columns in query
2015-06-08 18:56:09 UTC: --------------------------------------------

Kartones:
I see strange that if the limit is variable we hit exactly at 256 when you used simple columns (string? empty values right?). 

I raised the number in the past, I think that when we updated to PG 9.3, so I'm starting to "bet" that the issue is either our old Rails 3, or PostgreSQL Ruby gem or the ORM.

We use the following gem versions:
```
gem 'pg',                      '0.13.2'
gem 'sequel',                  '3.42.0'
gem 'sequel_pg',               '1.6.3'
```

[...]

PR #6267, opened 2015-12-14
Handle RangeError from sequel: too many columns for content guessing

iriberri:
Although Postgres and therefore [OGR2OGR allow up to 1600](https://github.com/CartoDB/cartodb/blob/86c42a7306a8198d61ceb1fe6b7c3497e4ce3ce4/services/importer/lib/importer/ogr2ogr.rb#L96) columns in a dataset (depending on type), `sequel_pg` [sets the maximum number to 256](https://github.com/jeremyevans/sequel_pg/blob/79c768d9508c55f55db3f20331b9631b85ff97f0/ext/sequel_pg/sequel_pg.c#L16) in `SPG_MAX_FIELDS = 256`. (More info: https://github.com/CartoDB/cartodb/issues/3975)

The `table_sampler` file used in content guessing was crashing when getting a sample from the table whenever the dataset contained more than 256 columns. This is now rescued by raising a new exception `TooManyColumnsGuessedError` (other naming suggestions are welcome!).

With guessing disables, the old `TooManyColumnsError` would take care of the number of columns to be inserted in the dataset.

Fixes https://github.com/CartoDB/cartodb/issues/6263

### DIFF
--- a/lib/cartodb/import_error_codes.rb
+++ b/lib/cartodb/import_error_codes.rb
@@ -153,7 +153,7 @@ module CartoDB
     },
     2004 => {
       title: 'Too many columns',
-      what_about: "Data has too many columns. You can only import up to 1600 columns. You can delete the columns you're not interested in, or split the file into smaller ones.",
+      what_about: "Data has too many columns. You can only import up to 250 columns. You can delete the columns you're not interested in, or split the file into smaller ones.",
       source: ERROR_SOURCE_USER
     },
     2005 => {
@@ -179,11 +179,6 @@ module CartoDB
     2009 => {
       title: 'KML without style Id',
       what_about: "The KML file you tried to import failed because a style element doesn't have an ID attribute. To fix this error, please open the file and add an ID to all the style tags.",
-      source: ERROR_SOURCE_USER
-    },
-    2010 => {
-      title: 'Too many columns for guessing',
-      what_about: "Data has too many columns to be guessed. The guessing process can only support up to 256 columns. Please, disable content guessing on import or reduce the number of columns in the file.",
       source: ERROR_SOURCE_USER
     },
     3007 => {
--- a/lib/cartodb/import_error_codes.rb
+++ b/lib/cartodb/import_error_codes.rb
@@ -181,6 +181,11 @@ module CartoDB
       what_about: "The KML file you tried to import failed because a style element doesn't have an ID attribute. To fix this error, please open the file and add an ID to all the style tags.",
       source: ERROR_SOURCE_USER
     },
+    2010 => {
+      title: 'Too many columns for guessing',
+      what_about: "Data has too many columns to be guessed. The guessing process can only support up to 256 columns. Please, disable content guessing on import or reduce the number of columns in the file.",
+      source: ERROR_SOURCE_USER
+    },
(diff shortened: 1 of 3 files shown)


iriberri:
@csobier could you please review the wording at my changes in `import_error_codes.rb`? :) Thanks! You can also see the same text in the image I pasted in this PR description.

iriberri:
cc @rafatower (or let me know if I should instead ping Mario to review)

lib/cartodb/import_error_codes.rb
```diff
@@ -181,6 +181,11 @@ module CartoDB
       what_about: "The KML file you tried to import failed because a style element doesn't have an ID attribute. To fix this error, please open the file and add an ID to all the style tags.",
       source: ERROR_SOURCE_USER
     },
+    2010 => {
+      title: 'Too many columns for guessing',
+      what_about: "Data has too many columns to be guessed. The guessing process can only support up to 256 columns. Please, disable content guessing on import or reduce the number of columns in the file.",
```
csobier: Suggested edits:
Is it related to the data, or the file size? That will determine if the below should say "The Data" or "The file". Also, is "content guessing on import" an actual setting?

title: "The Data contains too many columns"
description: "The importer process can only upload up to 256 columns from your file. Reduce the number of columns in the file, or disable the content guessing on import."

iriberri: Notice that the "too many columns" (the one that refers to 1600) is already defined above.

This is just in the case when content guessing is enabled. If content guessing is not enabled, you can surely import more than 256 columns (I have for example a table with 270).

rafatower:
I wouldn't say "for guessing" since sequel is part of our platform and it could fail at some other places. I.e: probably the editor won't work on such cases (but the SQL API will)

lib/cartodb/import_error_codes.rb
csobier: Ah, so if it is specific to this option, perhaps this title would work?

title: There are too may columns for the Connect Dataset importer to guess
description: The importer process can only upload up to 256 columns from your file (when the "Let CartoDB automatically guess data types and content on import" checkbox is enabled." You can disable this option from the Connect Dataset dialog box, or reduce the number of columns in the file.

iriberri:
@rafatower good point. I checked and (if I'm not wrong) the sampler is only being used right now for guessing (https://github.com/CartoDB/cartodb/blob/78922fb2e887ceda09f1acfd536ba2839d3ea08c/services/importer/lib/importer/content_guesser.rb#L152) although I totally agree that this should not be that specific.

What about a `TooManyColumnsRangeError` for the exception?

rafatower:
My point is changing the message because `Sequel` is used in other places. Particularly in the `UserTable` and `Table` service. See https://github.com/CartoDB/cartodb/blob/master/app/models/table.rb#L816

So probably updating from the editor (which uses that function) will fail due to the `Sequel` library limitation and it is not something that will change in the near future.

iriberri:
Paused until further limits discussions.

iriberri:
image

iriberri:
Finally tested in staging, @rafatower :-)

rafatower:
:+1:

Issue #6249, opened 2015-12-11
Importer error code 1010 still relevant?


csobier:
@iriberri , can you check if this [Importer Error code](http://docs.cartodb.com/cartodb-editor/error-codes-and-solutions/#importer-error-codes), and description, is still relevant?

![image](https://cloud.githubusercontent.com/assets/13319925/11749256/aad10836-9ffa-11e5-8626-7df2b3bfb45c.png)

PR #6268, opened 2015-12-14
Remove outdated GDrive exception

iriberri:
As CartoDB uses OAuth and doesn't rely anymore on the public sharing from GDrive (or Dropbox) this exception is outdated.

Closes https://github.com/CartoDB/cartodb/issues/6249


iriberri:
@ethervoid @rafatower any concerns about this?

ethervoid:
You have to remove this too https://github.com/CartoDB/cartodb/blob/master/lib/cartodb/import_error_codes.rb#L67

It seems ok, like there is no code using it

iriberri:
Oops. Done, thanks.

PR #6270, opened 2015-12-14
Changes permission show policy for read-permission grants

### LINKED ISSUE #6269: Shared tables set to 'private' cannot be opened in the Dashboard
When sharing a dataset with someone in the organization, if privacy is set to 'Private' the dataset is not accessible through the Dashboard. 

## Steps to reproduce

1. Go to Dashboard, Datasets
2. Click on 'Shared with you'
3. Try and open one that is private. 
4. The following appears: 
  
  >Looks like this map no longer exists or is set as private

## Context
1. Table is accesible through the SQL console. ```SELECT * FROM org_user.table_name```will produce the expected result. 
2. Table is also accessible through the URL

cc @gfiorav

gfiorav:
Fixes #6269

Now, the more details about the owner of a give permission are shown to users with (at least) read privileges in its acl. Before, they where only show to the owner of the permission.

Please CR @ethervoid 

cc/ @xavijam

### DIFF
--- a/app/controllers/carto/api/permission_presenter.rb
+++ b/app/controllers/carto/api/permission_presenter.rb
@@ -18,7 +18,7 @@ module Carto
       end
 
       def to_poro
-        return to_public_poro unless !@options[:current_viewer].nil? && @permission.is_owner_user?(@options[:current_viewer])
+        return to_public_poro unless !@options[:current_viewer].nil? && @permission.user_has_read_permission?(@options[:current_viewer])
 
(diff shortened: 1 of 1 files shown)


xavijam:
What about users with write permissions?

gfiorav:
A user with write permission also has read permission, no worries there :)

Let's check it all tomorrow

ethervoid:
:+1:

xavijam:
Tested in staging :+1:.

xavijam:
:fire:

Issue #6276, opened 2015-12-15
"ArgumentError: invalid byte sequence in UTF-8" error on ZIP import


iriberri:
````
2015-12-14 15:32:01 UTC: Unpacking /tmp/imports/xxxxx-1455-1gqe5as/xxxxxx.zip
2015-12-14 15:32:01 UTC: Errored importing data:
2015-12-14 15:32:01 UTC: ArgumentError: invalid byte sequence in UTF-8
2015-12-14 15:32:01 UTC: ----------------------------------------------------
2015-12-14 15:32:01 UTC: [\"/home/ubuntu/www/production.cartodb.com/releases/xxx/services/importer/lib/importer/unp.rb:180:in `=~'\", 
\"/home/ubuntu/www/production.cartodb.com/releases/xxx/services/importer/lib/importer/unp.rb:180:in `unp_failure?'\", 
\"/home/ubuntu/www/production.cartodb.com/releases/xxx/services/importer/lib/importer/unp.rb:91:in `extract'\", 
\"/home/ubuntu/www/production.cartodb.com/releases/xxx/services/importer/lib/importer/unp.rb:42:in `run'\", 
\"/home/ubuntu/www/production.cartodb.com/releases/xxx/services/importer/lib/importer/runner.rb:247:in `block (2 levels) in single_resource_import'\"

```

iriberri:
b57e8686-a383-4e10-81f7-da82f9f2fe0e

iriberri:
After everything we discussed time ago, I think I'm just gonna rescue the encoding error and return it to the user so that it is not an unknown error and people can sort it out. These would also start counting as user errors and not as platform errors.

iriberri:
![image](https://cloud.githubusercontent.com/assets/1730320/13315967/12b5f7ec-dbae-11e5-944e-6f79622ef03f.png)

:-)

PR #6283, opened 2015-12-15
Change string matching method to avoid encoding error

iriberri:
The previous string matching method that was used would not match the regexp if not all code points are valid.

This was provoking import errors in Shapefiles with names that contained special characters. Fixes https://github.com/CartoDB/cartodb/issues/6276.

Quick example:

````ruby
irb(main):001:0> a = "\xFF\xFF\xFFhello"
=> "\xFF\xFF\xFFhello"
irb(main):002:0> a.encoding
=> #<Encoding:UTF-8>
irb(main):003:0> a.valid_encoding?
=> false
irb(main):004:0> a.include? "hello"
=> true
irb(main):005:0> a =~ /hello/
ArgumentError: invalid byte sequence in UTF-8
````

### DIFF
--- a/services/importer/spec/unit/unp_spec.rb
+++ b/services/importer/spec/unit/unp_spec.rb
@@ -231,7 +231,7 @@ describe Unp do
 
   describe '#unp_failure?'  do
     it 'returns true if unp cannot read the file' do
-      Unp.new.unp_failure?('Cannot read', 0).should eq true
+      Unp.new.unp_failure?('Cannot read', -1).should eq true
     end
 
     it 'returns true if returned an error exit code' do
--- a/services/importer/lib/importer/unp.rb
+++ b/services/importer/lib/importer/unp.rb
@@ -12,7 +12,6 @@ module CartoDB
   module Importer2
     class Unp
       HIDDEN_FILE_REGEX     = /^(\.|\_{2})/
-      UNP_READ_ERROR_REGEX  = /.*Cannot read.*/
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by iriberri** [comment]
cc @rafatower

**Comment by iriberri** [comment]
I fixed one text matching, but there are quite a few more in the same file that were working in local but are raising errors in the staging environment.

**Comment by iriberri** [comment]
Ops. It seems that removing the error message comparison in the unp output is breaking some tests.

File: services/importer/spec/unit/unp_spec.rb [hunk 1]
```diff
@@ -231,7 +231,7 @@
 
   describe '#unp_failure?'  do
     it 'returns true if unp cannot read the file' do
-      Unp.new.unp_failure?('Cannot read', 0).should eq true
```
rafatower: this was the wrong assumption in the past.

iriberri:
retest this please

rafatower:
there are a number of tests broken, probably by me :(

let me know if you need help fixing it.

iriberri:
I have been working in this branch in a new attempt to run everything from scratch: https://github.com/CartoDB/cartodb/compare/6276-unp_filenames_encoding

Haven't PR'd yet, as I managed to "get it working" last day at 10PM and still wanted to think about it because the approach I followed is quite aggressive.

iriberri:
Closing without merging. We can handle the error conveniently (instead of raising an unknown error).

PR #6298, opened 2015-12-18
Store the user geocoding requests to redis

### LINKED ISSUE #6286: New geocodings metrics system
We need to implement a new usage metrics system for the geocoding tasks in order to improve the metrics, change a per-request system and integrate the geocoder api.

From now on the data will be stored in Redis that will be the storage system that could be access from all the systems (geocoder api, rails api, tiler and so on)

The tasks for this issue will be:

- [ ] Add a redundant system to store usage both metrics: old (database) and new type (redis)
- [ ] Add a new endpoint, v2, for the new kind of metrics
- [ ] Check if the v1 and v2 are consistent, despite its differences.

//cc @rafatower

ethervoid:
Closes #6286 

- [x] Store the geocoding metrics in Redis
- [x] When there is no rows to proccess don't store the results without service. Ie. "user:ethervoid-ded04::total_requests"
- [ ] When there is cache_hits in the payload, don't add it to geocoder. Add it to the cache keys

//cc @rafatower

### DIFF
--- a/app/models/geocoding.rb
+++ b/app/models/geocoding.rb
@@ -367,7 +367,8 @@ class Geocoding < Sequel::Model
   end
 
   def total_geocoded_rows(rows_geocoded_before)
-    rows_geocoded_after = table_service.owner.in_database.select.from(table_service.sequel_qualified_table_name).where('cartodb_georef_status is true and the_geom is not null').count rescue 0
+    rows_geocoded_after = table_service.owner.in_database.select.from(table_service.sequel_qualified_table_name).
+      where('cartodb_georef_status is true and the_geom is not null').count rescue 0
     rows_geocoded_after - rows_geocoded_before
   end
 
--- a/app/models/geocoding.rb
+++ b/app/models/geocoding.rb
@@ -1,17 +1,21 @@
 # encoding: UTF-8'
 require 'socket'
+require_relative './geocoding/geocoder_metrics_decorator'
 require_relative '../../services/table-geocoder/lib/table_geocoder_factory'
 require_relative '../../services/table-geocoder/lib/exceptions'
 require_relative '../../services/table-geocoder/lib/mail_geocoder'
 require_relative '../../services/geocoder/lib/geocoder_config'
+require_relative '../../services/metrics/lib/geocoder_metrics'
 require_relative '../../lib/cartodb/metrics'
 require_relative '../../app/helpers/bounding_box_helper'
 require_relative 'log'
 require_relative '../../lib/cartodb/stats/geocoding'
 
 class Geocoding < Sequel::Model
+  include CartoDB::GeocoderMetricsDecorator
 
   ALLOWED_KINDS   = %w(admin0 admin1 namedplace postalcode high-resolution ipaddress)
+  HIGH_RESOLUTION_GEOCODER = 'high-resolution'
 
 
   PUBLIC_ATTRIBUTES = [:id, :table_id, :table_name, :state, :kind, :country_code, :region_code, :formatter,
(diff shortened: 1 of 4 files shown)


app/models/geocoding.rb
```diff
@@ -168,6 +169,10 @@ def run_geocoding!(processable_rows, rows_geocoded_before = 0)
   def report(error = nil)
     payload = metrics_payload(error)
     CartoDB::Metrics.new.report(:geocoding, payload)
+    if !geocoder_type.blank?
```
rafatower: This should never be blank, maybe we shall report a warning or sth if it is.

rafatower:
we should listen to hound a little more for the sake of code's uniformity

ethervoid:
Yup, hound changes are the last I'm going to commit.

services/metrics/lib/geocoder_metrics.rb
```diff
@@ -0,0 +1,49 @@
...
+
+module CartoDB
+  class GeocoderMetricsError < StandardError; end
+  class GeocoderMetrics
+
+    SUCCESSFUL_RESPONSE_TYPE = "successful_response"
+    FAILED_RESPONSE_TYPE = "failed_response"
+    EMPTY_RESPONSE_TYPE = "empty_response"
+    TOTAL_RESPONSE_TYPE = "total_requests"
+    CACHE_HITS_TYPE = "cache_hits"
```
rafatower: I prefer to see cache as a service in itself

services/metrics/lib/geocoder_metrics.rb
```diff
@@ -0,0 +1,49 @@
...
+    private
+
+    def check_valid_type(type)
+      ALLOWED_TYPES.include?(type)
+    end
+
+    def build_key_prefix(type)
+      if @user.organization_user?
+        orgname = @user.organization.name
+        %{org:#{orgname}:#{@user.username}:#{@geocoder_type}:#{type}}
```
rafatower: to be changed as per our discussion in the feature doc.

rafatower: I think the interface to redis should be something like
```ruby
service = ServiceMetrics::get(ServiceMetrics::Here)
service_metrics.incr(@user, ServiceMetrics::Here::REQUESTS)
service_metrics.incr(@user, service, ServiceMetrics::SUCCESSFUL_RESPONSES)

service = ServiceMetrics::get(ServiceMetrics::Cache)
service_metrics.incrby(@user, ServiceMetrics::Cache::REQUESTS, 100)
service_metrics.incrby(@user, service, ServiceMetrics::Cache::SUCCESSFUL_RESPONSES, 100)
```

My points here are:
- it shouldn't assume all metrics will be available just at the end of the process
- the key prefix generation shall be identical up to the `[service]` part
- the service is to be hardcoded, each service defines its metrics (if it can be regular, much better)

We can go ahead with this solution but think about adding the metrics storage at the place where they "happen". This connects with my other comment about persistence of metrics in the model.

What do you think?

services/metrics/lib/geocoder_metrics.rb
ethervoid: I thought that but this is a geocoder cache we could have more caches. We could create GeocoderCacheMetrics for example

services/metrics/lib/geocoder_metrics.rb
ethervoid: So ServiceMetrics is a factory that return the GeocoderServiceMetrics for example?

I don't like to emulate Redis or 3rd party commands in the interface that's why I called it add_response (could think a better name of course)

rafatower: as we're measuring with counters... incr doesn't strike me as a bad option.

PR #6304, opened 2015-12-22
Segregated different kind of geocoder results for stats

ethervoid:
This PR is derived from #6286.

We need to add more stats to our current system to segregate by processed, failed and successful responses.


ethervoid:
@rafatower check this out, I need to add the GME part but I need validation for this work

db/migrate/20151222112300_add_success_and_failed_processed_rows_to_geocodings.rb
```diff
@@ -0,0 +1,15 @@
+Sequel.migration do
+  up do
+    alter_table(:geocodings) do
+      add_column :successful_processed_rows, :bigint
```
rafatower: I see you're adding a couple of attributes here. Are we going to use the existing ones yet? Worried a little about compatibility and breaking things in the interim

services/geocoder/lib/hires_batch_geocoder.rb
```diff
@@ -105,6 +106,8 @@ def cancel
       handle_api_error(response)
       @status         = extract_response_field(response.body, '//Response/Status')
       @processed_rows = extract_response_field(response.body, '//Response/ProcessedCount')
+      @successful_processed_rows = extract_response_field(response.body, '//Response/SuccessCount')
+      @failed_processed_rows = extract_response_field(response.body, '//Response/ErrorCount')
```
rafatower: WIN! :smile:

services/geocoder/lib/hires_geocoder.rb
```diff
@@ -99,8 +99,11 @@ def input_rows
     def process_row(input_row, output_csv_file)
       @processed_rows = @processed_rows + 1
       latitude, longitude = geocode_text(input_row["searchtext"])
-      if !(latitude.nil? || latitude == "" || longitude.nil? && longitude == "")
+      if !(latitude.blank? || longitude.blank?)
+        @successful_processed_rows = @successful_processed_rows + 1
         output_csv_file.add_row [input_row["searchtext"], 1, 1, latitude, longitude]
+      else
+        @failed_processed_rows = @failed_processed_rows + 1
```
rafatower: minor thing: anything against `+=` operator?

db/migrate/20151222112300_add_success_and_failed_processed_rows_to_geocodings.rb
ethervoid: Is backward compatible. `processed_rows` will be the same, I've changed it in the internal geocoder but we don't care about it. In the Nokia geocoder is the same but we have more info about it

services/geocoder/lib/hires_geocoder.rb
ethervoid: Nothing at all, I've used in other places. I'll change it

services/table-geocoder/lib/internal_geocoder.rb
```diff
@@ -116,17 +125,27 @@ def copy_results_to_table
...
         @temp_table_name ||= %Q{"#{@table_schema}".internal_geocoding_#{Time.now.to_i}}
-      end # temp_table_name
+      end
 
       def name
         'internal'
       end
 
-    end # Geocoder
+      def update_stats(response, number_of_terms)
```
rafatower: Instead of `number_of_terms` I'd use `search_input_length` to avoid confusion.

services/table-geocoder/spec/internal_Geocoder_spec.rb
```diff
@@ -1,17 +1,30 @@
 # encoding: utf-8
 require 'open3'
+require_relative '../../../spec/rspec_configuration.rb'
 require_relative 'factories/pg_connection'
 require_relative '../lib/internal_geocoder.rb'
-require_relative '../../../spec/rspec_configuration.rb'
+require 'byebug'
```
rafatower: probably to be removed.

db/migrate/20151222112300_add_success_and_failed_processed_rows_to_geocodings.rb
rafatower: ok :)

rafatower:
overall looks ok

db/migrate/20151222112300_add_success_and_failed_processed_rows_to_geocodings.rb
rafatower: Actually, we don't need persistence here. Think about it. I mean: what does this provide if we're gonna move to the new stats and they're going to be persisted in redis?

rafatower: it's more a phylosofical question: if it makes storing metrics easier for the legacy part, I'm 100% for it (as I think it does)

ethervoid: Indeed is not needed but for consistency we should store it. 
We could get the metrics from the geocoder "clients" and store directly into Redis but for now I like more to create iterations and this iteration should be "empower the legacy metrics storage". But that's only my opinion, if you think this could lead to an overhead in the database or something like that could be easily removed.

rafatower: no, no, it's ok

PR #6320, opened 2015-12-24
Improve HTTP headers handling in importer downloader

iriberri:
This started as an attempt to change the way we are retrieving the `Content-Type` header from downloads, as only exact matches would be supported later on in the code. We are supposing that the `Content-Type` would only contain the media type and not its parameters (such as  the charset: `Content-Type: text/html; charset=ISO-8859-4`).

* Content-Header definition in HTTP 1.1 - RFC 2616: https://tools.ietf.org/html/rfc2616#section-14.17
* Media types in HTTP 1.1 - RFC 2616: https://tools.ietf.org/html/rfc2616#section-3.7

In order to retrieve just the type, we are using a split with the `;` character which is the delimiter of parameters according to the RFC.

I noticed that we were commonly using different comparisons to retrieve the headers:

```ruby
      def content_type
        headers.fetch('Content-Type', nil) ||
          headers.fetch('Content-type', nil) ||
          headers.fetch('content-type', nil)
      end
```

but since Typhoeus 0.5.2 it is [downcasing the headers for easier access](https://github.com/typhoeus/typhoeus/issues/227). CartoDB uses `typhoeus 0.7.2` at the moment.

They [added a @sanitized variable](https://github.com/typhoeus/typhoeus/blob/v0.7.2/lib/typhoeus/response/header.rb#L18) that stores the downcased keys and then [set a fallback procedure](https://github.com/typhoeus/typhoeus/blob/v0.7.2/lib/typhoeus/response/header.rb#L20) for the key lookups. If the lookup doesn't find the key (`Content-Type`) then it would fallback to a downcased lookup `content-type` that will always work as they're storing this @sanitized downcase copy of the hash.

This way, if the header comes as `Content-type`, any combination as `headers.fetch('Content-type')` would work, `headers['Content-type']` would work, and the default downcased `headers['content-type']` would work. The lookup will be done with a downcased key, that will always match with the downcased key that Typhoeus stores in the sanitized header.

Therefore I edited the way that all headers are retrieved substituting them by a `headers['capitalized_key']` approach. The `content_type` function defined above ends up like:

```ruby
      def content_type
        media_type = headers['Content-Type']
        return nil unless media_type
        media_type.split(';').first
      end
``` 

**Note**: Even though lookups are case insensitive in Typhoeus Headers, I am using the capitalized header names.

### Testing

* Added new test `'ignores extra type parameters in Content-Type header' ` to test the output of the content type detection when a parameter is received
* These changes broke:
  * `'gets the file name from the Content-Disposition header if present'`: this test was generating directly the hash for the headers in a new variable, which ignores the behaviour that Typhoeus provides and the one that we are taking advantage of in this approach. I decided to downcase the key names to mimic Typhoeus' behaviour.
  * `'imports KML files from url'`: this test was working with a KML URL whose content type is  `text/plain; charset=utf-8`. Due to the charset parameter, this content type was never captured in the past and now the file was uploaded as a CSV (provoking the test to fail). I added `kml` to the mapping with `text/plain` content type, as Google says it can be used too officially. There are more details about this in the comments below.

Fixes https://github.com/CartoDB/cartodb/issues/5524

### DIFF
--- a/services/importer/lib/importer/downloader.rb
+++ b/services/importer/lib/importer/downloader.rb
@@ -322,24 +322,19 @@ module CartoDB
       end
 
       def content_length_from(headers)
-        content_length = headers.fetch('Content-Length', nil)
-        content_length ||= headers.fetch('Content-length', nil)
-        content_length ||= headers.fetch('content-length', -1)
+        content_length = headers['content-length']
+        return -1 unless content_length
         content_length.to_i
       end
 
       def etag_from(headers)
-        etag  =   headers.fetch('ETag', nil)
-        etag  ||= headers.fetch('Etag', nil)
-        etag  ||= headers.fetch('etag', nil)
-        etag  = etag.delete('"').delete("'") if etag
+        etag = headers['etag']
+        etag = etag.delete('"').delete("'") if etag
         etag
       end
 
       def last_modified_from(headers)
-        last_modified =   headers.fetch('Last-Modified', nil)
-        last_modified ||= headers.fetch('Last-modified', nil)
-        last_modified ||= headers.fetch('last-modified', nil)
+        last_modified = headers['last-modified']
         last_modified = last_modified.delete('"').delete("'") if last_modified
         if last_modified
           begin
@@ -374,15 +369,13 @@ module CartoDB
       end
 
       def content_type
-        full_content_type = headers['content-type']
-        return nil unless full_content_type
-        full_content_type.split(';').first
+        media_type = headers['content-type']
+        return nil unless media_type
+        media_type.split(';').first
       end
 
       def name_from_http(headers)
-        disposition = headers.fetch('Content-Disposition', nil)
-        disposition ||= headers.fetch('Content-disposition', nil)
-        disposition ||= headers.fetch('content-disposition', nil)
+        disposition = headers['content-disposition']
         return false unless disposition
         filename = disposition.match(CONTENT_DISPOSITION_RE).to_a[1]
         return false unless filename
@@ -410,7 +403,7 @@ module CartoDB
       end
 
       def gdrive_deny_in?(headers)
-        headers.fetch('X-Frame-Options', nil) == 'DENY'
+        headers['x-frame-options'] == 'DENY'
       end
 
       def md5_command_for(name)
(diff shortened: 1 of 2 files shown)


iriberri:
The old code was not giving precedence to filename VS content type. There's a spec failing with a KML URL import whose Content-Type is `text/plain; charset=utf-8`. The test was working before just because the Content-Type was not really being captured, as the charset didn't match with any approach. The code matches `text/plain` with a CSV file, so the KML file is not passed through the correct path and gets imported as if it was a CSV. The geometry is obviously not detected from the KML and the spec fails while checking the geometry type.

iriberri:
Google suggests using `text/plain` or `application/vnd.google-earth.kml+xml`. Content-Type for KMLs.
> When a call is made from the client (Google Earth) to the server, the server must (1) return a response code of HTTP 200 and (2) set the response's content type to text/plain or application/vnd.google-earth.kml+xml.

> The response must be valid KML.

Source: https://developers.google.com/kml/documentation/kml_tut?hl=en

iriberri:
Finally, I'm adding `kml` to map with `plain/text` content type.

```
[314, 323] in /home/ubuntu/www/production.cartodb.com/current/services/importer/lib/importer/downloader.rb
   314          file_extension = File.extname(name).split('.').last
   315          name_without_extension = File.basename(name, ".*")
   316          #If there is no extension or file extension match in the content type extensions, add content type
   317          #extension to the file name deleting the previous extension (if exist)
   318          debugger
=> 319          if (file_extension.nil? || file_extension.empty?) || !content_type_extensions.include?(file_extension)
   320            return "#{name_without_extension}.#{content_type_extensions.first}"
   321          else
   322            return name
   323          end
(rdb:1) content_type_extensions
["txt", "kml"]
(rdb:1) content_type_extensions.include?(file_extension)
true
(rdb:1) name
"one_layer.kml"
```

As the if would never be true for this case, it would always return the name which contains already the own extension, even though CartoDB is finding two possible matches. I don't know if it would be interesting to add a special case in the code to check .kml files and avoid the content_type_extensions array length to be > 1.

iriberri:
@rafatower please, CR :)

rafatower:
:+1:

iriberri:
Going to deploy this now. This will fix two immediate bugs:

* KMZ not being imported correctly because we didn't recognize `plain/text` as a valid Content Type for them
* URL files not being supported because of parameters in the Content-Type header

PR #6325, opened 2015-12-28
Sync message edit for sync errors

iriberri:
Using "sync failed" instead of "synced failed" in the copies that contained it.

cc @xavijam @viddo


xavijam:
:+1:

PR #6333, opened 2015-12-29
Print viz_json html safe

### LINKED ISSUE #6332: cartod1b map fails
Related with #6197, the embedded map fails due rails is escaping the vizjson.

acanimal:
reviewer @xavijam 

close #6332 

Modified the view to print the viz_json data html safe.


xavijam:
:+1:

PR #6346, opened 2016-01-04
Remove public profile mention

iriberri:
cc @ethervoid:

Any specific reason why "public profile" was mentioned here? This is not really true, as map views are measured from any source. If there's not any specific reason why the sentence is there, I would like to remove it as it can cause confusion for users with "with link" map privacies.


ethervoid:
Good for me but this is a copy that I got

iriberri:
@brittanymicek could you please take a quick look here? <3

brittanymicek:
Hey @iriberri 

Feel free to drop the "public profile" if it creates confusion. 

Wording should now read:
Your [map name] has received more than [number] map views.

PR #6362, opened 2016-01-07
Organization user quota UI

### LINKED ISSUE #5363: Organization user quotas
Related https://github.com/CartoDB/cartodb/issues/4884

Possible new feature: add custom twitter credits to user (until now credits limit is the same for all users).

cc/ @saleiva @javisantana @Kartones @juanignaciosl

MariaCheca:
Closes #5363 

untitled

CR @viddo cc/ @xavijam


lib/assets/javascripts/cdb
```diff
@@ -1 +1 @@
-Subproject commit 3798323b953318576fac7ef00dede0aeec81e978
+Subproject commit be76b55caeeb0dfc147e3fa36628d9e215717ce5
```
viddo: Why have cartodb.js changed?

lib/assets/javascripts/cartodb/organization/entry.js
```diff
@@ -4,7 +4,7 @@ var HeaderView = require('../common/views/dashboard_header_view');
 var SupportView = require('../common/support_view');
 var HeaderViewModel = require('./header_view_model');
 var LocalStorage = require('../common/local_storage');
-var QuotaProgressBar = require('./organization_progress_bar');
+var UserQuota = require('./organization_quota');
```
viddo: We have as a convention to name variables and filenames consistently and suffix them with the kind it represents, e.g. `OrganizationQuotaView` / `organization-quota-view.js`, makes it easier read/interpret/browse/search files and vars throughout the codebase.

lib/assets/javascripts/cartodb/organization/entry.js
```diff
@@ -146,20 +146,17 @@ $(function() {
...
-      )
+    // User quota main view
+    if (this.$('.js-userQuota').length > 0 && organization && this.$('#user_quota').length > 0) {
+      new UserQuota({
+        el: this.$('.js-userQuota'),
+        model: new cdb.core.Model({
+          orgQuota: organization.get('quota_in_bytes'),
+          orgUsedQuota: organization.get('used_quota_in_bytes'),
+          userUsedQuota: used_user_quota,
+          userQuota: assigned_user_quota
```
viddo: 1. Taking out values from an existing model is a bit of anti-pattern, since then we don't have a "single source of truth" anymore (in case some other view/model rely on this value to calculate something), why not use the organization model?
2. I presume the last two variables are set in the `window` context somewhere (rails template?), we should avoid have loose vars. The suffix hint that these values belong to the organization user, so why not provide them through `organizationUser`?

lib/assets/javascripts/cartodb/organization/entry.js
viddo: …just to clarif,: should be applied from now on going forwrad, so no need to fix all existing vars/files used in the file, but at least the changed/new files in this PR. :)

lib/assets/javascripts/cartodb/organization/organization_quota_slider.js
```diff
@@ -0,0 +1,68 @@
...
+        title: function() {
+          var orgAvailableQuota = view_model.get('orgQuota') - view_model.get('orgUsedQuota');
+          var quotaPer = (view_model.get('userQuota') * 100) / orgAvailableQuota;
+          return quotaPer.toFixed(0) + "% of the available org quota";
+        }
+      })
+    );
+  }
+
+})
```
viddo: As a general rule, the views should never have to know about domain knowledge (I.e. calculations of quotas, in this case). That would be better of in a model (or value object of some kind), that can be tested in isolation.

lib/assets/javascripts/cartodb/organization/organization_quota_slider.js
```diff
@@ -0,0 +1,68 @@
...
+    var userQuota = this.model.get('userQuota');
+    var orgAvailableQuota = this.model.get('orgQuota') - this.model.get('orgUsedQuota');
+    var assignedPer = (userQuota * 100) / orgAvailableQuota;
+
+    $('.ui-slider-range').css('width', assignedPer + '%');
+    $('.js-quotaProgressSlider').css('left', assignedPer + '%');
+  },
+  
+  _showTooltip: function() {
+    var view_model = this.model;
```
viddo: var names should be `camelCased`

lib/assets/javascripts/cartodb/organization/organization_quota_slider_input.js
```diff
@@ -0,0 +1,40 @@
...
+    this._initBinds();
+    this._initViews();
+  },
+
+  _initBinds: function() {
+    this.model.bind('change:userQuota', this._onQuotaChange, this);
+  },
+
+  _initViews: function() {
+    var quotaInMb = Math.floor(this.model.get('userQuota')/1024/1024).toFixed(0);
```
viddo: On similar lines like the prev comments re: "domain knowledge". I think we have a helper to translate bytes to Mb (or other sizes, e.g. [`Utils.readablizeBytes`](https://github.com/CartoDB/cartodb/blob/bdd85a755f67ee9bfbc1d52385e4cb917ee14bc6/lib/assets/javascripts/cartodb/organization/organization_users/organization_user_view.js#L32)

lib/assets/test/spec/cartodb/organization/organization_quota.spec.js
```diff
@@ -0,0 +1,37 @@
...
+
+    this.view.render();
+  });
+
+  it('should render properly', function() {
+    this.view.render();
+    expect(this.view.$el.hasClass('js-userQuota')).toBeTruthy();
+  });
+
+  it('should bind model changes', function() {
```
viddo: It's not really relevant that the model does a bind, that test may pass but the view might still be broken. Instead, test what the _side-effets_ should be when the model changes. Something like

```js
…
describe('when the User gets new quota set', function() {
  beforeEach(function(){
    this.model.set('userQuota', 123);
  });

  it('should render the new quota in the hidden input element', function() {
    expect(this.view.$('.js-hidden-input-whatever').val()).toEqual(123);
  });
});
```

[This slide](https://speakerdeck.com/skmetz/magic-tricks-of-testing-railsconf?slide=187) is a great "cheat-sheet" to remember what to focus on in a test.

viddo:
Overall :+1: well done! Wrote some comments on some details that needs some work though, let me know if you need a hand.

xavijam:
Ready to go?

xavijam:
Did @josecruz take a look?

josecruz:
Already reviewed, ping me again when you need me :*

xavijam:
Looks very good @MariaCheca :clap: :clap:. To take into account:

- Make added strings (and not numbers) fail in the input.
- If user has "used quota", set it as the min acceptable value.
- If user add more quota than available, set the max value in the slider.
- Regenerate api button should have ```type="button"``` avoiding problems when user presses enter key being within quota input.

josecruz:
After the review with @MariaCheca 

**Actual Behaviour:**
mariaje3

When you add an outta quota number like `999` it saves the first valids numbers and when you save it, it says that everything went correct.

**Expected Behaviour:**
It should display an error message when you exceed and save your quota

MariaCheca:
We'll make the changes @josecruz is proposing in another ticket.

@viddo please review the last changes to release it :)

cc @xavijam

xavijam:
Agree :+1:

viddo:
Except for the comments :+1: 
@MariaCheca I understand that this feature might be urgent, so even if not fixing them now at least keep the topics mentioned in there in mind for the future.

MariaCheca:
Improvements listed here: https://github.com/CartoDB/cartodb/issues/6404

MariaCheca:
:sparkles:

PR #6374, opened 2016-01-11
Don't round remote datasets size when they are bigger than 1GB

### LINKED ISSUE #6340: User can't import data library dataset even with enough data storage space & import limit
@rafatower 

S/B 7574555

With over 1GB of data storage available and just over 1GB import limit, a 1GB Data Library dataset appears in red so that user is unable to connect it.

## Context

data storage = 1.5GB total
data storage currently used = .2GB
max import row limit = 5,000,000
max import file size = 1073741824 bytes

Data Library dataset to be connected = US Block Groups (tl_2014_block_groups)

## Process

1. Set test account to the above limits in SuperAdmin 
2. Log into test account
3. Go to Datasets page & click on Data Library
4. Search for US Block Groups 

## Expected result

When US Block Groups is found in Data Library search results, it should be able to be connected instead of showing red/unable to be connected.

## Current result

screen shot 2015-12-29 at 2 58 57 pm

----
The dataset can be connected when I increase my team account's max import size limit to 1.5GB (with over 1GB data storage available, and 7,500,000 rows max). 

If the dataset is actually larger than 1GB, can we change the UI to show that?

[not sure if [this](https://github.com/CartoDB/cartodb/issues/5315) is related]

xavijam:
screen shot 2016-01-11 at 15 41 53

Fixes #6340.

cc @ztephm

### DIFF
--- a/lib/assets/javascripts/cartodb/common/dialogs/create/listing/datasets/remote_dataset_item_view.js
+++ b/lib/assets/javascripts/cartodb/common/dialogs/create/listing/datasets/remote_dataset_item_view.js
@@ -57,7 +57,10 @@ module.exports = DatasetItem.extend({
 
     var datasetSize = table.get('size');
     if (datasetSize >= 0) {
-      d.datasetSize = Utils.readablizeBytes(datasetSize, true);
+      d.datasetSize = Utils.readablizeBytes(
+        datasetSize,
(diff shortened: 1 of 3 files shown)


ztephm:
Thanks @xavijam !

PR #6334, opened 2015-12-29
Remove custom basemaps for google users

xavijam:
Don't let google users add a custom basemap.

Fixes https://github.com/CartoDB/cartodb-platform/issues/1132.
Fixes https://github.com/CartoDB/cartodb-platform/issues/1320.

cc @luisbosque @yasharora73
CR: @javierarce

PR #6375, opened 2016-01-11
Don't let google users add custom XYZ/TMZ/... basemaps

xavijam:
It comes from this merged [PR](https://github.com/CartoDB/cartodb/pull/6334) and then [reverted](https://github.com/CartoDB/cartodb/pull/6359) due to color was disabled too.

---

Standard user:
screen shot 2016-01-11 at 16 05 42

screen shot 2016-01-12 at 11 58 06

---
Google user (_*don't take into account small previews, they are wrong_):
screen shot 2016-01-11 at 16 05 55

screen shot 2016-01-12 at 11 59 37

---

cc @yasharora73 @javisantana @saleiva

### DIFF
--- a/lib/assets/javascripts/cartodb/table/basemap/basemap_dropdown/color_basemap.js
+++ b/lib/assets/javascripts/cartodb/table/basemap/basemap_dropdown/color_basemap.js
@@ -42,9 +42,9 @@ cdb.admin.BackgroundMapColorView = cdb.admin.BaseLayerButton.extend({
 
   _changeModel: function() {
     if (this.model) {
-      this.model.unbind(null, null, this);  
+      this.model.unbind(null, null, this);
     }
-    
+
     this.model = this.map.getBaseLayer();
     this._initBinds();
     this.render();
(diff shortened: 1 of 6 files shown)


xavijam:
@javierarce same you reviewed in https://github.com/CartoDB/cartodb/pull/6334, but now custom color/img is still available for any user type, could you take a look in any case?

lib/assets/javascripts/cartodb/table/basemap/basemap_dropdown/color_basemap.js
```diff
@@ -60,11 +60,11 @@ cdb.admin.BackgroundMapColorView = cdb.admin.BaseLayerButton.extend({
     this.color_picker = new cdb.admin.ColorPicker({
       className: 'dropdown color_picker basemap border vertical_offset',
       target: this.$el,
-      vertical_position: "down",
+      vertical_position: "up",
```
xavijam: Just moving the position of the basemap color picker.

javierarce: When you say 'is still available for any user type' … you mean that's the expected behaviour or is a bug?

xavijam: Set a colored basemap should be only enabled for standard users (no for google ones) but seems like it is being used, so we will leave as it is. We will manage that non-desired feature in the new editor, I guess.

javierarce:
:+1: Code looks good to me.

lib/assets/javascripts/cartodb/table/basemap/basemap_dropdown/color_basemap.js
javierarce: :ok_hand:

PR #6376, opened 2016-01-11
Avoid erroneous XLS format detection for XLSX files

### LINKED ISSUE #6361: XLSX file names get switched to XLS, causing format incompatibility import issues
From: https://github.com/CartoDB/cartodb/blob/0418ccd6bc640b92291c7a035e97b721bdca50db/services/importer/lib/importer/downloader.rb#L395

Should fix: https://github.com/CartoDB/cartodb/issues/1896 and https://github.com/CartoDB/cartodb/issues/6278

iriberri:
Fixes #6361 

In previous versions of the code, the order of the arrays in which we were storing the supported extensions was not taken into account.

This was causing that the format XLS was being considered for some XLSX files, as it was the first match of the `URL_FILENAME_REGEX`. This happened in a specific scenario in which the Content-Type for the file was empty.

Now the supported extensions array gets ordered by length of the extensions name so that the regex is greedy:

```ruby
CartoDB::Importer2::Unp::SUPPORTED_FORMATS
                                  .concat(CartoDB::Importer2::Unp::COMPRESSED_EXTENSIONS)
                                  .sort_by { |s| s.length }.reverse
````

This PR also contains code changes to move the regex related constants to class methods.

Finally, a new test has been added to ensure that the XSLX extension is being detected correctly in the scenario where it wasn't previously. The old version of the code makes this test to fail.

### DIFF
--- a/services/importer/lib/importer/downloader.rb
+++ b/services/importer/lib/importer/downloader.rb
@@ -26,6 +26,11 @@ module CartoDB
 
       SUPPORTED_EXTENSIONS = CartoDB::Importer2::Unp::SUPPORTED_FORMATS
                               .concat(CartoDB::Importer2::Unp::COMPRESSED_EXTENSIONS)
+                              .sort_by { |s| s.length }.reverse
+
+      URL_FILENAME_REGEX = Regexp.new(
+                              "[[:word:]]+(#{ SUPPORTED_EXTENSIONS.map{ |s| s.sub(/\./, "\\.")}.join("|") })+",
+                              true)
 
       DEFAULT_FILENAME        = 'importer'
       CONTENT_DISPOSITION_RE  = %r{;\s*filename=(.*;|.*)}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by iriberri** [comment]
Please CR @rafatower

File: services/importer/lib/importer/downloader.rb [hunk 1]
```diff
@@ -392,7 +398,7 @@ def name_from_http(headers)
       end
 
       def name_in(url)
-        url_name = URL_FILENAME_REGEX.match(url).to_s
+        url_name = Downloader.url_filename_regex.match(url).to_s
```
rafatower: use `self` instead of the class name:
```
url_name = self.url_filename_regex.match(url).to_s
```

rafatower:
Just had a minor comment, other than that :+1:

iriberri:
Will deploy tomorrow morning.

PR #6379, opened 2016-01-12
HOTFIX: Add usage metrics to the importer georeferencer

ethervoid:
Added usage metrics to the import georeferencer


rafatower:
Ok, but we should consider using the factory for that instead. Feel free to add a comment/ticket to remember in the future.

ethervoid:
I thought about it but I didn't want to couple it with the factory. I dunno...

PR #6395, opened 2016-01-13
Delete cdb_importer stale tables after import error

iriberri:
Whenever there's an exception in an import, `cdb_importer.importer_*` tables do never get moved to the user schema, so they remain at `cdb_importer`.

This code is aimed to find those and delete them if they exist, so that the databases don't keep storing useless data.

### DIFF
--- a/services/importer/lib/importer/runner.rb
+++ b/services/importer/lib/importer/runner.rb
@@ -196,8 +196,10 @@ module CartoDB
         else
           valid_table_names = loader.valid_table_names
           additional_support_tables = loader.additional_support_tables
-          clean_importer_tables(@job.db, @job.schema, valid_table_names)
         end
+
+        # Delete job temporary table from cdb_importer schema
+        @job.delete_job_table
 
         @job.log "Errored importing data from #{source_file.fullpath}:"
         @job.log "#{exception.class.to_s}: #{exception.to_s}", truncate=false
@@ -389,17 +391,6 @@ module CartoDB
                                                                user: user,
                                                                db: db
                                                              })
-      end
-
-      # This function cleans any stale cdb_importer.importer_* table
-      # related with the running process which should be discarded
-      def clean_importer_tables(database, schema, table_names)
-        table_names.each do |table|
-          CartoDB.notify_debug('Dropping cdb_importer table', { schema: schema, table: table, database: database })
-          database.execute(%Q{
-            DROP TABLE IF EXISTS "#{schema}"."#{table}"
-         })
-        end
       end
 
       def add_warning(warning)
--- a/services/importer/lib/importer/runner.rb
+++ b/services/importer/lib/importer/runner.rb
@@ -395,6 +395,7 @@ module CartoDB
       # related with the running process which should be discarded
       def clean_importer_tables(database, schema, table_names)
         table_names.each do |table|
+          CartoDB.notify_debug('Dropping cdb_importer table', { schema: schema, table: table, database: database })
           database.execute(%Q{
             DROP TABLE IF EXISTS "#{schema}"."#{table}"
          })
--- a/services/importer/lib/importer/runner.rb
+++ b/services/importer/lib/importer/runner.rb
@@ -196,6 +196,7 @@ module CartoDB
         else
           valid_table_names = loader.valid_table_names
           additional_support_tables = loader.additional_support_tables
+          clean_stale_importer_tables(@job.db, @job.schema, valid_table_names)
         end
 
         @job.log "Errored importing data from #{source_file.fullpath}:"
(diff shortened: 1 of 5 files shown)


iriberri:
Notes:

I have been trying about moving `valid_table_names` to the Job directly, but there's an additional variable used in the code (not used really, according to the comments) that does not leave in Job but in Loader: `additional_support_tables`.

Right now, code is:

````
  rescue => exception
        if loader.nil?
          valid_table_names = []
          additional_support_tables = []
        else
          valid_table_names = loader.valid_table_names
          additional_support_tables = loader.additional_support_tables
          clean_importer_tables(@job.db, @job.schema, loader.valid_table_names)
        end
````

If I moved `valid_table_names` to the job, but not `loader.additional_support_tables` , it would become something like: (taking into account that @job does always get generated, but loader can be nil)

```
  rescue => exception
        if loader.nil?
          valid_table_names = []
          additional_support_tables = []
        else
          additional_support_tables = loader.additional_support_tables
          valid_table_names = @job.valid_table_names
          clean_importer_tables(@job.db, @job.schema, valid_table_names)
        end 
```

`additional_support_tables` lives in Loader and is as follows:

````
      # Not used for now, but for compatibility with tiff_loader
      def additional_support_tables
        []
      end
```

It seems to be used [here](https://github.com/CartoDB/cartodb/blob/78922fb2e887ceda09f1acfd536ba2839d3ea08c/services/importer/lib/importer/tiff_loader.rb), which has access to the @job too. So I guess both methods could be inserted in @job.


As I have the @job, I could call use the cleaning function directly on @job.table_name while ignoring valid_table_name. This way I avoid the loop in the function.


cc @rafatower @ethervoid

services/importer/lib/importer/runner.rb
```diff
@@ -196,6 +196,7 @@ def import(source_file, downloader, loader_object = nil)
         else
           valid_table_names = loader.valid_table_names
           additional_support_tables = loader.additional_support_tables
+          clean_importer_tables(@job.db, @job.schema, valid_table_names)
```
rafatower: I'd move this outside of the if/else block

services/importer/lib/importer/runner.rb
```diff
@@ -390,6 +391,17 @@ def table_row_count_limit_instance(user, db)
                                                              })
       end
 
+      # This function cleans any stale cdb_importer.importer_* table
+      # related with the running process which should be discarded
+      def clean_importer_tables(database, schema, table_names)
```
rafatower: I'd move this to the Job class. The job object is the one that knows about the db connection, the schema used during import, the generated table names... and looks like the best candidate to keep those in sync.

In this file it will be sth like:
```
rescue => e
  @job.cleanup_last_created_table
```

services/importer/lib/importer/runner.rb
ethervoid: Valid table names comes from the loader if the loader is nil is not possible

services/importer/lib/importer/runner.rb
ethervoid: The data comes from the loader no the job isn't it?

iriberri: @rafatower  makes sense too.

@ethervoid from the initial code it seems like it comes from the loader, but the loader is really getting them from the job in the end.

rafatower: the data comes from the loader; the name of the table used everywhere comes from the job. re-architecting all the data flows of the importer is a bit beyond this ticket.

rafatower: (call it "importer metadata flow" so to speak)

jvillarf:
@iriberri @rafatower it seems the issue is still happening... :( Would you please take a look to my last comment on https://github.com/CartoDB/cartodb-platform/issues/1509? I haven't removed the temporal tables yet, just in case they give us any hint about the issue.

Thanks a lot!

iriberri:
I'll take a look at this asap.

jvillarf:
Hi, did you have the chance to take a look at this @iriberri? Thanks a lot.

rafatower:
@jvillarf iriberri is on vacation and we did not have a chance to look into that, sorry.

rafatower:
PS: I know there are remaining cases but I'm pretty confident a lot of them were covered by this fix.I also saw some killed resque processes and there's nothing we can do about clean up from ruby's side in those cases.

jvillarf:
It's ok Rafa, i was just checking :)

On Fri, Jan 29, 2016 at 4:59 PM, Rafa de la Torre <notifications@github.com>
wrote:

> @jvillarf <https://github.com/jvillarf> iriberri is on vacation and we
> did not have a chance to look into that, sorry.
>
> —
> <https://github.com/CartoDB/cartodb/pull/6395#issuecomment-176836808>.
>



-- 

*Javier Villar *
Site Reliability Engineer

*CartoDB*
Plaza de Callao 4, 2, 28013, Madrid, España
www.cartodb.com

jvillarf:
Hi again :), i'm about to clean around 80GB of temporal tables in dbd13. @rafatower  @iriberri  Do you want to take a look before i remove them? Last time i cleaned that server was around 25 days ago.
Thanks!
(more details in https://github.com/CartoDB/cartodb-platform/issues/1509)

rafatower:
cc/ @jgoizueta 

@jvillarf I think you can go ahead as unfortunately there will be more tables (it is just a matter of time, otherwise we wouldn't be discussing any issue).

jvillarf:
Ok @rafatower , i have just deleted them.

PR #6396, opened 2016-01-13
Fix missing meta on privacy update

### LINKED ISSUE #5667: Privacy change causes dataset to not show meta indicators in listing
dataset-properties

gfiorav:
Fixes #5667

The problem was that the JSON return was not done properly and nested info about synchronisation was lost. Now it should be fine.

This uses the new presenter (returns one thing or another depend on who is asking) but has to "convert" from member to viz, which might not be the best thing.

When CI pases I'll call for a CR.

P.D.: Although not in the scope of the ticket I corrected the return of create as well, as it had the same flaw.

cc/ @xavijam

### DIFF
--- a/app/controllers/api/json/visualizations_controller.rb
+++ b/app/controllers/api/json/visualizations_controller.rb
@@ -60,7 +60,10 @@ class Api::Json::VisualizationsController < Api::ApplicationController
 
         track_event(vis, 'Created')
 
-        render_jsonp(vis)
+        visualization = Carto::Visualization.find(vis.id)
+        presenter = Carto::Api::VisualizationPresenter.new(visualization, current_viewer, self)
+
+        render_jsonp(presenter.to_poro)
       rescue CartoDB::InvalidMember
         render_jsonp({ errors: vis.full_errors }, 400)
       rescue CartoDB::NamedMapsWrapper::HTTPResponseError => exception
--- a/app/controllers/api/json/visualizations_controller.rb
+++ b/app/controllers/api/json/visualizations_controller.rb
@@ -60,10 +60,7 @@ class Api::Json::VisualizationsController < Api::ApplicationController
 
         track_event(vis, 'Created')
 
-        visualization = Carto::Visualization.find(vis.id)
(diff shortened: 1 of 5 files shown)


gfiorav:
Hey, @juanignaciosl, could you have a look?

juanignaciosl:
Uhm. Maybe this fix works, but the approach feels wrong. That controller, `app/controllers/api/json/visualizations_controller.rb` is an "old" controller, so it should use `CartoDB::Visualization::Member`, not `Carto::Visualization`. In general "old" code (code not using ActiveRecord) should not depend to new. There're several reasons for that, from technical (avoid circular dependencies) to "procedural": we decided not to refactor old code but rewrite and delete.

I'd try to fix the old one first. If fixing it easy enough, I'd go for it.

gfiorav:
@juanignaciosl So you propose writing a different presenter to support the old model?

juanignaciosl:
Not really. Presenters were mostly introduced for the new models to avoid presentation logic inside the models code, I think there were no presenters before (at least there were not "mandatory").

I don't know as much as you do about this problem, but taking a look at the (wrong) JSON there's a `null` `synchronization` attribute, so it looks like sync is there but it`s not loaded. I'd first review the code, maybe you can fix the issue just loading the synchronization.

gfiorav:
@juanignaciosl Synchro is not null.

The `render_jsonp` is just not rendering it (I guess it has to do with it being nested).

juanignaciosl:
screen shot 2016-01-14 at 11 32 30

^ it is, isn't it?

gfiorav:
It is in the JSON, but just before entering `render_jsonp` it's not, that's what I meant.

juanignaciosl:
Ah, sorry :sweat_smile: 

Could you look into the code to check why `render_json`` renders `permission` but not `synchronization`?

gfiorav:
Hi @juanignaciosl , take a look now please!

juanignaciosl:
<3 :+1: Nice catch! I hope you enjoyed mining down the old models :-)

gfiorav:
:heart: :kiss: thanks @juanignaciosl 

Yes I enjoyed :laughing: :gun:

PR #6400, opened 2016-01-14
New method to get the geocoder used quota from Redis

ethervoid:
- [x] New methods to retrieve the current used geocoder quota using the new geocoder metrics system based in Redis
- [x] Rake task to compare the geocoder quota values between the legacy (DB) and the new (Redis) geocoder metrics system 

// @rafatower please check this out

### DIFF
--- a/lib/tasks/geocoder_metrics.rake
+++ b/lib/tasks/geocoder_metrics.rake
@@ -1,0 +1,28 @@
+# encoding: utf-8
+namespace :geocoder do
+  namespace :metrics do
+    # e.g. bundle exec rake geocoder:metrics:check_legacy_with_new_system
+    #      bundle exec rake geocoder:metrics:check_legacy_with_new_system[20]
+    desc 'Compare the geocoding quota for the current billing cycle between the legacy and the new system'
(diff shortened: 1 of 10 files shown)


lib/tasks/geocoder_metrics.rake
```diff
@@ -0,0 +1,28 @@
...
+        quota_delta_percentage = Float(args[:quota_delta]) / 100
+      rescue
+        raise "Quota value passed is not an integer"
+      end
+      users = ::User.overquota(quota_delta_percentage)
+      puts "There is no overquota users for today" if users.blank?
+      users.each do |user|
+        legacy_quota_value = user.get_geocoding_calls
+        new_system_quota_value = user.get_new_system_geocoding_calls
+        if legacy_quota_value != new_system_quota_value
```
rafatower: ok, let's run this as it is and we'll see later on if we need to add a tolerance for the checks :D

rafatower:
Looks neat! :+1:

lib/tasks/geocoder_metrics.rake
ethervoid: yup i thought the same

ethervoid:
retest this please

Issue #6413, opened 2016-01-18
icons resources for licenses are missing


matallo:
it would only work for Creative Commons, and even in that case it's not working anymore

**STR**
- create a new public dataset
- asign a license in the metadata (i.e. CC by attribution alone)

**expected behaviour**
- license icon to appear, related to the license chosen, as in the capture below
  <img width="201" alt="screen shot 2016-01-18 at 15 30 23" src="https://cloud.githubusercontent.com/assets/36676/12393885/d5787940-bdf8-11e5-9527-d90fd1e9c814.png">

**actual behaviour**
- no icon appears, not for CC, not for anyone else, as a compromise solution we can make the icon default to CC, but ideally it should relate to the license or display a more generic one

carlostallon:
Can you provide more background on this please?

matallo:
sure, on public datasets with license set in the metadata, the icon should be related to the license in each case (see attached an example with MIT which defaults to CC icon)

<img width="413" alt="screen shot 2016-01-21 at 16 56 44" src="https://cloud.githubusercontent.com/assets/36676/12485799/2aa47ab8-c060-11e5-9f75-9b8720cb103a.png">

we could also use a generic icon for all licenses, otherwise it would require producing a new icon for each license

carlostallon:
What about this icon? As a generic one:

![image](https://cloud.githubusercontent.com/assets/1712650/12721397/efe7e86c-c8ff-11e5-96c7-5c3ce606c754.png)

xavijam:
There has been no activity on this issue for more several months. We are closing it. If you think this still needs to be addressed please open a new issue.

matallo:
this is still happening

nobuti:
There has been no activity on this issue for more several months. We are closing it. If you think this still needs to be addressed please open a new issue.

matallo:
this is still happening

PR #6414, opened 2016-01-18
public datasets: fix license icons

matallo:
compromise solution for https://github.com/CartoDB/cartodb/issues/6413

I'll leave the issue open, to create the new icons or replace them with a more general one

CR @xavijam @juanignaciosl


xavijam:
Screenshot?

juanignaciosl:
Strictly speaking `license_info` can be nil, so it should be checked, but since old code hadn't the check I guess it can be skipped :+1:

matallo:
sure, ded01

screen shot 2016-01-18 at 16 21 23

xavijam:
:+1:

PR #6418, opened 2016-01-19
data-library breadcrumb fixes

### LINKED ISSUE #6256: data-library: link to data library index in breadcrumb
if feature flag "data_library" link to `/data-library`

matallo:
close #6256

when a `data_library_path` is set in the `app_config.yml` and the user has the `data_library` feature flag, breadcrumb changes from that for uses to one linking to data-library home

obv, this would need updating `app_config.yml`

CR @juanignaciosl @xavijam

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 -------------------
 * Update navigation in public pages.
 * New feature in organizations management: user quota slider.
+* Added `data_library_path` to the config. Usage instructions:
+  1. Configure data-library path. You can copy `data_library_path` config from `config/app_config.yml.sample` into your `config/app_config.yml`.
 
 3.12.0 (2016-01-05)
 -------------------
--- a/app/controllers/admin/visualizations_controller.rb
+++ b/app/controllers/admin/visualizations_controller.rb
@@ -176,6 +176,8 @@ class Admin::VisualizationsController < Admin::AdminController
     # Public export API SQL url
     @export_sql_api_url = "#{ sql_api_url("SELECT * FROM #{ @table.owner.sql_safe_database_schema }.#{ @table.name }", @user) }&format=shp"
 
+    @data_library_url = CartoDB.data_library_path.nil? ? nil : "#{request.protocol}#{CartoDB.account_host}#{CartoDB.data_library_path}"
+
     respond_to do |format|
       format.html { render 'public_dataset', layout: 'application_table_public' }
     end
--- a/app/controllers/admin/visualizations_controller.rb
+++ b/app/controllers/admin/visualizations_controller.rb
@@ -34,6 +34,8 @@ class Admin::VisualizationsController < Admin::AdminController
 
   before_filter :resolve_visualization_and_table_if_not_cached, only: [:embed_map]
 
+  before_filter :get_data_library_config, only: [:public_table]
+
   skip_before_filter :browser_is_html5_compliant?, only: [:public_map, :embed_map, :track_embed,
                                                           :show_protected_embed_map, :show_protected_public_map]
   skip_before_filter :verify_authenticity_token, only: [:show_protected_public_map, :show_protected_embed_map]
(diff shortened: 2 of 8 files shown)


app/views/data_library/index.html.erb
```diff
@@ -1,7 +1,6 @@
 <%= content_for(:title) do %>CartoDB · Data Library<% end %>
 
-<%# TODO: Description for data library %>
-<%= content_for(:description) do %><% end %>
+<%= content_for(:description) do %>This is the data library of CartoDB. Here you will find open resources to help you populate your maps.<% end %>
```
iriberri: Would this apply if the Data Library loaded is not our own but an organization one? ("the data library of CartoDB").

matallo: uhm good point, we've "CartoDB" pretty ingrained in views descriptions (i.e. https://github.com/CartoDB/cartodb/blob/master/app/controllers/admin/pages_controller.rb#L254-L265) so if necessary I'd open a new issue to fix this and create a new config field to set the name of the app

NEWS.md
```diff
@@ -2,6 +2,8 @@
 -------------------
 * Update navigation in public pages.
 * New feature in organizations management: user quota slider.
+* Added `data_library_path` to the config. Usage instructions:
+  1. Configure data-library path. You can copy `data_library_path` config from `config/app_config.yml.sample` into your `config/app_config.yml`.
```
juanignaciosl: How come is this path customizable?

app/views/admin/visualizations/public_dataset.html.erb
```diff
@@ -100,24 +100,40 @@
...
 
         <div class="PublicMap-leftBlock PublicMap-leftBlock--owner">
-          <ul class="Navmenu-list Navmenu-list--owner Navmenu-list--avatar">
-            <li class="Navmenu-item">
-              <a href="<%= CartoDB.url(self, 'public_user_feed_home', {}, @visualization.user) %>" target="_blank" class="UserAvatar">
-                <img class="UserAvatar-img--medium" src="<%= @visualization.user.avatar %>" alt="<%= @name %>" title="<%= @name %>" />
-              </a>
-            </li>
-          </ul>
+          <% if @viewed_user.has_feature_flag?("data_library") && @data_library_url %>
```
juanignaciosl: I wouldn't use feature flags here, a configuration parameter seems to be a better fit.

app/views/admin/visualizations/public_dataset.html.erb
```diff
@@ -100,24 +100,40 @@
...
+            <ul class="Navmenu-list Navmenu-list--owner Navmenu-list--avatar">
+              <li class="Navmenu-item">
```
juanignaciosl: I think there's too much code duplication here, isn't it? I think setting a variable either to `@ data_library_url ` or `artoDB.url(self, 'public_user_feed_home', {}, @visualization.user)` before would save many lines of code.

juanignaciosl:
I guess there'll be more discussion on this, ask for another CR if you agree upon doing changes.

matallo:
alright, now we have in the `app_config.yml`

```
  data_library:
    username: 'datalibrary'
    path: '/data-library'
```

and /data-library view would only render if username exists in the config (and it coincide with the viewed user), otherwise it will return 404

in that case only "Data Library" will appear as the name of the user

screen shot 2016-01-20 at 14 45 19

if username exists AND path exists both the avatar and the name will link to the data library path

screen shot 2016-01-20 at 14 44 21

for other users, breadcrumb will appear as is

screen shot 2016-01-20 at 14 35 01

CR @juanignaciosl

app/controllers/admin/visualizations_controller.rb
```diff
@@ -144,6 +144,15 @@ def public_table
     end
 
     @name = @visualization.user.name.present? ? @visualization.user.name : @visualization.user.username.truncate(20)
+    @user_url = CartoDB.url(self, 'public_user_feed_home', {}, @visualization.user)
+
+    has_data_library_username = Cartodb.config[:data_library].present? && !Cartodb.config[:data_library]['username'].blank?
```
juanignaciosl: Nothing that you need to change, but next time you need to read the configuration, you can use our nil-friendly `get_configuration` helper. Usage example: `Cartodb.get_config(:oauth, 'dropbox', 'app_key')`. It will return the value if all the keys are present, `nil` otherwise.

app/controllers/admin/visualizations_controller.rb
```diff
@@ -144,6 +144,16 @@ def public_table
...
+    @is_data_library = has_data_library_username && data_library_user?
```
juanignaciosl: Why not moving `has_data_library_username` check into `data_library_user?` as well?

juanignaciosl:
I have done some minor comments, take them into account if you want :+1:

matallo:
of course I'll take them into account, thanks mate :heart_eyes: `get_config`

PR #6422, opened 2016-01-19
Migrate user metadata config to redis

### LINKED ISSUE #6415: Migrate needed user data from metadata db to redis
- [x] Migrate the data from db to redis: soft_limit, nokia_quota, etc
- [x] Update user creation / update to update redis values too

ethervoid:
Closes #6415 

In order to use metadata config in the geocode API we need to store and synchronize some metadata fields in redis either for users and for organizations.

Also we have a new rake task to update all the existing users and organizations to have all this new fields available.

And finally there is a new redis database object in the editor that is going to be used to store the geocoder metrics although now is pointing to the same redis db than users_metadata.

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -340,7 +340,7 @@ class Organization < Sequel::Model
   end
 
   def period_end_date
-    owner.period_end_date
+    owner ? owner.period_end_date : nil
   end
 
   def public_vis_count_by_type(type)
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -80,12 +80,23 @@ class Organization < Sequel::Model
 
   def before_save
     super
+    @geocoding_quota_modified = changed_columns.include?(:geocoding_quota)
     self.updated_at = Time.now
     raise errors.join('; ') unless valid?
   end
 
   def before_destroy
     destroy_groups
+  end
+
+  def after_create
+    super
+    save_metadata
+  end
+
+  def after_save
+    super
+    save_metadata
   end
 
   # INFO: replacement for destroy because destroying owner triggers
@@ -282,6 +293,22 @@ class Organization < Sequel::Model
     display_name.nil? ? name : display_name
   end
 
+  # create the key that is used in redis
+  def key
+    "rails:orgs:#{name}"
+  end
+
+  # save orgs basic metadata to redis for other services (node sql api, geocoder api, etc)
+  # to use
+  def save_metadata
+    $users_metadata.HMSET key,
+      'id', id,
+      'geocoding_quota', geocoding_quota,
+      'google_maps_client_id', google_maps_key,
+      'google_maps_api_key', google_maps_private_key,
+      'period_end_date', period_end_date
+  end
+
   private
 
   def destroy_groups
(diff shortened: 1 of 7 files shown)


ethervoid:

app/models/organization.rb
```diff
@@ -278,6 +289,21 @@ def name_to_display
...
+  # create the key that is used in redis
+  def key
+    "rails:orgs:#{name}"
+  end
+
+  # save orgs basic metadata to redis for other services (node sql api, geocoder api, etc)
+  # to use
+  def save_metadata
+    $users_metadata.HMSET key,
+      'id', id,
```
rafatower: what's this `id`? I don't think we shall have the `key` with the org's `name` plus the id.

app/models/user.rb
```diff
@@ -736,12 +736,17 @@ def dashboard_viewed?
...
   def key
     "rails:users:#{username}"
   end
 
-  # save users basic metadata to redis for node sql api to use
+  # save users basic metadata to redis for other services (node sql api, geocoder api, etc)
+  # to use
   def save_metadata
     $users_metadata.HMSET key,
       'id', id,
```
rafatower: mmm, I see, this is legacy... wondering who is using that.

lib/tasks/db_maintenance.rake
```diff
@@ -1290,5 +1290,34 @@ namespace :cartodb do
...
+      else
+        user = ::User.where(username: args[:username]).first
+        update_user_metadata(user)
+      end
+    end
+
+    def update_user_metadata(user)
+      begin
+        if user.organization_owner?
+          user.organization.save_metadata
```
rafatower: maybe the decision and the save for organization metada is to be moved up, the name of the function turns out to be a bit misleading.

config/initializers/redis.rb
```diff
@@ -41,6 +41,7 @@
 $api_credentials     = Redis.new(redis_conf.merge(db: databases[:api_credentials]))
 $users_metadata      = Redis.new(redis_conf.merge(db: databases[:users_metadata]))
 $redis_migrator_logs = Redis.new(redis_conf.merge(db: databases[:redis_migrator_logs]))
+$geocoder_metrics    = Redis.new(redis_conf.merge(db: databases[:users_metadata]))
```
rafatower: this is not only just an alias but also a new object-connection and I'm not sure it is worth or even convenient

ethervoid: Yes after the conversation with rochoa and systems they told me could be a good idea to separate the config for the metrics part and the metadata part. In editor is easy, just make that and when the new config is available put it on the redis_conf part instead that.

If you think it don't worth it, I'll remove it

lib/tasks/db_maintenance.rake
ethervoid: Ok I'll split the method in two one for the user and other for the organization

config/initializers/redis.rb
rafatower: I missed that conversation and I'm getting a bit puzzled about what you want to achieve there. Also that conclusion seems to conflict with what we talked with rambo at the time if you remember.

rafatower: ok. for the reference: basically split metrics gathering and config storage. prepared in case we hit storage limits. qps or any other limit.

rafatower:
:+1:

ethervoid:
Splitter the update method of the rake in two: one for the user and other for the organization

Issue #6421, opened 2016-01-19
Check Excel files before passing them to in2csv


iriberri:
https://github.com/CartoDB/cartodb-management/issues/4383

It seems that a text file (CSV) with XLS extension might leave in2csv stuck.

iriberri:
https://github.com/CartoDB/cartodb/blob/master/services/importer/lib/importer/excel2csv.rb#L39

iriberri:
https://github.com/CartoDB/cartodb/pull/4389/files

PR #6430, opened 2016-01-20
Raise exception when CSV file with XLS extension is detected

iriberri:
Fixes https://github.com/CartoDB/cartodb/issues/6421 

* [x] Make sure that excel2csv specs are included in the Makefile

https://github.com/CartoDB/cartodb-management/issues/4383

### DIFF
--- a/services/importer/lib/importer/excel2csv.rb
+++ b/services/importer/lib/importer/excel2csv.rb
@@ -32,13 +32,12 @@ module CartoDB
         job.log "Converting #{@format.upcase} to CSV"
 
         Open3.popen3("file -b --mime-type #{filepath}") do |stdin, stdout, stderr, process|
-          @file_mime_type = stdout.read.delete("\n")
+          file_mime_type = stdout.read.delete("\n")
           job.log "Can't get the mime type of the file" unless process.value.to_s =~ /exit 0/
+          # CSV files with XLS extensions are considered malformed files
+          raise CartoDB::Importer2::MalformedXLSException.new if file_mime_type == "text/plain"
         end
-
-        # Take into account that here should come or csv files with xls extensions or xls documents
-        file_format = (@file_mime_type == "text/plain") ? "-f csv" : ""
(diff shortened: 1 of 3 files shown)


iriberri:
Please, @rafatower, take a look when you can :-)

rafatower:
Ok, but even better with a test, should be easy to write

iriberri:
The current tests for this should be actually failing as far as I've seen, but they aren't. Taking a look at them.

services/importer/lib/importer/excel2csv.rb
```diff
@@ -32,13 +32,12 @@ def run
...
+          file_mime_type = stdout.read.delete("\n")
           job.log "Can't get the mime type of the file" unless process.value.to_s =~ /exit 0/
+          # CSV files with XLS extensions are considered malformed files
+          raise CartoDB::Importer2::MalformedXLSException.new if file_mime_type == "text/plain"
         end
-
-        # Take into account that here should come or csv files with xls extensions or xls documents
-        file_format = (@file_mime_type == "text/plain") ? "-f csv" : ""
-        in2csv_command_line = %Q[in2csv #{filepath} #{file_format} | #{in2csv_warning_filter} | #{newline_remover_path} > #{converted_filepath}]
+        in2csv_command_line = %Q[in2csv #{filepath} | #{in2csv_warning_filter} | #{newline_remover_path} > #{converted_filepath}]
```
iriberri: Just for the record, `#{in2csv_warning_filter}` contains double quotes.

rafatower:
IIRC the test file needs to be added to the `Makefile` for CI to execute them, right?

iriberri:
Yes! I confirmed with Systems. Will update the Makefile for the CI to run again here and then we'll see. :-)

iriberri:
@luisbosque, I just made a spec addition into the Makefile. Could you please confirm that I won't be breaking anything or if there's something extra that needs to be done for it to run in CI? :heart:

luisbosque:
AFAIK that's the right way to add it. Go ahead!

iriberri:
Ready to be deployed then. Thanks!

Issue #6307, opened 2015-12-23
Set "Errno::EISDIR: Is a directory" as import user error


iriberri:
Sometimes, when using the API, users send incorrectly the file path and CartoDB is raising this error. This is not being detected as a user error at the moment.

Sample trace:

````
2016-01-20 01:37:27 UTC: Unpacking ./\n2016-01-20 01:37:27 UTC: Errored importing data:\n2016-01-20 01:37:27 UTC: Errno::EISDIR: Is a directory @ rb_sysopen - /tmp/imports/20160120-8589-x4cxv5/./.
2016-01-20 01:37:27 UTC: ----------------------------------------------------
2016-01-20 01:37:27 UTC: [\"/usr/lib/ruby/2.2.0/fileutils.rb:1392:in `initialize'\", 
\"/usr/lib/ruby/2.2.0/fileutils.rb:1392:in `open'\",
 \"/usr/lib/ruby/2.2.0/fileutils.rb:1392:in `block in copy_file'\", 
\"/usr/lib/ruby/2.2.0/fileutils.rb:1391:in `open'\", 
\"/usr/lib/ruby/2.2.0/fileutils.rb:1391:in `copy_file'\",
 \"/usr/lib/ruby/2.2.0/fileutils.rb:485:in `copy_file'\", 
\"/usr/lib/ruby/2.2.0/fileutils.rb:402:in `block in cp'\", 
\"/usr/lib/ruby/2.2.0/fileutils.rb:1570:in `block in fu_each_src_dest'\",
 \"/usr/lib/ruby/2.2.0/fileutils.rb:1584:in `fu_each_src_dest0'\",
 \"/usr/lib/ruby/2.2.0/fileutils.rb:1568:in `fu_each_src_dest'\", 
\"/usr/lib/ruby/2.2.0/fileutils.rb:401:in `cp'\", 
\"/home/ubuntu/www/production.cartodb.com/releases/x/services/importer/lib/importer/unp.rb:50:in `without_unpacking'\",
````

**Comment by iriberri** [comment]
@rafatower if you're not against, adding this to my list. This one seems to be punching import errors a lot, and it's indeed a user error and not a platform error (as it's detected now).

**Comment by iriberri** [comment]
```
[46, 55] in /home/ubuntu/www/production.cartodb.com/current/services/importer/lib/importer/unp.rb
   46:       end
   47: 
   48:       def without_unpacking(path)
[...]

PR #6433, opened 2016-01-20
Detect directories to notify incorrect use of Import API

iriberri:
* [x] Before merging, update docs accordingly: http://docs.cartodb.com/cartodb-editor/error-codes-and-solutions/

Fixes https://github.com/CartoDB/cartodb/issues/6307

### DIFF
--- a/services/importer/lib/importer/unp.rb
+++ b/services/importer/lib/importer/unp.rb
@@ -46,7 +46,7 @@ module CartoDB
       end
 
       def without_unpacking(path)
-        raise NotAFileError if File.directory?(path)
+        raise NotAFileError if !File.file?(path)
 
         local_path = "#{temporary_directory}/#{File.basename(path)}"
         FileUtils.cp(path, local_path)
--- a/services/importer/lib/importer/unp.rb
+++ b/services/importer/lib/importer/unp.rb
@@ -46,6 +46,8 @@ module CartoDB
       end
 
       def without_unpacking(path)
+        raise NotAFileError if File.directory?(path)
+
         local_path = "#{temporary_directory}/#{File.basename(path)}"
         FileUtils.cp(path, local_path)
         self.source_files.push(source_file_for(normalize(local_path)))
--- a/services/importer/spec/unit/unp_spec.rb
+++ b/services/importer/spec/unit/unp_spec.rb
@@ -31,7 +31,7 @@ describe Unp do
       unp.run(zipfile_factory)
       unp.source_files.length.should eq 2
     end
-  end #run
+  end
 
   describe '#without_unpacking' do
     it 'pushes a source file for the passed file path to the source files' do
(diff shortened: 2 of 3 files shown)


iriberri:
The details to reproduce the issue can be found: https://github.com/CartoDB/cartodb/issues/6307#issuecomment-173221350

iriberri:
Data Import result from reproduction in staging:

```json
{
id: "591109a2-62fb-4fde-b973-f4ee3b70a50a",
user_id: "868aad2f-e085-441a-afb8-7adb7c0802c6",
table_id: null,
data_type: "datasource",
table_name: null,
state: "failure",
error_code: 1010,
queue_id: "591109a2-62fb-4fde-b973-f4ee3b70a50a",
tables_created_count: null,
synchronization_id: null,
type_guessing: true,
quoted_fields_guessing: true,
content_guessing: false,
create_visualization: false,
visualization_id: null,
user_defined_limits: "{"twitter_credits_limit":0}",
get_error_text: null,
display_name: "591109a2-62fb-4fde-b973-f4ee3b70a50a",
success: false,
original_url: "",
warnings: null,
is_raster: false
}
```

iriberri:
Please, @rafatower, take a look when you can. Thanks! :-)

services/importer/lib/importer/unp.rb
```diff
@@ -46,6 +46,8 @@ def run(path)
       end
 
       def without_unpacking(path)
+        raise NotAFileError if File.directory?(path)
```
rafatower: better use
```ruby
raise NotAFileError if File.file?(path)
```
http://ruby-doc.org/core-2.2.3/File.html#method-c-file-3F

if that's precisely what you mean to do

services/importer/spec/unit/unp_spec.rb
```diff
@@ -41,7 +41,11 @@
       unp.without_unpacking(zipfile_factory)
       unp.source_files.size.should eq 1
     end
-  end #withount_unpacking
+
+    it 'raises if the path does not belong to a file' do
+      expect { Unp.new.without_unpacking('/var/tmp') }.to raise_error NotAFileError
+    end
+  end #without_unpacking
```
rafatower: I'd take advantage of this CR to remove the `#without_unpacking` line

rafatower:
basically ok, consider my comments (minor things)

services/importer/lib/importer/unp.rb
iriberri: Hm, not sure if I'm following.

Your line would raise an exception for any valid file, right? I want to raise an exception in the case that what is sent _is not_ a file (which could happen through the Import API).

I guess that a:

````
raise NotAFileError if !File.file?(path)
`````

could work too!

services/importer/spec/unit/unp_spec.rb
iriberri: Always a pleasure! :-)

services/importer/lib/importer/unp.rb
rafatower: correct, sorry my bad, BIG typo :dancer:

iriberri:
Just proposed a doc change [here](https://github.com/CartoDB/docs/pull/652). As soon as it's accepted this PR is ready to be deployed.

rafatower:
:+1:

PR #6443, opened 2016-01-22
Fixing JS tests

xavijam:
We have found several JS errors trying to make requires in several places, as in data_library, feed and explore "places".

REVIEWER: @javierarce


javierarce:
:+1:

Issue #6403, opened 2016-01-15
Detect error NoMethodError: undefined method close for 0:Fixnum true


iriberri:
We fixed the incorrect format detection, but this error can still be raised, for example, with a sheet that contains a graph in an Excel file.

PR #6445, opened 2016-01-22
Improve empty CSV files handling

iriberri:
For empty CSVs, which usually came from unsupported Excel files, such
as sheets with graphs, the normalization was crashing.

The first_line method was returning the `stream.rewind` value in case
that the file was empty, and therefore the first_line would be nil.

This commit also edits the CSV column number checking in order to
be able to admit empty files (otherwise, nil.length would raise an
exception).

* [x] Before merging, update docs.

Fixes https://github.com/CartoDB/cartodb/issues/6403

### DIFF
--- a/services/importer/lib/importer/csv_normalizer.rb
+++ b/services/importer/lib/importer/csv_normalizer.rb
@@ -170,11 +170,8 @@ module CartoDB
 
       def single_column?
         columns = ::CSV.parse(first_line, csv_options)
-        if !columns.any?
-          raise EmptyFileError.new
-        else
-          columns.first.length < 2
-        end
+        raise EmptyFileError.new if !columns.any?
+        columns.first.length < 2
       end
 
       def multiple_column(row)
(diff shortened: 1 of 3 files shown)


services/importer/lib/importer/csv_normalizer.rb
```diff
@@ -169,7 +169,12 @@ def needs_normalization?
       end
 
       def single_column?
```
rafatower: the else block is not needed. Typically it would be written like this:
```ruby
  def single_column?
       columns = ::CSV.parse(first_line, csv_options)
        raise EmptyFileError.new if !columns.any?
        columns.first.length < 2
    end
```
the point here is to separate "exceptional" flow from "normal flow" and keep the normal flow as "pure" as possible

iriberri: Thanks! :-) That was one of my attempts too, but I decided to go through the other way incorrectly, ha! Thanks for the feedback, very appreciated. :-)

rafatower:
:+1:

iriberri:
Docs edits PR'ed [here](https://github.com/CartoDB/docs/pull/654). As soon as the technical writes gives approval to my copy, this is ready to go.

PR #6473, opened 2016-01-27
Adds new TabPane

javierarce:
This PR:

- Adds the new `TabPane` component.
- Adds a factory to create two kinds of TabPanes (`createWithTextLabels` & `createWithIcons`)
- Adds a label and icon components (just to be added in the TabPane)
* Changes the paths (and name) of the components folder (lib  > assets > javascripts > cartodb3 > components)
* Updates the names of the tab pane files with the `tab-pane-` prefix
* Flattens the structure of the `tab-pane` components directory
* Lazy loading of the tab-pane content views.
* Removes the default model of the tab-pane component.
* Views within the factory now receive a model instead of a hash of options.
* Fixes specs and increased its coverage.
* Fixes spec system for the cartodb3 specs.

### Creating a TabPane directly

```js
var collection = new TabPaneCollection([
  {
    createButtonView: function() { return new cdb.core.View(); },
    createContentView: function () { return new createWadusView(arguments); }
  }, {
    createButtonView: createItemView,
    createContentView: createEmptyContentView
  }, {
    createButtonView: createItemView,
    createContentView: createEmptyContentView
  }, {
    createContentView: createEmptyContentView
  }
]);

var tabPaneView = new TabPaneView({
  collection: collection,
  template: require('./tab-pane-template')
});
```

### Creating a TabPane with a factory

In this case we use the factory `createWithTextLabels`:

```js
var tabPaneView = TabPaneFactory.createWithTextLabels([
  {
    label: 'Layers',
    createContentView: createEmptyContentView //= function() { return cdb.core.View() }
  }, {
    label: 'Elements',
    createContentView: createEmptyContentView
  }, {
    // Widget will be implemented from the start
    label: 'Widgets',
    selected: true,
    createContentView: function() {
      return new EditorWidgetsView(editorWidgetsCollection);
    }
  }
]);
```


lib/build/tasks/jasmine.js
```diff
@@ -24,12 +24,16 @@ exports.task = function () {
...
     cartodb3: {
-      src: [], // specs require their src files they need
+      src: [
+        '<%= assets_dir %>/javascripts/templates3.js',
+      ],
       options: {
         keepRunner: true, // do not delete the runner (added to gitignore anyway), makes sure the runner is up-to-date
         outfile: '_SpecRunner-cartodb3.html',
         summary: true,
+        keepRunner: true,
```
viddo: duplicated, see 3 lines :up: :

lib/build/tasks/jasmine.js
```diff
@@ -24,12 +24,16 @@ exports.task = function () {
...
         display: 'short',
+        helpers: ['http://maps.googleapis.com/maps/api/js?sensor=false&v=3.12'].concat(js_files._spec_helpers3),
```
viddo: why do we need to load google maps for the new specs?

lib/build/tasks/jst.js
```diff
@@ -20,6 +20,10 @@
             dest: "<%= assets_dir %>/javascripts/templates.js"
           },
           {
+            src: js_files._templates3,
+            dest: "<%= assets_dir %>/javascripts/templates3.js"
```
viddo: I think we can :fire: this templates3-thingy, since we should not need any jst templates for the new entry point.

lib/build/files/js_files.js
```diff
@@ -165,6 +165,11 @@ module.exports = files = {
     'lib/assets/test/spec/SpecHelper.js'
   ],
 
+  _spec_helpers3: [
+    'lib/assets/test/lib/sinon-1.3.4.js',
```
viddo: IMHO :fire:, we should avoid sinon if we can, jasmine already provide a spy mechanism that covers essentially all mocking/stubbing use-cases already.

lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane-collection.js
```diff
@@ -0,0 +1,55 @@
+var Backbone = require('backbone');
+var cdb = require('cartodb-deep-insights.js');
+var _ = require('underscore');
+
+module.exports = Backbone.Collection.extend({
+
+  model: cdb.core.Model,
```
javisantana: why cdb.core.Model and not a TabItemModel with `selected` set to false? is this model used in any other place?

lib/assets/test/spec/SpecHelper3.js
```diff
@@ -0,0 +1,248 @@
+/**
+ * Utilities to help the tests
+ * @type {Object}
+ */
+var TestUtil = {};
```
viddo: Do we really need _all_ these TestUtil stuff? From what I can see it's a lot of bagage to bring over from the old specs that we simply won't need for the new code.

lib/build/files/js_files.js
javisantana: then you would need to rewrite all the previous tests, right?

viddo: No, no need to touch/change any existing/current/old code. Note that the sinon lib here is added to the _spec_helpers3 which is only used for the new test bundle/runner, but these new tests don't even use sinon.

lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane-collection.js
```diff
@@ -0,0 +1,55 @@
...
+    this._setSelected();
+  },
+
+  getSelected: function() {
+    return this.find(function(model) {
+      return model.get('selected');
+    });
+  },
+
+  _checkSelected: function(model) {
```
viddo: No refs to this function, so :fire:

lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane-collection.js
```diff
@@ -0,0 +1,55 @@
...
+
+  _setSelected: function() {
+    var isSelected = this.getSelected();
+
+    if (!isSelected && this.size() > 0) {
+      this.at(0).set('selected', true)
+    }
+  },
+
+  _onItemSelected: function(itemModel, isSelected) {
```
viddo: This name is a bit misleading, why not us the convention and have it match the event? I.e. `_onSelectedChange`?

lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane-collection.js
```diff
@@ -0,0 +1,55 @@
...
+    if (!isSelected) {
+      return;
+    }
+
+    this.each(function(model) {
+      if (model.cid !== itemModel.cid) {
```
viddo: Why compare with .cid instead of the simpler `model !== itemModel`?

lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane-factory.js
```diff
@@ -0,0 +1,78 @@
...
+  },
+
+  /*
+   * {
+   *   icon: 'my-icon',
+   *   createContentView: cdb.core.View(),
+   *   selected: false
+   * }
+   */
+  createWithIcons: function(paneItems) {
```
viddo: I still make the argument that this won't be a generic implementation (which will be obvious once we implement the actual styles for the top-menu bar), but we can move it later I guess.

lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane-factory.js
```diff
@@ -0,0 +1,78 @@
...
+      });
+
+      return {
+        selected: paneItem.selected,
+        label: paneItem.label,
+        createButtonView: function() {
+          return new LabelView({ model: this });
+        },
+        createContentView: function() {
+          return paneItem.createContentView && paneItem.createContentView() || new cdb.core.View();
```
viddo: I'd change this to `createContentView: paneItem.createContentView` because:
1. can't control what context the fn will have
1. The fallback code could/should be handled by the TabPaneView, to not have to repeat it in multiple places

viddo: also, createContentView is already checked if it's truthy a few lines earlier, so should not need to guard for it to exist again here

lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane-view.js
```diff
@@ -0,0 +1,71 @@
...
+
+    if (model) {
+      this._renderSelected(model, true);
+    }
+
+    return this;
+  },
+
+  _renderSelected: function(model, isSelected) {
+    if (this._selectedView) {
```
viddo: This block should probably be wrapped in the `if (isSelected) {` too, otherwise a `selected:false`-change that comes after a `selected:true`-change will delete the content-view, no?

lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane-view.js
```diff
@@ -0,0 +1,71 @@
...
+  },
+
+  _renderTabPaneItemView: function(model) {
+    var tabPaneItemView = new TabPaneItem({ model: model });
+    this.addView(tabPaneItemView);
+    this.$('.js-menu').append(tabPaneItemView.render().$el);
+  },
+
+  _renderTabPaneContentView: function(model) {
+    var tabPaneContentView = new TabPaneContent({ model: model });
```
viddo: Looking at the implementation of `TabPaneContent` I don't see any need for it (it just wraps the createContentView call, why not just do `var tabPaneContentView = model.get('createContentView').call(model)` here?

lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane-view.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    this.template = this.options.template || Template;
+
+    this.collection.bind('change:selected', this._renderSelected, this);
+  },
+
+  render: function() {
+
+    this.$el.append(this.template());
```
viddo: Why `.append` instead of `.html` to inject the new DOM content?

viddo:
A bunch of comments, but overall :+1: 
I think we might need to :sparkles: some details when implementing the CSS/markup the actual use-cases, but they can be dealt with then. Good work!

Btw, please runt `grunt test` locally until the CI is working again.

lib/build/tasks/jasmine.js
javierarce: yes, we don't needed… we just copied from above. I'll remove it.

lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane-collection.js
javierarce: You're totally right.

lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane-view.js
javierarce: True :+1:

lib/assets/test/spec/SpecHelper3.js
javierarce: I've removed a bunch of the methods we aren't using. @xavijam: can you take a look too, please?

Cartofante:

lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane-collection.js
javierarce: Because we don't have (and don't really need) a TabItemModel :) I think we could get rid of that line and it would keep working. 

What do you think, @xavijam?

xavijam: :fire:

lib/build/tasks/jst.js
xavijam: Agree

lib/assets/test/spec/SpecHelper3.js
xavijam: :+1:

PR #6476, opened 2016-01-27
Changing map item url to 'editor-v3' instead of old vis entry

xavijam:
I have some doubts, what do you think?

RW: @viddo


viddo:
I'd skip this for now.

xavijam:
ook!

PR #6484, opened 2016-01-28
Overviews (create overviews on import)

jgoizueta:
This PR integrates overview creation with data imports.


services/importer/lib/importer/overviews.rb
```diff
@@ -0,0 +1,65 @@
...
+      end
+
+      private
+
+      def table_row_count(schema, table)
+        # TODO: TableRowCount is not intended to be used like this
+        # (to compute the table row count; method `get` is protected)
+        # could have a class (derived from TableRowCount for implementation)
+        # with this purpose
+        table_row_count = CartoDB::PlatformLimits::Importer::TableRowCount.new(
```
ethervoid: Maybe you could use this https://github.com/CartoDB/cartodb/blob/master/app/models/table.rb#L708

services/importer/lib/importer/overviews.rb
```diff
@@ -0,0 +1,65 @@
...
+    class Overviews
+
+      DEFAULT_MIN_ROWS = 1000000
+
+      def initialize(runner, user, options = {})
+        @runner = runner
+        @user = user
+        @database = options[:database] || user.in_database
+        @min_rows =  options[:min_rows] ||
+                     (Cartodb.config[:overviews] && Cartodb.config[:overviews]['min_rows']) ||
```
ethervoid: This could crash. I'd use https://github.com/CartoDB/cartodb/blob/599ec76c5f717f3d5f00ea92882f0d1112b3eecb/config/initializers/01_app_config.rb#L2

jgoizueta: Nice! I'll get to it

services/importer/lib/importer/overviews.rb
jgoizueta: ...And maybe should be passing a Table rather than the table name to `create!` ?

PR #6486, opened 2016-01-29
Interview Test cartodb broken sync

janayama:
Fix the issue with the synchronization when a external resource has changed the schema.

cc/ @juanignaciosl


Cartofante:

PR #6487, opened 2016-01-29
Disable geocoder api notifications for oss users

ethervoid:
We want to disable and remove noise for the OSS users of CartoDB that don't have, by now, a complete instance of Geocoder API

### DIFF
--- a/app/models/user/db_service.rb
+++ b/app/models/user/db_service.rb
@@ -415,7 +415,14 @@ module CartoDB
               SET search_path TO #{build_search_path}") if @user.organization_user?
           return true
         rescue => e
-          CartoDB.notify_error('Error installing and configuring geocoder api extension', error: e.inspect, user: @user)
+          # For now CartoDB.com is the only host with Geocoder API installed
+          # so we want to avoid noise to OSS users
+          if Cartodb.config[:cartodb_com_hosted] == false
+            CartoDB.notify_error(
+              'Error installing and configuring geocoder api extension',
+              error: e.inspect, user: @user
+            )
+          end
           return false
       end
 
--- a/app/models/user/db_service.rb
+++ b/app/models/user/db_service.rb
@@ -408,31 +408,32 @@ module CartoDB
       end
 
       def install_and_configure_geocoder_api_extension
-          install_geocoder_api_extension
+          geocoder_api_config = Cartodb.get_config(:geocoder, 'api')
+          # If there's no config we assume there's no need to install the
+          # geocoder client as it is an independent API
+          return if geocoder_api_config.blank?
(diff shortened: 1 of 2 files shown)


rafatower:
as per our conversation, I'd rather prefer to change this:

https://github.com/CartoDB/cartodb/blob/master/app/models/user/db_service.rb#L1332

with that:
```ruby
# If there's no config we assume there's no need to install the geocoder client as it is an independent API
return if config.blank?
```

ethervoid:
It's done in the last commit :|

app/models/user/db_service.rb
```diff
@@ -415,7 +415,14 @@ def install_and_configure_geocoder_api_extension
               SET search_path TO #{build_search_path}") if @user.organization_user?
           return true
         rescue => e
-          CartoDB.notify_error('Error installing and configuring geocoder api extension', error: e.inspect, user: @user)
+          # For now CartoDB.com is the only host with Geocoder API installed
+          # so we want to avoid noise to OSS users
+          if Cartodb.config[:cartodb_com_hosted] == false
```
rafatower: I don't agree with this change

config/app_config.yml.sample
```diff
@@ -171,11 +171,6 @@ defaults: &defaults
       base_url: ''
       api_key: ''
       table_name: ''
-    api:
```
rafatower: same must be done in app_config.yml.testing

app/models/user/db_service.rb
ethervoid: removed

config/app_config.yml.sample
ethervoid: This config is not in the testing file

ethervoid:
@rafatower check it to deploy if its ok

rafatower:
Looks ok, but can you please check these errors in the tests?
```
13:43:41 CartoDB::Visualization::TableBlender
13:43:43 error: Error installing and configuring geocoder api extension
13:43:43 {:error=>"#<TypeError: wrong argument type nil (expected String)>", [...]
```
(btw missing a backtrace there, is `inspect` what we want?)

config/app_config.yml.sample
rafatower: my bad, it was in my PR :dancer:

rafatower:
By looking into the code I'd say we should skip entierly `install_and_configure_geocoder_api_extension` if it is not in app_config

ethervoid:
I come to the same conclusion, context switching isn't good for focus xD

ethervoid:
Solved the error traces. Tomorrow I'll deploy it

rafatower:
:+1: great!

PR #6493, opened 2016-01-29
public map: fix buttons and caret

### LINKED ISSUE #6490: Mising styles in sign up buttons
In public pages
image

matallo:
close #6490

screen shot 2016-01-29 at 18 44 01

CR @piensaenpixel @xavijam

### DIFF
--- a/lib/build/files/css_files.js
+++ b/lib/build/files/css_files.js
@@ -41,7 +41,9 @@ module.exports = css_files = {
   ],
 
   public_dashboard: [
-    '<%= assets_dir %>/stylesheets/public_dashboard/**/*.css'
+    '<%= assets_dir %>/stylesheets/public_dashboard/**/*.css',
+    '<%= assets_dir %>/stylesheets/public_map/public_map_buttons.css',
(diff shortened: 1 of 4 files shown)


xavijam:
caretDown is a new icon from CartoAssets or it is already there?

matallo:
Checked and it was already there, it's in ded02

matallo:
so, is it a :+1: ? :)

lib/build/files/css_files.js
```diff
@@ -41,7 +41,9 @@ module.exports = css_files = {
   ],
 
   public_dashboard: [
-    '<%= assets_dir %>/stylesheets/public_dashboard/**/*.css'
+    '<%= assets_dir %>/stylesheets/public_dashboard/**/*.css',
+    '<%= assets_dir %>/stylesheets/public_map/public_map_buttons.css',
```
xavijam: Why do you add them to the public dashboard?

matallo: that is the view where those styles were missing

xavijam:
I don't know if we should move those buttons to a public common view, but in any case, :+1:. Let's face that problem in another place.

Issue #6459, opened 2016-01-25
Replace temporary solution for CSS for new entry point


viddo:
In #6458 we added the bundle generated by deep-insights to the `vendor/stylesheets` dir, AFAIK we should really use the https://github.com/CartoDB/CartoAssets once it's ready, so as part of integrating that this temporary CSS should be :fire:

viddo:
Reopening because we still need to :fire: the temporary files as parto f adding cartoassets https://github.com/CartoDB/cartodb/commit/76d3f19072f3d50f5ca6dc4b4d848e4c676e471b 
cc @xavijam

javierarce:
:ok_hand:

xavijam:
It should be covered in branch: ```issue-6455```.

PR #6495, opened 2016-02-01
Creating new editor stylesheets from deep-insights and cartodb.js#v4 modules

xavijam:
Basically we have added 'cartodb.js#v4' as a module in the package.json. With that, we need to avoid compatibility problems with current cartodb.js version so we have renamed all old requires from cartodb.js with ```require('cartodb-v3')```.
That said, we can get all necessary stylesheets from both modules (deep-insights and cartodb.js#v4) in order to create the proper editor-v3 stylesheet :).

Fixes https://github.com/CartoDB/cartodb/issues/6459.
CR: @viddo, reviewing in RUI servers.


viddo:
Too big to comment inline so I'll reference file here:
- why execute `jasmine:cartodb3',` on `grunt dev`? Adds additional load time on startup that should not be necessary
- I think it would be more to rename the shim to `cartodb.js-v3` to make it clear that it's an older version of cartodb**.js** that's being required.

Apart from that looks :+1:

xavijam:
- [x] why execute jasmine:cartodb3', on grunt dev? Adds additional load time on startup that should not be necessary
- [x] I think it would be more to rename the shim to cartodb.js-v3 to make it clear that it's an older version of cartodb.js that's being required.
- [x] Updated version of compass for cartodb.js repository too. Having the same gem for both, winner.

viddo:
:dancer:

PR #6500, opened 2016-02-01
Widgets html

xavijam:
screen shot 2016-02-01 at 18 04 09

First and basic structure for new editor HTML.

CR: @javierarce.
cc @viddo


javierarce:
:+1: thanks!

app/assets/stylesheets/editor-3/editor.css.scss
```diff
@@ -0,0 +1,38 @@
...
+// --------------------------------------------------
+
+@import 'cdb-utilities/mixins';
+@import 'cdb-variables/colors';
+
+@mixin block-shadow() {
+  @include css3-prefix(box-shadow, inset 10px 0 10px -10px rgba(0, 0, 0, 0.3));
+};
+
+.Editor {
```
javierarce: If this is going to be temporal, maybe it'd be a good idea to mark it with a TODO, right?

xavijam: It is temporal because we will implement a better CSS, but for the moment it does the trick.

PR #6509, opened 2016-02-02
Check import URLs against a blacklist of IP addresses

iriberri:
Currently, the importer does not perform any kind of check when a
URL is requested to be imported. In production environments, this
blacklist can be used in order to block imports from specific IPs
to improve security measures.

This commit adds a new entry in the app_config that allows to specify
specific IP addresses or ranges of IP addresses from which
imports will not be allowed.

Notes:

`Resolv` is being used to get the IPs from hostnames given that Typhoeus does not have a resolving function:  https://github.com/typhoeus/typhoeus/issues/482

### DIFF
--- a/config/app_config.yml.sample
+++ b/config/app_config.yml.sample
@@ -177,6 +177,7 @@ defaults: &defaults
       dbname: ''
       user: ''
   importer:
+    blacklisted_ip_addr: []
     content_guessing:        # Depends on geocoding
       enabled:         false # Disabled if false or not present
       threshold:       0.9   # 90% or more matches
--- a/lib/carto/url_validator.rb
+++ b/lib/carto/url_validator.rb
@@ -1,4 +1,6 @@
 require 'uri'
+require 'resolv'
+require 'ipaddr'
 
 module Carto
   module UrlValidator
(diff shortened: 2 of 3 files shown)


iriberri:
cc @ethervoid

lib/carto/url_validator.rb
```diff
@@ -18,9 +20,21 @@ def validate_url!(str, valid_ports = DEFAULT_URL_VALID_PORTS)
...
-      return uri.is_a?(URI::HTTP) && valid_ports.include?(uri.port)
+      return uri.is_a?(URI::HTTP) && valid_ports.include?(uri.port) && !blacklisted_ip?(uri)
     rescue URI::InvalidURIError
       return false
     end
+
+    def blacklisted_ip?(uri)
+      # The uri includes regularly a hostname and is not directly the IP address
+      # so we need to resolve it to compare it with the IP blacklist
+      uri_ip = IPAddr.new(Resolv.getaddress(uri.host))
```
iriberri: Rescue possible timeouts/failures on resolve

lib/carto/url_validator.rb
```diff
@@ -1,8 +1,10 @@
 require 'uri'
+require 'resolv'
 
 module Carto
   module UrlValidator
     DEFAULT_URL_VALID_PORTS = [80, 443]
+    BLACKLISTED_IP_RANGES = Cartodb.config[:importer]['blacklisted_ip_addr']
```
ethervoid: This could lead to an error if the element doesn't exists. Better use https://github.com/CartoDB/cartodb/blob/master/config/initializers/01_app_config.rb#L2 instead and dont put it as a constant

config/app_config.yml.sample
```diff
@@ -177,6 +177,7 @@ defaults: &defaults
       dbname: ''
       user: ''
   importer:
+    blacklisted_ip_addr: []
```
ethervoid: This could be a quick start but if we have more than 2 or 3 blacklisted ips we should move it to db

lib/carto/url_validator.rb
ethervoid: Yes, we have to take into account that

ethervoid:
Despite de comments, I miss some tests for this new functionality

iriberri:
Thanks for the review, mate!

Working on all that :-) Tests too, ha! ;-)

config/app_config.yml.sample
iriberri: Interesting :-) At first, we're just thinking about a single IP, but definitely something to take into account.

iriberri:
Hey @ethervoid, I tried to apply your feedback :-)

Now tests are included, the IP blacklist is no longer a constant and I've changed a bit the logic too.

ethervoid:
Looks good to me. :+1:  :smile:

lib/carto/url_validator.rb
```diff
@@ -18,9 +20,34 @@ def validate_url!(str, valid_ports = DEFAULT_URL_VALID_PORTS)
...
+        begin
+          uri_ip = IPAddr.new(Resolv.getaddress(uri.host))
+        rescue Resolv::ResolvError, Resolv::ResolvTimeout
+          return false
+        rescue IPAddr::AddressFamilyError, IPAddr::InvalidAddressError
+          raise InvalidUrlError, str
+        end
+
+        blacklisted_ip_ranges.each do |ip_range|
+          return true if IPAddr.new(ip_range).include?(uri_ip)
```
iriberri: Going to change this. `blacklisted_ip_ranges` should return an array of IP objects instead of an array of strings. Too much `new` here :-)

iriberri:
I pushed an extra test to make explicit that IP ranges can be used. Also, I tweaked the @blacklisted_ip_ranges generation so that new IPAddr is instantiated just once, not each time that the code was run, which was happening before.

iriberri:
retest this please

rafatower:
:+1:

PR #6521, opened 2016-02-04
Port cdb.admin.Table model (and its underlying concepts) for the new editor

### LINKED ISSUE #6516: Implement table model, needed for new form inputs
Provide means to access a user's tables, and their schema (columns, types etc.)
Will be used for #6508 

My initial idea is to port the minimal things necessary from old models, i.e.:
```js
// setup root model in entry file (e.g. user)
var user = new User({…})
var tables = user.tablesCollection // => Backbone.Collection
var columns = tables.columnsCollection // => Backbone.Collection
console.log(columns.chain().pluck('name')) // => ['id', 'cartodb_id', 'homes', 'some_text', …]
```

cc @xavijam @javierarce makes sense? something missing that you need?

--- 

TODOs
- [x] implement `User` model (the root model that owns a `tables` collection) 
- [x] implement `Table` model + collection
- [x] implement `Column` model + collection

viddo:
resolves #6516

Due to a lot of legacy (e.g. missing models/collections, logic in views etc.), and strong attachment to cartodb.js v3 it was not doable to just copy-paste the old models and be done with. Instead, I migrated just enough for what we need under current requirements. This way we'll have a good base to continue from too, since some concepts were plainly missing (e.g. columns, logic is spread out in various views).

```js
userModel.tablesCollection.fetch()
// …

t=userModel.tablesCollection.first()
// to get actual table data, e.g. the schema (columns)
t.fetch()
// …

c=t.columnsCollection.first()
c.attributes // => {name: "cartodb_id", type: "number"}
```

I also take the opportunity to clarify how URLs should be generated following Backbone conventions/best-practices, I wrote it on the [internal doc](https://github.com/CartoDB/cartodb-management/wiki/Backbone-best-practices) (perhaps we should move that under `docs/` in this repo instead?):
> - If a model belongs to a collection the URL is implicitly generated from it automatically, i.e. no need to implement any url method.
- query strings should _not_ go in the url function, but in the [fetch call](http://backbonejs.org/#Collection-fetch) as `model.fetch({ data: { foo: 'bar', baz: 123 }})} // => /url?foo=bar&baz=123`. 
  - If you want defaults override the fetch method and set sensible defaults there.
- _Exception:_ In the case that a model have some special path and/or can't follow the convention, it's correct to implement/override the url method.

I have some concerns about some details, I added TODOs for those in the code, I'll comment further in the changeset so we can have threaded discussions on those topics.

@xavijam can you review?
cc @javisantana @alonsogarciapablo @acanimal @javierarce FYI


lib/assets/javascripts/cartodb3/data/tables-collection.js
```diff
@@ -0,0 +1,71 @@
...
+  },
+
+  // Overrides the default fetch, to use the internal methods to construct parmas
+  fetch: function (opts) {
+    opts = opts || {
+      data: {
+        // If reaches this code path it's because there were no opts given, i.e. should do a 'full fetch'
+        // Since there is no current way to really do a full fetch let's just set a really high number to get all…
+        // TODO this is obviously bad for organization users, how can we do this differently
+        per_page: 1000
```
viddo: @javisantana @xavijam This is a funny one.
Despite that we have no `/tables` endpoint, but utilizing /viz/, we can hide away that detail and have a proper tables collection like I've done here. 

The only caveat is that the vis data is obviously incomplete, so to actually have table data one need to fetch the table data one-by-one. See the parse method below for details on how I set this up for now.

Btw, now that we use Backbone v1.x we could use something like https://github.com/backbone-paginator/backbone.paginator perhaps, to have a consistent way of implementing pagination of large collections instead of having to maintain custom in-house code.

viddo: Btw, in the current editor it's limited to the default (9), so one do actually not have all tables on the `user.tables` array… related to this, since some UI elements might not be able to display _all_ items (e.g. tables), this needs to be taken into account for the new designs, cc @urbanphes @saleiva

lib/assets/javascripts/cartodb3/data/map-definition-model.js
```diff
@@ -0,0 +1,28 @@
...
+    this._baseUrl = opts.baseUrl;
+
+    this.layerDefinitionsCollection = new LayerDefinitionsCollection([], {
+      // The layer definition URL is under the path of this model's API url
+      // id might not be set from the beginning though, so the url needs to be constructed when needed
+      baseUrl: this.url.bind(this)
+    });
+  },
+
+  url: function () {
```
xavijam: It is likely we could replace it by ```urlRoot``` method.

lib/assets/javascripts/cartodb3/data/table-model.js
```diff
@@ -0,0 +1,47 @@
...
+    schema: []
+  },
+
+  initialize: function (models, opts) {
+    if (!opts.baseUrl) throw new Error('baseUrl is required');
+    if (!opts.configModel) throw new Error('configModel is required');
+    this._baseUrl = opts.baseUrl;
+    this._configModel = opts.configModel;
+
+    this.columnsCollection = new ColumnsCollection([], {
```
xavijam: Could there be a change in this collection instead of the model schema? If so, how would the connection be between ```schema``` attribute and the ```columnsCollection```?

lib/assets/javascripts/cartodb3/editor.js
```diff
@@ -81,24 +63,29 @@ dashboard.vis.done(function () {
...
 
-visualizationModel.dashboard = dashboard;
-window.visualizationModel = visualizationModel;
+// Expose the root stuff to be able to inspect and modify state from developer console
+visDefinitionModel.dashboard = dashboard;
+window.configModel = configModel;
+window.userModel = userModel;
+window.dashboard = dashboard;
+window.vis = dashboard.vis;
+window.visDefinitionModel = visDefinitionModel;
```
xavijam: everything public :dancer:

xavijam:
Two :+1: as petition of @javierarce.

lib/assets/javascripts/cartodb3/data/table-model.js
viddo: When table is fetched, the schema gets updated which would update the columns collection; works, implemented, in `.parse`.

We don't need to modify a table right now, but it would be easy to migrate/copy-paste the old code for that to work once that time come. It's not a prio right now though.

PR #6530, opened 2016-02-08
BB - IP Blacklisting - Check final URL after Typhoeus request

iriberri:
The original URL validator checks the initial URL that
is provided by the user, but this validation check could
be tricked in case that the checked URL performs a redirect.

Now, the redirect_count is checked after Typhoeus's first
request to know if the URL performs a redirect. In case it does,
the final URL (effective_url) is checked again against the
url_validator.

By checking the redirect_count value we avoid running the
validator in those cases which are already covered by the validation
performed by the import controller.

### DIFF
--- a/services/importer/lib/importer/downloader.rb
+++ b/services/importer/lib/importer/downloader.rb
@@ -14,10 +14,13 @@ require_relative './url_translator/google_docs'
 require_relative './url_translator/kimono_labs'
 require_relative './unp'
 require_relative '../../../../lib/carto/http/client'
+require_relative '../../../../lib/carto/url_validator'
 
 module CartoDB
   module Importer2
     class Downloader
+
+      extend Carto::UrlValidator
 
       # in seconds
       HTTP_CONNECT_TIMEOUT = 60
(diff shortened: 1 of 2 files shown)


iriberri:
retest this please

rafatower:
I don't like much the idea of "converting" `InvalidUrlError` into a `CouldntResolveDownloadError` but much better than nothing. As usual, there's a certain asymmetry between syncs and regular imports.

I'd take advantage of this PR to add `maxredirs` param: https://github.com/typhoeus/typhoeus/issues/326

Aside, I'm checking Typhoeus / libcurl to see if there's a better way of doing that, instead of after-the-fact checking.

rafatower:
> Aside, I'm checking Typhoeus / libcurl to see if there's a better way of doing that, instead of after-the-fact checking.

Oh, since it is `on_headers` it should be ok.

iriberri:
@rafatower, about the exception: agree. Actually, I used this one because of the description that is already defined [here](https://github.com/CartoDB/cartodb/blob/e3725044e192d74df3c5fef597b1616315fa2074/lib/cartodb/import_error_codes.rb#L132), which sounds appropriate. Although the naming of the exception is confusing.

rafatower:
:+1: fits the bill, ok

iriberri:
@rafatower with respect to `maxredirs`, which would you think that could be a secure value but at the same time, not break anything? 3?

rafatower:
3 is (arbitrary) and fine by me. 5 is also ok. I'd add it as a constant `MAX_REDIRECTS = 5`. Anything better than infinite is good. :smiley: 

On a second thought, this PR is ready so you can add it to a different one. Up to you.

iriberri:
Sounds fine! Thanks! :revolving_hearts:

PR #6559, opened 2016-02-10
[wip] Adapt Varnish invalidation keys to multi-database format

zenitraM:
With FDWs, tables can come from multiple databases.

This is specified on the X-Cache-Channel by:
`database_name:table1,table2;;other_database_name:table2,table3`

This changes the regex in the invalidation triggers to match this new schema and ensure that tables called the same on different databases don't invalidate each other: 
http://rubular.com/r/TSwsUWnRtL

Also, ensures that X-Cache-Channel keys involving schemas with special characters are fully-qualified (related with https://github.com/CartoDB/cartodb-postgresql/issues/198)

cc @rafatower

### DIFF
--- a/app/models/table.rb
+++ b/app/models/table.rb
@@ -593,9 +593,9 @@ class Table
 
   def varnish_key
     if owner.db_service.cartodb_extension_version_pre_mu?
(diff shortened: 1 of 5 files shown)


zenitraM:
also this probably will need to be changed on other things that invalidate things @rochoa

zenitraM:
Superseded by #6805

PR #6562, opened 2016-02-11
fixes avatar urls, now they can point to an absolute url or a local url.

javisantana:
- this change requires to modify product chef script so base_url contains a "//" at the begging, rails no longer gets the url as a absolute one cc @santisaez @azamorano 
- requires changes in onpremise configuration, in app_config, avatars section should be like
```
 avatars:
    base_url: '/assets/unversioned/images/avatars'
    kinds: ['star', 'mountain', 'heart', 'ghost', 'planet']
    colors: ['green', 'yellow', 'red', 'orange']

```

- I removed image_tag because was doing magic (prepending assets path when the image is not absolute)

cc @xavijam


xavijam:
Fine by me, default avatars to a different folder :+1:. Not sure if we were using those old urls in the frontend (I don't think so).

PR #6568, opened 2016-02-11
Stack layout component

### LINKED ISSUE #6548: Create stack-layout component
In order to create the first approach to the hierarchy of the new editor.

xavijam:
stack-layout

Implemented the stack layout component (version 1, no animation yet) and applied for the current editor3 layout.

Fixes #6548.
CR: @javierarce @viddo (men in pair programming, men in double code reviewing).


lib/assets/javascripts/cartodb3/editor/editor-widget-view.js
```diff
@@ -11,12 +10,18 @@ module.exports = cdb.core.View.extend({
...
     'click .js-remove': '_onRemove',
-    'click .js-edit': '_onEdit'
+    'click': '_onEdit'
   },
 
   initialize: function (opts) {
-    if (!opts.tableModel) throw new Error('tableModel is required');
-    this._tableModel = opts.tableModel;
+    if (!opts.tableModel) {
+      throw new Error('Table model should be provided');
```
javierarce: minor stuff, we are using this kind of error messages already:

`throw new Error('tableModel should be provided');`

ditto for `nextStackItem`

xavijam: Changed :dancer:

viddo:
- Extract the state logic from the the stack layout view to a view model.
- StackLayoutView to listen to `change:position` events on the view model, and clean/render view(s) accordingly (i.e. the `gotoPrev/NextStep` logic of removing/creating views can stay in the view)
- …with these changes, we can pass the `viewModel` when creating a stack item instead of individual closures, which now have the semantic methods on the view model i.e.
```js
// …
{
  createStackItemView: function(stackViewModel) {
    // since first only can go fwd:
     stackViewModel.gotoNext();
  },
  createStackItemView: function(stackViewModel) {
   // since in the middle can go both back and fwd
    stackViewModel.gotoPrev();
    stackViewModel.gotoNext();
  },
  createStackItemView: function(stackViewModel) {
    // Last step, can only do:
    stackViewModel.gotoNext()
  }}
```
- The template should not be necessary, going prev/next could/should be translated into `this.$el.append(stackItemView.el)` or `this.$el.append(stackItemView.el)` depending on direction. The markup/style will depend on the context and as such should be implementing by the individual stack item views.

xavijam:
What do you think now?

lib/assets/javascripts/cartodb3/components/stack-layout/stack-layout-model.js
```diff
@@ -0,0 +1,53 @@
...
+
+    if (nextPos >= stackLayoutItemsSize) {
+      throw new Error('There is no ' + nextPos + ' stack view in the collection');
+    } else {
+      this.set({
+        position: nextPos
+      }, {
+        silent: true
+      });
+      this.trigger('change:position', arguments);
```
viddo: Considering `change:attr` is a standard backbone thing I would find it very surprising that the params to the callback to not match the standard [`(model, newVal, options)`](http://backbonejs.org/#Events-catalog), instead I'd suggest to either…
```js
// …use a completely different, custom event, e.g.:
model.on('positionChanged', function(position, customArgs) {
//…

// or pass as after the default args: 
model.on('change:position', function(m, newPosition, options, customArgs) {
//…
```

lib/assets/javascripts/cartodb3/editor.js
```diff
@@ -74,6 +81,34 @@ var userModel = new User(userData, {
   tablesCollection: tablesCollection
 });
 
+// Setup root view
+// TODO for now we only render widgets, but at some point this should be "wrapped" by a EditorView
+this.collection = new Backbone.Collection([
+  {
+    createStackView: function (stackLayoutModel, opts) {
+      return new EditorWidgetsView({
+        nextStackItem: stackLayoutModel.nextStep,
```
viddo: why not pass the stackLayoutModel directly to the view? so the view internally calls `stackLayoutModel.nextStep(a,b,c)`, which would also remove the need to bind the functions to a certain scope ahead of time (like done using the _.bindAll in the model right now).

viddo:
a part from the minor comments :shipit:

lib/assets/javascripts/cartodb3/components/stack-layout/stack-layout-model.js
xavijam: Better custom event, :+1:.

lib/assets/javascripts/cartodb3/editor.js
xavijam: yep, makes sense, implemented.

xavijam:
I hope this is the final change :fingers_crossed:.

xavijam:
Frontend tests passed, looks ok now @javierarce @viddo?

javierarce:
:ok_hand: fine for me

lib/assets/javascripts/cartodb3/editor/editor-widget-view.js
```diff
@@ -28,11 +30,7 @@ module.exports = cdb.core.View.extend({
   },
 
   _onEdit: function () {
-    var view = new WidgetsFormContentView({
-      widgetDefinitionModel: this.model,
-      tableModel: this._tableModel
-    });
-    this.$el.append(view.render().$el);
+    this.stackLayoutModel && this.stackLayoutModel.nextStep(this.model, this.tableModel);
```
viddo: Why not have `this.stackLayoutModel.nextStep(this.model, this.tableModel);`? The spec is currently passing because it's not testing this, and passing an arg that's not used (`nextStackItem`).

xavijam: Totally agree, fixing it. No more checks that `stackLayoutModel` exists, thanks.

xavijam: Fixed the thing, thanks again!

xavijam:
Waiting for tests, but I feel this is a good one, @javierarce @viddo?

viddo:
:shipit:

xavijam:
There was conflicts with master, waiting for frontend tests and I'll merge it.

PR #6574, opened 2016-02-11
Adds prerequisite checks to allow/disallow changing widget types

### LINKED ISSUE #6563: Add prerequisite checks to allow/disallow changing widget types
Time-series widget requires at least one date column to work. If prerequisites are not met the type should not be available/displayed in the carousel selector.

TODOs
- [x] time-series type: only available for table if there is at least one date column
- [x] formula: only allow selecting columns of type integer

javierarce:
Fixes #6563


javierarce:
cc: @xavijam

xavijam:
Fine by me ;)

PR #6578, opened 2016-02-12
Make non-number columns not be selectable on histogram widget type

### LINKED ISSUE #6577: Non-number columns should not be selectable on histogram widget type
Right now we offer _all_ columns in the table. Non-number columns (e.g. date, string) columns won't work with a histogram, so should not be available in the column select.

javierarce:
Fixes #6577 

CR: @xavijam


xavijam:
Ahhh, this is music to my ears, :+1:.

javierarce:
:notes: :notes: :notes: :notes: :ear:

xavijam:
:joy:

Issue #6511, opened 2016-02-02
Apply new tab pane on new editor


viddo:
Apply the new tab pane component to the two use-cases there are:

![screen shot 2016-02-02 at 16 59 36](https://cloud.githubusercontent.com/assets/978461/12755215/d627ab40-c9ce-11e5-9237-03146ae785da.png)

- [x] render widgets in a tabpane (text labels)
- [x] render map tabpane in the top-tabpane (icons)

PR #6586, opened 2016-02-12
Render map tab pane 

javierarce:
maptabpane

In order to style the TabPane we now allow to pass a hash with options in the create methods of the factory (https://github.com/CartoDB/cartodb/pull/6586/files#diff-24dacefe9307afde9ade195f67cf5f19R20)

Closes: #6511


javierarce:
CR: @viddo
/cc: @xavijam

lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane-factory.js
```diff
@@ -49,7 +51,7 @@ module.exports = {
    *   selected: false
    * }
    */
-  createWithIcons: function (paneItems) {
+  createWithIcons: function (options, paneItems) {
```
viddo: IMHO this `createWithIcons` method should not be located on the component factory because it's too tied to this specific use-case ([I already mentioned that in the last PR](https://github.com/CartoDB/cartodb/pull/6473#discussion_r50998142)). The setup would be better located in `cartodb3/editor`.

viddo:
Overall OK, I think we should iterate a little bit on how the tabpane is setup/created, because we'll need more control over menu vs. content contexts (e.g. on the create-widgets modal).

lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane-factory.js
javierarce: I'm still unsure about the specificity of the method (I think it makes sense to have it here), but as you said in your previous review, let's leave it here until we have to deal with the styles and things get a bit more complex, ok?

viddo: ok

xavijam: IMHO, `createWithTextLabels` and `createWithIcons` satisfy needs for tab-pane use cases. If one is there, the other should be there too. If not, in my opinion, factory doesn't make any sense, because it would be a factory with one method, the most used one.

javierarce:
I've added the tab pane for layers, elements & widgets, which is contained in a view called "map".

map_pane

I've also reorganized the structure of directories, and now we have this:

screen shot 2016-02-15 at 12 05 39

So in the future the map folder will contain a `layers` and `elements` folder too.

javierarce:
Ready to be reviewed :dog:

lib/assets/javascripts/cartodb3/editor/map/widgets/widgets-form/widgets-form-content-view.js
```diff
@@ -53,7 +53,7 @@ module.exports = cdb.core.View.extend({
       }
     }];
 
-    this.tabPaneView = TabPaneViewFactory.createWithTextLabels(this.tabPaneItems);
+    this.tabPaneView = TabPaneViewFactory.createWithTextLabels({}, this.tabPaneItems);
```
viddo: This is clearly a sign that the args should be inverted, options is not compulsory so the default case could be just `createWithTextLabels(this.tabPaneItems);`

lib/assets/test/spec/cartodb3/editor/map/widgets/widget-view.spec.js
```diff
@@ -1,8 +1,8 @@
-var EditorWidgetView = require('../../../../javascripts/cartodb3/editor/editor-widget-view');
-var WidgetDefinitionModel = require('../../../../javascripts/cartodb3/data/widget-definition-model');
+var EditorWidgetView = require('../../../../../../javascripts/cartodb3/editor/map/widgets/widget-view');
+var WidgetDefinitionModel = require('../../../../../../javascripts/cartodb3/data/widget-definition-model');
 var cdb = require('cartodb-deep-insights.js');
 
-describe('editor/editor-widget-view', function () {
+describe('editor/map/widgets/widget-view', function () {
```
viddo: I don't see how this directory change makes sense, widgets are not a concept that is located under the _map_.

javierarce: Looking at the following image I thought it made sense to organise all three elements inside the map:

screen shot 2016-02-15 at 12 37 23

lib/assets/javascripts/cartodb3/editor/map/widgets/widgets-form/widgets-form-content-view.js
javierarce: But that goes against the general pattern (options first, data second) we are using in other places. For example: 
https://github.com/CartoDB/cartodb/blob/1f955ad835540020cb0f56087d04a23bc4796a9a/lib/assets/javascripts/cartodb3/editor/widgets-form/widgets-form-factory.js#L49

lib/assets/test/spec/cartodb3/editor/map/widgets/widget-view.spec.js
viddo: Technically it's a _vis_ if if the name indicates otherwise. The hierarchy is a design, not necessarily how things are exactly mapped on the tech side (e.g. in the same way that widgets are technically stored under layers right now).

lib/assets/javascripts/cartodb3/editor/map/widgets/widgets-form/widgets-form-content-view.js
viddo: That's not a general pattern but one imposed by Backbone library. The general pattern is to leave _optional_ args at the end, since one should be able to leave it out if none is provided, e.g. `someMethod(compulsoryArg1, …, compulsoryArgN, options)`

viddo: Actually, even the Backbone.Model constructor is following that structure, `new Model(data, options)`

lib/assets/test/spec/cartodb3/editor/map/widgets/widget-view.spec.js
javierarce: :ok_hand: I'll move it outside the folder

lib/assets/javascripts/cartodb3/editor/map/widgets/widgets-form/widgets-form-content-view.js
javierarce: :ok_hand: ok, I'll change it then

javierarce:
Ok, I've made the changes that you suggested, @viddo.

PR #6591, opened 2016-02-15
Routing metrics and billing

ethervoid:
**THIS PR [HAS MIGRATION](https://github.com/CartoDB/cartodb/blob/routing_billing/db/migrate/20160212170700_add_routing_quota_and_block_price_fields.rb) SO THE MIGRATION SHOULD BE DEPLOYED BEFORE THIS PR**

This PR has all the need logic to configure and charge the users for the new routing services:

- Metrics for the new routing services
- Overquota logic for routing service
- Refactor to make it easy to add new services
- New field for the organization users to set the soft routing limit
- Tests


ethervoid:
@rafatower please check it out :)

ethervoid:
@juanignaciosl could you please check this out? thank you!

juanignaciosl:
_Added "do not merge yet" label because of the migration comment, beginning CR..._

app/controllers/admin/organization_users_controller.rb
```diff
@@ -131,13 +132,17 @@ def regenerate_api_key
...
-    extra_geocodings_enabled? || extra_tweets_enabled?
+    extra_geocodings_enabled? || extra_routings_enabled? || extra_tweets_enabled?
   end
 
   def extra_geocodings_enabled?
     !Cartodb.get_config(:geocoder, 'app_id').blank?
   end
 
+  def extra_routings_enabled?
+    true
```
juanignaciosl: Shouldn't this use `soft_routing_limit` column?

ethervoid: Take a look to the geocoding part. Is not using the `soft_geocoding_limit` is something strange

app/helpers/data_services_metrics_helper.rb
```diff
@@ -0,0 +1,58 @@
...
+  private
+
+  def get_geocoding_data(user, from, to)
+    orgname = user.organization.nil? ? nil : user.organization.name
+    usage_metrics = CartoDB::GeocoderUsageMetrics.new(user.username, orgname)
+    geocoder_key = user.google_maps_geocoder_enabled? ? :geocoder_google : :geocoder_here
+    cache_hits = 0
+    countable_requests = 0
+    from.upto(to).each do |date|
+      success = usage_metrics.get(geocoder_key, :success_responses, date)
```
juanignaciosl: Instead of `get` + addition with `nil` check you could do an straightforward addition by converting `get` return value to integer with `to_i`, which is nil-safe (`nil.to_i == 0`). Minor, though.

app/helpers/data_services_metrics_helper.rb
```diff
@@ -0,0 +1,58 @@
...
+      countable_requests += success unless success.nil?
+      empty = usage_metrics.get(geocoder_key, :empty_responses, date)
+      countable_requests += empty unless empty.nil?
+      hit = usage_metrics.get(:geocoder_cache, :success_responses, date)
+      cache_hits += hit unless hit.nil?
+    end
+    countable_requests + cache_hits
```
juanignaciosl: How come are you using two different accumulators if you add both in the end?

ethervoid: Because I want to differenciate cache from countable_request for clarity and future changes

app/helpers/data_services_metrics_helper.rb
```diff
@@ -0,0 +1,58 @@
+# encoding: utf-8
+
+require_relative '../../services/dataservices-metrics/lib/geocoder_usage_metrics'
+require_relative '../../services/dataservices-metrics/lib/routing_usage_metrics'
+
+module DataServicesMetricsHelper
```
juanignaciosl: I'm not sure about merging this into the same helper. Those methods present really low cohesion, it makes me think that keeping `GeocoderMetricsHelper` and creating the new `RoutingMetricsHelper ` might make more sense.

I'm also thinking about the opposite: generalizing, `get_*_data` methods, because they look really similar, but maybe you come up with something too complicated.

Just give this a thought if you haven't already, whatever decision you come up with will be fine.

app/models/carto/organization.rb
```diff
@@ -41,6 +41,12 @@ def get_new_system_geocoding_calls(options = {})
       get_organization_geocoding_data(self, date_from, date_to)
     end
 
+    def get_routing_calls(options = {})
+      date_to = (options[:to] ? options[:to].to_date : Date.today)
```
juanignaciosl: Maybe a `to_and_from_dates_from_options` method returning both to avoid duplication? Not 100% sure either.

app/models/carto/user.rb
```diff
@@ -239,6 +240,15 @@ def remaining_geocoding_quota(options = {})
     (remaining > 0 ? remaining : 0)
   end
 
+  def remaining_routing_quota(options = {})
+    if organization.present?
+      remaining = organization.routing_quota.to_i - organization.get_routing_calls(options)
```
juanignaciosl: This logic belongs to `Organization`.

app/helpers/data_services_metrics_helper.rb
ethervoid: The reason to not have two helpers was to avoid multiples helpers for the same functionality. This way I can include just one helper in detriment of have a repeated or similar code

app/models/carto/organization.rb
```diff
@@ -41,6 +41,12 @@ def get_new_system_geocoding_calls(options = {})
       get_organization_geocoding_data(self, date_from, date_to)
     end
 
+    def get_routing_calls(options = {})
```
juanignaciosl: Those `options` hashes everywhere are cursed. I know you've replicated previous design, but maybe you could at least document allowed keys (`:to` and `:from`, afaik).

app/models/carto/organization.rb
ethervoid: I've always disliked the methods with multiple elements returned

app/models/carto/organization.rb
ethervoid: Sure, no problem

app/models/organization.rb
```diff
@@ -40,14 +40,16 @@ class Organization < Sequel::Model
...
 
   DEFAULT_GEOCODING_QUOTA = 0
+  DEFAULT_ROUTING_QUOTA = 0
 
   def validate
     super
     validates_presence [:name, :quota_in_bytes, :seats]
     validates_unique   :name
-    validates_format   /\A[a-z0-9\-]+\z/, :name, message: 'must only contain lowercase letters, numbers & hyphens'
+    validates_format   (/\A[a-z0-9\-]+\z/), :name, message: 'must only contain lowercase letters, numbers & hyphens'
```
juanignaciosl: Are parenthesis needed or suggested by Hound? (just curiosity).

juanignaciosl: screen shot 2016-02-16 at 17 26 42

Hahaha, Hound-trolling found afterwards.

ethervoid: Hound...

ethervoid: I hate it...

app/models/user.rb
```diff
@@ -835,6 +863,15 @@ def remaining_geocoding_quota
     (remaining > 0 ? remaining : 0)
   end
 
+  def remaining_routing_quota
+    if organization.present?
+      remaining = organization.routing_quota - organization.get_routing_calls
```
juanignaciosl: Same here.

db/migrate/20160212170700_add_routing_quota_and_block_price_fields.rb
```diff
@@ -0,0 +1,18 @@
+Sequel.migration do
+  up do
+    add_column :users, :routing_quota, :integer, null: false, default: 0
+    add_column :users, :routing_block_price, :integer
+    add_column :users, :soft_routing_limit, :boolean
```
juanignaciosl: Could you consider default values and not null constraint for those columns, please?

services/dataservices-metrics/lib/service_usage_metrics.rb
```diff
@@ -5,23 +5,9 @@
...
-      :geocoder_internal,
-      :geocoder_here,
-      :geocoder_google,
-      :geocoder_cache
-    ]
-
-    def initialize(redis, username, orgname = nil)
+  class ServiceUsageMetrics
+
+    def initialize(username, orgname = nil, redis=$geocoder_metrics)
```
juanignaciosl: Since `ServiceUsageMetrics` looks generic, I wouldn't favor `$geocoder_metrics` over others.

services/dataservices-metrics/lib/service_usage_metrics.rb
```diff
@@ -49,14 +35,14 @@ def get(service, metric, date = DateTime.current)
...
     end
 
-    private
+    protected
 
     def check_valid_data(service, metric, amount = 0)
-      raise ArgumentError.new('Invalid service') unless VALID_SERVICES.include?(service)
-      raise ArgumentError.new('Invalid metric') unless VALID_METRICS.include?(metric)
-      raise ArgumentError.new('Invalid geocoder metric amount') if !amount.nil? and amount < 0
+      raise NotImplementedError.new("You must implement check_valid_data in your metrics class.")
```
juanignaciosl: Oh, so java-ish :heart: :-)

app/controllers/admin/organization_users_controller.rb
juanignaciosl: The only usage of this method is at `extras_enabled?` :-\

app/helpers/data_services_metrics_helper.rb
juanignaciosl: Oki

app/helpers/data_services_metrics_helper.rb
juanignaciosl: Well, let's keep it this way.

app/models/carto/organization.rb
juanignaciosl: Ok

juanignaciosl:
My major concerns were that `true` returned value at a method and the nullable columns, since you've already answered about the first one, :+1: after whatever you decide with the migration. Let me know if you want a second CR or not.

app/controllers/admin/organization_users_controller.rb
ethervoid: Yes, I'm puzzle too but I don't want to break something in the other side of the code jajaja

app/models/carto/user.rb
ethervoid: Right, I'll change it

db/migrate/20160212170700_add_routing_quota_and_block_price_fields.rb
ethervoid: I did but then some parts of the code breaks because have logic for it XD so I removed it

services/dataservices-metrics/lib/service_usage_metrics.rb
ethervoid: We could call it $service_metrics hahaha

app/models/carto/user.rb
ethervoid: Done

app/models/user.rb
ethervoid: Done

app/models/carto/account_type.rb
```diff
@@ -35,5 +35,15 @@ def soft_geocoding_limit?(user)
         user[:soft_geocoding_limit]
       end
     end
+
+    def soft_here_isolines_limit?(user)
+      if user[:soft_here_isolines_limit].nil?
+        # Due is an expensive service and the tiler could make multiples request
```
rafatower: Don't hate me but there's a more "elegant" way of expressing the same idea. The typical ruby idiom to return a boolean is:
```ruby
# if true the user can request isolines past their quota
# set to false by default to avoid unexpected charges
def soft_here_isolines_limit?(user)
  !!user[:soft_here_isolines_limit]
end
```

ethervoid: I have to say that I dislike that kind of cleverish type of code but I'll add it

db/migrate/20160212170700_add_routing_quota_and_block_price_fields.rb
```diff
@@ -0,0 +1,18 @@
...
+    add_column :user_creations, :soft_here_isolines_limit, :boolean, default: true
+  end
+
+  down do
+    drop_column :users, :here_isolines_quota
+    drop_column :users, :here_isolines_block_price
+    drop_column :users, :soft_here_isolines_limit
+    drop_column :organizations, :here_isolines_quota
+    drop_column :organizations, :here_isolines_block_price
+  end
```
rafatower: Is the `user_creations` column missing here?

services/dataservices-metrics/lib/geocoder_usage_metrics.rb
```diff
@@ -0,0 +1,37 @@
...
+module CartoDB
+  # The purpose of this class is to encapsulate storage of usage metrics.
+  # This shall be used for billing, quota checking and metrics.
+  class GeocoderUsageMetrics < ServiceUsageMetrics
+
+    VALID_METRICS = [
+      :total_requests,
+      :failed_responses,
+      :success_responses,
+      :empty_responses,
```
rafatower: I love trailing commas but to be compliant with our style guide...

Also probably you should `CONSTANT = [ :sym1, :sym2 ].freeze` as hound suggests

db/migrate/20160212170700_add_routing_quota_and_block_price_fields.rb
ethervoid: Yes, good catch!

spec/requests/signup_controller_spec.rb
```diff
@@ -278,6 +281,7 @@
           end
 
           it 'triggers user creation with organization' do
+            Cartodb::Central.stubs(:sync_data_with_cartodb_central?).returns(false)
```
rafatower: why is this needed now?

PR #6592, opened 2016-02-15
update big numbers in public pages footer

matallo:
CR @javierarce


javierarce:
:+1: Spread the :heart:!

Issue #6550, opened 2016-02-10
Set maxredirects option in Typhoeus downloader


iriberri:
From https://github.com/CartoDB/cartodb/pull/6530#issuecomment-181827348

cc @rafatower

PR #6595, opened 2016-02-15
Set maxredirs Typhoeus value

iriberri:
This change add a new option for Typhoeus in the importer downloader
that could avoid infinite loops on it. The MAX_REDIRECT constant has
been set to 5.

Closes https://github.com/CartoDB/cartodb/issues/6550

When tested at development and exceeded the maximum, the following
import error is being reported:

image

Find Typhoeus options [here](http://www.rubydoc.info/gems/typhoeus/0.4.2/Typhoeus/Curl).


iriberri:
cc @rafatower

rafatower:
:+1:

xavijam:
ready to go? :dancer:

Issue #6565, opened 2016-02-11
Implement "new widget" modal


viddo:
The "new widget" the scope will focus on implementing the _functional_ requirements, any visual and nice-to-have details will be pushed out of scope for this first iteration. 
Minimum functional requirements include:
- Can have a modal working (same as for current dashboard)
- Column title + selected state
- A type should only be visible if there is at least one option available (e.g. times-series will only work if there's at least one date column)

Must-haves:
- [x] Port modal code to new editor #6601 
- [x] PoC of adding a widget, using a single type (formula) #6617
- [x] Implement remaining types (histogram, category, time-series, …)

All nice-to-have features (that doesn't fit the scope of this milestone) needs to be dealt with later, see https://github.com/CartoDB/cartodb/issues/6635

viddo:
FYI @urbanphes I'll track progress here

viddo:
Postponed until the new designs are ready, so moving to backlog…
#6601 contains the stuff finished so far at least (the modal part).

viddo:
@urbanphes the basic functionality is done (once [the last PR](https://github.com/CartoDB/cartodb/pull/6632) is merged), meanwhile I encountered some issues, see the _"Needs further discussion"_ in the desc of this PR. Let me know when we can talk about these, meanwhile I'll continue with the bugs/remaining features.

PR #6601, opened 2016-02-16
6565 modals api PoC

viddo:
Waypoint for #6565 
PoC of adding widgets modal, not finished though since we decided to change the scope due to the problems encountered.

The old modal implementation is too tied to cartodb.js v3 (extending lower-level concepts…), so not possible to use or port as-is. Instead, I implemented it as a new component with a more semantic API, that should be more legible, extensible and easier to maintain.

The `.Dialog` styles assumes the Lato font to be loaded, but the new UI seems to use open-sans(?). For now not loading Lato, so the elements are falling back to "sans-serif". How should we deal with the CSS mismatch between old vs. new styles? cc @piensaenpixel 

Also added a loading component, to encapsulate the view/logic of control flow of having some async prerequisite for a particular view, we had this pattern all over the place in the old code so should come handy here too.

@javierarce review please
cc @xavijam FYI


lib/assets/test/spec/cartodb3/components/view-factory.spec.js
```diff
@@ -0,0 +1,86 @@
...
+      spyOn(this.view.$el, 'html');
+      this.view.render();
+    });
+
+    it('should not have leaks', function () {
+      expect(this.view).toHaveNoLeaks();
+    });
+
+    it("should inject the rendered results into the view's element", function () {
+      expect(this.view.$el.html).toHaveBeenCalled();
```
javierarce: Isn't this call to `toHaveBeenCalled` a bit redundant having `toHaveBeenCalledWith` in the following line?

viddo: I usually do this because the `toHaveBeenCalledWith` is much noisier if the arg is missing and doesn't indicate if the method was called at all. Compare `this method should have been called` vs. `this method should have been called with {asdasdasdasdasdasdasdasdasdasdasdasdasdasdasdasdas` (which might or might not been due to any call at all).

lib/assets/javascripts/cartodb3/components/modals/modal-view.js
```diff
@@ -0,0 +1,69 @@
...
+  _onDestroy: function () {
+    this.hide();
+    this._delayDueToAnimation(function () {
+      this.clean();
+    });
+  },
+
+  _delayDueToAnimation: function (fn) {
+    // timeout value ought to match the .Dialog.is-closing animation duration.
+    setTimeout(fn.bind(this), 120);
```
javierarce: Would it make sense to extract the number to a a global variable of the model?

lib/assets/test/spec/cartodb3/components/view-factory.spec.js
javierarce: Oh, it makes sense. Good to know!

javierarce:
:+1: Looks good to me

lib/assets/javascripts/cartodb3/components/modals/modal-view.js
viddo: I wasn't sure if it's worth doing, I thought the comment above should be explanatory enough. :/ What would you name that variable to make the intention and comment content more clear?

javierarce: `closingDelay` for example?

PR #6617, opened 2016-02-19
6565 add formula widgets options

viddo:
Waypoint for #6565, PoC of creating a formula widget.

The code related to the actual widget creation is minor and straight forward, the big changes are mainly on how the models and collections are setup, which were kind of prerequisite to finalize the parent issue. 

We were complicating things for ourselves by not following Backbone idioms, in summary:
- Models
  - should _only_ deal with their own business and only depend on what's essential
  - The internal structures are always flat now, please use the [`parse`](http://backbonejs.org/#Model-parse) + [`toJSON`](http://backbonejs.org/#Model-toJSON) methods to translate from/to the REST API format. This makes it much easier to use them.
- Collections
  - utilizes the [Collection#model prop](http://backbonejs.org/#Collection-model) to define _how_ a model of its collection should be created. This way we can utilize the Backbone methods (`create`, `add`, etc.) only providing attrs (see widget creation below).
  - Sometimes a good place to coordinate with related, external deps, e.g. widget definition changes are synced to the dashboard models through the collection instead of convoluting the model.

Other noteworthy changes:
- Introduced the `LayerTable` data model, which encapsulated a layer's data source state, i.e. basically the source table + query (which affects what columns are available).
  - The logic was ported from the olds models (cartodb-layer, CartoDBTableMetadata, and CartoDBTableData), but tried to keep the models here more focused and not exposing internal details to widget creation/editing that only really cares about the columns.
- The `baseUrl` is common for all models/collections, so set it to the configModel, which is used to create the appropriate API URL.
- :fire: a lot of unnecessary boilerplate and ceremony related to the forms.

E.g. creating widgets now just require knowing the attrs (compare to previous snippets…):
```js
widgetDefinitionsCollection.create({
  type: 'formula',
  title: 'AVG homes',
  layer_id: layerDefinitionsCollection.at(1).id,
  column: 'homes',
  operation: 'avg'
}, {
  wait: true /* do not create widget model until definition is OK'd by API*/
})
```

There are still some smells related to passing around dependencies through several layers (of backbone views, not the cartodb/map concept, mind you!), I'll comment on that in a separate PR though (probably the next one).

@xavijam pretty please review? :bow: 
@javierarce @acanimal @alonsogarciapablo FYI


package.json
```diff
@@ -70,7 +70,7 @@
   "browserify": {
     "transform": [
       "browserify-shim",
-      "jstify"
+      ["jstify", { "minifierOpts": false }]
```
viddo: Enables conditionals in `.tpl` templates, e.g. `<input <% if (isSome) { %>checked="checked"<% }%> />`

viddo:
retest this please

viddo:
bump for review
I'm continuing in a separate branch for the time being, but would be good to get  feedback ASAP, if there's something serious I need to redo.

lib/assets/javascripts/cartodb3/data/table-query-model.js
```diff
@@ -0,0 +1,69 @@
+var _ = require('underscore');
+var cdb = require('cartodb-deep-insights.js');
+var Backbone = require('backbone');
+
+// var MAX_GET_LENGTH = 1024;
```
javierarce: I guess we can remove this comment.

javierarce:
Maybe this has been talked offline, but why are there changes on the TabPaneViewFactory in this PR?

xavijam:
Hey, we agreed to keep tab-pane factory, what happened?

viddo:
The only difference is
```js
var createTextLabelsTabPane = require('../../components/tab-pane/create-text-labels-tab-pane');
…
createTextLabelsTabPane(…)
```

instead of :

```js
var tabPanFactory = require('../../components/tab-pane/tab-pane-factory');
…
tabPaneFactory.createTextLabelsTabPane(…)
```

i.e. split up the files in two separate files, since the two functions are static and have no logic in common. For the new use-case here I don't need anything from the icons tab-pane so why have to require any of that?

javierarce:
:+1: LGTM… let's go!

PR #6618, opened 2016-02-22
Upgraded dataservices api extension version

ethervoid:
Upgrade the dataservices API version to 0.2.0 that has the isolines feature


ethervoid:
retest this please

PR #6634, opened 2016-02-23
Use dataset as filename for downloads from organization users

### LINKED ISSUE #6409: When you download a dataset from a public page, the file is named for the user account not the table
This happens at least on multiuser accounts

https://team.cartodb.com/u/andrew/tables/stuartlynn.l_walking_routes_with_block_group_data_and_lodes/public/map?redirected=true

If you download that dataset as a CSV or a GeoJSON, you'll get 

stuartlynn.csv and stuartlynn.geojson

javitonino:
Fixes #6409 
We pass "user.dataset" to the SQL-API, and it only uses the name up to the first point. With this patch, we pass "dataset" directly as the filename.

CR @xavijam

### DIFF
--- a/lib/assets/javascripts/cartodb/common/dialogs/export/public_export_view.js
+++ b/lib/assets/javascripts/cartodb/common/dialogs/export/public_export_view.js
@@ -55,6 +55,11 @@ module.exports = ExportView.extend({
     var options = {};
     options.filename = this.model.get('name');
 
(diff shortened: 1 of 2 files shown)


xavijam:
Yay \o/!

xavijam:
:+1

viddo:
retest this please (CI job stalled, see https://github.com/CartoDB/cartodb/issues/6638 for details)

Issue #2917, opened 2015-03-25
Avoid leaving stale cdb_importer tables


Kartones:
Approaches:
- Cron/Job to sweep all DBs and DROP tables older than X hours (gather metrics of longest imports)
- Better capture errors, removing involved tables in the process (see [importer Runner class](https://github.com/CartoDB/cartodb/blob/master/services/importer/lib/importer/runner.rb#L190)). 
  - Must take into account scenario of single compressed file [spawning multiple files to import](https://github.com/CartoDB/cartodb/blob/master/services/importer/lib/importer/runner.rb#L306), out of quota erros and the like.
  - Must take into account non-S3 uploads to '/public/uploads' 

Second approach is better and the most appropiate. 1st approach can be done as a mere old data cleaning after implementing 2nd.

**Goal**: Leave no traces of files 'and cdb_importer.' tables in case of failed/errored imports.

rafatower:
@zenitraM what was the approach taken when you did a "manual" clean-up on the user DB's?

Exception handling in the importer deserves a review.

zenitraM:
It was quite crude, I went and dropped all `cdb_importer` tables for all users which didn't have an active connection on `pg_stat_activity` (so no importer activity, well, there could actually be in some point between ogr2ogr and ruby but it would be very uncommon)

iriberri:
Just a reminder to myself: we noticed at https://github.com/CartoDB/cartodb/issues/6276 that we are leaving files in there.

rafatower:
The log id of one of the stale tables: `328b6ff1-0491-42a3-9011-f11309f96902` (thanks to @zenitraM)

rafatower:
More debugging info: it is a sync table: `synchronization_id: abb7afc2-b3ee-11e5-b0e0-0e787de82d45`

Some interesting exception:
```
true: Renaming column wkb_geometry TO the_geom
true: Enabling autovacuum for "cdb_importer"."importer_057fb3beb61b11e5a2780ef7f98ade21_1"
true: Georeferenced
true: Errored importing data from /tmp/imports/20160108-22765-19ltbrq/building_1015.shp:
false: CartoDB::Importer2::TooManyTableRowsError: importer_057fb3beb61b11e5a2780ef7f98ade21_1
```

and the beautified backtrace:
```
services/importer/lib/importer/runner.rb:184:in `block in import'
lib/cartodb/stats/aggregator.rb:45:in `block in timing'
[...]

PR #6639, opened 2016-02-23
Let exceptions bubble up to base_job and delete temp stale tables

### LINKED ISSUE #6628: Notify errors at base_job

iriberri:
Fixes #6628 
Fixes https://github.com/CartoDB/cartodb/issues/2917

While checking the importer flow, I've seen that there are multiple exceptions which get rescued and never get notified.

This PR adds several `raise` for those exceptions that were getting hidden and notifies them with our error notifier.

When those exceptions happened, we weren't cleaning the stale tables in the `cdb_importer` schema, which was also treated at #2917, so this new iteration tries to get rid of the tables that get stuck when an exception occurs.

### DIFF
--- a/app/connectors/importer.rb
+++ b/app/connectors/importer.rb
@@ -76,7 +76,7 @@ module CartoDB
         runner.log.append("Table '#{name}' registered")
       rescue => exception
         if exception.message =~ /canceling statement due to statement timeout/i
-          drop("#{ORIGIN_SCHEMA}.#{current_name}")
+          drop("#{ORIGIN_SCHEMA}.#{result.table_name}")
           raise CartoDB::Importer2::StatementTimeoutError.new(
             exception.message,
             CartoDB::Importer2::ERRORS_MAP[CartoDB::Importer2::StatementTimeoutError]
--- a/app/connectors/importer.rb
+++ b/app/connectors/importer.rb
@@ -117,7 +117,7 @@ module CartoDB
       def drop(table_name)
         database.execute(%Q(DROP TABLE #{table_name}))
       rescue
-        runner.log.append("Couldn't drop table #{table_name}: #{exception}. Backtrace: #{exception.backtrace.to_s}. ")
+        runner.log.append("Couldn't drop table #{table_name}: #{exception}. Backtrace: #{exception.backtrace} ")
         self
       end
 
--- a/app/connectors/importer.rb
+++ b/app/connectors/importer.rb
@@ -116,7 +116,7 @@ module CartoDB
 
       def drop(table_name)
         database.execute(%Q(DROP TABLE #{table_name}))
-      rescue
+      rescue => exception
         runner.log.append("Couldn't drop table #{table_name}: #{exception}. Backtrace: #{exception.backtrace} ")
         self
       end
@@ -205,8 +205,7 @@ module CartoDB
           RENAME TO "the_geom_#{UUIDTools::UUID.timestamp_create.to_s.gsub('-', '_')}"
         })
       rescue => exception
-        runner.log.append("Silently failed rename_the_geom_index_if_exists from #{current_name} to #{new_name} with exception #{exception}. Backtrace: #{exception.backtrace} ")
-        raise exception
+        runner.log.append("Silently failed rename_the_geom_index_if_exists from #{current_name} to #{new_name} with exception #{exception}. Backtrace: #{exception.backtrace}. ")
       end
 
       def persist_metadata(result, name, data_import_id)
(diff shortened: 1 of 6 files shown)


iriberri:
In my initial commit I added a raise exception in [this resque](https://github.com/CartoDB/cartodb/blob/6af94f5c44a9cbd5dbbf9572770626cbdd37e003/app/connectors/importer.rb#L201), but according to the tests this part is not expected to raise an import error, so I'm getting rid of it.

iriberri:
I added a change in tests, [here](https://github.com/CartoDB/cartodb/blob/6628-notify_errors_base_job/spec/connectors/importer_spec.rb#L249-L250) so that they pass with the new exception raising additions.

````
    # It raises #<RuntimeError: Error: User doesn't have private tables enabled>
    expect{data_import.run_import!}.to raise_error(RuntimeError)
````

It comes from [here](https://github.com/CartoDB/cartodb/blob/9a4d3d696629720752f223a749a33ab313b36de6/app/models/table.rb#L424), where it is not treated as a specific CartoDB exception but simply `raised`.

**Comment by iriberri** [comment]
Please, @ethervoid, could you take a look? :)

File: app/connectors/importer.rb [hunk 1]
```diff
@@ -76,6 +76,7 @@ def register(result)
         runner.log.append("Table '#{name}' registered")
       rescue => exception
         if exception.message =~ /canceling statement due to statement timeout/i
+          drop("#{ORIGIN_SCHEMA}.#{current_name}")
```
ethervoid: There is no `current_name` variable in the method

app/models/data_import.rb
```diff
@@ -208,6 +208,7 @@ def run_import!
     stacktrace = exception.to_s + exception.backtrace.join
     Rollbar.report_message('Import error', 'error', error_info: stacktrace)
```
ethervoid: Change this to use `CartoDB.report_exception(e, message = nil, additional_data = {})` . We have to remove all the Rollbar direct calls

app/models/table_registrar.rb
```diff
@@ -52,6 +52,7 @@ def set_metadata_from_data_import_id(table, data_import_id)
       end
     rescue => e
       Rollbar.report_exception(e)
```
ethervoid: Same here

lib/resque/base_job.rb
```diff
@@ -32,10 +32,12 @@ def self.run_action(options, queue_name, action)
             retry
           else
             raise e
-          end  
+          end
         else
           raise e
         end
+      rescue => e
+        CartoDB.notify_exception(e)
```
ethervoid: Here is well done

ethervoid:
Take a look to the comments

iriberri:
I'll do asap. I just saw your issue about the Rollbar change in the list, so I'll change whatever mentions to it that still appear in these files.

iriberri:
I made a new commit for Hound's recommendations, only style fixes in a test.

I guess that this is good to go next week then. :-)

iriberri:
I found another one [here](https://github.com/CartoDB/cartodb/blob/ce8834ffac1d738aad08d1b12710d7eb3da9694b/app/models/data_import.rb#L598), although unrelated with the cdb_import stale tables. This is affecting: https://github.com/CartoDB/cartodb/issues/6707

app/connectors/importer.rb
```diff
@@ -198,7 +205,7 @@ def rename_the_geom_index_if_exists(current_name, new_name)
           RENAME TO "the_geom_#{UUIDTools::UUID.timestamp_create.to_s.gsub('-', '_')}"
         })
       rescue => exception
-        runner.log.append("Silently failed rename_the_geom_index_if_exists from #{current_name} to #{new_name} with exception #{exception}. Backtrace: #{exception.backtrace.to_s}. ")
+        runner.log.append("Silently failed rename_the_geom_index_if_exists from #{current_name} to #{new_name} with exception #{exception}. Backtrace: #{exception.backtrace}. ")
```
rafatower: are you sure logging directly `backtrace` instead of `backtrace.to_s` is not gonna cause trouble? I remeber some issues logging exceptions because of not being able to serialize when sending to rollbar (breaking when trying to log something)

cc/ @ethervoid

iriberri: Oh, good catch. I removed the `to_s` because of Hound complaints, but it would make sense in this case.

lib/resque/base_job.rb
rafatower: Please raise the exception again.

By quickly looking into the code, this class has two purposes: retrying when there's a connection error and reporting resque stats properly. It was done by systems. In case of doubt ask systems.

By catching the exception and not re-throwing you're modifying the behavior

luisbosque:
retest this please

PR #6657, opened 2016-02-24
Use defaults instead of extend

viddo:
To not accidentally overwrite values, e.g. set the id of a namedmap
instead of the keeping the layer id.
cc @xavijam


xavijam:
:+1:  thanks and well spotted!

viddo:
frontend tests passed, so merged

xavijam:
:dancer:

PR #6669, opened 2016-02-24
Adds widget list style

javierarce:
screen shot 2016-02-24 at 15 53 49

CR: @xavijam 
CC: @piensaenpixel


javierarce:
Not sure what should be the description of each element, but I've chosen the name of the layer.

xavijam:
Why have we chosen BlockList/Item as name?

lib/assets/javascripts/cartodb3/editor/widgets/widgets-view.js
```diff
@@ -29,6 +31,7 @@ module.exports = cdb.core.View.extend({
   _addWidgetItem: function (m) {
     var view = new EditorWidgetView({
       model: m,
+      layerName: this._layerDefinitionsCollection.get(m.get('layer_id')).getName(),
```
xavijam: _OPTIONAL: I'd pass the whole layer model and then I'd decide what to render in the item view._

javierarce:
We needed a generic name, since the same design will be used for layers. After some discussion, we chose that name.

xavijam:
Oook! no more to add :+1:

javierarce:
retest this please

PR #6670, opened 2016-02-24
6654 different add time series widget

### LINKED ISSUE #6654: Time-series type should be treated differently in the add-widgets modal
When there are is no time-series widget yet
- the none-option should pre-selected
- selecting a column and continue should _create_ a new time-series widget

When a time-series already exists:
- the column of the existing widget should be pre-selected in the modal
- selecting none and continue should _remove_ the existing widget
- selecting another column and continue should _update_ the existing widget with new values (column, start and end basically)

viddo:
Resolves #6654 

time-series-modal

- Had to refactor the internals a little bit to meet the new requirements. Now not only does the modal _create_ widgets, but might also _update_ or _destroy_ some too (only time-series at the moment though). 
- Any existing time-series widget will be pre-selected in the modal window.
- Renamed the "none" option from "no-option" to "none" for consistency, hence there appear to be more changes here than there really is. ;)

@javierarce can you review (since you reviewed the last PR)?
cc @urbanphes @xavijam


javierarce:
:+1: LGTM. The only thing that I don't like is the name of the method `createUpdateOrOtherwise`… but after reviewing the code and reading the description I'm not sure what alternative would be better.

viddo:
Yep, me neither, forgot to mention that I didn't come up with a better name… I'm all ears for a better name. :kissing_heart:

viddo:
`createUpdateOrSimilar`? :joy:

PR #6673, opened 2016-02-24
6664 fix sync values in style form

### LINKED ISSUE #6664: False values on sync_on_* attrs are not reflected in style form
It's because the forms expect string values, where as the model have booleans.

We need to a translation from/to the definition <-> form model of those attributes. It's basically the same issue as I solved for the [time-series use-case, where we needed to transform from/to unix_timestamp <-> Date object, for the start/end attrs](https://github.com/CartoDB/cartodb/pull/6632/files#diff-ea1dc6cdfdc40622c04c304cba85917b).

viddo:
The backbone forms expect another kind of format to work

Fixes #6664

cc @xavijam


xavijam:
Dale :mans_shoe:!

PR #6683, opened 2016-02-25
Raise encoding error for filenames with incorrect codepoints

### LINKED ISSUE #6276: "ArgumentError: invalid byte sequence in UTF-8" error on ZIP import

iriberri:
Fixes #6276 

In previous attempts of fixing this bug, we tried to normalize and rename the
incorrect user input to build new names that would not raise an exception.
After a discussion time ago, I think it's better to raise an understandable error
that could allow users to fix their files without us having to struggle with it and
manipulate the files.

In this new approach I am adding a new function that checks the validity of the
code points in the filenames that are being crawled after decompression.

This `unp.rb` file is the only place were we are trying to match filenames with 
other strings, such as in:
```ruby
next if hidden?(subpath)
next if subpath =~ /.*readme.*\.txt/i		   next if subpath =~ /.*readme.*\.txt/i
next if subpath =~ /\.version\.txt/i		   next if subpath =~ /\.version\.txt/i
````
where `ArgumentError: invalid byte sequence in UTF-8` is raised if the `subpath` 
string is not valid -- in terms of encoding.

This PR also fixes one of the tests, which was working before only because of the check in `UNP_READ_ERROR_REGEX`, as when an error occurs in `unp` the error_code is `!= 0`.

Now, when a bad filename is detected, we will be raising `EncodingError` (defined 
[here previously](https://github.com/CartoDB/cartodb/blob/09bbff02590d5e0c4eb6c11178cfa59c65428fb1/services/importer/lib/importer/exceptions.rb#L124) -- definition [here](https://github.com/CartoDB/cartodb/blob/2882df5a561aff60bcce3575ef403e9fc8dab12b/lib/cartodb/import_error_codes.rb#L162), which I might probably edit as you might want to open QGIS
instead of your text editor to fix the encoding of your Shapefile).

The result of uploading these files now is:

image

Until now, it was:
image

### DIFF
--- a/services/importer/spec/unit/unp_spec.rb
+++ b/services/importer/spec/unit/unp_spec.rb
@@ -235,7 +235,7 @@ describe Unp do
 
   describe '#unp_failure?'  do
     it 'returns true if unp cannot read the file' do
-      Unp.new.unp_failure?('Cannot read', 0).should eq true
+      Unp.new.unp_failure?('Cannot read', -1).should eq true
     end
 
     it 'returns true if returned an error exit code' do
--- a/services/importer/lib/importer/unp.rb
+++ b/services/importer/lib/importer/unp.rb
@@ -137,7 +137,7 @@ module CartoDB
       end
 
       def filename_valid_encoding?(filename)
-        return filename.force_encoding("UTF-8").valid_encoding?
+        filename.force_encoding("UTF-8").valid_encoding?
       end
 
       def normalize(filename)
--- a/services/importer/lib/importer/unp.rb
+++ b/services/importer/lib/importer/unp.rb
@@ -138,7 +138,7 @@ module CartoDB
       end
 
       def filename_valid_encoding?(filename)
-        return filename.valid_encoding?
+        return filename.force_encoding("UTF-8").valid_encoding?
       end
 
       def normalize(filename)
--- a/services/importer/lib/importer/unp.rb
+++ b/services/importer/lib/importer/unp.rb
@@ -39,7 +39,6 @@ module CartoDB
 
       def run(path)
         return without_unpacking(path) unless compressed?(path)
-        raise EncodingError unless filename_valid_encoding?(path)
         extract(path)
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by iriberri** [comment]
cc @rafatower :smile:

File: services/importer/spec/unit/unp_spec.rb [hunk 1]
```diff
@@ -235,7 +235,7 @@
 
   describe '#unp_failure?'  do
     it 'returns true if unp cannot read the file' do
-      Unp.new.unp_failure?('Cannot read', 0).should eq true
+      Unp.new.unp_failure?('Cannot read', -1).should eq true
```
rafatower: Is this test in the Makefile?

rafatower:
:+1:

iriberri:
Yep! The test is in the Makefile. It was passing before my change too just because it was sending the string "Cannot read", even though the error code was incorrect.

PR #6702, opened 2016-02-26
Removes svg errors from the console

javierarce:
The problem was a negative height.

CR: @viddo


lib/assets/javascripts/cartodb3/editor/add-widgets/histogram/histogram-option.tpl
```diff
@@ -18,8 +18,8 @@
           </g>
 
           <% for (var i = 0; i < 9; i++) { %>
-          <% y = Math.random() * 14 + 6; %>
-          <rect id="Bar-Copy-7" fill="#9DE0AD" sketch:type="MSShapeGroup" x="<%- 29 * i %>" y="<%- y %>" width="27" height="<%- 16 - y %>"></rect>
+          <% y = Math.round(Math.random() * 12); %>
```
viddo: I'd suggest to use
```js
var MAX_VAL = 12;
var MIN_VAL = 1;
Math.floor(Math.random() * MAX_VAL + MIN_VAL)
```
`Math.round()` works ofc, but would yield MAX_VALUE + 1 too

javierarce: Really? For a fake thing? :|

viddo: See it as a minor comment since you're fixing the issue and validating that negative values won't happen here again I guess.

From a larger perspective, fake piece or not, it's still code that goes out in production and that may throw errors if wrong (which already happened), which is why I advocate to clarify the intents.

javierarce: :+1: I see your point. I'll fix it. Thanks!

PR #6705, opened 2016-02-26
Hides navigation when editing the widget

javierarce:
widgets


xavijam:
I feel it is a good solution, the only thing I don't like so much is, after getting back from the widget form it should show the widgets list (not the first one, layers), I think this will require changes over tab-pane, but from my point of view, we will need it. What do you think?

xavijam:
Also, are we including the "title view" in both views or it is out of the stack layout? Because that view should only belong to the first "stack view", in the widget form view we should show the widget title.

javierarce:
> Also, are we including the "title view" in both views or it is out of the stack layout? Because that view should only belong to the first "stack view", in the widget form view we should show the widget title.

I'll fix that and add the right title for the widget :)

xavijam:
P*** amo.

javierarce:
widgets

Ok that's done, onto the back navigation part now.

javierarce:
Can you check this again, @xavijam? I'm opening a new ticket for the new tab-pane feature to go to the last used position.

lib/assets/javascripts/cartodb3/editor/tab-pane-template.tpl
```diff
@@ -0,0 +1,17 @@
+<div class="CDB-HeaderInfo">
```
xavijam: I would not add these things within a tab pane template, I'd create another view.

xavijam:
Aren't we adding the logic to choose the tab-pane position instead of always showing the first element?

javierarce:
> Aren't we adding the logic to choose the tab-pane position instead of always showing the first element?

Are you asking if the logic is already implemented but not used?

xavijam:
Well, I'm wondering If we already have it or if we are going to implement it.

javierarce:
Sorry, the problem is actually in the stack layout. I guess we'll need a way to store a reference to a tab pane whenever we invoke the `nextStep` method, so when we do a  `prevStep` we can go back to the stored position.

I've separated the header title from the tab pane.

xavijam:
Ok, so let's face that thing in another ticket, sounds good?
About the current code, looks great :+1: :dancer: :sun_with_face:

Issue #6707, opened 2016-02-26
"URL doesn't look as from ArcGIS server" returns 99999 import error code


iriberri:
Instead of a controlled error.

Import log as follows (no stacktrace on it):

````
2016-02-26 12:22:30 UTC: Running on server one_server with PID: 29549
2016-02-26 12:22:30 UTC: new_importer()
2016-02-26 12:22:30 UTC: Fetching datasource <CartoDB::Datasources::Url::ArcGis @url= @metadata= @ids_total=0 @ids_retrieved=0 current_block_size=0> metadata for item id an_invalid_arcgis_url
2016-02-26 12:22:30 UTC: Additional error info: Url doesn't looks as from ArcGIS server
2016-02-26 12:22:30 UTC: After dispatch
2016-02-26 12:22:30 UTC: ERROR!
===LOG END===
```

Exception is being hidden here: https://github.com/CartoDB/cartodb/blob/ce8834ffac1d738aad08d1b12710d7eb3da9694b/app/models/data_import.rb#L598

````ruby

    rescue => ex
      had_errors = true
      manual_fields = {
        error_code: 99999,
        log_info: ex.to_s
      }
    end
````

PR #6708, opened 2016-02-26
Rescue InvalidInputDataError and assign error_code 1012

iriberri:
Fixes https://github.com/CartoDB/cartodb/issues/6707

The exception `InvalidInputDataError` was previously defined in the platform but it was not being rescued when raised from the ArcGIS Server URL connector.

This code rescues the specific exception so that the user receives a proper error code and message, instead of an Unknown 99999 code.

While working in this issue, I noticed that there's a rescue in [this piece of code which is hiding exceptions](https://github.com/CartoDB/cartodb/blob/ce8834ffac1d738aad08d1b12710d7eb3da9694b/app/models/data_import.rb#L598), as we found out at https://github.com/CartoDB/cartodb/pull/6639

### DIFF
--- a/app/models/data_import.rb
+++ b/app/models/data_import.rb
@@ -589,6 +589,12 @@ class DataImport < Sequel::Model
         error_code: 1013,
         log_info: ex.to_s
       }
+    rescue InvalidInputDataError => ex
+      had_errors = true
+      manual_fields = {
+        error_code: 1012,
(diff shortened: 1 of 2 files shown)


iriberri:
@ethervoid please, could you take a look here?

ethervoid:
Looks good to me

Issue #6619, opened 2016-02-22
Error cases on modals


urbanphes:
Identify error cases on modals.

urbanphes:
We need identify and redesign this cases.
Only with screenshot we can do it.

javitonino:
You can find the list of error texts for the first modal at https://github.com/CartoDB/cartodb/blob/master/lib/cartodb/import_error_codes.rb

urbanphes:
On this kind of modal:
![](https://cloud.githubusercontent.com/assets/14546701/13217958/a6c23bf2-d966-11e5-9095-e818271fd8c6.png)

How many errors can be chained on the same window?

javitonino:
Just one error per window, with:
- Title and number
- Longer description (box 1)
- Error id (box 2)

xavijam:
There has been no activity on this issue for more several months. We are closing it. If you think this still needs to be addressed please open a new issue.

PR #6709, opened 2016-02-26
Fix 6697 When opens add-widget modal, it silently fails but window doesn't say anything 

### LINKED ISSUE #6697: When opens add-widget modal, it silently fails but window doesn't say anything
## STR

- Create or use a map with a private dataset on it.
- Go to the new editor url.
- Open the "new widget" modal.
- It loads forever.
- Check browser console, it returned a 401 error (Unauthorized). 

Looks like we have forgotten the `api_key` in those requests. But in any case we should provide an error state, shouldn't we?

viddo:
Closes #6697 

See the commits for the details, should be self-explanatory.

The style of the error view is the same as we currently have on the modals. I've seen that there's some [planned work #6619](https://github.com/CartoDB/cartodb/issues/6619) to change the styles of the error dialogs, but until that is finalized I think we should go with the current styles. cc @urbanphes 

screen shot 2016-02-26 at 14 03 12

@xavijam can you review


lib/assets/javascripts/cartodb3/components/error/error.tpl
```diff
@@ -0,0 +1,5 @@
+<div class="LayoutIcon LayoutIcon--negative">
+  <i class="CDB-IconFont CDB-IconFont-cockroach"></i>
+</div>
+<h3 class="IntermediateInfo-title"><%- title %></h3>
+<p class="DefaultParagraph"><%= desc %></p>
```
xavijam: <%= xx %> because it could contain html, i guess?

lib/assets/javascripts/cartodb3/components/modals/add-widgets/add-widgets-view.js
```diff
@@ -96,12 +105,17 @@ module.exports = cdb.core.View.extend({
       }, []);
   },
 
-  _onLayerTableFetched: function () {
+  _onLayerTableFetched: function (m) {
```
xavijam: Seems like we don't need the `m` argument.

lib/assets/javascripts/cartodb3/components/error/error.tpl
viddo: Yes, correct… probably should make it clearer(?)

xavijam: Could we use the html sanitizer here?

lib/assets/javascripts/cartodb3/editor.js
```diff
@@ -45,7 +45,8 @@ var layersCollection = dashboard.vis.map.layers;
 var configModel = new ConfigModel(
   _.defaults(
     {
-      base_url: userData.base_url
+      base_url: userData.base_url,
+      api_key: userData.api_key
```
xavijam: :+1: :kiss:

lib/assets/javascripts/cartodb3/components/modals/add-widgets/add-widgets-view.js
viddo: Indeed, removed

lib/assets/javascripts/cartodb3/components/error/error.tpl
viddo: Indeed, applied!

lib/assets/javascripts/cartodb3/data/layer-table-model.js
```diff
@@ -33,6 +33,13 @@ module.exports = cdb.core.Model.extend({
 
     this.listenTo(this._originalTableModel, 'sync', this._onOriginalTableModelSync);
     this.listenTo(this._tableQueryModel, 'sync', this._onTableQuerySync);
+
+    // Pass on certain events
+    [this._originalTableModel, this._tableQueryModel].forEach(function (m) {
```
xavijam: Don't you like underscore :D ?

xavijam:
Good to go :heart: .

lib/assets/javascripts/cartodb3/data/layer-table-model.js
viddo: We don't really [need it](http://caniuse.com/#search=foreach) nowadays, and [youmightnotneedjquery(.com)](http://youmightnotneedjquery.com/) either for that matter ;)

xavijam: :joy: come on, we have already loaded, anyway, it was a joke. Don't take the opportunity to blame in jQuery :dancer:.

Issue #960, opened 2014-10-29
Importing a CSV with an empty line fails with wrong feedback


Kartones:
Importing a CSV with an empty line (which makes it invalid) fails (which is correct) but it gives the following feedback to the user:

```
Unsupported/Unrecognized file type (1002)
Should we support this filetype? Let us know in our support email!
```

Confirmed by @rafatower that it is not a problem of OGR2OGR

iriberri:
Still happening `3512653e-3806-4871-94de-788d57f78fa9`.

jgoizueta:
Current behaviour (was changed in https://github.com/CartoDB/cartodb/pull/6152) is as follows:

If we import this file:

```csv
A,B,C
100,"xyz","uvw"
,,
200,"abc","mno"
300,"aaa","bbb"

400,"ccc","ddd"
,"",""
500,"uuu","vvv"
```

No error occurs and we get this table: (blanks are NULLs)

 cartodb_id | the_geom | the_geom_webmercator |  a  |  b  |  c  
------------|----------|----------------------|-----|-----|-----
          1 |          |                      | 100 | xyz | uvw
          2 |          |                      |     |     | 
          3 |          |                      | 200 | abc | mno
          4 |          |                      | 300 | aaa | bbb
          5 |          |                      | 400 | ccc | ddd
          6 |          |                      |     |     | 
          7 |          |                      | 500 | uuu | vvv

I.e.: a blank line such as the second row in the input is ignored, and rows will all blank fields are preserved with NULL values.

We still have to look at the `3512653e-3806-4871-94de-788d57f78fa9` occurrence mentioned by @iriberri which could have been caused by an special kind
of blank line such as those of https://github.com/CartoDB/cartodb/pull/6535
and which is probably now solved by that PR.

see also: #6178

jgoizueta:
Confirm problem still persists in cases like `3512653e-3806-4871-94de-788d57f78fa9`: csv files consisting only of empty lines (i.e. line separators).
Doesn't happen if any line is non-empty (even if it is blank, i.e. all columns are blank).

jgoizueta:
The proper error when an imported file is empty is 1005 (Empty file), but we should, as discussed in #6178, keep the 1002 (Unsupported/unrecognized file) for files containing malformed lines, which are now treated as blank lines.

jgoizueta:
[...]

Issue #6178, opened 2015-11-24
CSV imports treat bad/blank lines inconsistently


jgoizueta:
When a CSV file is imported:
- Malformed rows and completely empty lines (lines that are only line separators)
  are silently ignored (they are omitted from the generated table and we don't tell the user).
- Lines consisting of whitespace and possibly column separators are kept with all field
  having null values.

I think the most sensible behaviour would be instead to:
- Omit all-null rows as well as empty/malformed lines
- Tell the user how many non-blank rows were dropped because they were invalid
  (i.e. how many malformed lines that were not essentially empty) but don't bother
  her with blank lines dropped because that's a minor detail.

jgoizueta:
@rafatower @iriberri  @ethervoid what do you think?

rafatower:
Personally I prefer to improve our documentation about what's acceptable and what's not and do not deal with that many special cases.

E.g: before changing the UI I'd check how often that situation of mostly-ok-but-invalid-rows. I'm reluctant to changing the UI as those rows do not contain information anyway.

That said, I'm open to discussion.

iriberri:
This issue reminds me of that "Rome" line that I imported from GeoNames and included a single quote in one of the fields of the CSV. That broke the entire line and we weren't aware that at least a row failed.

ethervoid:
My opinion:

- For empty/null lines remove and don't notify the user
- Malformed (non-blank rows) use the warning in the import to notify the user some lines hasn't been imported

xavijam:
There has been no activity on this issue for more several months. We are closing it. If you think this still needs to be addressed please open a new issue.

PR #6732, opened 2016-02-29
Fix CSV import errors for empty files

jgoizueta:
This addresses #960

cc @rafatower @iriberri do you think this would be the right spec?


iriberri:
Check `all_empty_lines.csv`, it's just a single empty line at the moment.

jgoizueta:
@iriberri has noticed we have an specific 'Malformed CSV' error message (2003).
So we can use it instead of 1002 (Unsupported/Unrecognized file type)

jgoizueta:
The problem with having malformed CSVs to fail (currently malformed lines are ignored, then the import fails only for an empty result, i.e. all lines are malformed or empty) is that some users may prefer a file with an odd malformed line to import.

The alternative, discussed in #6178 would be to import them as now, ignoring malformed lines and then show the user a message about 'some malformed lines were ignored'.

It would be simpler for us to just return the malformed CSV error.

iriberri:
> The alternative, discussed in #6178 would be to import them as now, ignoring malformed lines and then show the user a message about 'some malformed lines were ignored'.

I guess that the problem here is that at the end you would need to show those lines to the user (or the line number of the first incidence, at least), or the warning could become annoying for the user. Not sure what your thoughts are, but just to contribute a bit. :-)

On the other hand I also think that if something is malformed we should just raise an error for it. Perhaps a combination of "import error" + "malformed entry at line 92" would be nice, but afaik we don't have any "customized" error input at the moment so this would be a new thing.

jgoizueta:
I also think we should raise an error, with or without the additional information about affected lines.

But I've also remembered why it's not easy to fix this: we may or may not normalize not normalize the csv before sending it to ogr2ogr based on the detected encoding and separator. If the file goes through the CSV normalizer it's easy to detect the 
(non-blank) malformed lines, but otherwise...

rafatower:
I'm confused: I just see changes in test files.

For the sake of the discussion: this is trying to deal with malformed CSV files, defined as "CSV file with empty lines". To avoid confusion with some malformed line geometry. Am I right?

I think it's better not to mix end of line delimiters with empty lines. One of the delimiters is not supported by `ogr2ogr` so something to be documented as part of the CSV requirements cc/ @iriberri:

http://www.gdal.org/drv_csv.html
> At least two fields per line must be present. Lines may be terminated by a DOS (CR/LF) or Unix (LF) style line terminators

jgoizueta:
@rafatower: yes, I've committed only tests so far, expecting to define the spec before implementing it.

And we're trying to discern the malformed CSV cases (no relation to line geometries) from the empty file case which are now treated identically. And invalid end of lines are  a case of malformed CSV (as treated but the CSV Ruby lib, at least).

Cartofante:

rafatower:
closing it, to be reopened some day?

PR #6741, opened 2016-03-01
Use column name as values for select list

### LINKED ISSUE #6738: Unable to edit widgets if map is created from a query
I have a map whose points come from a sql query, no from a dataset. 
When showing this map on new editor, I can add new widgets to the map but I'm unable to edit them afterwards. If I remove the sql query and use the associated dataset to paint the points, widgets work properly.

cc'ed @xavijam

viddo:
Resolves #6738

When using a query result the columns don’t have ids, which caused the form to set the form model with an empty column (thus don’t trigger changes afterwards).

cc @xavijam


viddo:
retest this please

luisbosque:
retest this please

PR #6749, opened 2016-03-02
No longer install schema-triggers extension from user-mover

zenitraM:
Fixes specs in 9.5


azamorano:
Checked and it works :+1:

PR #6766, opened 2016-03-03
Organization users list 'Add user' button arrow icon & users list email fixed

MariaCheca:
screen shot 2016-03-03 at 12 55 15

CR @xavijam

### DIFF
--- a/app/controllers/carto/api/user_presenter.rb
+++ b/app/controllers/carto/api/user_presenter.rb
@@ -51,6 +51,7 @@ module Carto
         end
 
         if @options[:current_user] && @options[:current_user].organization_owner?
+          poro[:email] = @user.email
           poro[:table_count] = @user.table_count
           poro[:all_visualization_count] = @user.all_visualization_count
         end
--- a/app/controllers/carto/api/user_presenter.rb
+++ b/app/controllers/carto/api/user_presenter.rb
@@ -50,7 +50,7 @@ module Carto
           poro.merge!(groups: @user.groups ? @user.groups.map { |g| Carto::Api::GroupPresenter.new(g).to_poro } : [])
         end
 
-        if @options[:current_user] && @options[:current_user].organization_owner?
+        if @options[:current_user] && @options[:current_user].organization_owner? && @user.belongs_to_organization?(@options[:current_user].organization)
           poro[:email] = @user.email
           poro[:table_count] = @user.table_count
           poro[:all_visualization_count] = @user.all_visualization_count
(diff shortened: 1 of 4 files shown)


xavijam:
@gfiorav could you take a look to the Ruby changes?

xavijam:
Ping!

app/controllers/carto/api/user_presenter.rb
```diff
@@ -51,6 +51,7 @@ def to_public_poro
         end
 
         if @options[:current_user] && @options[:current_user].organization_owner?
+          poro[:email] = @user.email
```
gfiorav: hey @xavijam we removed emails from public view remember? If this is available to the public it can't have the email.

xavijam: That was the reason I want you reviewing this :). So the thing is, we'd like to have emails in the organization users view, there is no other workaround for that?  (Not touching user presenter of course).

gfiorav: I'd also check if @user actually belongs to current_user's org and if so i'd allow for the email to be present. It's no big deal that owners see their employees email

xavijam: @MariaCheca, how do you like? ^^^

xavijam:
Sounds good this solution?

app/models/carto/user.rb
```diff
@@ -323,6 +323,10 @@ def organization_user?
     self.organization.present?
   end
 
+  def belongs_to_organization?(organization)
```
gfiorav: We need this method also in the `models/user.rb` just in case

app/models/carto/user.rb
```diff
@@ -323,6 +323,9 @@ def organization_user?
...
+    self.organization_user? && organization != nil && self.organization_id == organization.id
```
gfiorav: I agree with the Hound :grin: 

Also, if you can remove the trailing spaces in the other chunks of code it would be nice (there is a sublime package for detecting trailing specs :wink: )

app/models/carto/user.rb
MariaCheca: That method already exists there :) https://github.com/CartoDB/cartodb/blob/master/app/models/user.rb#L1376

gfiorav: Ah! Didn't check, sorry!

xavijam:
Frontend part -> :+1:

gfiorav:
Backend part -> Once you take care of trailing spaces and my other comment, ship it! :package:

xavijam:
:shipit:

MariaCheca:
\o/
Will merge on Monday :D

PR #6775, opened 2016-03-03
polygon gamma suggestions for simple, choropleth, category, and density default settings

makella:
@andrewxhill please review

The changes proposed in this pull request are:
- to add `polygon-gamma: 0.2` to simple, choropleth, category, and density polygons
- for each type of polygon map, the `line-width` has been set to `0`
- in addition, I've adjusted the default size of the hexbins from 15 to 5

Explanation and Examples:

By default, `polygon-gamma` (polygon-antialiasing) is set to `1`. If someone is designing a map and removes the outline, there is a noticeable gap in between each polygon. This is especially true with choropleth and category maps that vary in color over a variety of shapes and sizes. Here are two examples of this effect:

screen shot 2016-03-03 at 3 08 57 pm
screen shot 2016-03-03 at 3 09 06 pm

One solution is to modify the `line-color` of each polygon fill to get a smooth appearance but this can be tedious especially when there are a large number of classes/categories.

A default solution is to adjust the `polygon-gamma` setting to `0.2`. In doing so, the antialiasing is minimized and each polygon appears to have an outline that is a shade of its fill color and makes for an overall smoother looking map:
screen shot 2016-03-03 at 3 12 54 pm
screen shot 2016-03-03 at 3 12 38 pm
screen shot 2016-03-03 at 3 14 37 pm
screen shot 2016-03-03 at 3 14 48 pm

Density Now:
screen shot 2016-03-03 at 3 19 48 pm

Proposed (with `polygon-gamma` and default size of `5`):
screen shot 2016-03-03 at 3 20 42 pm

### DIFF
--- a/lib/assets/javascripts/cartodb/table/menu_modules/carto_wizard_forms.js
+++ b/lib/assets/javascripts/cartodb/table/menu_modules/carto_wizard_forms.js
@@ -15,7 +15,7 @@ cdb.admin.forms.get = function(what) {
 var polygon_stroke = {
    name: 'Polygon Stroke',
    form: {
-      'line-width': { type: 'width', value: 0.5 },
+      'line-width': { type: 'width', value: 0 },
       'line-color': { type: 'color' , value: '#FFF' },
       'line-opacity': { type: 'opacity', value: 1 }
    }
@@ -43,7 +43,8 @@ var polygon_fill = {
    name: 'Polygon Fill',
    form: {
       'polygon-fill': { type: 'color' , value: '#FF6600', extra: { image_property: 'polygon-pattern-file', image_kind: 'pattern' }},
-      'polygon-opacity': { type: 'opacity_polygon' , value: 0.7 }
+      'polygon-opacity': { type: 'opacity_polygon' , value: 0.7 },
+      'polygon-gamma': { type: 'gamma' , value: 0.2 }
   }
 };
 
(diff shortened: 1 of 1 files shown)


andrewxhill:
lgtm!
we'll need someone to test this. 
also, i'm not sure the behavior of 'type' and what will happen with the new type, 'gamma'
let's see!

@saleiva can someone spare a few minutes to try this out? i know it isn't the future :) buuuut

Cartofante:

makella:
After further review, for the new editor, let's not use `plygon-gamma` instead, we'll have a modified proposal for each map type's cartocss. 

cc: @andrewxhill

xavijam:
Do you feel we could close this PR due to new builder changes?

makella:
yes, that sounds OK. @xavijam the only thing i would want to be sure of is that we get default styling for polygons (simple, choropleth, etc.) with some specific specs that i have to produce the best looking map while taking the `polygon-gamma` into account.

xavijam:
There is no default styling for polygons or lines, the get the same simple style as points, for the moment.

makella:
totally

i'm working on all that this week!

reference:
https://github.com/CartoDB/camshaft/issues/86
https://github.com/CartoDB/cartodb/issues/8163

xavijam:
Closing this one then :kiss:

xavijam:
Not deleting the branch by the way, just in case.

PR #6838, opened 2016-03-10
Improves localization namespacing

### LINKED ISSUE #6831: Fix translation namespaces
e.g. group top-level tab-panel labels together (layers/elements,widgets)

javierarce:
Closes #6831 

This PR improves the namespacing of our localization file. Instead of using a pure semantic structure, we'll now use the path of the elements to define the localization strings too.

As @viddo proposes:
Instead of having 

```json
"editor": {
    "layers": "Layers",
    "elements": "Elements",
    "widgets": "Widgets"
  },
…
```

we'll now have:

```json
"editor": {
  "tab-pane-menu": {
    "layers-label": "Layers",
    "elements-label": "Elements",
    "widgets-label": "Widgets"
  }
…```

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/editor-view.js
+++ b/lib/assets/javascripts/cartodb3/editor/editor-view.js
@@ -39,7 +39,7 @@ module.exports = cdb.core.View.extend({
     } ]);
 
     var tabPaneTabs = [{
-      label: _t('editor.layers.title'),
+      label: _t('editor.tab-pane.layers.title-label'),
       selected: true,
       createContentView: function () {
         return new LayersView({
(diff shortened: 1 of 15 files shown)


javierarce:
CR: @viddo 
CC: @xavijam

viddo:
:+1: 
But I'd also change the structure slightly, to make it clear _where_ the different keys are used. Keep in mind that a translator reading this file will have no idea of context so that needs to be codified in this translation file. 

E.g. the tab pane labels should be clear that they are for that context, so I'd change that [these](https://github.com/CartoDB/cartodb/blob/6831-loc-namespace/lib/assets/locale/en.json#L28-L36) to something like:
```json
// en.json
"editor": {
  "tab-pane-menu": {
    "layers-label": "Layers",
    "elements-label": "Elements",
    "widgets-label": "Widgets"
  },
  …
}

javierarce:
You're totally right. I fixed the thing you mentioned.

PR #6850, opened 2016-03-11
Do not recreate viz table in collection_spec

javitonino:
`collection_spec` creates copies of the visualization tables to perform its tests. However, some tests (namely, testing the visualizations locator) use the main table, but still recreate it every time.

This PR avoids this situation by never recreating the main table and ensuring that the visualization repository is left pointing to the main table (instead of one of the temp tables). This fixes Maps API controller tests in master CI.

It also updates the Migrator schema to the current schema so tests are run against the real table.

See #6848

### DIFF
--- a/spec/models/visualization/collection_spec.rb
+++ b/spec/models/visualization/collection_spec.rb
@@ -10,15 +10,7 @@ include CartoDB
 
 describe Visualization::Collection do
   before(:each) do
-    db_config   = Rails.configuration.database_configuration[Rails.env]
-    # Why not passing db_config directly to Sequel.postgres here ?
-    # See https://github.com/CartoDB/cartodb/issues/421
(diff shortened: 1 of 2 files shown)


javitonino:
:+1:

javitonino:
CR @gfiorav

spec/models/visualization/collection_spec.rb
```diff
@@ -47,6 +40,10 @@
     @user_2.destroy
   end
 
+  def restore_vis_backend_to_normal_table_so_relator_works
```
gfiorav: I know you inherited this, but maybe it might be a good time to rename this method!

gfiorav:
Except for that single comment about the method name, :package: it!

Nice work :)

Issue #6765, opened 2016-03-03
Test fail sometime with duplicate usernames


javitonino:
Example: https://ci.cartodb.net/job/CartoDB-master-testing/56/console

```
07:54:14 
07:54:14 Failures:
07:54:14 
07:54:14   1) CartoDB::Visualization::Collection#fetch counts total liked
07:54:14      Failure/Error: @user_1 = FactoryGirl.create(:valid_user, quota_in_bytes: 524288000, table_quota: 500, private_tables_enabled: true)
07:54:14      Sequel::DatabaseError:
07:54:14        PG::Error: ERROR:  duplicate key value violates unique constraint "users_username_key"
07:54:14        DETAIL:  Key (username)=(user8515) already exists.
07:54:14      # ./spec/support/factory_girl.rb:4:in `save!'
07:54:14      # ./spec/models/visualization/collection_spec.rb:37:in `block (2 levels) in <top (required)>'
07:54:14 
07:54:14 Failures:
07:54:14 
07:54:14   1) CartoDB::Visualization::Collection#fetch counts total liked
07:54:14      Failure/Error: @user_1 = FactoryGirl.create(:valid_user, quota_in_bytes: 524288000, table_quota: 500, private_tables_enabled: true)
07:54:14      Sequel::DatabaseError:
07:54:14        PG::Error: ERROR:  duplicate key value violates unique constraint "users_username_key"
07:54:14        DETAIL:  Key (username)=(user8515) already exists.
07:54:14      # ./spec/support/factory_girl.rb:4:in `save!'
07:54:14      # ./spec/models/visualization/collection_spec.rb:37:in `block (2 levels) in <top (required)>'
07:54:14 
07:54:14 Finished in 21 minutes 9.79 seconds
07:54:14 741 examples, 1 failure, 28 pending
07:54:14 
07:54:14 Failed examples:
07:54:14 
07:54:14 rspec ./spec/models/visualization/collection_spec.rb:402 # CartoDB::Visualization::Collection#fetch counts total liked
07:54:15 
07:54:15 Failures:
07:54:15 
07:54:15   1) CartoDB::Visualization::Collection#fetch counts total liked
07:54:15      Failure/Error: @user_1 = FactoryGirl.create(:valid_user, quota_in_bytes: 524288000, table_quota: 500, private_tables_enabled: true)
07:54:15      Sequel::DatabaseError:
07:54:15        PG::Error: ERROR:  duplicate key value violates unique constraint "users_username_key"
[...]

PR #6851, opened 2016-03-11
Increase randomness for :valid_user factory

### LINKED ISSUE #6849: Increase randonmness in :valid_user factory
The `:valid_user` factory generates random 5-digit usernames and emails for the users. However the factory is used over 60 times in the test suite and the users are not always destroyed. This creates a relatively high collision chance and makes the test fail from time to time.

javitonino:
Closes #6849 #6765


javitonino:
CR @gfiorav

gfiorav:
:package: it!

Issue #6854, opened 2016-03-11
GDrive: KeyError: key not found: "downloadUrl"


iriberri:
````
Request URL:https://xxxxx.cartodb.com/api/v1/imports/service/gdrive/list_files
Request Method:GET
Status Code:400 Bad Request
````
Response:

````json
{"errors":{"imports":"key not found: \"downloadUrl\""}}
```

Line of code: https://github.com/CartoDB/cartodb/blob/b376ec6fc5f346539be8a0c83811f4f34fefc86e/services/datasources/lib/datasources/url/gdrive.rb#L336

iriberri:
We found out the issue. 

When you have have in Google Drive a file that was shared by other person, and this person enabled the option to "not allow viewers to download the file" this file won't return any download link or related field. So, basically, when this happens we can re-use the current code for debugging that I added so ignore these non-downloadable files. I'll edit the debug message for it to say that the file is not downloadable so that in the future we don't lose tracks of non-listed datasets.

PR #6855, opened 2016-03-11
Add debug information for "key not found: downloadurl" @gdrive

iriberri:
Related: https://github.com/CartoDB/cartodb/issues/6854

We need some extra debug information for the issue reported in #6854. We know that gdrive file listing is failing because the item that we are parsing does not have a `downloadUrl` neither an `exportLink` (for gdrive files).

We're using a pretty old Ruby library for the Google Drive client which perhaps has something to do with the exception that we were facing, but we would need to know which are the details of the elements that make the list crash.

This PR basically adds a notfiy_debug for "malformed" items and ignores them when listing to workaround the issue.

### DIFF
--- a/services/datasources/lib/datasources/url/gdrive.rb
+++ b/services/datasources/lib/datasources/url/gdrive.rb
@@ -182,7 +182,7 @@ module CartoDB
           end
 
           @client.execute(batch_request)
-          all_results
+          all_results.compact
         rescue Google::APIClient::InvalidIDTokenError => ex
           raise TokenExpiredOrInvalidError.new("Invalid token: #{ex.message}", DATASOURCE_NAME)
         rescue Google::APIClient::BatchError, Google::APIClient::TransmissionError, Google::APIClient::ClientError, \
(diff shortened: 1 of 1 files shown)


iriberri:
After some issues when testing in staging produced by using `compact!`, which was getting rid of the whole list, the new code seems to be working as expected (it doesn't break listing).

iriberri:
retest this please

iriberri:
retest this please

iriberri:
cc @jgoizueta

services/datasources/lib/datasources/url/gdrive.rb
```diff
@@ -332,11 +332,14 @@ def format_item_data(item_data)
...
             data[:filename] = clean_filename(item_data.fetch('title')) + '.csv'
             data[:size] = NO_CONTENT_SIZE_PROVIDED
-          else
+          elsif item_data.include?('downloadUrl')
             data[:url] = item_data.fetch('downloadUrl')
             # For Drive files, title == filename + extension
             data[:filename] = item_data.fetch('title')
             data[:size] = item_data.fetch('fileSize').to_i
+          else
+            CartoDB.notify_debug('downloadURl key not found @gdrive', item: item_data.to_s, user: @user)
```
jgoizueta: I think `inspect` would be more appropriate here than `to_s` since it may be more informative for debugging ('nil' vs string special characters, etc.)

jgoizueta:
:+1:

PR #6878, opened 2016-03-15
Integrate camshaft reference for analyses

viddo:
Instead of having individual models for each we can just have one, and use the reference to parse and export toJSON based on a type’s definition.

Based some of my implementation on what @alonsogarciapablo already had [implemented in cartodb.js](https://github.com/CartoDB/cartodb.js/pull/1184/files#diff-fe19d2ba7b4811d600fb51d127b51a9aR1) (thx!)

Note that not all analyses types are ready yet, so we'll need to update the dependency (and thus the shrinkwrap) when new types are added, see [latest reference version (0.1.0 at time of writing](https://github.com/CartoDB/camshaft/tree/master/reference/versions) for current types available.

edit: also fixed some things encountered while reviewing this:
- [x] render analysis nodes linear, instead of nested c5ae0cb
- [x] added a "composite" view for nodes with multiple sources
- [x] fix bug with non-source nodes provoking creation of new analysis on node creation

New snippets for creation:

```js
m1 = analysisDefinitionsCollection.createNode({
  type: 'trade-area',
  kind: 'walk',
  time: 300,
  source_id: l.get('source')
});
l.save({ source: m1.id })

// or source from node "owned" by the other layer
m1 = analysisDefinitionsCollection.createNode({
  id: 'c1',
  type: 'trade-area',
  kind: 'walk',
  time: 300,
  source_id: l1.get('source')
});
l.save({ source: m1.id })
```

@javierarce can you review?
cc @xavijam @rochoa


xavijam:
:clap:

lib/assets/javascripts/cartodb3/data/camshaft-reference.js
```diff
@@ -0,0 +1,45 @@
...
+      throw new Error('source names for analysis of type ' + analysisType + " couldn't be found");
+    }
+
+    return sourceNames;
+  },
+
+  getParamNamesForAnalysisType: function (analysisType) {
+    var paramNames = ANALYSIS_TYPE_TO_PARAM_NAMES_MAP[analysisType];
+    if (!paramNames) {
+      throw new Error('source names for analysis of type ' + analysisType + " couldn't be found");
```
javierarce: I think here it should be ```*param names* for analysis of type``` instead of ```source names for analysis of type```.

lib/assets/javascripts/cartodb3/data/analysis-definition-node-model.js
```diff
@@ -0,0 +1,84 @@
+var _ = require('underscore');
+var cdb = require('cartodb.js');
+var camshaftReference = require('./camshaft-reference');
+
+/**
+ * Base model for an analysis definition node.
+ * May point to
```
javierarce: May point to… ? :)

rochoa:
Quick/off-topic question: will you find useful to know what is the output geometry from an analysis?

xavijam:
Absolutely! geometry and schema would be necessary.

rochoa:
What do you mean by schema?

xavijam:
Columns and type of those columns.

rochoa:
Sorry if I mislead you, I was talking about camshaft-reference.

xavijam:
Ahhhhh, ok, then geometry is ok.

viddo:
Yes, geometry will probably be minimum, to be able to decide which analysis can be added on existing head (TBD in next task, #6886)

PR #6883, opened 2016-03-15
Ghost Tables Manager

### LINKED ISSUE #5580: Ghost tables can be triggered twice
There's been a race condition on ghost tables and some tables have been registered twice.

cc @CartoDB/backend @javisantana

gfiorav:
Fixes #5580 #6133 #5581 

This PR adds a new `GhostTablesManager` that contains all the logic for ghost tables management. ~~Along with a general refactor, it bring per user concurrency using a [redlock algorithm](http://redis.io/topics/distlock) (gem license will be added to licenses file).~~ (Deployed separately on old GT code through https://github.com/CartoDB/cartodb/pull/7035 [bolt] and https://github.com/CartoDB/cartodb/pull/7037 [bolt applied on old gt code])

It also provides a method to detect wether there are stale tables (tables linked by the editor whose sql tables have been deleted) that can cause errors when the user clicks on them. Right now, there is a POC where ghost tables won't be executed in the background if stale tables are present (to avoid errors when user clicks on these tables) although it up to @javisantana if we leave this in.

Test for the manager and in extension ghost tables are now in CartoDB.

Please CR, @juanignaciosl


juanignaciosl:
Ping me when tests are ok, please.

gfiorav:
Tests are passing @juanignaciosl !

app/controllers/admin/visualizations_controller.rb
```diff
@@ -439,14 +439,19 @@ def is_logo_hidden(vis, parameters)
   private
 
   def link_ghost_tables
-    return true unless current_user.present?
+    return unless current_user.has_feature_flag?('ghost_tables')
```
juanignaciosl: Keep the nil check, I think this filter is used by unauthenticated users (or check that it's _actually_ not needed, of course).

app/controllers/admin/visualizations_controller.rb
```diff
@@ -439,14 +439,19 @@ def is_logo_hidden(vis, parameters)
...
-    if current_user.search_for_modified_table_names && current_user.has_feature_flag?('ghost_tables')
-      # this should be removed from there once we have the table triggers enabled in cartodb-postgres extension
-      # test if there is a job already for this
-      if !current_user.link_ghost_tables_working
-        ::Resque.enqueue(::Resque::UserJobs::SyncTables::LinkGhostTables, current_user.id)
-      end
+    ghost_tables_manager = Carto::GhostTablesManager.new(current_user.id)
+
+    # Check if any tables need to be linked, relinked or unlinked
+    return if ghost_tables_manager.consistent?
```
juanignaciosl: I think that this "if consistent, if stale - else" can be refactored to "if stale, elsif !consistent". It'd be clearer and I think it's equivalent.

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,192 @@
...
+  class GhostTablesManager
+    MUTEX_REDIS_KEY = 'ghost_tables_working'.freeze
+    MUTEX_TTL = 2000
+
+    def initialize(user_id)
+      @user = ::User.where(id: user_id).first
+    end
+
+    def link
+      bolt = Redlock::Client.new(["redis://#{Cartodb.config[:redis]['host']}:#{Cartodb.config[:redis]['port']}"])
```
juanignaciosl: Since all connections to Redis are placed at `config/initializers/redis.rb` there's probably some value on linking this to that file. I'm not sure about the precise way to do that though. _Maybe_ (alternatives)...
- `$redlock_metadata = Redlock::Client.new(["redis://#{Cartodb.config[:redis]['host']}:#{Cartodb.config[:redis]['port']}"])`?
- `$redlock_metadata_conf = ["redis://#{Cartodb.config[:redis]['host']}:#{Cartodb.config[:redis]['port']}"]`?

Open to discussion. Final code should be CR'd by devops, since this involves Redis.

Also, I'm missing Redis database number, can't you configure one? Take a look at `default_databases` at the initializer.

juanignaciosl: Also: Redlock gem implements distributed locking, which is an overkill in our case, since we'll only have one Redis. Have you consider just using `SETNX`?

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,192 @@
+# encoding utf-8
+
+module Carto
+  class GhostTablesManager
+    MUTEX_REDIS_KEY = 'ghost_tables_working'.freeze
+    MUTEX_TTL = 2000
```
juanignaciosl: Add the time unit (`_MS`, I guess) to the name, please.

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,192 @@
...
+
+    def link
+      bolt = Redlock::Client.new(["redis://#{Cartodb.config[:redis]['host']}:#{Cartodb.config[:redis]['port']}"])
+
+      bolt.lock(mutex_redis_key, MUTEX_TTL) do |locked|
+        next unless locked
+
+        # Lock aquired, inside the critical zone
+
+        # NOTE: Order DOES matter, FIRST renamed, THEN new and deleted LAST
```
juanignaciosl: Explain why order does matter, please.

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,192 @@
...
+
+    def initialize(user_id)
+      @user = ::User.where(id: user_id).first
+    end
+
+    def link
+      bolt = Redlock::Client.new(["redis://#{Cartodb.config[:redis]['host']}:#{Cartodb.config[:redis]['port']}"])
+
+      bolt.lock(mutex_redis_key, MUTEX_TTL) do |locked|
+        next unless locked
```
juanignaciosl: Wouldn't this provoke a CPU-consuming block because of looping without sleeping?

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,192 @@
...
+
+        # Lock aquired, inside the critical zone
+
+        # NOTE: Order DOES matter, FIRST renamed, THEN new and deleted LAST
+        unless all_tables.blank?
```
juanignaciosl: What's the purpose of this check on `all_tables`? Since `all_tables` is not iterated or passed as an argument it's not clear, maybe this is not needed.

app/helpers/carto/ghost_tables_manager.rb
gfiorav: About the later: If you only provide one database in the initialiser it's not an overkill :)

app/helpers/carto/ghost_tables_manager.rb
gfiorav: No, it just exits the lock, no retries.

juanignaciosl: In fact, maybe we don't want to serialize two ghost table processes. Maybe it's enough discarding every one but the first.

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,192 @@
...
+      # Left the critical zone, bolt automatically unlocked
+    end
+
+    # determine linked tables vs cartodbfied tables consistency; i.e.: needs to run link
+    def consistent?
+      non_linked_tables.empty? && stale_tables.empty?
+    end
+
+    # checks if sql-api deleted/renamed tables that are still linked
+    def has_stale_linked_tables?
```
juanignaciosl: `has_` probably is left over, since this is a manager, not a `User`.

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,192 @@
...
+      (null_table_id_user_tables - all_tables).each do |user_table|
+        t = Table.new(table_id: user_table[:id])
+
+        t.keep_user_database_table = true
+        t.destroy
+      end
+    end
+
+    # Grabs the Table associated with a UserTable identified by its table_id. Also reports duplicate UserTables.
+    def fetch_table_for_user_table(table_id)
```
juanignaciosl: This `table_id` is actually a PostgreSQL `oid`, isn't it? It _looks_ like a normal, surrogate key. I think keeping oid in the name might make this clearer.

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,192 @@
...
+
+    def mutex_redis_key
+      "rails:users:#{@user.username}:#{MUTEX_REDIS_KEY}"
+    end
+
+    def relink_renamed_tables
+      non_linked_tables.each do |t|
+        table = fetch_table_for_user_table(t[:id])
+
+        next if table.nil? # UserTable hasn't been created yet; take care of it in link_new_tables
```
juanignaciosl: I can't find this check in the old code, how come it's now needed? Does this fix a previous bug?

app/helpers/carto/ghost_tables_manager.rb
gfiorav: Right now it works in a much simpler way: if your late to the party you don't come in. No queuing what so ever.

app/helpers/carto/ghost_tables_manager.rb
juanignaciosl: Oh, I see. I think this is because `non_linked_tables` has both renamed and new tables, am I right? Maybe iterating over `renamed_tables` and `created_tables` is clearer.

PS: _reading between lines_, I find next really annoying, I'd try to get rid of it.

app/helpers/carto/ghost_tables_manager.rb
gfiorav: You're right, this could be swapped for `non_linked_tables.empty?`

I've deliberately duplicated the `all_tables` code from the `::User` code thinking in the future refactor. We'll just have to change the code in there to check pg_class in another non User dependent way.

app/helpers/carto/ghost_tables_manager.rb
gfiorav: Yes, indeed :+1:

app/helpers/carto/ghost_tables_manager.rb
juanignaciosl: PS2: the use of `next` at the top of a loop, as a precondition, is somehow accepted by many people. I don't like it but it might be a matter of taste ;-)

app/helpers/carto/ghost_tables_manager.rb
gfiorav: I've tried hard to keep the naming consistent. I think that table_id makes it clearer since thats what you want for fetch: a user table with that `table_id`.

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,192 @@
...
+        rescue Sequel::DatabaseError => e
+          raise unless e.message =~ /must be owner of relation/
+        end
+      end
+    end
+
+    def link_new_tables
+      non_linked_tables.each do |t|
+        begin
+          CartoDB.notify_debug('ghost tables', action: 'registering table', new_table: t[:name])
```
juanignaciosl: You've improved renaming action to `link renamed`, changing this to `link new` might also make sense.

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,192 @@
...
+
+          table.user_id = @user.id
+          table.name = t[:name]
+          table.table_id = t[:id]
+          table.register_table_only = true
+          table.keep_user_database_table = true
+
+          table.save
+        rescue => e
+          puts e
```
juanignaciosl: This is old but this is a good moment to improve this by logging the error.

app/controllers/admin/visualizations_controller.rb
gfiorav: `before_filter :login_required, only: [:index]` does guarantee that `current_user` is not nil right?

app/helpers/carto/ghost_tables_manager.rb
gfiorav: changing!

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,192 @@
...
+    end
+
+    def unlink_deleted_tables
+      syncs = CartoDB::Synchronization::Collection.new.fetch(user_id: @user.id).map(&:name).compact
+
+      # Remove tables with oids that don't exist on the db
+      stale_tables.each do |user_table|
+        # Sync tables replace contents without touching metadata DB, so if method triggers meanwhile sync it will fail
+        # TODO: Flag running syncs to distinguish between deleted table syncs and running syncs
+        next if syncs.include?(user_table[:name])
```
juanignaciosl: _Refactor suggestion_.

This code looks like...

```
let's iterate through all tables
    if X, next
    if Y, next
    do
```

In this scenarios (this probably applies to previous loop as well), using `select` is much clearer:
```
relevant tables = given all the tables, discard those which X, discard those which Y
do for each relevant table
```

app/controllers/admin/visualizations_controller.rb
gfiorav: If it's consistent it can't be stale though. Could you elaborate on this please?

app/helpers/carto/ghost_tables_manager.rb
gfiorav: hehe, it's just a possibility that could indicate a bug!

I'm split on using next, but it's the most elegant way I found of doing this. If you have a better suggestion i'm all in for it :)

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,192 @@
...
+
+        table = fetch_table_for_user_table(user_table[:id])
+
+        next if table.nil? # TODO: Report this if buggy.
+
+        CartoDB.notify_debug('ghost tables', action: 'dropping table', dropped_table: user_table[:name])
+
+        table.keep_user_database_table = true
```
juanignaciosl: I know this is old. IMHO being set makes no sense, it should be a `destroy` argument. But ignore this comment if you don't want to refactor more code in this PR.

app/helpers/carto/ghost_tables_manager.rb
gfiorav: oops you're right! missed that, changing!

app/helpers/carto/ghost_tables_manager.rb
gfiorav: Will do!

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,192 @@
...
+      @user.in_database(as: :superuser)[sql].all.map { |table| table[:table_name] }
+    end
+
+    # Tables viewable in the editor; in users user_tables
+    def linked_tables
+      @user.tables.select(:name, :table_id).map { |table| { id: table.table_id, name: table.name } }
+    end
+
+    # May not be viewed in the editor; carotdbyfied_only
+    def all_tables
```
juanignaciosl: Since this code is in the line between the physical tables and table metadata `all_tables` is confusing, and I think that it excludes not cartodbfied tables. Maybe `cartodbfied_real_tables`? Also, document what it returns, please.

And this returns a hash

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,192 @@
...
+      real_tables.select { |table| search_for_cartodbfied_tables.include?(table[:relname]) }
```
juanignaciosl: You're looping with `search_for_cartodbfied_tables` inside the `select`, so it will be run many times. I think you can extract it.

app/helpers/carto/ghost_tables_manager.rb
juanignaciosl: PS: old code just returned the oid, Why is name needed now?

app/helpers/carto/ghost_tables_manager.rb
gfiorav: Seems big enough right now :/ 

I'll remember it for the refactors

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,192 @@
...
+module Carto
+  class GhostTablesManager
+    MUTEX_REDIS_KEY = 'ghost_tables_working'.freeze
+    MUTEX_TTL = 2000
+
+    def initialize(user_id)
+      @user = ::User.where(id: user_id).first
+    end
+
+    def link
```
juanignaciosl: I think every method but `initialize` and `link` can be `private`, isn't it?

app/helpers/carto/ghost_tables_manager.rb
gfiorav: Table metadata has nothing to do with this whole process :sweat_smile: that's why I changed the naming, it was confusing. 

I though of all tables as all tables that are of any business to the editor. If it's not cartodbfied, it's not understood by the editor. 

I can change the naming if you want, though

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,192 @@
...
+        t.destroy
+      end
+    end
+
+    # Grabs the Table associated with a UserTable identified by its table_id. Also reports duplicate UserTables.
+    def fetch_table_for_user_table(table_id)
+      user_tables = ::UserTable.where(table_id: table_id, user_id: @user.id)
+
+      if user_tables.count > 1
+        CartoDB.notify_error("#{@user.username} has duplicate UserTables", duplicated_table_id: table_id)
```
juanignaciosl: Not 100% sure about this: I think this will create an ungrouped Rollbar error. Maybe you can use the brand new Rollbar wrappers to log duplicated tables as well.

app/helpers/carto/ghost_tables_manager.rb
gfiorav: Oh, so used to RAILS model caching :gun: 

Will extract, sorry

lib/resque/user_jobs.rb
```diff
@@ -59,21 +59,17 @@ def self.perform(user_creation_id, common_data_url=nil, organization_owner_promo
...
     module SyncTables
-
       module LinkGhostTables
         extend ::Resque::Metrics
         @queue = :users
 
         def self.perform(user_id)
-          u = ::User.where(id: user_id).first
-          u.link_ghost_tables
+          Carto::GhostTablesManager.new(user_id).link
```
juanignaciosl: `link` is not accurate. What about `sync_user_schema_and_tables_metadata`? Not important though.

app/helpers/carto/ghost_tables_manager.rb
gfiorav: ah, wait, it will only run 1 time for all_tables call no?

app/helpers/carto/ghost_tables_manager.rb
gfiorav: I could definitely punch out al the syncs at the beginning so I don't iterate over them but the `if Y, next` depends on `fetch_table_for_user_table` so not so straight forward in that case. Still could do the former though.

spec/helpers/carto/ghost_tables_manager_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+module Carto
+  describe 'GhostTablesManager' do
+    before(:all) do
+      @user = FactoryGirl.create(:carto_user)
+
+      @ghost_tables_manager = Carto::GhostTablesManager.new(@user.id)
+    end
+
+    after(:all) do
+      @user.destroy
```
juanignaciosl: Don't use destroy on a `Carto::User`, you should do `::User[@user.id].destroy` so cleanup is performed. Maybe we could override `Carto::User#destroy` so it raises an error.

gfiorav: Totally snapped there, changing!

spec/helpers/carto/ghost_tables_manager_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    end
+
+    it 'should be consistent when no new/renamed/dropped tables' do
+      @ghost_tables_manager.consistent?.should be true
+    end
+
+    it 'should be inconsistent when renamed/dropped tables' do
+      @ghost_tables_manager.stubs(:stale_tables).returns [id: 3, name: 'ManoloEscobar']
```
juanignaciosl: I wouldn't stub in this test, because you're 100% bounding the test to the implementation. In spite of being slower and painful to code, I'd do actual integration tests with actual tables.

gfiorav: ok, no problem, i'll write integration tests, CR no more of the test part :stuck_out_tongue_winking_eye:

juanignaciosl:
@gfiorav I've done some comments. Don't take this the wrong way, this is a great and brave effort :clap: Take into the account as many comments as you want, and ask me about anything that is not clear. Ping me again when you do it and I'll CR this again. I have two major concerns: the lock (comments about the gem, Redis connection and the next loop) and the tests, which lack actual integration.

gfiorav:
Thanks @juanignaciosl !

I'll take these comments into account and ping you when it's refactored :grin:

app/helpers/carto/ghost_tables_manager.rb
juanignaciosl: I meant that using the gem is an overkill, since the gem implements much more than we need, and it forces us to skip our current Redis connection definition. I'd give a try to a simpler, `SETNX`-based implementation, but up to you.

lib/resque/user_jobs.rb
gfiorav: I'd insist in avoiding the table metadata stuff... nothing to do with `cdb_tablemetadata`. What about `sync`?

app/helpers/carto/ghost_tables_manager.rb
juanignaciosl: Oh, sure. I thought that `.lock` looped until he managed to get the lock. :+1: to this then.

app/helpers/carto/ghost_tables_manager.rb
juanignaciosl: I'll ask in small steps to get to your point: where does `table_id` comes from?

app/controllers/admin/visualizations_controller.rb
juanignaciosl: Oh, you're right because of `before_filter :link_ghost_tables, only: [:index]`. This is the dark side of the filters.

gfiorav: :joy:

app/helpers/carto/ghost_tables_manager.rb
gfiorav: `stale_tables_linked?` is used from the outside to determine synchronous or background execution

app/controllers/admin/visualizations_controller.rb
juanignaciosl: This is what I propose:

```
if ghost_tables_manager.has_stale_linked_tables?
  # You might see phantom tables that generate erros upon clicking them; needs relink now
  ghost_tables_manager.link
elsif !ghost_tables_manager.consistent?
  # Worst case you'll refresh for an SQL API genearted table to appear
  ::Resque.enqueue(::Resque::UserJobs::SyncTables::LinkGhostTables, current_user.id)
end
```

This approach has the following advantages:
- Less complexity because it has one less conditional and less code.
- Has a clear link between triggering the job and being unconsistent.

app/helpers/carto/ghost_tables_manager.rb
gfiorav: Yeah, i'll change this report for sure

app/helpers/carto/ghost_tables_manager.rb
juanignaciosl: No, I didn't mean to replace it, I proposed removing the conditional and just call to
```
relink_renamed_tables
link_new_tables
```

I propose this because code doesn't expose the actual relationship between what `all_tables.blank?` (or `non_linked_tables.empty?`, I don't mind) means.

In other words, this would be clearer (this is not my proposal but my explanation about the code not being obvious enough):

```
unless all_tables.blank?
  relink_renamed_tables(all_tables)
  link_new_tables(all_tables)
end
```

But since those methods have no arguments, it looks like they work if whatever they need is empty.

app/helpers/carto/ghost_tables_manager.rb
juanignaciosl: Yup. I'd replace the three first lines with the following:

```ruby
non_linked_tables.map { |t| fetch_table_for_user_table(t[:id]) }.compact.each do |table|
  ...
end
```

And, inside the block, use only `table` instead of using `t` and `table`. If `t` is strictly needed `map` could return both, though, and use a `select` instead of `compact`.

app/helpers/carto/ghost_tables_manager.rb
juanignaciosl: :ok_hand: Since we're reaching a good pace maybe you could open a small refactor issue for this.

app/controllers/admin/visualizations_controller.rb
gfiorav: Ah! I see! Changing!

app/helpers/carto/ghost_tables_manager.rb
gfiorav: I'll experiment with it!

app/helpers/carto/ghost_tables_manager.rb
juanignaciosl: With "table metadata" I meant "the content of `user_tables`", not `cdb_tablemetadata` ;-) My problem with "table" is that it can mean both "physical table" and "dataset" (which is indeed a final user term, not used inside). "all_tables" don't mean "all tables that are of any business to the editor" at all to me :sweat_smile: But maybe my naming is not better. Have you got a proposal that satisfies both? :-)

app/helpers/carto/ghost_tables_manager.rb
juanignaciosl: No, `select` loops, so it will be run one time per `real_tables`. Since `table[:relname]` is not needed for `search_for_cartodbfied_tables` you can do `cartodbfied_tables = search_for_cartodbfied_tables` before.

app/helpers/carto/ghost_tables_manager.rb
juanignaciosl: Proposal (sorry, I often use natural language when Ruby would work better):

```ruby
stale_tables.
  select { |ut| !syncs.include?(ut[:name]) }.
  map { |ut| [ut, fetch_table_for_user_table(ut[:id]) }
  select { |table_and_ut| !table_and_ut[1].nil? }
```

lib/resque/user_jobs.rb
juanignaciosl: For outside readers: we've had a conversation about this naming offline. Using "metadata" is confusing because it means "metadata database" (`visualizations`, `user_tables`...) for me but it means `cdb_tablemetadata`) for @gfiorav . I think he'll try to find a declarative name which is unambiguous for both of us :_)

juanignaciosl: PS: "sync" is an _state_, not a "thing", it's not that clear to me :-\

app/helpers/carto/ghost_tables_manager.rb
juanignaciosl: Ok, three public methods, everything else under `private` then? :-)

gfiorav: It's already under protected, better private?

juanignaciosl: I don't think you need to expose this to classes extending this, better private. Minor anyway, I didn't realize that you need some more public methods.

gfiorav:
@juanignaciosl I've changed the code in two ways:

+ I've refactored to your wonderful comments :heart: 
+ I've dropped the `redlock-rb` gem (license removed from license excel) and went for our own implementation: the `Carto::Bolt` 

Name sucks but it's better than doing `Lock.lock` plus it's a bit more original IMHO.

Please re-CR this :)

app/helpers/carto/bolt.rb
```diff
@@ -0,0 +1,49 @@
+# encoding utf-8
+
+module Carto
```
juanignaciosl: `app/helpers` is meant for Rails _view_ helpers, not general helpers, so this doesn't belong here. I suggest `lib/carto`.

app/helpers/carto/bolt.rb
```diff
@@ -0,0 +1,49 @@
...
+  class Bolt
```
juanignaciosl: How come `Bolt`? :-)

app/helpers/carto/bolt.rb
```diff
@@ -0,0 +1,49 @@
...
+    DEFAULT_REDIS_OBJECT = $users_metadata
+    DEFAULT_TTL_MS = 10000
```
juanignaciosl: This is the timeout for the blocked area, isn't it? 10 seconds can be a little low, since this might involve IO operations, but I'm really guessing, it might actually be the other way around. Have you got any kind of metric? Maybe you could add one.

app/helpers/carto/bolt.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    def initialize(bolt_key, redis_object: DEFAULT_REDIS_OBJECT, ttl_ms: DEFAULT_TTL_MS)
+      @bolt_key = add_namespace_to_key(bolt_key)
+      @redis_object = redis_object
+      @ttl_ms = ttl_ms
+    end
+
+    def lock
+      is_locked = @redis_object.set(@bolt_key, true, px: @ttl_ms, nx: true)
```
juanignaciosl: Big things first: why not just provide something like `run_locked`, receiving a block that will run locked for that key? `Bolt` interface and code would be much simpler and still complete. As you provide a lock/unlock mechanism you open the door to not releasing the resource.

app/helpers/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,210 @@
...
+
+    def initialize(user_id)
+      @user = ::User.where(id: user_id).first
+    end
+
+    def sync_user_schema_and_tables_metadata
+      bolt = Carto::Bolt.new("#{@user.id}:#{MUTEX_REDIS_KEY}", ttl_ms: MUTEX_TTL_MS)
+
+      bolt.lock do |locked|
+        next unless locked
```
juanignaciosl: Related to the `Bolt` design comment: if you wouldn't have added `next unless locked` the dangerous code would've been run, isn't it? Consider simplifying this to avoid errors by design, maybe the block shouldn't be run at all if the lock is not achieved. The method might return `true` / `false` to tell the client whether the lock was retrieved (and the block run) if needed.

juanignaciosl:
I'll wait for Bolt design discussion conclusions to resume CR-ing.

app/helpers/carto/bolt.rb
gfiorav: It allows for both:

+ If you pass a block it will run the block and unlock.
+ If you don't it will just lock if it can, returning true/false depending on success.

app/helpers/carto/ghost_tables_manager.rb
gfiorav: If you mean that this will retry and hence it's dangerous it won't. It's basically it to desist if locked. The cool thing about this is that you can choose to do a different thing if locked if true instead of just passing like now (maybe queue?).

app/helpers/carto/bolt.rb
juanignaciosl: That's actually my question. If there's no need for the second, don't do it. It's error prone.

app/helpers/carto/bolt.rb
gfiorav: Will change this!

app/helpers/carto/bolt.rb
gfiorav: It's hard to pick a time for this. Systems suggested a lot more time. Let's discuss this and post conclusions.

app/helpers/carto/ghost_tables_manager.rb
juanignaciosl: No, I'm not worried about retrying this time :-) I mean that checking _inside_ the block whether locking has been achieved or not feels a little strange to me, but it might because of the use of `next`. I was thinking about this design:

```ruby
locked = bolt.lock do
  <mutual exclusion area>
end
unless locked
   <fallback>
end
```

This way the "happy path" (getting the lock) is 100% clean, and you only need to get the `locked` status if you actually need it. I was also thinking about passing one block for each situation, but I dislike the result.

What do you think?

app/helpers/carto/bolt.rb
gfiorav: You can either set the lock, do what ever and unlock it or set the lock passing whatever you want it to do and let it be unlocked automatically.

Maybe an error you could make is think that it's not unlocked automatically and unlock it again, but it's no big deal. The other thing that could happen is that you set it and for some reason never unset it but it has a TTL for that. I think it's nice to have the duality :(

app/helpers/carto/bolt.rb
juanignaciosl: I would also suggest more time indeed, but I'm blind guessing. Maybe doing a reasonable guess and adding metrics that allow us to check if they were right is the way to go.

app/helpers/carto/ghost_tables_manager.rb
gfiorav: I can totally 100% see your point. Absolutely reasonable. The thing is that I can't see this as "cleaner". still prefer:

```ruby
bolt.lock do |locked|
  if locked
    <do your thing>
  else
    <save your thing for later>
  end
```

app/helpers/carto/bolt.rb
juanignaciosl: I _know_ that you _can_ do it with that API. My question is: does current code need it? Isn't the block version good enough?

[TIMTOWTDI](http://c2.com/cgi/wiki?ThereIsMoreThanOneWayToDoIt) is not only a cool ancronym but also a smell. You have two versions. If you only need one, the other should not exist. And block one is superior because it handles resource management for free.

I think that not needing both and one being superior is an stronger argument than a nice duality, but I'm open to further discussion =D

app/helpers/carto/ghost_tables_manager.rb
juanignaciosl: I don't love mine, indeed. Let me give an example of my main concern:

```ruby
bolt.lock do
  <do your thing>
end
```

It's perfectly valid, and a naive programmer who hasn't read `Bolt` documentation might think that it's code is lock-protected. Depending on documentation is bad. On the contrary, the proposal without `locked` parameter for the block makes failing impossible. Maybe you'll end up hating me, but Japanese have another cool name for "designing something in a way that makes failure impossible", [poka-yoke](https://en.wikipedia.org/wiki/Poka-yoke) :sweat_smile:

gfiorav: Looking at a USB-C connector...

:stuck_out_tongue: ok i'll change the damn thing!

juanignaciosl: :tada:

gfiorav:
@juanignaciosl have a look! I've refactored the `.lock` method to only accept blocks and moved things to `lib/`.

gfiorav:
Tests are passing @juanignaciosl !

app/controllers/admin/visualizations_controller.rb
```diff
@@ -439,14 +439,17 @@ def is_logo_hidden(vis, parameters)
...
 
-    if current_user.search_for_modified_table_names && current_user.has_feature_flag?('ghost_tables')
-      # this should be removed from there once we have the table triggers enabled in cartodb-postgres extension
-      # test if there is a job already for this
-      if !current_user.link_ghost_tables_working
-        ::Resque.enqueue(::Resque::UserJobs::SyncTables::LinkGhostTables, current_user.id)
-      end
+    ghost_tables_manager = Carto::GhostTablesManager.new(current_user.id)
+
+    # Check if any tables need to be linked, relinked or unlinked
```
juanignaciosl: First: you've finally kept the sync/async choice in this PR, I thought you said the contrary a few minutes ago. Not a problem anyway.

I understand this because you've explained it to me in person, but maybe you can add a more explicit message about the intention of this, something like `# We'll trigger ghost tables synchronously if there's risk of displaying a stale table, or asynchronously otherwise`.

But.

While I read that, I thought about this scenario:
1.- User opens the dashboard. There're stale tables, so they begin being fixed synchronously.
2.- Let's assume that it takes > 3 s and user is anxious to enjoy his beloved CartoDB account and he refreshes the browser. There're still stale tables (process is taking long because of disk IO) and, thanks to `Bolt` it doesn't even try to fix it because the lock is acquired by the request spawned at 1. So user dashboard appears with stale tables.

Of course locking the second request is much worse, I think this is the best possible effort.

I think I'm ok with this since in spite of not being perfect, it still mitigates the problem. Or maybe I'm just wrong, I think that sharing this thought with you might be useful.

app/controllers/admin/visualizations_controller.rb
```diff
@@ -439,14 +439,17 @@ def is_logo_hidden(vis, parameters)
...
+    if ghost_tables_manager.stale_tables_linked?
+      # You might see phantom tables that generate erros upon clicking them; needs relink now
+      ghost_tables_manager.sync_user_schema_and_tables_metadata
+    elsif !ghost_tables_manager.consistent?
+      # Worst case you'll refresh for an SQL API genearted table to appear
+      ::Resque.enqueue(::Resque::UserJobs::SyncTables::LinkGhostTables, current_user.id)
```
juanignaciosl: Uhm. Proposed refactor, in two steps:

1.- Since enqueuing `LinkGhostTables` job is 100% related to "ghost tables managing" maybe
```ruby
::Resque.enqueue(::Resque::UserJobs::SyncTables::LinkGhostTables, current_user.id)
```

can be refactor with "extract method" to

```ruby
ghost_tables_manager.sync_user_schema_and_tables_metadata(async: true)
```

2.- Suddenly the relation between the 5 code lines (I'll omit the comments for the sake of shortness) becomes evident, since there're 4 calls to `ghost_tables_manager` methods:

```ruby
if ghost_tables_manager.stale_tables_linked?
  ghost_tables_manager.sync_user_schema_and_tables_metadata
elsif !ghost_tables_manager.consistent?
  ghost_tables_manager.sync_user_schema_and_tables_metadata(async: true)
end
```

Which makes a final "extract method" refactor obvious:

```ruby
Carto::GhostTablesManager.new(current_user.id).link_ghost_tables
```

Probably `link_ghost_tables` can be the only public method of `GhostTablesManager `.

What do you think?

lib/carto/bolt.rb
```diff
@@ -0,0 +1,47 @@
...
+    def initialize(bolt_key, redis_object: DEFAULT_REDIS_OBJECT, ttl_ms: DEFAULT_TTL_MS)
+      @bolt_key = add_namespace_to_key(bolt_key)
+      @redis_object = redis_object
+      @ttl_ms = ttl_ms
+    end
+
+    def lock
+      raise 'no block given' unless block_given?
+
+      is_locked = @redis_object.set(@bolt_key, true, px: @ttl_ms, nx: true)
```
juanignaciosl: I guess you have checked that this works with the same methodology that you used to check that the gem didn't work. Did you see why the gem didn't work as expected?

lib/carto/bolt.rb
```diff
@@ -0,0 +1,47 @@
...
+      ensure
+        unlock if is_locked
+      end
+    end
+
+    def unlock
+      removed_keys = @redis_object.del(@bolt_key)
+
+      if removed_keys > 1
+        CartoDB.notify_error('Removed bolt key was duplicated', bolt_key: @bolt_key, amount: removed_keys)
```
juanignaciosl: What are the scenarios where this can happen?

lib/carto/bolt.rb
```diff
@@ -0,0 +1,47 @@
+# encoding utf-8
+
+module Carto
+  class Bolt
```
juanignaciosl: The best thing of this class is that it's 100% agnostic about the use, not related to ghost tables at all, so we can use it in several places if we want. This kind of refactors are key for having smaller classes, good work :muscle:

lib/carto/bolt.rb
```diff
@@ -0,0 +1,47 @@
...
+
+      if removed_keys > 1
+        CartoDB.notify_error('Removed bolt key was duplicated', bolt_key: @bolt_key, amount: removed_keys)
+      end
+
+      removed_keys > 0 ? true : false
+    end
+
+    def info
+      { ttl_ms: @ttl_ms, bolt_key: @bolt_key }
```
juanignaciosl: I think you only use this method for a test, is it really necessary?

juanignaciosl: Suggestion: if you want to test that those arguments are configurable you can pass a [spy](http://www.martinfowler.com/bliki/TestDouble.html) `redis_object` to the constructor and check the values that `set` is called with.

lib/carto/bolt.rb
gfiorav: That's what I wanted, thanks :__)

lib/carto/bolt.rb
gfiorav: A previous lock was somehow corrupted and the key was never removed... due to expire times, maybe it would require redis itself to fail. Very unlikely, still nice to know.

lib/carto/bolt.rb
gfiorav: Yeah, it's more future oriented... thought it was nice to have.

lib/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,213 @@
...
+    def initialize(user_id)
+      @user = ::User.where(id: user_id).first
+    end
+
+    def sync_user_schema_and_tables_metadata
+      bolt = Carto::Bolt.new("#{@user.id}:#{MUTEX_REDIS_KEY}", ttl_ms: MUTEX_TTL_MS)
+
+      got_locked = bolt.lock do
+        # Lock aquired, inside the critical zone
+        unless non_linked_tables.empty?
```
juanignaciosl: First a known comment: is it necessary calling `unless non_linked_tables.empty?` before? I'd prefer calling the three in a row. Not important, though.

Second. This is the original code:
```
    link_renamed_tables unless no_tables
    link_deleted_tables
    link_created_tables(search_for_cartodbfied_tables) unless no_tables
```

How come you've change the original order (renamed-deleted-created to (renamed-created-deleted)? I'm not sure whether this is better or not, but this is the kind of subtle change that can have unwanted side effects.

lib/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,213 @@
...
+          relink_renamed_tables
+          link_new_tables
+        end
+
+        unlink_deleted_tables
+      end
+      # Left the critical zone, bolt automatically unlocked
+
+      unless got_locked
+        # TODO: Handle requests that failed to run due to bolt being locked
```
juanignaciosl: Is this an actual `TODO`? I don't think there's anything we can do, isn't it? If there's nothing that we can do you can even delete `got_locked` ;-)

lib/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,213 @@
...
+
+    def initialize(user_id)
+      @user = ::User.where(id: user_id).first
+    end
+
+    def sync_user_schema_and_tables_metadata
+      bolt = Carto::Bolt.new("#{@user.id}:#{MUTEX_REDIS_KEY}", ttl_ms: MUTEX_TTL_MS)
+
+      got_locked = bolt.lock do
+        # Lock aquired, inside the critical zone
```
juanignaciosl: That's the classical example of redundant comment ;-). Suggestion: `# We use Bolt to prevent two ghost table link processes to run concurrently, since it can generate duplicated metadata`.

lib/carto/bolt.rb
gfiorav: Yes! This is the methodology:

1. Create a table, cartodbfy it, view it in the dataset view (GT ran)
2. Set a breakpoint inside the critical zone
3. Create a table and cartodbfy it
3. Refresh datasets, background job is stopped at breakpoint, lock taken.
4. Drop first table with the SQL API and refresh the dataset view
5. GT are launch but not in the background, lock is taken by halted background process, GT doesn't run

Point proven.

lib/carto/bolt.rb
juanignaciosl: Ok. Then add a comment with your guesses, please. You've been really into this and they'll be valuable for future readers :-)

gfiorav: Will do!

lib/carto/bolt.rb
juanignaciosl: Nah, testing the usage of the parameters would be supported by any TDD-er, but an spy is a much better test.

lib/carto/ghost_tables_manager.rb
gfiorav: Yes, right now order doesn't matter anymore since I've added some checks that make order irrelevant. 

About the `non_linked_tables' be aware that a SQL-API deleted table IS linked! We have to unlink it :)

lib/carto/bolt.rb
juanignaciosl: 7. Dashboard still displays the first table, isn't it?

Great :-)

I guess that you didn't dig into the gem to check the problem then.

lib/carto/ghost_tables_manager.rb
gfiorav: Well, it would be nice to enqueue failed attempts just in case...

lib/carto/ghost_tables_manager.rb
gfiorav: Will change!

lib/carto/ghost_tables_manager.rb
juanignaciosl: If you really think that we should enqueue failing attempts, write down it in the comment.

lib/carto/ghost_tables_manager.rb
juanignaciosl: About the order: why was it relevant before and now it's not? _I'm not doubting about your words, I want to understand the details about this._

About the `non_linked_tables`: then, since "we don't have anything to relink or link" that check is not necessary, isn't it? Example with arrays: you wouldn't do this:

```ruby
unless array.empty?
  array.map { |i| whatever(i) }
end
```

you'd just

```ruby
array.map { |i| whatever(i) }
```

Obviously this is not important, but I want to be sure about what's going on here.

lib/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,213 @@
...
+        # TODO: Handle requests that failed to run due to bolt being locked
+      end
+    end
+
+    # determine linked tables vs cartodbfied tables consistency; i.e.: needs to run sync
+    def consistent?
+      non_linked_tables.empty? && dropped_tables.empty?
+    end
+
+    # checks if sql-api deleted/renamed tables that are still linked
```
juanignaciosl: `# checks if there're sql-api deleted/renamed tables still linked`

lib/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,213 @@
...
+    end
+
+    private
+
+    def mutex_redis_key
+      "rails:users:#{@user.username}:#{MUTEX_REDIS_KEY}"
+    end
+
+    def relink_renamed_tables
+      # exlude nil table because if UserTable hasn't been created yet we take care of it in link_new_tables
```
juanignaciosl: exclude (typo).

What do you mean with "excluding nil table"?

lib/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,213 @@
...
+        linked_table.table.keep_user_database_table = true
+        linked_table.table.destroy
+      end
+
+      clean_user_tables_with_null_table_id
+    end
+
+    # this method searchs for tables with all the columns needed in a cartodb table.
+    # it does not check column types, and only the latest cartodbfication trigger attached (test_quota_per_row)
+    def search_for_cartodbfied_tables
```
juanignaciosl: I see that previously this method excluded metadata tables, but now it doesn't, isn't it? How come?

lib/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,213 @@
...
+    def non_linked_tables
+      all_cartodbyfied_tables - linked_tables
+    end
+
+    # Tables that have been dropped via API but have an old UserTable
+    def dropped_tables
+      (linked_tables - all_cartodbyfied_tables).select { |t| !renamed_tables.map(&:id).include?(t.id) }
+    end
+
+    # Tables that have been renamed throught the SQL API
```
juanignaciosl: Here and 155, "through".

lib/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,213 @@
...
+      non_linked_tables.select { |metadata_table| !metadata_table.table.nil? }
+    end
+
+    # Tables that have been created trhought the SQL API
+    def new_tables
+      non_linked_tables.select { |metadata_table| metadata_table.table.nil? }
+    end
+
+    # Remove tables with null oids unless the table name exists on the db
+    def clean_user_tables_with_null_table_id
```
juanignaciosl: Move this method just after `unlink_deleted_tables`, please, since it's the only place where it's used. You've extracted some pieces of code to new methods, which is nice because it provides intention meaning, but it makes this more difficult to read. Maybe I'll add more comments like this, sorry, I'm trying to compare previous version and new code so I can find issues.

lib/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,213 @@
...
+
+    def initialize(id, name, user_id)
+      @id = id
+      @name = name
+      @user_id = user_id
+    end
+
+    # Grabs the Table associated with this LinkedTable.
+    def table
+      user_tables = ::UserTable.where(table_id: id, user_id: user_id)
```
juanignaciosl: Strange: `name` is a constructor argument but it's not used for the search, which could lead to inconsistencies (different name). Is this ok? Renamed tables or deleted + created with the same name might interfere? Is this a fact we'd like to log?

lib/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,213 @@
...
+
+      if user_tables.count > 1
+        CartoDB.notify_error("#{@user.username} has duplicate UserTables", duplicated_table_id: table_id)
+      end
+
+      user_tables.first ? Table.new(user_table: user_tables.first) : nil
+    end
+
+    def eql?(other)
```
juanignaciosl: Shouldn't `eql?` compare `user_id` as well? Also, after checking [Object docs about it](http://ruby-doc.org/core-2.3.0/Object.html#method-i-eql-3F), I'd also implement `==`, probably based in `eql?` as well.

spec/lib/carto/bolt_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+require_relative '../../../lib/carto/bolt.rb'
+
+module Carto
+  describe Bolt do
+    before(:each) do
+      @bolt = Carto::Bolt.new('manolo')
+    end
+
+    it 'should allow for custom ttl_ms' do
+      Carto::Bolt.new('manolo', ttl_ms: 1234).info.should include(ttl_ms: 1234)
```
juanignaciosl: Repeating myself, but the nice way to test this is using an spy.

lib/carto/bolt.rb
```diff
@@ -0,0 +1,47 @@
...
+
+      begin
+        yield if is_locked
+        !!is_locked
+      ensure
+        unlock if is_locked
+      end
+    end
+
+    def unlock
```
juanignaciosl: Now that `lock` can only receive a module, `unlock` should be private. I'd probably inline it inside `lock`.

lib/carto/bolt.rb
```diff
@@ -0,0 +1,47 @@
...
+    DEFAULT_REDIS_OBJECT = $users_metadata
+    DEFAULT_TTL_MS = 10000
+
+    def initialize(bolt_key, redis_object: DEFAULT_REDIS_OBJECT, ttl_ms: DEFAULT_TTL_MS)
+      @bolt_key = add_namespace_to_key(bolt_key)
+      @redis_object = redis_object
+      @ttl_ms = ttl_ms
+    end
+
+    def lock
```
juanignaciosl: Now that it receives a block this is actually more `run_locked`, specially since you don't have to call `unlock`.

spec/lib/carto/bolt_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      expect { @bolt.lock.should_raise }.to raise_error('no block given')
+    end
+
+    it 'should allow access in block if locked and unlocked automatically' do
+      @bolt.lock {}.should be_true
+
+      @bolt.unlock.should be_false
+    end
+
+    it 'should not allow access if locked' do
```
juanignaciosl: Oh, nice check :-)

juanignaciosl: Maybe using a `@bolt_2` might be interesting, since it will make clearer that bolt is acquired outside `@bolt` (in Redis).

spec/lib/carto/bolt_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      got_locked.should be_true
+
+      @bolt.unlock.should be_false
+    end
+
+    it 'should expire a lock after ttl_ms' do
+      bolt = Carto::Bolt.new('manolo', ttl_ms: 200)
+
+      got_locked = bolt.lock do
+        sleep(0.5.second)
```
juanignaciosl: Magic number. Use a `ttl_ms = 200` and then `ttl_ms * 2`, for example.

spec/lib/carto/ghost_tables_manager_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    def run_in_user_database(query)
+      ::User[@user.id].in_database.run(query)
+    end
+
+    it 'should be consistent when no new/renamed/dropped tables' do
+      @ghost_tables_manager.consistent?.should be_true
+    end
+
+    it 'should link sql created table, relink sql renamed tables and unlink sql dropped tables' do
```
juanignaciosl: This test is crystal clear. Nevertheless, I'd like a couple of integration tests with some other scenarios (more tables, registered/unregistered...). It might be redundant, but writing these is really quick and we don't want loose ends on this.

juanignaciosl:
I've done some comments. Some are open to discussion and there're no major ones (code looks great), but there're a couple of changes I'd really want. IMHO you could take them into account, spend a while testing today, and try to merge & deploy this tomorrow.

Let me know when this is ready for testing, I'd like to help.

app/controllers/admin/visualizations_controller.rb
gfiorav: Yeah I thought of that scenario. In any case (until we decide to treat the "not locked" case if we do) we go back to the previous situation, so it's best effort as you state. Regarding the comments, I do think it's clear as is no?

```ruby
# Check if any tables need to be linked, relinked or unlinked
if ghost_tables_manager.stale_tables_linked?
  # You might see phantom tables that generate erros upon clicking them; needs relink now
  ghost_tables_manager.sync_user_schema_and_tables_metadata
elsif !ghost_tables_manager.consistent?
  # Worst case you'll refresh for an SQL API genearted table to appear
  ::Resque.enqueue(::Resque::UserJobs::SyncTables::LinkGhostTables, current_user.id)
end
```

I can add your comment anyway!

app/controllers/admin/visualizations_controller.rb
gfiorav: Like it!

lib/carto/bolt.rb
gfiorav: I suspect I might have been using the wrong redis database, /shrug!

lib/carto/ghost_tables_manager.rb
gfiorav: About why order was important before:

The way stale tables were identified was broad: renamed tables and deleted tables where the same, so it relied on running code for relink first and whatever couldn't be relinked would be erased.

lib/carto/ghost_tables_manager.rb
gfiorav: Will do!

lib/carto/ghost_tables_manager.rb
gfiorav: Oh god, I really need to up my comment game :sweat_smile:

lib/carto/ghost_tables_manager.rb
gfiorav: Yeah the comment is poor, will change.

What I meant, is that we will exclude tables with no UserTable since these are tables that need to be linked for the first time, not relinked.

lib/carto/ghost_tables_manager.rb
gfiorav: It was previous-flow oriented. It didn't really look for cartodbified tables as the name suggested, it only grabbed those not already linked. It's better this way and causes less problems on relink since renamed tables would appear twice (as metadata and as non linked) when the truth is that it is linked but needs it's table viz to be renamed. Now it make more sense IMHO and the method names suggest what the method actually does.

lib/carto/ghost_tables_manager.rb
gfiorav: There is only one MetadataTable per physical table and that's intended.

When the physical table is renamed via SQL API, this MetadataTable will contain the new name. MetadataTable might be a misleading name in this case, but I can't figure out a better name?

lib/carto/ghost_tables_manager.rb
gfiorav: Not sure about `==`. 

Since this runs per user, user_id is assured to be the same, but I see your point. Worth changing?

app/controllers/admin/visualizations_controller.rb
juanignaciosl: In fact I can't know, since I already know it. Keep your comments then and if you're ever asked about it, add the comment :-)

spec/lib/carto/bolt_spec.rb
gfiorav: good one!

gfiorav:
Thanks a ton @juanignaciosl !

I'll do as you say, can't wait to refactor to the interesting ideas you proposed!

lib/carto/bolt.rb
juanignaciosl: :+1:

lib/carto/ghost_tables_manager.rb
juanignaciosl: There're two questions on this thread, let "A)" be the order and "B)" the `non_linked_tables ` check.

A) Well, renamed tables where the ones matching oid but not matching name, and deleted tables where metadata tables without matching real table oid, so I can't see the difference. Could you be a little more specific, so I can see this? I'm sorry if I'm getting a little wearying about this, but I want to be as sure as possible.
Also, keeping the original order would also work with the new code, isn't it? If there's no known advantage on the order, I'd keep it.

B) It's true that the check might save maybe one query when it's not needed (since the check query would be moved to the first one and something similar should be done inside the second). Consider a refactor similar to this:

```ruby
non_linked = non_linked_tables
relink_renamed_tables(non_linked)
link_new_tables(non_linked)
unlink_deleted_tables
```

That way you still get the performance benefit but still explains the relationship between `non_linked_tables` and the following methods.

lib/carto/ghost_tables_manager.rb
juanignaciosl: :ok_hand:

lib/carto/ghost_tables_manager.rb
juanignaciosl: Well, it looked for tables with `cartodb_columns`, which is ~ equivalent to search for cartodbified tables, what's the difference between old and now about it?

lib/carto/ghost_tables_manager.rb
juanignaciosl: Uhm, either I've not managed to explain myself or I don't understand your point.

I'm not talking about `MetadataTable` class naming. I'm talking about `name` constructor argument. At first sight it feels strange that it's not used in the search, so search can return a `UserTable` with a different `name`.  I don't know if it's intended or overlooked. Again, this is probably ok, but I'm double-checking that I understand it.

lib/carto/ghost_tables_manager.rb
juanignaciosl: A) About `==`: since docs say that both `eql?` and `==` means approximately the same, I'd override both unless you explicity want to give `==` another meaning.

B) About `user_id`: absolutely. Reading `MetadataTable` class you can't make any assumption on how it is run. Equality doesn't care at all about it.

juanignaciosl:
I guess you're trolling me or being ironic xD.

There're still some doubts to be solved, but most of the comments have been minor. I'll take a last look and do some testing when you ping me again.

gfiorav:
No no, no troll. I genuinely believe the refactors suggested will really improve the code.

juanignaciosl:
Ey, just kidding :-)

spec/lib/carto/ghost_tables_manager_spec.rb
gfiorav: I'm pretty certain that this tests every possible case. If you start with a renamed table, you're in square 2 of this test. If you start with a deleted one, square 3.

I can make this same test use more tables.

Can you come up with a cool scenario to test that might cause any errors?

juanignaciosl: No, not really. Let's keep this untouched unless you come up with any :+1:

lib/carto/ghost_tables_manager.rb
gfiorav: Answering A), take a look at these 4 key methods:

```ruby
# Not viewable in the editor
def non_linked_tables
  all_cartodbyfied_tables - linked_tables
end

# Tables that have been dropped via API but have an old UserTable
def dropped_tables
  (linked_tables - all_cartodbyfied_tables).select { |t| !renamed_tables.map(&:id).include?(t.id) }
end

# Tables that have been renamed through the SQL API
def renamed_tables
  non_linked_tables.select { |metadata_table| !metadata_table.table.nil? }
end

# Tables that have been created trhought the SQL API
def new_tables
  non_linked_tables.select { |metadata_table| metadata_table.table.nil? }
end
```

Please pin point here what method seems weird to you so we can discuss it :)

B) 

It seems too much of a hassle TBH. I could do:
```ruby
relink_renamed_tables unless renamed_tables.empty?
link_new_tables unless new_tables.empty?
unlink_deleted_tables unless dropped_tables.empty?
```

but I wanted to make it more query efficient. Having returns and passing parameters complicates the code more IMHO, but I've been doing this for a week so I might be crazy by now.

lib/carto/ghost_tables_manager.rb
gfiorav: The old code exclude metadata tables assuming that you already had them through UserTables. I disliked it because:

a) It had a dishonest name (it didn't look for all the cartodbfied tables)
b) It made assumptions that might not always work (I might want all the tables to check consistency with actual metadata tables)

lib/carto/ghost_tables_manager.rb
gfiorav: OK, changing

lib/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,213 @@
...
+
+    # Tables viewable in the editor; in users user_tables
+    def linked_tables
+      @user.tables.select(:name, :table_id)
+           .map { |table| Carto::MetadataTable.new(table.table_id, table.name, @user.id) }
+    end
+
+    # May not be viewed in the editor; carotdbyfied_only
+    def all_cartodbyfied_tables
+      real_tables.select { |table| search_for_cartodbfied_tables.include?(table[:relname]) }.compact
```
juanignaciosl: Oh, I've spotted this (again, isn't it? did we day this was ok? :-\). This runs a query inside of a loop, which is really inefficient (one more query doesn't use to matter, one more query inside a loop does). You can instead...

```ruby
cartodbfied_tables = search_for_cartodbfied_tables
real_tables.select { |table| cartodbfied_tables.include?(table[:relname]) }.compact
...
```

lib/carto/ghost_tables_manager.rb
juanignaciosl: A) No, I'm not asking about the new methods but about your "renamed tables and deleted tables where the same" affirmation. Those methods look ok indeed.

B) Your proposal has the same issue than the current code IMHO, since method call is still depending on information that is not passed to him. But if you think that passing a parameter is more complicated just let it be.

lib/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,213 @@
...
+      @user.in_database(as: :superuser)[sql].all.map { |table| table[:table_name] }
+    end
+
+    # Tables viewable in the editor; in users user_tables
+    def linked_tables
+      @user.tables.select(:name, :table_id)
+           .map { |table| Carto::MetadataTable.new(table.table_id, table.name, @user.id) }
+    end
+
+    # May not be viewed in the editor; carotdbyfied_only
```
juanignaciosl: Typo :-)

lib/carto/ghost_tables_manager.rb
gfiorav: Oops! Missed that, changing

lib/carto/ghost_tables_manager.rb
juanignaciosl: Ok. I see. Previously that logic was at the query, now you do it here:

```ruby
    def non_linked_tables
      all_cartodbyfied_tables - linked_tables
    end
```

Much better indeed :+1: I was missing the big picture of this.

lib/carto/ghost_tables_manager.rb
gfiorav: Damn it! Fixed!

juanignaciosl:
Ok, the conversation about the order of the operations is still open (I'd suggest to keep the old order if your refactor supports that though), but everything else looks fine. Since this kind of code is tricky, reading it is not enough for me, so I'll do some tests before the scheduled smoke testing.

Added the "do not merge yet" label because of the smoke testing to be scheduled.

gfiorav:
Ok, sounds great. I'd like to preserve this order since it let's me group operations that depend on the existence of non linked tables. I'll deploy this to staging and have it set up so we can test it.

juanignaciosl:
I've been doing some tests and it looks fine. Nevertheless I still have some concerns about fixing deleted and renamed tables synchronously. Although from a user experience it's not likely to be a problem (worst case scenario, refreshing the browser, will work without triggering a second ghost table process), it could still block some Unicorn workers. Please have a quick talk with @javisantana about it if you haven't done it yet.

Only smokes remaining, AFAIK :-)

:clap:

gfiorav:
Thanks! How would this block a worker?

juanignaciosl:
This code blocks unicorn workers attending user requests when the user has deleted or renamed tables. For cases where it takes long, several workers might get blocked. I _think_ that's a tradeoff that we can assume in exchange of displaying a more accurate dashboard list for most scenarios, but it's something that we should have in mind and monitor. If workers begin to be too busy I'd async all ghost table linkages.

gfiorav:
So the interface is blocked while synchronous ghost tables work right? I mean, asynchronous request that find the lock set won't be blocked, just discarded.

juanignaciosl:
Yes. Let me explain with the worst possible scenario: let's say we have 4 workers and 4 users with deleted tables open the dashboard at the same time. All workers will be locked until ghost tables are linked. I don't think this is a likely scenario at all, but I think this is worth some monitoring after the deploy.

gfiorav:
Exactly, I wanted to confirm I was understanding your concerns. 

It certainly is something to worry about. Ghost tables do run synchronously in some cases right now in production, but not as "commonly" as it will now. All things considered, I do think that the worst case would be some user who creates a lot of tables through the SQL API *and* cartodbyfies them, which I've heard from systems is bound to be a rare case.

I guess we'll find out sooner or later!

juanignaciosl:
:sparkles:

gfiorav:
@juanignaciosl, please take a look.

It has been heavily changed, but now supports mismatch oids so I think it's worth it. Testing all previous dangerous scenarios like syncs and drop-create works like a charm.

Thanks!

app/connectors/importer.rb
```diff
@@ -166,7 +166,7 @@ def rename(result, current_name, new_name, rename_attempts=0)
             # this is likely caused by a table deletion, so we run ghost tables to cleanup and retry
             if rename_attempts == 1
               runner.log.append("Triggering ghost tables for #{user_id} because collision on #{new_name}")
-              ::User.where(id: user_id).first.link_ghost_tables
+              Carto::GhostTablesManager.new(user_id).link_ghost_tables_synchronously
```
juanignaciosl: A method is synchronous by definition (in Ruby, at least), so a better naming would renaming `link_ghost_tables_synchronously` to `link_ghost_tables` and `link_ghost_tables` to `enqueue_link_ghost_tables_job`. Not important though, up to you.

juanignaciosl: PS: oh, I see that `link_ghost_tables` makes the (a)sync decision, so my renaming suggestion isn't accurate. Not sure whether there's a better alternative then.

app/models/table.rb
```diff
@@ -70,6 +70,7 @@ class Table
 
   def initialize(args = {})
     if args[:user_table].nil?
+      # TODO: This won't work, you need to UserTable.new.set_fields(args, args.keys)
```
juanignaciosl: Is this TODO meant to stay after this PR?

lib/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,254 @@
...
+module Carto
+  class GhostTablesManager
+    MUTEX_REDIS_KEY = 'ghost_tables_working'.freeze
+    MUTEX_TTL_MS = 60000
+
+    def initialize(user_id)
+      @user = ::User.where(id: user_id).first
+    end
+
+    def link_ghost_tables
```
juanignaciosl: I have mixed feelings about hiding the fact of (a)sync triggering behind this method. There's a smell at `def safe_async?` definition, the comment: "Check if any unsafe stale (dropped or renamed) tables will be shown to the user". This reveals that this couples ghost table manager to the visualization controller, a manager should not know anything about "showing". Suggestion:

1. Rename `safe_async?`to `dropped_or_stale_table?`.
2. Rename `link_ghost_tables_synchronously` to `link_ghost_tables` (it's synchronous by definition).
3. Create a `trigger_link_ghost_tables_job` that enqueues.
4. Current `link_ghost_tables` method disappears.
5. Check `return if user_tables_synced_with_db?` at both link methods.

I think that this approach decouples manager from controller and makes the async decision explicit for the client.

Of course, I'm open to discussion on this as usual. What do you think?

lib/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,254 @@
...
+      got_locked = bolt.run_locked { sync }
+
+      CartoDB::Logger.info(message: 'Ghost table race condition avoided', user: @user) unless got_locked
+    end
+
+    def sync
+      cartodbfied_tables = fetch_cartobfied_tables
+
+      # Update table_id on UserTables with physical tables with changed oid. Should go first.
+      cartodbfied_tables.select(&:regenerated?).each(&:regenerate_user_table)
```
juanignaciosl: These lines feel dangerous, because you iterate on `cartodbfied_tables`, which represent the state of the database in that moment, and then you both change the state and do new checks. I'll take a deeper thought on this but this is code is hard to reason about.

lib/carto/ghost_tables_manager.rb
```diff
@@ -0,0 +1,254 @@
...
+                 tg.tgrelid reloid,
+                 count(column_name::text) cdb_columns_count
+          FROM information_schema.columns c, pg_tables t, pg_trigger tg
+          WHERE
+            t.tablename = c.table_name AND
+            t.schemaname = c.table_schema AND
+            c.table_schema = '#{@user.database_schema}' AND
+            t.tableowner = '#{@user.database_username}' AND
+            column_name IN (#{cartodb_columns}) AND
+            tg.tgrelid = (quote_ident(t.schemaname) || '.' || quote_ident(t.tablename))::regclass::oid AND
```
juanignaciosl: Previous code excluded ` c.table_name NOT IN (#{metadata_table_names}) AND` (if there were `metadata_table_names`), how come it's no longer needed?

app/connectors/importer.rb
gfiorav: Yeah, I totally see what you mean, but it's harder to name a method to imply that it might or might not run asynchronously so you could have this implicitly mean it's synchronous. Unless we come up with a better naming, I'd leave it as is.

app/models/table.rb
gfiorav: Yes, it's outside of the scope of the refactor and when I changed it a lot of stuff broke. I've even seen some code that depends on this not working :/

lib/carto/ghost_tables_manager.rb
gfiorav: I see what you mean. The comment reveals the final result of either action, but if it read "determine priority of sync" or something like that and the name changed accordingly, would it still be a smell?

Probably, in the sense that you're right that this might be to much responsibility for a manager. 

In this same PR 2 weeks ago, the decision was made outside, but we refactored it to only have one public method. It seemed like a good idea back then, but I see what mean.

I'll play around with the code a bit.

app/models/table.rb
juanignaciosl: 👌

lib/carto/ghost_tables_manager.rb
gfiorav: I do take a picture on the cartodbfied tables a this point and reuse it. It's not accidental. I think it's safer to play through instead of updating the snapshot.

I *don't* change their state though, I just change the UserTables, never the physical tables directly.

I don't feel it's dangerous and from testing it's pretty solid. I guess the only week spot is if you go and remove a table through sql api in the mean time, but that just means another run.

lib/carto/ghost_tables_manager.rb
juanignaciosl: Yes, I could even defend the opposite position. As I've said, I'm not sure about any of both alternatives.

lib/carto/ghost_tables_manager.rb
gfiorav: Previous code lied: It wasn't fetching all cartodbfied tables, just those that weren't already linked. That gave way to some undetectable errors, like when a linked table had a bad oid or something like that.

This manager changed that from the start. I use full comparison to determine what the SQL API has done since the last time the Editor check more accurately.

gfiorav: Also, another change is that I also grab the oids now, not just the name. This helps detect "regenerated" tables better.

juanignaciosl:
I've done several comments. I feel a little like running in circles with this, so feel free to ignore my comments if you feel confident enough. Code style comments are not major at all, and the "feels risky" ones are due to the nature of the change, so the only thing I have to stress before merge is _test this thoroughly, please_.

gfiorav:
Thanks! I've tested it thoroughly. I'll get a couple of people to test it a bit more and deploy.

PR #6894, opened 2016-03-16
Service list item file type ellipsis

### LINKED ISSUE #6890: Geojson name on file listing does not fit in the format box
image

(When using the Box connector, which is the only one right now that can retrieve GeoJSON files)

Probably not worthy, but at least to take into account if a similar file listing is to be implemented/is implemented.

cc @xavijam

MariaCheca:
Fixes #6890 

Fixed overflow adding ellipsis to the element.

screen shot 2016-03-16 at 10 57 33

CR @xavijam cc @iriberri


xavijam:
:+1:

MariaCheca:
:sparkles:

Issue #6904, opened 2016-03-17
Mixed content warnings for embedded maps - avatar url


alexkrolick:
The iframe for published map embeds loads the user's avatar over HTTP, causing mixed content warnings on HTTPS sites

javitonino:
This should be fixed now, although it could take up to a day to propagate due to caching. If more than 24h passed and you still see some embeds with mixed contents, feel free to reopen this issue including a link, so we can check what we missed. Thanks for reporting!

PR #6905, opened 2016-03-17
Regex change /^http:/ to 'https:' in user avatar url

whokilledtheelectricmonk:
- User avatars are being served over HTTP even if the map embed is HTTPS, causing mixed content warnings on sites with SSL (#6904)
- Assets on S3 should always have a HTTPS equivalent, so this should not break anything


Cartofante:

michellechandra:
hey @rochoa - this is related to a Support ticket, can you take a look?

xavijam:
Please, @javitonino, could you take a look?

javitonino:
Hi @whokilledtheelectricmonk and thanks for the patch.

Unfortunately, there is a problem with unconditionally replacing `http` with `https`. Some of our older users have custom URLs for their avatars, and changing the URL for them may break things for them.

I think it would be better to solve this at creation time, so new avatars will always use https, and users could simply fix this for any maps they own by reuploading their avatars (or we could run a task to update the database).

whokilledtheelectricmonk:
So some avatars are hosted off-site? What about importing them to your S3
bucket?

whokilledtheelectricmonk:
Closing this in favor of making sure assets are HTTPs at the asset level

PR #6906, opened 2016-03-17
Update version of cartodb-postgresql to 0.14.3

iriberri:
Includes a commit of `table_spec.rb` in which a test has been commented as it represented legacy behavior -- I'm not deleting it directly in order to keep track of information that might be interesting in the future for further changes on the extension. This test is no longer needed because incorrect data in the `cartodb_id` column will not generate a new invalid column, but it will suppose an error in the import process.


iriberri:
New user in staging `mr-importer-ded-07` has been created correctly with the new version.

iriberri:
@rafatower could you approve? Tests are running right now as I had to commit a change in NEWS.md because of conflicts but they passed previously.

rafatower:
:+1:

PR #6907, opened 2016-03-17
Strong passwords for organization signups

gfiorav:
Adds "strong" password validation to user signups.

By default, it requires:

+ At least 8 characters
+ At most 64 characters
+ Both letters and numbers or symbols
+ At least 1 number or 1 symbol 

Recognised as symbols are:

`{ } [ ] , . < > ; : ‘ “ ? / | \ `` ~ ! @ # $ % ^ & * ( ) _ - + =`

Limitations: 

+ Only applies to organization signups.
+ Doesn't take into account dictionaries or entropy (weaker than other implementations)

Is this what you had in mind, @vhamer ?

Please CR @juanignaciosl, @javitonino 
cc/ @javisantana


javisantana:
Is this for all the organizations? Should we have a flag to enable it? What do you think @vhamer @saleiva

gfiorav:
It could easily be hidden behind a feature flag, indeed. Just give me a name for the feature flag.

app/helpers/carto/strong_password_validator.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      if @password.length < @min_length
+        @errors << "must be at least #{@min_length} #{'character'.pluralize(@min_length)} long"
+      end
+
+      if @password.length > @max_length
+        @errors << "must be at most #{@max_length} #{'character'.pluralize(@max_length)} long"
+      end
+
+      unless @password =~ /[a-zA-Z]{#{@min_letters},}/
```
javitonino: Careful! This detects consecutive letters. It works for `@min_letters = 1` but is incorrect otherwise.
`'abc34def' =~ /[a-zA-Z]{5,}/` => nil`

javitonino: I suggest `'abcdef'.scan(/[[:alpha:]]/).size < @min_letters`

gfiorav: Ha! Good one!

app/helpers/carto/strong_password_validator.rb
```diff
@@ -0,0 +1,68 @@
...
+    DEFAULT_MIN_LETTERS = 1
+
+    SYMBOLS = %w({ } [ ] , . < > ; : ‘ “ \? \/ \| \ ` ~ ! @ # $ % ^ & \* ( ) _ - \+ =).freeze
+
+    def initialize(password,
+                   min_length = DEFAULT_MIN_LENGTH,
+                   max_length = DEFAULT_MAX_LENGTH,
+                   min_letters = DEFAULT_MIN_NUMBERS,
+                   min_symbols = DEFAULT_MIN_SYMBOLS,
+                   min_numbers = DEFAULT_MIN_LETTERS)
```
javitonino: I'd use keyword arguments (`min_length: DEFAULT_MIN_LENGTH`) for this. Much easier to read when calling it.

app/helpers/carto/strong_password_validator.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      @min_length  = min_length.nil?  ? DEFAULT_MIN_LENGTH  : min_length
+      @max_length  = max_length.nil?  ? DEFAULT_MAX_LENGTH  : max_length
+      @min_letters = min_letters.nil? ? DEFAULT_MIN_NUMBERS : min_letters
+      @min_symbols = min_symbols.nil? ? DEFAULT_MIN_SYMBOLS : min_symbols
+      @min_numbers = min_numbers.nil? ? DEFAULT_MIN_LETTERS : min_numbers
+
+      @password = password.nil? ? '' : password
+
+      valid?
```
javitonino: Is there a reason for this? I cannot think of one right now, but I do not know if Ruby does some kind of magic in "constructors".

app/helpers/carto/strong_password_validator.rb
gfiorav: Changed for `unless @password.scan(/[[:alpha:]]/).size > @min_letters`

app/helpers/carto/strong_password_validator.rb
```diff
@@ -0,0 +1,68 @@
...
+      end
+
+      @errors.empty?
+    end
+
+    def errors
+      @errors
+    end
+
+    def message
```
javitonino: Nitpicking: I like `error_message` better.

app/helpers/carto/strong_password_validator.rb
gfiorav: It's a bit ugly, but I want to avoid having to run `.valid?` and then `.message` wherever this is used... I could do it inside `.message` with a flag or something but it's unnecessarily complicated IMHO.

I'll put in the `after_initialize` though.

app/helpers/carto/strong_password_validator.rb
gfiorav: Meh! I wanted to avoid the hash but I guess you're right. I'll change it.

spec/helpers/carto/strong_password_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+
+    it 'should be invalid when password too long' do
+      validator = Carto::StrongPasswordValidator.new(PASSWORD, max_length: PASSWORD.length - 1)
+
+      validator.valid?.should be_false
+      validator.message.should == "must be at most #{PASSWORD.length - 1} characters long"
+    end
+
+    it 'should be invalid when password not enough numbers or symbols' do
```
javitonino: Please add a test that succeeds because it does have enough symbols OR numbers (but not both). It would have caught the boolean logic problem we have just talked about.

app/helpers/carto/strong_password_validator.rb
gfiorav: I was sort of following the rails model validations where you ask for full_message to get the validation errors as a string. Let's have nacho break the tie.

spec/helpers/carto/strong_password_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      validator = Carto::StrongPasswordValidator.new(PASSWORD, min_symbols: 3, min_numbers: 2)
+
+      validator.valid?.should be_false
+      validator.message.should == "must contain at least 3 symbols or 2 numbers"
+    end
+
+    it 'should be invalidate a nil password' do
```
javitonino: Also missing an explicit test for minimum number of letters (although it will be caught by this test). Maybe a good opportunity to test that `ñ` really counts as a letter.

gfiorav: good point

javitonino:
:+1: LGTM, pending those shiny new tests.

gfiorav:
Thanks for those excellent points, @javitonino :heart:

juanignaciosl:
Ping me when test pass, please. _Writing down doing a leapfrog about notifying people on green_.

gfiorav:
Test passing! @juanignaciosl

app/helpers/carto/strong_password_validator.rb
```diff
@@ -0,0 +1,70 @@
...
+  class StrongPasswordValidator
+    DEFAULT_MIN_LENGTH = 8
+    DEFAULT_MAX_LENGTH = 64
+    DEFAULT_MIN_NUMBERS = 1
+    DEFAULT_MIN_SYMBOLS = 1
+    DEFAULT_MIN_LETTERS = 1
+
+    SYMBOLS = %w({ } [ ] , . < > ; : ‘ “ \? \/ \| \ ` ~ ! @ # $ % ^ & \* ( ) _ - \+ =).freeze
+
+    def initialize(password,
```
juanignaciosl: I think that `password` should not be a constructor argument but a `valid?` one. Being _data_ instead of _configuration_, like other arguments are, is a smell. The need for `after_initialize` call to `valid?` is another. Validating several passwords might require several instantiations, that also smells weird to me :-\

app/helpers/carto/strong_password_validator.rb
juanignaciosl: Related to the comment about `password` at the constructor, I'd really transform `def valid?` into `def validate(password)`, returning the `errors` array. Give that design a try, you'll see it's much simpler. This object essentially models a parametrized algorithm ("validating a password"), having state (`@password`, `@errors`) is not useful and makes it harder to use. I see that you've designed this after ActiveRecord, but since this is not a model same rules not apply.

juanignaciosl: That would make `message` unnecessary as well. Rails accept several errors to be added to an attribute, so, being `password_errors` the result of `validate` and `errors` the `User` validate errors list (inside `User#validate`, `password_errors.each { |e| errors.add(:password, e) } will bind them to the user password.

spec/helpers/carto/strong_password_validator_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      validator.valid?.should be_true
+    end
+
+    it 'should be valid when password has enough symbols but not enough numbers' do
+      validator = Carto::StrongPasswordValidator.new(PASSWORD, min_symbols: 2, min_numbers: 3)
+
+      validator.valid?.should be_true
+    end
+
+    it 'should be invalidate a nil password' do
```
juanignaciosl: Fix the sentence grammar and add an empty string test, please :-)

juanignaciosl:
I've made some major comments about `StrongPasswordValidator` design that we should discuss before closing this.

Also, I _really_ doubt that this should be enabled by default for all organizations. IMHO this is clearly an authentication configuration parameter for organizations, such as enabling Google login is. This is clearly _not_ a feature flag, which is a development tool, not a configuration one.

app/helpers/carto/strong_password_validator.rb
gfiorav: Yeah, I went to far with the `ActiveRecord` design, I see that. I'll make the suggested changes, but as for message, I think we should keep it. It's the only way to get a easy-readable message to the front, i've tried passing the the elements separated like you suggest but I had worse results.

app/helpers/carto/strong_password_validator.rb
gfiorav: ActiveRecord influence... changing

spec/helpers/carto/strong_password_validator_spec.rb
gfiorav: ah, typo, fixing.

app/helpers/carto/strong_password_validator.rb
juanignaciosl: Could you paste an screenshot? I tried that with a username in a different screen and worked fine, and I think I've already used at signup.

gfiorav:
Hey @juanignaciosl , here are the screenshots:

screen shot 2016-03-18 at 12 32 13
vs
screen shot 2016-03-18 at 12 32 56

lib/carto/strong_password_validator.rb
```diff
@@ -0,0 +1,75 @@
...
+
+      if symbols_in(password) < @min_symbols && numbers_in(password) < @min_numbers
+        errors << "must contain at least #{@min_symbols} #{'symbol'.pluralize(@min_symbols)} or " +
+                  "#{@min_numbers} #{'number'.pluralize(@min_numbers)}"
+      end
+
+      errors
+    end
+
+    def formatted_error_message(errors)
```
juanignaciosl: _resuming previous conversation that I can't find_

After seeing the screenshots, I still think that having a method that is 100% unrelated to password validation only for a couple of whitespaces and an 'and' that Rails? (I'm not sure about where the problem is) should handle doesn't really pays off, but up to you, not really relevant.

gfiorav: I see your point, but for readability, I think this is better. We can refactor one line if we need the separated messages in the future :)

gfiorav:
@juanignaciosl I've added the toggle for the auth settings and a migration for the org owner to choose to opt in on strong passwords. Please look into the solution!

lib/carto/strong_password_validator.rb
juanignaciosl: /shrug

app/views/admin/organizations/auth.html.erb
```diff
@@ -71,7 +71,22 @@
...
+          <div class="FormAccount-rowLabel">
+            <label class="FormAccount-label">Strong Passwords</label>
+          </div>
+          <div class="FormAccount-rowData">
+            <div class="Toggler">
+              <%= f.check_box :strong_passwords_enabled, :id => "strong_passwords_enabled" %>
+              <%= label_tag(:strong_passwords_enabled, '') %>
+            </div>
+            <div class="FormAccount-rowInfo FormAccount-rowInfo--marginLeft">
+              <p class="FormAccount-rowInfoText">Require users to use strong passwords</p>
```
juanignaciosl: Please add information about what a "strong" password is.

juanignaciosl:
I've added a minor UI suggestion.

:+1:

gfiorav:
Thanks, @juanignaciosl, changed!

I'm pinging @csobier here so she can approve or suggest a better way of putting it.

csobier:
"Passwords must contain 8 characters, and require a combination of letters, numbers, and/or special characters."

gfiorav:
Hey @csobier , i've just realised that I haven't posted a screenshot of the actual message. I think you're proposing that for the password validation dialogue.

Here is what I wanted you to review:

screen shot 2016-03-18 at 14 26 06

csobier:
Ah, okay. Yeah, I just read the code.  This image helps.  I would still edit the Strong Password description as follows: " Must contain 8 characters, and requires a combination of letters, numbers, and/or special characters." 

Other than than, looks good! (Make sure you create a related Docs issue for me, so I can update on my end when the Product changes). Thanks!

gfiorav:
Thanks @csobier ! It looks like this now:

screen shot 2016-03-18 at 14 42 44

csobier:
:+1:

vhamer:
so great guys, thanks for working on this feature! it will make our big
customers happy.

On Fri, Mar 18, 2016 at 9:50 AM, csobier <notifications@github.com> wrote:

> [image: :+1:]
>
> —
> <https://github.com/CartoDB/cartodb/pull/6907#issuecomment-198369489>
>



-- 
Vanessa Hamer
Business Development & Operations
CartoDB <http://cartodb.com> | vhamer@cartodb.com
919.389.7768 | skype: vhamer

gfiorav:
Tests are back to :green_apple: 

If I get a final :+1: I'll :package: it on Monday

gfiorav:
Separated migration to #6939

gfiorav:
@vhamer feature is in production

vhamer:
fantastic @gfiorav thank you!

PR #6908, opened 2016-03-17
Fix layers' letter representations

viddo:
Especially on reset/adding multiple layers, to yield correct letters.

Removed the need to track letters separately, derive from collections layers at the time it's needed instead.

Also fix bug with basemap being a non-tiled layer (e.g. background/plain one)

@javierarce please review


javierarce:
:+1:

viddo:
Frontend tests passed, so merging

PR #6910, opened 2016-03-17
Fix analysis definition update

viddo:
Didn’t send correct node id on subsequent nodes being added

cc @javierarce


javierarce:
:+1:

PR #6914, opened 2016-03-17
Use branch heads for internal deps

viddo:
Ditto https://github.com/CartoDB/cartodb.js/pull/1199

A plain `npm install` will install dependencies defined in the shrinkwrap file, regardless of what's stated in package.json. To update an arbitrary dependency you would need to modify the `package.json`, and then do a `npm update name-of-pkg && npm shrinkwrap --dev`.

To ease development for ourselves, I've added a npm script that basically does the above for our own owned libs:

```bash
$ npm run update-internal-deps
> npm update camshaft-reference cartoassets cartodb-deep-insights.js cartodb.js && npm shrinkwrap --dev
…
```

Running the command above will change `npm-shrinkwrap.json`, review it before the commit/PR, so the changes you want are included in there.

cc @javisantana @fdansv @alonsogarciapablo @xavijam @javierarce  @rochoa (did I miss someone? feel free to mention them)

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -18,13 +18,13 @@
           "resolved": "https://registry.npmjs.org/num2fraction/-/num2fraction-1.2.2.tgz"
         },
         "caniuse-db": {
-          "version": "1.0.30000430",
+          "version": "1.0.30000431",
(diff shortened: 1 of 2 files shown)


xavijam:
:clap:

PR #6915, opened 2016-03-17
Take into account letters from layers data on initial reset

viddo:
E.g. if some persisted layers have letters set, and the last one
happens to be the one with the letter a, it needs to be taken into
account when creating the models.

@javierarce


javierarce:
:sunglasses: :+1:

PR #6926, opened 2016-03-18
Importer using batch api instead of batch query

ethervoid:
- [x] Batch SQL Api lib
- [x] Use of batch sql API for the importer column query batch parts
- [x] Use of batch sql API for the cartodbfy function
- [ ] Check when an error occurs in the Batch SQL API
- [ ] Test the Batch SQL API
- [ ] Change all the tests that import datasets or use the importer to mock the Batch SQL Api

// @iriberri @rafatower


services/importer/lib/importer/batch_api_query.rb
```diff
@@ -0,0 +1,39 @@
...
+          api_key: api_key
+        })
+      end
+
+      def execute(query)
+        @response = @sql_api_client.execute(query)
+        until ['done', 'failed', 'cancelled'].include? @response["status"]
+          sleep POLLING_TIME
+          @response = status(@response["job_id"])
+          debugger
```
iriberri: :red_circle:

services/importer/lib/importer/batch_api_query.rb
```diff
@@ -0,0 +1,39 @@
...
+          puts 'lala'
```
iriberri: "lalalalalá"

services/importer/lib/importer/georeferencer.rb
```diff
@@ -72,7 +72,7 @@ def create_the_geom_from_geometry_column
         geometry_column_name = geometry_column_in
         return false unless geometry_column_name
         job.log "Creating the_geom from #{geometry_column_name} column"
-        column = Column.new(db, table_name, geometry_column_name, schema, job)
+        column = Column.new(db, table_name, geometry_column_name, user, schema, job)
```
iriberri: `user.username`

ethervoid:
VACUUM FULL was added [here](https://github.com/CartoDB/cartodb/commit/4e047214f82b0e54694679d3fbc68c1b4ac1e57b#diff-6863e1a510b5ab1531e0e1e906c48b0cR396), and there is no explanation for the use of it. I think, that the reason was to compact the table to free space.

The [explanation](https://wiki.postgresql.org/wiki/VACUUM_FULL#When_to_use_VACUUM_FULL_and_when_not_to) to use VACUUM FULL is:

```VACUUM FULL is only needed when you have a table that is mostly dead rows - ie, the vast majority of its contents have been deleted. It should not be used for table optimization or periodic maintenance, as it's generally counterproductive. In most cases the freed space will be promptly re-allocated, possibly increasing file-system-level fragmentation and requiring file system space allocations that are slower than just re-using existing free space within a table.````

So I think we should change to VACUUM alone to improve the performance of the importer in big tables like 12M points that spends 20 minutes doing the FULL VACUUM

// @rafatower

rafatower:
Just adding a bit more context to the issue: https://github.com/CartoDB/cartodb-platform/issues/260

As mentioned, we'd need to connect ogr directly to pg, or else create a big-ass transaction.

That basically means adding a bit more config (that I think went back and forth and don't know why). My feeling is that if we solve the ogr config problem then the other one can be solved with a `user.direct_db_connection.execute_with_timeout` or similar w/o adding an extra layer of complexity.

javisantana:
One of the main problems with those operations is not the way they are run (batch api or from resque), the actual problem is we run a full update many times when we could do just one. 

(it's fine to move to the batch api but keep the real problem in mind please)

ethervoid:
We have one problem solved, this way we don't have to touch the timeouts. Eventually we have a dataset that even if we don't make full updates, surpasses the limit of timeout.

But you're right we have to check the Cartodbfy function and improve it.

With this change we are able to import a 12M points dataset in 30 minutes which is a good start point don't you think? :)

rafatower:
Closing this PR FTM, I'll create a ticket to work on the configs and the `user.direct_db_connection.execute_with_timeout` or similar.

PR #6935, opened 2016-03-21
Get correct kind value for layer type setting

viddo:
Get kind from original response data instead of attrs, which has filtered it out. 

Updated test case to have it covered

cc @javierarce


javierarce:
:+1:

viddo:
frontend tests pass, so merging

PR #6938, opened 2016-03-21
Move node destroy method to correct model

viddo:
Nodes don’t have API endpoint, so the custom destroy (i.e. remove from
collection) should be on the node, not on the analysis definition.

@javierarce


javierarce:
:+1:

viddo:
Frontend tests pass, so merging

viddo:
Accidentally closed… passing build: https://travis-ci.org/CartoDB/cartodb/builds/117422265

PR #6946, opened 2016-03-22
1125 error updating user in central

javitonino:
This ensures that no local changes are made before they can be saved at Central, so the state is kept in sync. Adds tests to check this behaviour (by stubbing `update_in_central` to fail).

Closes cartodb/cartodb-central#1125

CR @gfiorav

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -66,8 +66,8 @@ class Admin::UsersController < Admin::AdminController
       @user.set_fields(attributes, [:email])
     end
 
-    @user.save(raise_on_failure: true)
     @user.update_in_central
+    @user.save(raise_on_failure: true)
 
     update_session_security_token(@user) if password_change
 
--- a/app/controllers/carto/api/organization_users_controller.rb
+++ b/app/controllers/carto/api/organization_users_controller.rb
@@ -47,12 +47,14 @@ module Carto
         # ::User validation requires confirmation
         params_to_update[:password_confirmation] = params_to_update[:password]
 
-        if !(soft_limits_validation(@user, params_to_update) && @user.update_fields(params_to_update, params_to_update.keys))
+        @user.set_fields(params_to_update, params_to_update.keys)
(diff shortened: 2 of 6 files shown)


app/controllers/admin/users_controller.rb
```diff
@@ -66,8 +66,8 @@ def account_update
       @user.set_fields(attributes, [:email])
     end
 
-    @user.save(raise_on_failure: true)
     @user.update_in_central
+    @user.save(raise_on_failure: true)
```
gfiorav: I know it's old code, but why is this specified manually? shouldn't it be true by default?

app/controllers/carto/api/organization_users_controller.rb
```diff
@@ -47,12 +47,14 @@ def update
         # ::User validation requires confirmation
         params_to_update[:password_confirmation] = params_to_update[:password]
 
-        if !(soft_limits_validation(@user, params_to_update) && @user.update_fields(params_to_update, params_to_update.keys))
+        @user.set_fields(params_to_update, params_to_update.keys)
+        if !(soft_limits_validation(@user, params_to_update) && @user.valid?)
```
gfiorav: I always prefer 

```ruby
unless this_condition && that_condition

end
```

than 

```ruby
if !(this_condition && that_condition)

end
```

I think it's easier to read!

spec/requests/admin/organization_users_controller_spec.rb
```diff
@@ -83,6 +83,17 @@
           @existing_user.reload
           @existing_user.quota_in_bytes.should eq new_quota
         end
+
+        it 'does not update users in case of Central failure' do
```
gfiorav: Very nice test!

gfiorav:
Minor comments, :package: ready to me!

Very good tests :clap: !

app/controllers/admin/users_controller.rb
javitonino: In the user model we have `self.raise_on_save_failure = false` and most of the code checks the return code of save, instead of handling exceptions. This appears to converts the false return to an exception, so it can be handled easier in this method.

PR #6959, opened 2016-03-23
Fixing problem related with limits reached dialog

### LINKED ISSUE #6958: Add some warning when adding a layer over the limit
In the editor, map view, when you try to add a layer and you are already at the maximum number allowed by the plan, the button does nothing. It should display a warning saying that the maximum number of layers has been reached (maybe upselling?). Alternatively, maybe we could permanently display a layer counter (3/4 layers in use).

cc @urbanphes @xavijam

xavijam:
cc @javitonino @urbanphes 
Fixes #6958.


javitonino:
That is what I call a quick fix! :+1:

PR #6966, opened 2016-03-28
Open analysis modal from edit-analysis-workflow

### LINKED ISSUE #6951: Check and connect all "add analysis" buttons available under analysis form views
screen shot 2016-03-22 at 15 02 38
screen shot 2016-03-22 at 15 02 51

viddo:
Opens analysis from the editing analysis

Resolves #6951

@xavijam can you review?

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/modals/modals-service-model.js
+++ b/lib/assets/javascripts/cartodb3/components/modals/modals-service-model.js
@@ -41,6 +41,12 @@ module.exports = cdb.core.Model.extend({
     return this._modalView;
   },
 
+  destroy: function () {
+    if (this._modalView) {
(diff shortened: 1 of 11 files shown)


xavijam:
Easy and straight-forward :clap:.
Also, I feel it fixes #6951, doesn't it?

viddo:
Missed that one, fixed in the last commit.

"Easy and straight-forward", IMHO should be easier, I still think passing dependencies several layers still smell, but I'll address that separately.

xavijam:
From what we have, it is easy. About if it could be easier, yes. But as you have said, let's tackle it in a separate issue.

viddo:
Frontend tests passing, merging

xavijam:
:+1:

PR #6972, opened 2016-03-28
add button

piensaenpixel:
@javierarce i suppose you need add now some logic, all yours (when you can)

image


javierarce:
@xavijam: can you review my javascript changes, please? I added the logic to create widgets, layers and elements (this two last items will be available in the future)

javierarce:
Ping?

lib/assets/javascripts/cartodb3/components/tab-pane/create-text-labels-tab-pane.js
```diff
@@ -30,6 +30,7 @@ module.exports = function (paneItems, options) {
     });
 
     return {
+      kind: paneItem.kind,
```
xavijam: Shouldn't it be 'context' more than `kind`? Tab panes only knows about content or context, not sure :disappointed:.

xavijam: or name!

lib/assets/javascripts/cartodb3/editor/editor-view.js
```diff
@@ -24,21 +28,47 @@ module.exports = cdb.core.View.extend({
     this._mapStackLayoutModel = opts.mapStackLayoutModel;
   },
 
+  _addItem: function () {
+    var selectedTab = this._mapTabPaneView.collection.getSelected();
+
+    switch (selectedTab.get('kind')) {
+      case 'widgets':
+        this._editorWidgetsView.addWidget();
```
xavijam: We should be able to open the dialog in this view, shouldn't we?

xavijam: Not sure if this is the best way, modals master @viddo ?

viddo: > We should be able to open the dialog in this view, shouldn't we?

Yes, and the add-side-effect should not rely on _content views'_ state at all, the `addWidget` logic should be moved to this view instead.

lib/assets/javascripts/cartodb3/editor/editor-view.js
```diff
@@ -24,21 +28,47 @@ module.exports = cdb.core.View.extend({
...
+      case 'elements':
+        // TODO: trigger element creation
+        break;
+    }
+  },
+
   render: function () {
     var self = this;
 
+    this._editorWidgetsView = new EditorWidgetsView({
```
viddo: Related to my [prev comment](https://github.com/CartoDB/cartodb/pull/6972/files#r58026109), these views should not need to be statically defined here.

javierarce: :ok_hand:

lib/assets/javascripts/cartodb3/editor/editor-view.js
javierarce: :ok_hand:

lib/assets/javascripts/cartodb3/editor/tab-pane.tpl
```diff
@@ -1,2 +1,9 @@
-<ul class="CDB-NavMenu-Inner CDB-Text is-semibold CDB-Size-medium js-menu"></ul>
+<nav class="CDB-NavMenu">
```
xavijam: Add button shouldn't be added by default in the template.

xavijam: Add button shouldn't be added by default in the template.

xavijam: I'd try to keep this file as easy as possible. If we need a different template for another purpose, let's create a new one and pass it as a new argument.

viddo:
Now that the add-button is added to the tab-pane context, shouldn't the add-button in the widgets listing be removed? AFAIK that was just a temporary thing until this was to be done.

javierarce:
You are right, @viddo, I'll get rid of those buttons.

lib/assets/javascripts/cartodb3/editor/tab-pane.tpl
javierarce: But this one is just that, isn't it? It's a copy of `lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane.tpl` with the button and the classes for this particular view.

xavijam: Ahhh, sorry, I saw the name and I mixed up both. Could we rename it? :sweat_smile:

xavijam:
BTW, I've realized that button should appear only when there is any element already added to the current context, right @piensaenpixel?

piensaenpixel:
Yes @xavijam

viddo:
What I was referring to was the [temporary add-widget-button](https://github.com/CartoDB/cartodb/blob/add-button/lib/assets/javascripts/cartodb3/editor/widgets/widgets-view.js#L71-L81), not only the layer one

javierarce:
@viddo: I'm afraid that method is called from the other button
 
screen shot 2016-03-31 at 17 22 55

javierarce:
items

lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane-view.js
```diff
@@ -42,6 +51,7 @@ module.exports = cdb.core.View.extend({
         this.removeView(this._selectedView);
       }
 
+      this.trigger('changeTab', model, this);
```
xavijam: Why do we need this?

lib/assets/javascripts/cartodb3/editor/editor-view.js
```diff
@@ -100,9 +111,82 @@ module.exports = cdb.core.View.extend({
...
+  _showAddButton: function () {
+    this.$('.js-add').removeClass('is-hidden');
+  },
+
+  _updateAddButtonState: function () {
+    this._hideAddButton();
+
+    switch (this._mapTabPaneView.getSelectedTabPaneName()) {
+      case 'widgets':
+        if (this._widgetDefinitionsCollection.length) {
```
xavijam: If it is a collection, this .lenght thing shouldn't work. We should use .size() instead.

lib/assets/javascripts/cartodb3/editor/editor-view.js
```diff
@@ -100,9 +111,82 @@ module.exports = cdb.core.View.extend({
...
+          this._showAddButton();
+        }
+        break;
+      case 'layers':
+        if (this._layerDefinitionsCollection.length) {
```
xavijam: ^^^

xavijam:
LGTM :ship: :it: after reviewing my last comment about .length :palm_tree:

lib/assets/javascripts/cartodb3/components/tab-pane/tab-pane-view.js
javierarce: To know that the user changed the tab and toggle the visibility of the add button.

xavijam: we could know it when tabs collection changed the selected item.

javierarce:
Merging this since the frontend builds are :green_apple:

xavijam:
:+1:

PR #6985, opened 2016-03-29
Upgrade SQL extension version to 0.14.4

jgoizueta:
The upgraded versions is a bugfix to support Overviews


jgoizueta:
Can you check this? @ethervoid / @iriberri

iriberri:
:+1:

PR #6986, opened 2016-03-29
Fixed some problems with tests due date issues

ethervoid:


rafatower:
:+1: this seems to be a blocker for other teams to pass their tests so please merge and let them know that they should probably update their branches with master

PR #6984, opened 2016-03-29
Fix drag

piensaenpixel:
@javierarce  could you take a look?

drag

PR #6990, opened 2016-03-29
add drag icon

piensaenpixel:
Move https://github.com/CartoDB/cartodb/pull/6984
cc @javierarce


javierarce:
Should we make the whole item draggable or just use the handler instead?

piensaenpixel:
Really the icon is only information

javierarce:
Which means… :)

piensaenpixel:
hahaha, that the whole item :) not the icon

javierarce:
Ok, :+1: then :)

Issue #6999, opened 2016-03-30
Check rollbar reporting failures


javitonino:

javitonino:
Fixed by https://github.com/CartoDB/cartodb/pull/7075

PR #7000, opened 2016-03-30
Rollbar failsafe

javitonino:
Fixes some failures on Rollbar reporting that fail due to including too much information (>128kb) or I/O objects (including the request object) in the additional data.

Fixes part of #6999 

CR @gfiorav

### DIFF
--- a/app/controllers/admin/organization_users_controller.rb
+++ b/app/controllers/admin/organization_users_controller.rb
@@ -63,7 +63,7 @@ class Admin::OrganizationUsersController < Admin::AdminController
     @user.organization.notify_if_seat_limit_reached
     redirect_to CartoDB.url(self, 'organization', {}, current_user), flash: { success: "New user created successfully" }
   rescue Carto::UnprocesableEntityError => e
-    CartoDB.report_exception(e, "Validation error", request: request, user: current_user)
+    CartoDB::Logger.error(exception: e, message: "Validation error")
     set_flash_flags
     flash.now[:error] = e.user_message
     render 'new', status: 422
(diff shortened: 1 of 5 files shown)


gfiorav:
Looks good but please explain:

Did `caller` really get that long? More than `128K`? What other things are going to be trimmed?

javitonino:
I have no hard data on stack traces, as when the failsafe is triggered, no additional info is reported. I capped the stacked as the lower parts of it are all Rails and it is the only change that I can think of that adds additional information, thus causing the report to fail.

The stack can get pretty big, I have been experimenting a bit, and I've seen stacks up to ~20k (they include the entire path for each file). Capping to 25 frames, should keep the stack under 4k.

gfiorav:
How about parsing the caller? For example: only calls related to project (routes that match app/...) and replacing others with only top and then (...). For example:

```
app/models/table.rb: 'method missing'
bin/gem/xxx/extractor.rb: 'parse arguments'
(...)
app/controllers/table_controller.rb: 'update'
services/importer/xxx....
```

I think it might be safer than cutting at 25, sometimes there important project related info "after" a lot of internal gem trace.

What do you think?

javitonino:
I wanted to push a quick fix to see if it would solve anything (I'm not sure if the problem is the stack). I think I still prefer to push this and see if it fixes anything. 25 frames of stack is still more than we had before.

You may be right that if I don't do this the good way now, it may stuck as it is. But I wanted to avoid the complexity of parsing the stack (e.g: removing prefix path or non-project frames) for what should be simple: error reporting.

gfiorav:
If 25 is more than what we had before :package: it!

I wanted to make sure no nasty bug was going to catch us with a short stack trace :)

:+1:

javitonino:
It was 0 before :P The last rollbar patches added the full stack to all reports that did not explictly add it, and that's when the reporting errors started.

gfiorav:
As long as we don't loose backtrace info we had before, fin with me :stuck_out_tongue:

Issue #6967, opened 2016-03-28
Migrate "add layer" dialog


xavijam:
## 1st part [COMPLETE]
- [x] Migrate the add-layer-view.
- [x] Migrate necessary models + specs.
  - [x] Add layer model + specs.
  - [x] User model (removed unnecessary stuff, like group models, layers stuff?).
  - [x] Organization model. 
  - [x] Visualizations fetch model.
  - [x] ~~Table model (CartoDBTableMetadata).~~ => already migrated :tada:
  - [x] Map model => Using visMap from `layerDefinitionsCollection`.
  - [x] Tables collection.
  - [x] Upload model.
  - [x] Add necessary dependencies:
      - ~~backbone-model-file-upload~~.
      - ~~moment~~ => already as npm module :tada:.
      - ~~tipsy~~.
  - [x] Think about what to do with `cdb.config`. => Working as It used to.
- [x] Migrate footer views + specs + i18n + styles.
- [x] Migrate navigation view + specs + i18n + styles.

## 2nd part
- [x] Migrate content views.
  - [x] Datasets views.
  - [x] Pagination views.
  - [x] Likes component.
  - [x] Permission model.
  - [x] Pagination component.
- [x] Add new CSS.
- [x] Make tabpane work without menu (setActive).
- [x] Start using localization templates.
- [x] Migrate specs.
- [x] Visualization table model.
- [x] Visualizations collection.

## 3rd part
- [x] Add layer to current layers collection (there are some methods that we will need to implement, e.g. `this._visMap.addCartodbLayerFromTable(…)`).
- [x] Populate all cdb config keys, as in: `cdb.config.get('oauth_gdrive')`.
- [x] Keep using `cdb.god` as we have, create ticket to implement it in another way.
- [x] Import views.
- [x] ~Migrate import views specs~. -> Issue created https://github.com/CartoDB/cartodb/issues/7140.
- [x] Start using localization strings within import views.

### Nice to have:
- [x] Metrics within add-layer-modal. -> issue created -> 
- [x] Missing specs: -> issue created -> https://github.com/CartoDB/cartodb/issues/7140
  - datasets-paginator-view
  - add-layer-view
  - dataset-item-view
  - remote-dataset-item-view
  - datasets-list-view
  - datasets-view
  - dialog-footer view

xavijam:
@javierarce also wants to know how to manage `cdb.god` in the new editor, let's discuss it @viddo @javierarce.

[...]

PR #7003, opened 2016-03-30
Migrating add-layer-modal, part 1

xavijam:
screen shot 2016-03-30 at 14 48 19

In short:

- [x] Migrate the add-layer-view.
- [x] Migrate necessary models + specs.
  - [x] Add layer model + specs.
  - [x] User model (removed unnecessary stuff, like group models, layers stuff?).
  - [x] Organization model. 
  - [x] Visualizations fetch model.
  - [x] ~~Table model (CartoDBTableMetadata).~~ => already migrated :tada:
  - [x] Map model => Using visMap from `layerDefinitionsCollection`.
  - [x] Tables collection.
  - [x] Upload model.
  - [x] Add necessary dependencies:
      - ~~backbone-model-file-upload~~.
      - ~~moment~~ => already as npm module :tada:.
      - ~~tipsy~~.
  - [x] Think about what to do with `cdb.config`. => Working as It used to.
- [x] Migrate footer views + specs + i18n + styles.
- [x] Migrate navigation view + specs + i18n + styles.

This information comes from #6967.

Notes:
- Task has been basically migration, just some important parts changes, commented already in the code.
- Internationalization is great, some problems from the previous dialog have been solved with that, like pluralize helper.
- Console logs doesn't represent anything, they are personal reminders although there are sub-issues created for them.

Could you take a look @viddo @javierarce?


lib/assets/javascripts/cartodb3/data/tables-collection.js
```diff
@@ -0,0 +1,78 @@
...
+    opts.data = _.extend({}, DEFAULT_FETCH_OPTIONS, opts.data);
+    return Backbone.Collection.prototype.fetch.call(this, opts);
+  },
+
+  getTotalStat: function (attribute) {
+    return this._stats[attribute] || 0;
+  },
+
+  parse: function (res) {
+    this._stats = _.omit(res, 'visualizations');
```
xavijam: Not sure about this change, but in order to have stats about user visualizations I did that.

lib/assets/javascripts/cartodb3/data/tables-collection.js
```diff
@@ -0,0 +1,78 @@
...
+        id: dt.id,
+        geometry_types: dt.geometry_types,
+        name: dt.name,
+        privacy: dt.privacy,
+
+        // From embedded table with same value, but different key
+        // so re-map them to match what's returned from a GET /tables/t-id
+        rows_counted: dt.row_count,
+        table_size: dt.size,
+        table_type: d.type // table or remote
```
xavijam: !Important: type of the table could be different, in this case we need to store if it is a `table` or a `remote` table for later (list rendering, behaviours,...).

lib/assets/javascripts/cartodb3/data/upload-model.js
```diff
@@ -0,0 +1,259 @@
...
+    if (!opts.configModel) throw new Error('configModel is required');
+
+    this._userModel = opts.userModel;
+    this._configModel = opts.configModel;
+    this._initBinds();
+    // We need to validate entry attributes
+    this._validate(this.attributes, { validate: true });
+  },
+
+  setUpload: function (d) {
```
xavijam: !important: this is new because new Backbone version doesn't validate by default `set` operations (it does for `save` operations :disappointed: ), so I decided to create this custom method in order to "set a new upload" wherever instead of `uploadModel.set('whatever', 'whatever', { validate: true });`.

lib/assets/javascripts/cartodb3/components/modals/add-layer/add-layer-model.js
```diff
@@ -0,0 +1,220 @@
...
+    //   success: function () {
+    //     // layers need to be saved because the order may changed
+    //     console.log("TODO: :scream: _visMap.layers.saveLayers");
+    //     self._visMap.layers.saveLayers();
+    //     self.trigger('addLayerDone');
+    //   },
+    //   error: function () {
+    //     self.set('contentPane', 'addLayerFailed');
+    //   }
+    // });
```
viddo: Is it worth leaving commented out code like this?

lib/assets/javascripts/cartodb3/editor/layers/layers-view.js
```diff
@@ -16,16 +18,27 @@ var BasemapLayerView = require('./basemap-layer-view');
  */
 module.exports = cdb.core.View.extend({
 
+  tagName: 'ul',
+  className: 'BlockList',
```
viddo: I noticed yesterday that this was removed by @piensaenpixel in #6997, is it really necessary to add again?

viddo: …although I guess it's possible this appears as a new change due to this branch not being up-to-date?

xavijam: Ouch, my bad, thanks for the warning!
we only need the events object.

lib/assets/javascripts/cartodb3/components/modals/add-layer/content/navigation-view.js
```diff
@@ -0,0 +1,234 @@
...
+
+  _onCreateEmptyClick: function () {
+    if (this._userModel.canCreateDatasets()) {
+      this._createModel.createFromScratch();
+    }
+  },
+
+  // Selection actions
+  _onSearchClick: function (e) {
+    if (e) {
```
javierarce: There's no need for this conditional, right?

lib/assets/javascripts/cartodb3/components/modals/add-layer/add-layer-model.js
xavijam: Yes, sorry for that, but it gives me an idea what was doing the dialog at that point and what we need to accomplish, although I pointed out as a sub-issue in the main issue.

lib/assets/javascripts/cartodb3/components/modals/add-layer/content/navigation-view.js
```diff
@@ -0,0 +1,234 @@
...
+      q: '',
+      tag: '',
+      shared: 'no',
+      library: this._createModel.showLibrary()
+    });
+    this.model.set('isSearchEnabled', false);
+  },
+
+  _submitSearch: function (e) {
+    if (e) {
```
javierarce: Same thing as above with the conditional.

lib/assets/javascripts/cartodb3/components/modals/add-layer/add-layer-model.js
xavijam: Also, I did the same in the tests. There won't be any commented code in the "final release".

package.json
```diff
@@ -22,6 +22,7 @@
     "aws-sdk": "2.0.0-rc11",
     "backbone": "1.2.3",
     "backbone-forms": "^0.14.0",
+    "backbone-model-file-upload": "^1.0.0",
```
viddo: shrinkwrap needs to be updated for this to be used really.
I think `npm shrinkwrap --dev` should suffice (verify the dependency being present afterwards)

xavijam: ouch, checking, thanks.

package.json
```diff
@@ -78,6 +79,9 @@
       ]
     ]
   },
+  "browser": {
```
xavijam: What do you think about this change?

lib/assets/javascripts/cartodb3/components/modals/add-layer/add-layer-view.js
```diff
@@ -0,0 +1,68 @@
...
+  render: function () {
+    this.clearSubViews();
+    this.$el.html(template());
+    this._initViews();
+  },
+
+  _initBinds: function () {
+    this._createModel.bind('addLayerDone', this.close, this);
+    this._createModel.bind('change:contentPane', this._onChangeContentView, this);
+    // cdb.god.bind('importByUploadData', this.close, this);
```
javierarce: We can get rid of this comment, right?

package.json
xavijam: BTW, I decided to installed because we have our `backbone-model-file-upload` customized with our dependencies, so better to not touch that one and start using the npm one.

lib/assets/javascripts/cartodb3/components/modals/add-layer/add-layer-view.js
xavijam: Not really, well, yes, but I'd like to keep them because they remind me what was doing as I pointed to @viddo too. Also, those comments represent a sub-item in the main issue, they are tracked of course.

lib/assets/javascripts/cartodb3/components/modals/add-layer/footer/guessing-toggler-view.js
```diff
@@ -0,0 +1,48 @@
+var cdb = require('cartodb-deep-insights.js');
+var template = require('./guessing-toggler.tpl');
+
+/**
+ * Manages if upcoming import should guessing or not.
```
javierarce: "Manages if upcoming import should guess or not."

lib/assets/javascripts/cartodb3/components/modals/add-layer/footer/privacy-toggler-view.js
```diff
@@ -0,0 +1,83 @@
...
+  },
+
+  render: function () {
+    this.clearSubViews();
+    this.$el.empty();
+
+    if (this._createModel.showPrivacyToggler()) {
+      var canChangePrivacy = this._userModel.canCreatePrivateDatasets();
+      var privacy = this.model.get('privacy');
+      var nexPrivacy = privacy === 'PUBLIC' ? 'PRIVATE' : 'PUBLIC';
```
javierarce: `nexPrivacy`… it should be `nextPrivacy`, right?

Also, `PUBLIC` and `PRIVATE` strings are repeated in this file… maybe they could be extracted as a global variables for this view.

lib/assets/javascripts/cartodb3/data/upload-model.js
viddo: why not just use the set with the `{validation: true} wherever it's necessary instead?

package.json
viddo: As long as the tipsy thing is compatible with the new jQuery version(?) it should be fine

lib/assets/javascripts/cartodb3/data/upload-model.js
xavijam: Not sure about that, because that method is used outside this model and I felt it was a good moment to create a public method and obey it to be validated.

lib/assets/javascripts/cartodb3/components/modals/add-layer/footer/privacy-toggler-view.js
xavijam: :disappointed_relieved: thanks!

viddo:
Some minor comments, since I guess the code is mostly migrated from existing code not much to say. :+1:

lib/assets/javascripts/cartodb3/data/upload-model.js
viddo: Either way `set` still exists, which means that if someone uses that instead you still have the same problem anway. If the intention is to enforce validation to be done I think it's better to override the default set behavior to always set this flag in this case.

package.json
xavijam: :+1:

package.json
xavijam: It was working so, yay!

xavijam:
Frontend tests passed, merging with analysis.

PR #7004, opened 2016-03-31
Fix widgets ui

piensaenpixel:
image


javierarce:
I think this could have been extracted and merged to master (since there's not directly related to analysis), but anyways, it's :+1:

PR #7021, opened 2016-04-04
Remove overquota calculation because we don't charge for it anymore

ethervoid:
Now we don't charge by mapviews anymore so we have to remove this slow operation to improve the response time


ethervoid:
@rafatower check it out

rafatower:
:+1:

Issue #6950, opened 2016-03-22
Remaining work on add-analysis modal


viddo:
Postponed stuff from #6886, that were considered nice-to-have at the moment. Most of them have impediments that needs to be discussed and resolved

- [x] Options groups in UI; _"data transformation", "relationship analysis", etc._ #7626 
- [x] ~~Collapse/expand option groups in UI~~ #7628
- [x] ~~Search~~ #7629

viddo:
Closing since the things originally listed here were either resolved, or have separate issues created to be done later (considered "easy")

PR #7031, opened 2016-04-04
6886 point in polygon

### LINKED ISSUE #6886: Analysis modal
User to be able to:
- add analysis from layers view, i.e. always add on top of an existing analysis head
- only show analysis that are compatible with input node as source

---

- base #6960
  - [x] base boilerplate etc.
  - [x] buffer option
  - [x] trade area option 
- [x] add analysis from layer content view #6966
- [x] disable options that are incompatible with input source's output geometry  
- [x] add point-in-polygon type #7031

viddo:
Closes #6886 (analysis modal), adds the point-on-polygon node type*

- will create a failing issue initially, showing the user the analysis-form and indicating the source input that needs to be fixed. 
- user can select (another) layer as input source, or a raw table (creates a new source node)

Noteworthy changes:
- removed the `_id` suffix on source attributes, to remove some unnecessary code that doesn't really add any value
- the tables+layer sources available varies depending on which source is considered secondary, can be either points or source depending on prev node, for which an additional editor-only param is set (`primary_source_name`) to indicate which one it is. the composite analysis node view needed to be changed to take this into account
- since both tables and layer's source geometries need to be fetched async, I'm using d3-queue to coordinate the fetch state, to only render options once all fetches returned… ([`queue-async`](https://github.com/CartoDB/cartodb/blob/45410bc06e5023772d622f4cd3aee4c3cac6de89/package.json#L63) wḣich we already have was renamed to d3-queue, but considering the changes are quite big I didn't want to touch that one).
- don't wait for server-side sync to create models, to avoid complicating cascading side-effects (might need to revisit this later)

This also depends on https://github.com/CartoDB/cartodb.js/issues/1216, since the initial node is created in incomplete state, and selecting the secondary source required to update the existing node. cc @alonsogarciapablo 

@xavijam can you review?
cc @javierarce who I pair programmed with on the the first ~half of the commits


viddo:
Encountered issues that needs further review was added to the existing issue #6950, to not loose track of them.

lib/assets/javascripts/cartodb3/data/camshaft-reference.js
```diff
@@ -81,6 +81,11 @@ module.exports = {
       case 'trade-area':
         outputGeometry = 'polygon';
         break;
+      case 'point-in-polygon':
+        // TODO but it's still the camshaft reference should yield output geometry based on definition params,
+        //      this param is still missing though so for now always yields point
+        outputGeometry = 'point';
```
xavijam: :+1:

xavijam: :+1:

xavijam: :+1:

lib/assets/javascripts/cartodb3/data/get-simple-geometry-type.js
```diff
@@ -0,0 +1,16 @@
+var MAP = {
```
xavijam: I could use this "class" within table-model too.

viddo:
retest this please

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses-form-types/trade-area/analysis-trade-area-form.tpl
```diff
@@ -8,25 +8,17 @@
...
         <p class="CDB-Legend u-upperCase u-iBlock CDB-Text is-semibold CDB-Size-small u-rSpace--m"><%- _t('editor.layers.analysis-form.input') %></p>
-        <div class="CDB-Text CDB-Size-medium u-iBlock" data-editors="source_id"></div>
+        <div class="CDB-Text CDB-Size-medium u-iBlock" data-editors="source"></div>
       </div>
     </div>
   </div>
   <div class="CDB-HeaderInfo">
     <div class="CDB-HeaderNumeration CDB-Text is-semibold u-rSpace--m">3</div>
-    <div class="CDB-HeaderInfo-Inner CDB-Text">
+    <div class="CDB-HeaderInfo-Inner CDB-Text" data-fields="kind,time">
```
xavijam: oh!

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses-form-types/trade-area/analysis-trade-area-form-model.js
```diff
@@ -1,19 +1,18 @@
...
+  initialize: function () {
+    BaseAnalysisFormModel.prototype.initialize.apply(this, arguments);
 
-  initialize: function (attrs, opts) {
     this.schema = {
-      source_id: {
+      source: {
         type: 'Select',
-        options: [ this.get('source_id') ]
+        options: [ this.get('source') ]
```
xavijam: Shouldn't it be disabled?

xavijam:
No much to add, If you (both) decided the best way to implement it, I'm 100% with you.

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses-form-types/trade-area/analysis-trade-area-form-model.js
viddo: I only changed this name due to the prop changes in the node model. No problem adding the disabled state, but why wasn't it added when the model was first created if it should have had that attribute from the beginning?

viddo:
@javierarce any comments?

btw, FYI, the node analysis doesn't work yet, since cartodb.js doesn't support changing source(s) of a node just yet

xavijam:
ticket for that in cartodb.js? cc @alonsogarciapablo

javierarce:
For me this looks good :+1:

viddo:
Scratch that, we're using the ``analyse(…)` on changes, so it should work, will investigate further…

viddo:
@xavijam @alonsogarciapablo with the last commits (updating deps) the map is update on changing nodes. :tada: 

BTW, went with the simplest solution to analyses persistence for now: save+analyse all analyses on a node change, we can always optimize it later

viddo:

xavijam:
:clap:

alonsogarciapablo:
:clap: :clap:

PR #7039, opened 2016-04-05
Cache overquota users and do it in background

ethervoid:
The ::User.overquota method is a heavy proccess due to the need
of iterate all the users checking if he/she is near of his/her quota.
Right now we are doing it without Resque, we are moving it to a Resque
process.

Also other processes use this method but most of the only need it daily
so we have created a new method to cache it daily in order to be able
to execute it one time by, for example, a rake task


ethervoid:
@rafatower check this out

rafatower:
:+1:

PR #7042, opened 2016-04-05
1683 Execute with timeout & direct connections to user db - take 2

iriberri:
https://github.com/CartoDB/cartodb/pull/6975

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -518,6 +518,9 @@ class User < Sequel::Model
     self.database_host
   end
 
+  # Obtain a db connection through the default port. Allows to set a statement_timeout
+  # which is only effective in case the connection does not use PGBouncer or any other
+  # PostgreSQL transaction-level connection pool which might not persist connection variables.
   def in_database(options = {}, &block)
     if options[:statement_timeout]
       in_database.run("SET statement_timeout TO #{options[:statement_timeout]}")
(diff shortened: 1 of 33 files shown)


iriberri:
@rafatower can you review this commit? https://github.com/CartoDB/cartodb/pull/7042/commits/65303ea80398a1c253863b6b4ec0664e933be6e9

app/models/user/db_service.rb
```diff
@@ -435,7 +435,7 @@ def in_database_direct_connection(statement_timeout:)
         raise 'need block' unless block_given?
 
         configuration = db_configuration_for
-        configuration[:port] = configuration.fetch(:direct_port, configuration["direct_port"])
+        configuration[:port] = configuration.fetch(:direct_port, configuration["direct_port"]) || configuration[:port]
```
rafatower:
```
configuration[:port] = configuration.fetch(:direct_port, configuration["direct_port"]) || configuration[:port] || configuration["port"]
```

app/models/user.rb
```diff
@@ -600,6 +592,19 @@ def transaction_with_timeout(options)
     end
   end
 
+  def get_connection(options = {}, configuration)
+  connection = $pool.fetch(configuration) do
+      db = get_database(_opts = {}, configuration)
```
javitonino: Hi @iriberri 

I was reviewing this PR because it was making some tests fail in master CI (fixed already :smile:) and I noticed this.

This is passing always an empty `options` to `get_database`, but before it was passing the `options` hash. Is there any reason for this change?

I see that the `options` parameter is still being used in [`get_database`](https://github.com/CartoDB/cartodb/pull/7042/files#diff-4676c008b11a5480d73d4a6de01e45b9R618) to determine if it is necessary set the search path. Can you double check, please?

rafatower: most likely a bug and we did not notice. Thanks!

@iriberri can you please take care of this?

iriberri: Sure, I'm on it.

Thanks @javitonino!

PR #7048, opened 2016-04-05
Fix rake task after updating the models

javitonino:
Part of #7013 

A few fixes for the Rake task:
- Fix `Carto::Visualization.entity_id` for use in the Rake. This is a quick hack that will be deleted once we drop the `entity_*` fields from the permissions table.
- Load all data in the first query, before it performed a `count` and then an `all`, doubling the work.
- Handle the cases where the `user_id` is null. Previously it only handled the case when the user didn't exist but the uuid was valid.

### DIFF
--- a/app/models/carto/permission.rb
+++ b/app/models/carto/permission.rb
@@ -40,12 +40,14 @@ class Carto::Permission < ActiveRecord::Base
     self.owner_id == user.id
   end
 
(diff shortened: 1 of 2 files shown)


javitonino:
Can you CR this @gfiorav?

gfiorav:
:package: ready!

PR #7049, opened 2016-04-05
Change deprecated PostGIS ST_Force_2D function to ST_Force2D

ethervoid:
Since PostGIS 2.1 the function `ST_Force_2D` is deprecated so since our PostGIS dependency is now 2.2 for all we have to change it to avoid warnings or future problems


ethervoid:
@jgoizueta check this out

PR #7059, opened 2016-04-06
Fixes tests for FK migration

javitonino:
Part of #7013 

Tests were failing with the migration. The main problem was the modification of the Visualizations repository in some tests, creating a new connection to the database instead of using Rails' one. The problem with that and the FK was that inserting the permission and the visualization was being done in different transactions/connections, and so, the permission was not seen by the visualization transaction.

I also removed the permission destruction from organization deletion, as they will be deleted when destroying the visualizations of the users. Destroying the permissions before the visualizations causes and integrity failure.

There are some more minor changes for specific tests.

Tests pass:
- [With migration](https://ci.cartodb.net/job/CartoDB-PR-testing/1315/)
- [Without migration](https://ci.cartodb.net/job/CartoDB-PR-testing/1329/)

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -107,21 +107,12 @@ class Organization < Sequel::Model
   # organization destroy
   def destroy_cascade
     destroy_groups
-    destroy_permissions
     destroy_non_owner_users
     if self.owner
       self.owner.destroy
     else
       self.destroy
     end
-  end
-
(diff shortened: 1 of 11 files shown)


javitonino:
Can you CR @gfiorav ?

gfiorav:
Looking good! :package: it!

PR #6970, opened 2016-03-28
Add scroll to the left-hand column

javierarce:
scroll

Closes: #6940 

CR: @xavijam

PR #7065, opened 2016-04-06
Revert "Add scroll to the left-hand column"

javierarce:
Reverts CartoDB/cartodb#6970


javierarce:
Frontend is :green_heart:, so merging

PR #7066, opened 2016-04-06
changed the way zoom is calculated to the upper zoom level instead of…

javisantana:
… the lower one

cc @rafatower


rafatower:
@ethervoid please review, I'll push as soon as tests passes.

PR #7072, opened 2016-04-07
Fixes temp point-in-polygon creation through drag and drop

javierarce:
CR: @xavijam


xavijam:
No test needed?

xavijam:
:+1:

xavijam:
As it is temporal, it is fine, thanks man. :)

PR #7073, opened 2016-04-07
Add geojson to the text/plain headers

ethervoid:
Some urls, for example raw.github.com, return geojson files with the
text/plain header and as we have it, we are going to try to process it
as a .txt file (CSV) and fails.


ethervoid:
@rafatower give me your blessing

rafatower:
:+1:

PR #7080, opened 2016-04-07
Cleanup code fk viz permissions

javitonino:
This cleans and refactors the permission code after removing the entity fields.


javitonino:
assas

javitonino:
dsadasdasd

javitonino:
retest this please

javitonino:
Please CR, @gfiorav

javitonino:
retest this please

app/controllers/carto/admin/visualization_public_map_adapter.rb
```diff
@@ -36,7 +36,7 @@ def to_vizjson(options = {})
       end
 
       def is_owner?(user)
-        @visualization.is_owner_user?(user)
+        @visualization.owner_user?(user)
```
gfiorav: Why not just `owner?(user)`? I think it sound better 😁

javitonino: Reads easier and doesn't appear to conflict with anything: changing for Carto models.

app/controllers/carto/api/tables_controller.rb
```diff
@@ -13,8 +13,8 @@ class TablesController < ::Api::ApplicationController
 
       def show
         return head(404) if table == nil
-        return head(403) unless table.table_visualization.has_permission?(current_user, Carto::Permission::ACCESS_READONLY)
-        render_jsonp(table.public_values({request:request}, current_user).merge(schema: table.schema(reload: true)))
+        return head(403) unless table.table_visualization.has_permission?(current_user, CartoDB::Permission::ACCESS_READONLY)
```
gfiorav: Why access the old model's constant? Should it be better to move stuff to the new models?

javitonino: Initially I changed this to use `has_write_permission?`. However, `table` here is using the old models which I have not changed here, so I had to use `has_permission?`.

Given that this is returning an old `CartoDB::Permission` it makes more sense to use the constant in this class. The constant is actually replicated in the new model, the old code would still work, but I think this makes it easier to realize that this is using old models, even when it is a Carto controller.

gfiorav: Divided on this to be honest. On one side, I think if it's replicated we should use the new one. On the other side, it's true that this is more coherent. 

Up to you!

gfiorav:
Made some minor comments. 📦 ready for me!

app/controllers/carto/api/tables_controller.rb
javitonino: I like leaving it as it is and making a new issue to migrate this controller to use the new models, as it should.

PR #7081, opened 2016-04-07
Riak s3

akatuluru:
Adding options to AWS.config to add s3 endpoint, s3 port and force path style urls


Cartofante:

xavijam:
There has been no activity on this issue for more several months. We are closing it. If you think this still needs to be addressed please open a new issue.

Issue #7047, opened 2016-04-05
All connections for sync operations should go through the same port


luisbosque:
For some async operations (resque) we are able to send queries to the DB through a different port.
All the queries sent through the editor (sync) are supposed to go through the same port. 
Otherwise is difficult perform some maintenance operations

/cc @rafatower

rafatower:
Please check R/B 17316 @iriberri

iriberri:
The current occurrences are coming from:

````
/app/models/table.rb:1048:in `georeference_from!
````

I'm going to move this one to a transaction too.

iriberri:
The change in table.rb#georeference_from! is already in production, so occurrences should be reduced (and, fortunately, don't exist) from now on. I'll keep an eye on it to see if there's still something we're missing.

iriberri:
![image](https://cloud.githubusercontent.com/assets/1730320/14486246/207cf55a-015c-11e6-86bc-674f00f1240a.png)

No more occurrences. :-)

iriberri:
@luisbosque I think you can close these connections now :)

PR #7084, opened 2016-04-08
#7047 Use transaction_with_timeout instead of direct conn

rafatower:
Fixing the one case we know it's being launched from rui machines.

I'll add some code to log the cases from other parts of the code and then we can clean up them systematically.

@jgoizueta can you please review?

### DIFF
--- a/app/models/user/db_service.rb
+++ b/app/models/user/db_service.rb
@@ -437,6 +437,9 @@ module CartoDB
         configuration = db_configuration_for
         configuration[:port] = configuration.fetch(:direct_port, configuration["direct_port"]) || configuration[:port] || configuration["port"]
 
+        # Temporary trace to be removed once https://github.com/CartoDB/cartodb/issues/7047 is solved
+        CartoDB.notify_debug 'Direct connection not used from queue', {backtrace: Kernel.caller} unless Socket.gethostname =~ /^que/
+
         connection = @user.get_connection(_opts = {}, configuration)
 
         begin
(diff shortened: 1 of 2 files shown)


rafatower:
retest this, please

app/models/user/db_service.rb
```diff
@@ -437,6 +437,9 @@ def in_database_direct_connection(statement_timeout:)
         configuration = db_configuration_for
         configuration[:port] = configuration.fetch(:direct_port, configuration["direct_port"]) || configuration[:port] || configuration["port"]
 
+        # Temporary trace to be removed once https://github.com/CartoDB/cartodb/issues/7047 is solved
+        CartoDB.notify_debug 'Direct connection not used from queue', {backtrace: Kernel.caller} unless Socket.gethostname =~ /^que/
```
rafatower: @ethervoid do you know if there's a way for rollbar to group this kind of trace as if it were an exception, by backtrace? I mean, have rollbar clasify occurrences instead of having to do it manually or putting all of them in the same bag.

ethervoid: We should start using the logger https://github.com/CartoDB/cartodb/blob/master/lib/cartodb/logger.rb where you can pass the exception and should group them

ethervoid:
LGTM

jgoizueta:
:+1:

app/models/user/db_service.rb
```diff
@@ -438,7 +438,7 @@ def in_database_direct_connection(statement_timeout:)
         configuration[:port] = configuration.fetch(:direct_port, configuration["direct_port"]) || configuration[:port] || configuration["port"]
 
         # Temporary trace to be removed once https://github.com/CartoDB/cartodb/issues/7047 is solved
-        CartoDB.notify_debug 'Direct connection not used from queue', {backtrace: Kernel.caller} unless Socket.gethostname =~ /^que/
+        CartoDB::Logger.warning(message: 'Direct connection not used from queue') unless Socket.gethostname =~ /^que/
```
jgoizueta: shouldn't be info or debug rather than warning?

PR #7085, opened 2016-04-08
Fix connection pool tests and drop global users from spec_helper

javitonino:
Fixes the connection pool tests by using its own users. They were broken when tests in make spec 2 started using spec_helper which creates/destroys the global users this test used.

Delete $user_* users as they were not being used and were recreated for each spec using spec_helper. Saves about 20 minutes of testing in master (much less in parallel).


javitonino:
CR @gfiorav

gfiorav:
:package: it!

PR #7097, opened 2016-04-11
fix ui common data - add layer

piensaenpixel:

image

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/modals/add-layer/content/datasets/remote-dataset-item-view.js
+++ b/lib/assets/javascripts/cartodb3/components/modals/add-layer/content/datasets/remote-dataset-item-view.js
@@ -15,7 +15,7 @@ var template = require('./remote-dataset-item.tpl');
 
 module.exports = DatasetItem.extend({
   tagName: 'li',
-  className: 'DatasetsList-item',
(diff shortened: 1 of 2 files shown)


xavijam:
:+1:

lib/assets/javascripts/cartodb3/components/modals/add-layer/content/datasets/remote-dataset-item-view.js
```diff
@@ -68,11 +68,7 @@ module.exports = DatasetItem.extend({
 
   _setItemClasses: function () {
     // Item selected?
-    this.$el.toggleClass('is--selected', !!this.model.get('selected'));
-    // Check if it is selectable
-    this.$el.toggleClass('DatasetsList-item--selectable', !!this._canImportDataset());
```
xavijam: Hey, why have you removed those things? They are important to know if the user can import or not the dataset. Those states are a must please.

xavijam:
Check my comment in the code, they should be fixed before merge please.

piensaenpixel:
fixed @xavijam you're right, sorry i deleted when change the className.

javierarce:
👍

PR #7106, opened 2016-04-11
Add fix header while scrolling

piensaenpixel:
@javierarce could you take a look, we need add .is-fixed while scrolling

image


Cartofante:

PR #7107, opened 2016-04-11
Add debug message to inspect response when error 1001 @ downloader.rb

iriberri:


iriberri:
@rafatower this is to try to get out something from https://github.com/CartoDB/cartodb-management/issues/4435#event-622440494, do you 👍?

rafatower:
Ok, I guess to be deleted when the problem is sorted out.

iriberri:
Yep!

PR #7109, opened 2016-04-12
Use transaction with timeout instead of DB direct connection in lat/long table geocoding

iriberri:
From what is being discussed at: https://github.com/CartoDB/cartodb/issues/7047

cc @rafatower

### DIFF
--- a/app/models/table.rb
+++ b/app/models/table.rb
@@ -1045,8 +1045,8 @@ class Table
     if !options[:latitude_column].blank? && !options[:longitude_column].blank?
       set_the_geom_column!('point')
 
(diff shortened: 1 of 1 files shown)


rafatower:
That call should never be made synchronously from the editor, this will be a workaround for it. I was positive there was already an issue for editor to change it. But I seem unable to find it.

iriberri:
I don't remember an issue like that either, but if you haven't created yet I can go for it.

iriberri:
This just passed CI so I'm deploying asap.

rafatower:
:+1:

iriberri:
retest this please

PR #7112, opened 2016-04-12
Run the Rake task only once on migration

javitonino:
Loading the Rakefile (already loaded as the migrations run inside a Rake) made the task run twice unnecessarily.

Also added a check so this task is not run on new database creations, as they are empty by default.


javitonino:
CR @gfiorav

gfiorav:
📦 it!

PR #7131, opened 2016-04-14
Report production errors on the console if Rollbar is disabled

javitonino:
This enables logging errors to the console even in production, if Rollbar is nor configured. Currently, there is no other way than Rollbar to see error logs in production environments, which is a PITA when debugging problems in on-premise installations.

I took the chance to clean up a little the generated console trace so it is easier to read.

### DIFF
--- a/config/app_config.yml.sample
+++ b/config/app_config.yml.sample
@@ -58,8 +58,7 @@ defaults: &defaults
     primary:          ''
     embeds:           ''
     domain:           ''
-  rollbar:
-    token: ''
+  rollbar_api_key: ''
   tumblr:
     api_key: ''
   trackjs:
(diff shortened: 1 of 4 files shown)


config/app_config.yml.sample
```diff
@@ -58,8 +58,7 @@ defaults: &defaults
     primary:          ''
     embeds:           ''
     domain:           ''
-  rollbar:
-    token: ''
+  rollbar_api_key: ''
```
javitonino: `rollbar->token` is unused, the actual configuration value is `rollbar_api_key`.

javitonino:
CR @gfiorav

gfiorav:
📦 it!

PR #7135, opened 2016-04-14
Correctly check for uniqueness in related_tables

### LINKED ISSUE #7132: Visualization with two layers of one dataset is not deleted correctly
To reproduce:
 1. Create a dataset
 2. Create a map from the dataset
 3. Add a second layer from the dataset (the map has two layers from one dataset)
 4. Delete the dataset

The visualization will not be deleted, and no error will be shown.  The visualization does not work as the dataset is deleted.

In some on-premise (1.2.1 with modifications) machines, a 500 error is generated when trying to delete the dataset and it remains as a ghost table (UserTable exists, physical table doesn't). This error does not replicate in current master, but the visualization is not deleted anyway.

javitonino:
Fixes #7132 

The `related_tables` check for uniqueness was broken, so if a visualization has two layers to the same UserTable, this returns 2 instead of 1. This breaks checks for visualization dependence, causing #7132.


javitonino:
CR please @gfiorav

gfiorav:
📦 ready!

PR #7149, opened 2016-04-15
7119 Add exception to report download error after a 200 response

iriberri:
We've seen some scenarios in which the HTTP code may be 200 OK but then the download seems to be interrupted (for example caused by a connection issue).

In a previous PR we added a Logger trace in this scenario in order to be able to retrieve more data when this happens, but still it as reported to the user as a Download Error (typically related with 4XX/5XX codes). This new exception takes into account the successful code and returns a "temporal issue, try again" message so that the user can know that the file is available and the URL is okay but there was a problem in the download.

Closes https://github.com/CartoDB/cartodb/issues/7119

### DIFF
--- a/services/importer/lib/importer/downloader.rb
+++ b/services/importer/lib/importer/downloader.rb
@@ -300,8 +300,12 @@ module CartoDB
           elsif error_response.code == 404
             raise NotFoundDownloadError.new(error_response.body)
           else
-            CartoDB::Logger.debug(message: "Generic download error", response: error_response.inspect)
-            raise DownloadError.new("DOWNLOAD ERROR: Code:#{error_response.code} Body:#{error_response.body}")
+            if error_response.code = 200
+              CartoDB::Logger.debug(message: "Generic download error", response: error_response.inspect)
+              raise IncompleteDownloadError.new("DOWNLOAD ERROR: Server responded with #{error_response.code} code but download failed")
+            else
+              raise DownloadError.new("DOWNLOAD ERROR: Code:#{error_response.code} Body:#{error_response.body}")
+            end
           end
         end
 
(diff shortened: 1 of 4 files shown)


iriberri:
@ethervoid could you take a look here too? You can find the history of it in the related issues, but basically this is the code in Typhoeus that marks the response as success:

````ruby
def success?
  (mock || return_code == :ok) && response_code && response_code >= 200 && response_code < 300
end
````

It seems that for some scenarios, even though the response_code is a 200, we're seeing that `success?` is false, probably due to issues in the connection itself. We're adding this new exception and restricting more the scope of the logger trace I put in a previous PR so that at least when this happens the user can know that this is kind of a "temporal issue" and the URL that they're trying to use is valid, but yet something happened there. With this restricted logger I expect to be able to get more accurate traces that could let us know what is really happening behind.

File: services/importer/lib/importer/downloader.rb [hunk 1]
```diff
@@ -299,8 +299,10 @@ def download_and_store
             raise UnauthorizedDownloadError.new(error_response.body)
           elsif error_response.code == 404
             raise NotFoundDownloadError.new(error_response.body)
+          elsif error_response.code == 200
```
ethervoid: Check also the return_code == :partial_file (http://www.rubydoc.info/github/typhoeus/ethon/Ethon/Easy:return_code)

iriberri:
retest this please

iriberri:
@ethervoid I applied what we talked about, could you check? :)

services/importer/lib/importer/downloader.rb
```diff
@@ -299,8 +299,9 @@ def download_and_store
             raise UnauthorizedDownloadError.new(error_response.body)
           elsif error_response.code == 404
             raise NotFoundDownloadError.new(error_response.body)
+          elsif error_response.return_code == :partial_file
```
ethervoid: You should make a test for this part. You've to mock the typhoeus response and check if this part works. I don't think that :partial_file symbol exists in the downloader

iriberri: Just pushed. I had to "mock the :mock", otherwise it would never be checked because of `(mock || return_code == :ok)`.

ethervoid:
LGTM

PR #7153, opened 2016-04-18
Use service_item_id or table_name in mail_notifier 

iriberri:
Fixes https://github.com/CartoDB/cartodb/issues/6296

The mail notifier was always trying to retrieve `service_item_id` to get the name of the table being imported, but tables created from query or table duplicates do not have this value and the table name is instead being obtained from `table_name`.


iriberri:
Hey @ethervoid, could you give me your thoughts here, please? :-)

ethervoid:
LGTM

PR #7163, opened 2016-04-18
6983 remove url options

javitonino:
Drops `url_options` from visualization. This was replaced by the overlays system two years ago in https://github.com/CartoDB/cartodb/pull/672.

No reason to keep this around, and eliminates unnecessary cruft from the new import/export functionality. I have not bumped version there as import/export is not yet finished/accessible by the user, and anyways, importing old exports with extra fields will still work.

Migration in #7164


javitonino:
Can you CR @gfiorav? Also check out the migration (#7164) which will be applied with this.

gfiorav:
📦 , 📦 , 📦 !

javitonino:
Uhm, definitely yesterday wasn't my best day. I forgot to remove the column from the AR model (Carto::Visualization), and this would have caused the visualization listing to stop working in the space of time between migration and restarting the Rails server.

Can you check the latest changes @gfiorav?

gfiorav:
📦 it!

PR #7168, opened 2016-04-18
Use cartodb_id as FID in ogr2ogr and handle exceptions

iriberri:
ogr2ogr2 uses by default a primary key called `ogc_fid` (integer, non-null, with a sequence as default value). This is exactly what we do with `cartodb_id`, which we are later generating/converting/checking in the cartodbfication.

By using `cartodb_id` as the name of the ogr2ogr primary key we directly obtain the column. Also, it recognizes `cartodb_id` columns from the imported file, and has exactly the same constraints we have on it: it detects null, repeated and non integer values. In case the `cartodb_id` is incorrect, the import process would fail straight away instead of going through the whole process and failing when CartoDBfying, as it does now.

### DIFF
--- a/app/models/data_import.rb
+++ b/app/models/data_import.rb
@@ -212,6 +212,12 @@ class DataImport < Sequel::Model
     CartoDB::notify_warning_exception(invalid_cartodb_id_exception)
     handle_failure(invalid_cartodb_id_exception)
     self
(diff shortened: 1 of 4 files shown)


Cartofante:

iriberri:
I'm gonna close this for the moment until we decide to keep the work on this.

PR #7176, opened 2016-04-19
geocodings calls returns 0 for google geocoder users

ethervoid:


ethervoid:
@rafatower check this temporary fix

rafatower:
mmm ok for the moment, but maybe to be checked in central rather than here?

ethervoid:
Yes is a wip PR but we need more time to fulfill correctly

PR #7187, opened 2016-04-19
Reenable tests

javitonino:
This reenables a bunch of tests that were disabled because they were missing from the Makefile. Most of them work with minor or no changes at all.

There are still some disabled tests that are old, using minitest, and probably not worth to fix right now. I have deleted the ones that obviously don't do anything, but left the rest in, just in case.


javitonino:
retest this please

javitonino:
retest this please

javitonino:
retest this please

javitonino:
Also should fix tests in master.

Please CR @gfiorav

gfiorav:
Looks 📦 

Have you checked why they were removed?

javitonino:
Most of them were never enabled in the first place, so probably, forgot to add to the Makefile. Some of them were added as a directory, which works for `make check` but not for parallel execution. Didn't found any test that was purposely disabled, but it is hard to find something in the Makefile log.

gfiorav:
Ok, I guess 🍏 is 🍏 

📦 it when tests pass!

PR #7197, opened 2016-04-20
Observatory snapshot integration

ethervoid:
- [x] Metrics for observatory snapshot: overquota, number of uses, remanining quota and so on
- [x] Quota for observatory snapshot service
- [x] Soft limit for observatory snapshot
- [x] Tests for the observatory snapshot part


javisantana:
as a comment, I see there is a big "data observatory" service. It's **perfectly fine** for now but inside DO we could have different set of services being charged in a different way in the future.

ethervoid:
Yes, I have it in mind and is going to be a pain in the ass change it in
the future, but I don't know how to name this service to differentiate from
the others that we don't know what are going to do.

Any idea? Maybe we could call this do_augment or do_snapshot... @rafatower
El 20/4/2016 19:33, "javi santana" <notifications@github.com> escribió:

as a comment, I see there is a big "data observatory" service. It's *perfectly
fine* for now but inside DO we could have different set of services being
charged in a different way in the future.

—
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub
<https://github.com/CartoDB/cartodb/pull/7197#issuecomment-212527964>

vendor/assets/javascripts/cartodb.mod.odyssey.uncompressed.js
```diff
@@ -1,5 +1,5 @@
-// version: 3.15.8
```
rafatower: @ethervoid I think this is a mistake to correct

ethervoid:
@rafatower final review and deploy

jgoizueta:
:+1:

PR #7206, opened 2016-04-21
Pass down update_metadata parameter in import_user for orgs

zenitraM:
Fixes issue where orgs were not moved in dry run

cc @azamorano


azamorano:
:+1:

PR #7225, opened 2016-04-25
Require rspec_configuration as in executor, avoids mocha errors

javitonino:
This fixes tests in master. Basically, this is the same problem that #dataservices had some weeks ago when adding tests, but this time is happening in master CI instead of PR.

Basically, just passing `rspec_configuration` with the require flag to rspec does the right thing (loads this as the first file). Before, it was added as a normal test and was loaded in alphabetical order, which failed for some test groups when adding new tests. This should be safer when adding new tests.

Also disables `log_spec` because it fails sometimes as some of the assertions are based on timestamps, which is not such a good idea.

Tests passing for PR and master testing, hopefully this stabilizes the situation for now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -275,17 +275,17 @@ endif
 
 # TODO: Ongoing removal of groups, that's the reason of holes in numbering
 check-1:
-	CHECK_SPEC=1 RAILS_ENV=test bundle exec rspec --require spec/rspec_configuration.rb $(WORKING_SPECS_1)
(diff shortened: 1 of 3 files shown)


javitonino:
retest this please

javitonino:
retest this, please

javitonino:
Waiting for tests to pass for on-premise (had a problem with the test-slave configuration), but can you review meanwhile @gfiorav? This is just a couple of fixes after the latest testing PR to make everything work in master CI.

gfiorav:
📦 !

PR #7227, opened 2016-04-25
Ghost tables manager won't drop any user tables that are part of maps

gfiorav:
Please CR, @juanignaciosl !


gfiorav:
On a second thought, this is to defensive. We're better off without this.

PR #7228, opened 2016-04-25
public pages: update navigation

matallo:
updates navigation and footer on public pages as per commercial site

CR @javierarce


javierarce:
👍  🎉

PR #7232, opened 2016-04-26
Fix cartodb.js and DI.js dependencies

rochoa:
Key changes:

 - Updates deep-insights.js dependency with no public available cartodb.js API
 - Replaces all cartodb-deep-insights.js requires with cartodb.js ones when apply.
 - Uses DI.js just for dashboard creation.
 - Browserify resolutions is explicitly declared.
 - Dependencies are declared splitting dev from library ones.


Cartofante:

xavijam:
:+1:

PR #7255, opened 2016-04-26
7016 migrate background importer

javierarce:
The previous PR was against the now defunct `analysis` branch


Cartofante:

PR #7259, opened 2016-04-27
Getting tipsy element size with getBoundingClientRect

### LINKED ISSUE #7256: Mapviews tooltip is not rendered in the correct position

xavijam:
There is a problem right now with the map views tooltip, and it is due to the calculation of the size of the element. We use within tipsy:

https://github.com/CartoDB/cartodb/blob/master/vendor/assets/javascripts/jquery.tipsy.js#L48

But for any reason, over svg elements, that function doesn't work properly (it returns 0 for width and height). Using `getboundingClientRect` fixes the bug.

Tested in several browsers (firefox, safari, chrome) and in several places.

#### Before:

screen shot 2016-04-27 at 10 30 59

#### After:

screen shot 2016-04-27 at 10 28 23
screen shot 2016-04-27 at 10 28 01
screen shot 2016-04-27 at 10 27 50

CR: @javierarce 
cc @saleiva 
Fixes #7256.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cartodb-ui",
-  "version": "3.23.65",
+  "version": "3.23.66",
   "description": "CartoDB UI frontend",
   "repository": {
     "type": "git",
--- a/vendor/assets/javascripts/jquery.tipsy.js
+++ b/vendor/assets/javascripts/jquery.tipsy.js
@@ -45,8 +45,8 @@
 
             //  Fixed by Arce
             var pos = $.extend({}, this.$element.offset(), {
-              width: this.$element[0].offsetWidth,
-              height: this.$element[0].offsetHeight
+              width: this.$element[0].getBoundingClientRect().width,
+              height: this.$element[0].getBoundingClientRect().height
             });
 
             var actualWidth = $tip[0].offsetWidth,


xavijam:
Checked in ded02 that the fix is working properly :).

vendor/assets/javascripts/jquery.tipsy.js
```diff
@@ -45,8 +45,8 @@
 
             //  Fixed by Arce
```
javierarce: Could we get rid of this incriminating evidence? :P

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cartodb-ui",
-  "version": "3.23.65",
+  "version": "3.23.66",
```
javierarce: This is now the current version of the front, so it should probably be bumped.

javierarce:
👍  +2 minor comments

package.json
xavijam: ok

vendor/assets/javascripts/jquery.tipsy.js
xavijam: :joy:

xavijam:
Thanks!

xavijam:
Passing frontend tests, merging! :)

PR #7261, opened 2016-04-27
Don't persist source nodes

viddo:
Source nodes are created if needed when the layer definitions models
are instantiated (see layer-definitions-collection), but should not be
persisted unless there is an actual analysis.

This is necessary to avoid cartodb.js throwing exceptions due to
missing analysis nodes of a persisted source id that doesn’t have a
corresponding analysis

@xavijam can you review?
cc @javierarce

### DIFF
--- a/lib/assets/test/spec/cartodb3/data/layer-definition-model.spec.js
+++ b/lib/assets/test/spec/cartodb3/data/layer-definition-model.spec.js
@@ -43,6 +43,14 @@ describe('data/layer-definition-model', function () {
         }
       });
     });
+
+    it('should not contain source if it is a source node', function () {
+      this.model.set('source', 'a1');
+      expect(this.model.toJSON().options.source).toEqual('a1');
+
+      this.model.set('source', 'a0');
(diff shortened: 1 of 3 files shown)


xavijam:
👍

lib/assets/test/spec/cartodb3/data/layer-definition-model.spec.js
```diff
@@ -43,6 +43,14 @@ describe('data/layer-definition-model', function () {
         }
       });
     });
+
+    it('should not contain source if it is a source node', function () {
+      this.model.set('source', 'a1');
+      expect(this.model.toJSON().options.source).toEqual('a1');
+
+      this.model.set('source', 'a0');
+      expect(this.model.toJSON().options.source).not.toEqual('a0');
```
xavijam: options.source should undefined right?

viddo: yes, that's true

PR #7265, opened 2016-04-27
Revamp parallel testing scripts

javitonino:
A little cleanup of the CI scripts and a better way to disable problematic specs from parallel execution directly from the scripts.


Cartofante:

javitonino:
CR please, @gfiorav

gfiorav:
Much better! 📦 when tests pass

PR #7267, opened 2016-04-27
Fixes the URL generation for the geocoding & import models

javierarce:
The URLs used to check the status of the imports and geocodings weren't adding the `baseURL` component in the path, and therefore failing in some cases.

### DIFF
--- a/lib/assets/javascripts/cartodb3/data/background-importer/geocoding-model.js
+++ b/lib/assets/javascripts/cartodb3/data/background-importer/geocoding-model.js
@@ -33,15 +33,11 @@ module.exports = cdb.core.Model.extend({
     }
   },
 
-  url: function (method) {
+  urlRoot: function () {
     var version = this._configModel.urlVersion('geocoding');
     var baseUrl = this._configModel.get('base_url');
-    var geocodingsPath = '/api/' + version + '/geocodings';
 
-    if (this.isNew()) {
-      return baseUrl + geocodingsPath;
-    }
-    return baseUrl + geocodingsPath + '/' + this.id;
+    return baseUrl + '/api/' + version + '/geocodings';
   },
 
   setUrlRoot: function (urlRoot) {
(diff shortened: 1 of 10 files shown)


javierarce:
CR: @viddo

lib/assets/javascripts/cartodb3/data/background-importer/geocoding-model.js
```diff
@@ -35,13 +35,13 @@ module.exports = cdb.core.Model.extend({
...
-    var base = '/api/' + version + '/geocodings/';
+    var baseUrl = this._configModel.get('base_url');
+    var geocodingsPath = '/api/' + version + '/geocodings';
 
     if (this.isNew()) {
-      return base;
+      return baseUrl + geocodingsPath;
     }
-    return base + this.id;
+    return baseUrl + geocodingsPath + '/' + this.id;
```
viddo: I know this model was ported from the old code, but I think it's a good opportunity to fix this kind of maluse of url methods; define the method as [urlRoot](http://backbonejs.org/#Model-urlRoot) instead and you don't have to check isNew and append the id manually

lib/assets/javascripts/cartodb3/data/background-importer/import-model.js
```diff
@@ -23,12 +23,14 @@ module.exports = cdb.core.Model.extend({
...
+    var baseUrl = this._configModel.get('base_url');
+
+    var importsPath = '/api/' + version + '/imports';
 
     if (this.isNew()) {
-      return base;
+      return baseUrl + importsPath;
     }
-    return base + '/' + this.id;
+    return baseUrl + importsPath + '/' + this.id;
```
viddo: (Ditto above) I know this model was ported from the old code, but I think it's a good opportunity to fix this kind of maluse of url methods; define the method as [urlRoot](http://backbonejs.org/#Model-urlRoot) instead and you don't have to check isNew and append the id manually

viddo:
👍

lib/assets/javascripts/cartodb3/data/background-importer/geocoding-model.js
javierarce: Oh, true. Thanks!

lib/assets/javascripts/cartodb3/data/background-importer/import-model.js
javierarce: 👌

PR #7272, opened 2016-04-28
Exclude grunt-timer from load-grunt-tasks

rochoa:
Stop receiving '>> Local Npm module "grunt-timer" not found. Is it installed?'

See https://github.com/leecrossley/grunt-timer/issues/2


xavijam:
💃

PR #7275, opened 2016-04-28
Offline frontend tests

javitonino:
This makes frontend tests work offline:
- Google maps JS is loaded externally. If this fails to load (offline) a couple of tests will fail. I stubbed one case and changed to use LeafletMapView instead of GoogleMaps for other.
- There was an external image being loaded (from cartodb CDN). Changed to an inline image.

Continues #7240

### DIFF
--- a/lib/assets/test/spec/cartodb/table/map/export_image_view.spec.js
+++ b/lib/assets/test/spec/cartodb/table/map/export_image_view.spec.js
@@ -2,12 +2,11 @@ describe("ExportImage", function() {
 
   describe("cdb.admin.ExportImageView", function() {
     beforeEach(function() {
-
       this.map = new cdb.geo.Map();
 
       this.container = $('<div>').css({ width: '500px', height: '500px' });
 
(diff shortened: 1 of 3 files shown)


javitonino:
Please CR @xavijam

lib/assets/test/spec/cartodb/table/map/export_image_view.spec.js
```diff
@@ -2,12 +2,11 @@ describe("ExportImage", function() {
 
   describe("cdb.admin.ExportImageView", function() {
     beforeEach(function() {
-
       this.map = new cdb.geo.Map();
 
       this.container = $('<div>').css({ width: '500px', height: '500px' });
 
-      this.mapView = new cdb.geo.GoogleMapsMapView({
+      this.mapView = new cdb.geo.LeafletMapView({
```
xavijam: If there is no other side effect 👍

xavijam:
Looks good man! :)

PR #7282, opened 2016-04-29
7120 keep raster base table

rafatower:
What this changeset does:
- Force output band of raster overviews to be Int16, as Float32 is not supported for rendering
- Create an additional overview table with factor = 1 called `o_1_original_table_name` from the reprojected and adjusted base table
- Import "as-is" the original raster file and keep it as the base table

It is only partially compatible with current tiler/mapnik without needing the changes in pgraster plugin: https://github.com/CartoDB/mapnik/pull/14 meaning that new imports won't be rendered when the factor/scale is <= 2 (high zoom levels relative to raster scale/resolution).

It works as intended with the mapnik changes.

there's some things remaining:
- a small refactor
- some more smoke testing in staging (with the pgraster plugin change)
- some "unit test"

for the test I was thinking of using something in the lines of:
```
WITH foo AS (
    SELECT the_raster_webmercator rast FROM test_borneo
) SELECT (stats).* FROM (SELECT ST_summarystatsagg(rast, FALSE, 1) stats FROM foo) bar;
```
and comparing with gdalinfo output on the original raster. I already did that manually and don't expect surprises. A bit more elaborated:

```
SELECT (pvc).value, SUM((pvc).count) as count
FROM (SELECT st_valuecount(the_raster_webmercator) AS pvc FROM test_borneo) AS foo
GROUP BY (pvc).value ORDER BY count DESC;
```

@javisantana can you please review? if it is really urgent I guess somebody from dataservices can take over these changes and deploy them.

### DIFF
--- a/services/importer/lib/importer/raster2pgsql.rb
+++ b/services/importer/lib/importer/raster2pgsql.rb
@@ -13,6 +13,7 @@ module CartoDB
       SQL_FILENAME                  = '%s%s.sql'
       OVERLAY_TABLENAME             = 'o_%s_%s'
       RASTER_COLUMN_NAME            = 'the_raster_webmercator'
+      GDALWARP_COMMON_OPTIONS       = '-co "COMPRESS=LZW" -co "BIGTIFF=IF_SAFER"'
 
       def initialize(table_name, filepath, pg_options, db)
         self.filepath             = filepath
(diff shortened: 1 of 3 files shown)


javisantana:
so this would need to deploy the tiler (and obviously repackage the node bindings)

rafatower:
yes, it would @javisantana

javisantana:
:es:

PR #7285, opened 2016-05-03
Disabling compression in pg_dump while exporting users

azamorano:
This PR disables the compressing of the pg_dump file that should make exporting users much faster.

Please @zenitraM :eyes:


zenitraM:
LGTM :+1:

zenitraM:
Oh wait, don't enable `-Fc` on dump_schema as that will break things

PR #7289, opened 2016-05-03
7271 fix embed widgets

### LINKED ISSUE #7271: When a widget is added to a map, analyses/nodes should be persisted
Taking a map (test) with only one layer (layer1) with source (table1) and source id (a0)  by default.

When a widget is added to that map (test) (and there is no analyses applied), the source of that widget/dataview is pointing to the layer source node (a0), but the problem is we don't have the analysis definition stored in the backend. That means, named map template will not contain that analysis info and the dataview request will fail as it does in the public view.

### STR
- Create a map from the dashboard with one layer of points.
- It will redirect you to the old editor, but let's go to the new editor.
- Add a new widget, for example, a category based on the `cartodb_id` column.
- Check the public embed, widget load will fail.
- Check error: `"There is no associated analysis for the dataview source id"`

cc @rochoa in case I missed something or my explanation is not that good :).

viddo:
Fixes #7271 

The issue was that analysis source nodes are created ad-hoc for the editor context (if not already existing), but were not persisted until an non-source node was created on top of it. So when vis was loaded as embed the widgets reference non-existing nodes. 
So, changed the editor to persist the save when the widget is created, which in turn persist the referenced analysis and its source node.

- [x] ~~This however causes cartodb.js to break in some cases, so need to investigate how that should be fixed, I'll need you @alonsogarciapablo to discuss this.~~, confirmed working with latest version of Windshaft/camshaft as commented below (already available in master and deployed)

While at it also fixed a bug with the `nodeModel.toJSON` that didn't serialize `options` param on nested nodes…

@xavijam can you review meanwhile?


xavijam:
:+1:

lib/assets/javascripts/cartodb3/editor.js
```diff
@@ -223,13 +225,28 @@ window.snippets = {
    * Delete all analysis, and update layers
    */
   resetAnalyses: function () {
-    _.clone(window.analysisDefinitionsCollection.models).forEach(function (m) {
-      m.destroy();
-    });
+    // Set all layers to point to their source node
     window.layerDefinitionsCollection.each(function (m) {
```
xavijam: Snippets FTW.

alonsogarciapablo:
Checking out the branch and investigating this issue.

lib/assets/test/spec/cartodb3/data/layer-definitions-collection.spec.js
```diff
@@ -56,6 +64,10 @@ describe('data/layer-definitions-collection', function () {
       expect(this.collection.getAnalysisDefinitionNodesCollection()).toEqual(this.analysisDefinitionNodesCollection);
     });
 
+    it('should create a definition for the node', function () {
+      expect(this.analysisDefinitionsCollection.findByNodeId).toBeDefined();
```
alonsogarciapablo: Is this test right?

alonsogarciapablo:
>  - [ ] This however causes cartodb.js to break in some cases, so need to investigate how that should be fixed, I'll need you @alonsogarciapablo to discuss this.

This should NOT "break" with the latest version of Windshaft-CartoDB. Please update your environment and you shouldn't see any errors :)

lib/assets/test/spec/cartodb3/data/layer-definitions-collection.spec.js
viddo: No, I'll fix it, well spotted!

PR #7298, opened 2016-05-04
Place overwrite option at the end of ogr2ogr command line

### LINKED ISSUE #7297: ogr2ogr2.1 ignores overwrite option
When the `-overwrite` option is added for ogr2ogr retries it is ignored. But it works if we place it at the end of the command line.

Happens with ogr2ogr2.1 and not with ogr2ogr2.

jgoizueta:
fixes #7297


iriberri:
I'm gonna do some tests, but I :+1: with this change. Please don't forget to squash the commits whenever this is merged, as some of them were tests and could provoke confusions if someone checks the history of the file in the future. :-)

rafatower:
Fine by me as long as the testing goes ok

iriberri:
My local tests are being successful. Adding the overwrite at the final of the string is respecting the update option with the prelude and closure statements. :+1: I'll try to reproduce a real scenario in staging as soon as I get a problematic file and after that we should be good to go.

iriberri:
It's working! :-) @jgoizueta is gonna remove the testing commits of the branch and we can merge afterwards.

iriberri:
FTR: No occurrences in Rollbar staging.

PR #7299, opened 2016-05-04
Split editor bundles

rochoa:
This is the first attempt to split editor into several bundles using https://github.com/substack/factor-bundle

For more information check https://github.com/substack/browserify-handbook#partitioning

In the future this will require a bit of work on the entry points for editor/public editor so it's possible to split it into more bundles. For instance having a bundle for vendor stuff: jquery, underscore, and friend that usually don't change at all.

This includes an update in cartodb.js including: https://github.com/CartoDB/cartodb.js/compare/3d2f8087b5087f0ea88871df28729ca8edbf4508...c88d5ec5979a343ca54178676ca277d8507ff5ea

cc @xavijam @viddo


xavijam:
It sounds super good when I listened @rochoa talking about it, so 👍 . You guys have more experience and knowledge about browserify so 100% trust on you @rochoa and @viddo.

viddo:
Actually had this on my private "someday" TODO list already, so all good for me :shipit:

rochoa:
This is still far from ideal but it's better to start now than wait until we can reach *"ideal"*.

As *"ideal"* I would expect to have something like:

 - vendor_editor3_bundle.js
   * underscore
   * jquery
   * perfect-scrollbar
   * d3
   * polyglot
   * ...
 - common_editor3_bundle.js
   * cartodb.js
   * deep-insights.js
 - editor3.js
 - public_editor3.js

And obviously some optimisations because, for instance, public_editor3.js is just a call to deep-insights.js' `createDashboard`.

xavijam:
👍

PR #7310, opened 2016-05-05
Add update_metadata method on ImportJob

santisaez:
This PR adds a new method required to support rollback feature in user-mover tool.

@azamorano and @zenitram, can you review, please? thanks!

### DIFF
--- a/services/user-mover/import_user.rb
+++ b/services/user-mover/import_user.rb
@@ -180,6 +180,10 @@ module CartoDB
           end
         end
 
+        if @options[:update_metadata]
+          update_metadata
+        end
+
         log_success
         remove_user_mover_banner(@pack_config['user']['id']) if @options[:set_banner]
       end
--- a/services/user-mover/import_user.rb
+++ b/services/user-mover/import_user.rb
@@ -180,9 +180,6 @@ module CartoDB
           end
         end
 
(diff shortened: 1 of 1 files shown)


services/user-mover/import_user.rb
```diff
@@ -539,6 +529,20 @@ def update_redis_database_name(user, db)
         metadata_redis_conn.hset("rails:users:#{user}", 'database_name', db)
       end
 
+      def update_metadata(target_dbhost, target_dbname)
```
azamorano: Is there need for this parameters? As they are class parameters there shouldn't, right?

santisaez: @azamorano yes, they are required since we're going to call `update_metadata` method "externally" as the rollback procedure, something like this:

```ruby
import=CartoDB::DataMover::ImportJob.new(import_args)
(..)
logger.error "Migrated #{user_type} #{name} successfully but with consistency check errors, executing ROLLBACK procedure"
import.update_metadata(source_database_host, database_name)
```

We need a different IP so the parameter is required.

santisaez:
@azamorano just applied your suggestions, double check it - thanks! :)

luisbosque:
This PR does change the behavior by itself, right?
There's just code enclosed in a new method

PR #7311, opened 2016-05-05
Update cartodbjs dependency

xavijam:
And adding trade-area time limitation, from 0 to 3600 (1 hour).

cc @viddo @dgaubert


dgaubert:
LGTM 👍

xavijam:
Frontend tests passed, merging!

PR #7323, opened 2016-05-06
Legends toggle fix

mbektas:
This is to fix the Legends toggle on Options menu not working.


mbektas:
@javisantana this was a request that I fixed on our fork. you can merge to main branch with this PR.

javisantana:
I don't know if this is the way to fix this.

The legend should be shown or hidden based on the model, a user action should never change a view directly (specially in this case where we are using a global variable `table`). The view shoudl listen to map changes and hide/show the legend

mbektas:
OK, I will update the fix and have it work based on model changes

mbektas:
Fix based on model change is at PR #7363

Issue #6742, opened 2016-03-01
Change layer definition table/query setup


viddo:

viddo:
Closing since will be resolved as part of #7294

PR #7325, opened 2016-05-09
7294 widgets sources

### LINKED ISSUE #7294: Widgets forms using wrong columns
When widgets were changed to rely on analysis nodes (#7217) upon creation, the widgets editing was missed. So if creating a widget to a particular node that has a particular schema, and later editing it the available columns are not correct.

Basically we'll need to:
- get schema (columns) from source node instead
- :fire: layerTableModel (conceptually replaced by analysis source node)

cc @javierarce @xavijam

viddo:
Resolves #7294 and #6742

Schema/columns are now always retrieved through an analysis' node query schema instead of associated layer.

Related to do this change I also did some housekeeping:
- :fire: layerTableModel since now longer used (replaced by the node schema)
- rename TableQueryModel to QuerySchemaModel since that's what it represents
- Removes the duplicated status/query state from the node-definition-model, query state moved to 
- extracted and consolidated deep-insights/vis integration to a single, separate object, making collections/models simpler in test env since they should not require to test/integrate there

Although widgets conceptually no longer are associated with layers directly, there's still code (in particular on the server-side) that still relies on this attr to be present, so didn't remove the layer_id attr in this PR. Perhaps we should tackle this separately to tidy things up?

@xavijam can you review review, please? I think going through the commits one by one is easiest to follow the trail of changes
cc @javierarce


lib/assets/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -0,0 +1,74 @@
+/**
```
xavijam: This is a good one.

lib/assets/javascripts/cartodb3/data/analysis-definition-node-source-model.js
```diff
@@ -19,6 +19,11 @@ module.exports = AnalysisDefinitionNodeModel.extend({
     this.bind('change:query', function () {
       this._deferredOutputGeometryType = null;
     }, this);
+
+    this.analysisTableModel.set({
+      status: 'ready',
```
xavijam: as a default status ready?

lib/assets/javascripts/cartodb3/components/modals/add-widgets/add-widgets-view.js
```diff
@@ -94,7 +94,7 @@ module.exports = cdb.core.View.extend({
     return this._analysisDefinitionNodesCollection
     .chain()
     .reduce(function (memo, m) {
-      m = m.analysisTableModel;
+      m = m.querySchemaModel;
       if (m.get('query') && m.get('status') === 'ready') {
```
xavijam: Shouldn't we start using `m.isReady()` or `m.hasFailed()`?

lib/assets/javascripts/cartodb3/data/widget-definition-model.js
```diff
@@ -1,7 +1,7 @@
 var _ = require('underscore');
 var cdb = require('cartodb.js');
 
-var ATTRS_AT_TOP_LEVEL = ['id', 'layer_id', 'title', 'type', 'order'];
+var ATTRS_AT_TOP_LEVEL = ['id', 'layer_id', 'source', 'title', 'type', 'order'];
```
xavijam: Shouldn't we get rid of `layer_id` finally?

lib/assets/javascripts/cartodb3/components/modals/add-widgets/add-widgets-view.js
```diff
@@ -150,6 +109,7 @@ module.exports = cdb.core.View.extend({
   },
 
   _createOptionsModels: function () {
+    this._optionsCollection.reset([]);
```
xavijam: We don't need to pass a [] argument.

lib/assets/javascripts/cartodb3/data/query-schema-model.js
```diff
@@ -22,29 +22,49 @@ module.exports = cdb.core.Model.extend({
     this._configModel = opts.configModel;
 
     this.columnsCollection = new Backbone.Collection([]);
```
xavijam: Same here, [] not needed.

lib/assets/javascripts/cartodb3/data/query-schema-model.js
```diff
@@ -1,3 +1,4 @@
+var $ = require('jquery');
```
xavijam: Q about this class: if there is a query applied from the beginning (`query` set in the layer attributes), which would be the `status` of this model when it is initialized?

xavijam:
👍 , just some comments. 🍧

lib/assets/javascripts/cartodb3/data/query-schema-model.js
viddo: it would be set to `unfetched`, the status and state is reset [when the query changes](https://github.com/CartoDB/cartodb/pull/7325/files#diff-0ac5b38d10b58405f35a161c0fd7b276R90)

PR #7337, opened 2016-05-09
Browserify editor3

rochoa:
Deprecates factor-bundle, uses browserify's external/require internals to use several bundles.

Contrary to previous solution it works with linked modules.

cc @alonsogarciapablo @viddo @xavijam


viddo:
Does this have the same output (bundles/results) as factor-bundle yielded?

> Contrary to previous solution it works with linked modules.

(For future reference) what does _"works with linked modules"_ really mean?

What goes for the changes look legit so 👍

alonsogarciapablo:
I'm giving this a try, but it looks 👍

rochoa:
@viddo If you were doing:

```
npm link cartodb.js
npm link cartodb-deep-insights.js
```

It was not generating the same bundles, missing some dependencies in common bundle, as when you were just `npm install`ing them.

Some refs:
 - https://github.com/substack/factor-bundle/issues/66
 - https://github.com/nodejs/node/issues/3402

alonsogarciapablo:
Working 👍

Issue #7327, opened 2016-05-09
New custom select doesn't go up/down correctly with keys


xavijam:
![screen shot 2016-05-09 at 11 28 56](https://cloud.githubusercontent.com/assets/132146/15108971/3e288c5e-15d9-11e6-805c-6aa840b34688.png)

### STR: 

- Go to new editor.
- Try any of the new forms with a new select on it (try to use one with several items and search enabled).
- Search for something that returns several results.
- Go down with the key.
- Selection doesn't work properly.

nobuti:
Closed by #7341

PR #7341, opened 2016-05-10
Removing the key binding properly

nobuti:
This PR implements #7327.

The problem was the way we add the `keydown` handler. We used the `bind` method, that returns a new function. But when we tried to remove the binding, that function didn't exist. The implemented solution set a reference to the binded function, and we use it to add and remove the handler.

Also, in order to avoid unnecessary operations, we only respond to the keyup handler only if the select is open.

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/custom-list/custom-list-view.js
+++ b/lib/assets/javascripts/cartodb3/components/custom-list/custom-list-view.js
@@ -13,6 +13,7 @@ module.exports = cdb.core.View.extend({
   tagName: 'div',
 
   initialize: function () {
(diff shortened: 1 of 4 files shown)


nobuti:
Review this please @xavijam

nobuti:
Implemented highlight on show @xavijam

xavijam:
Marvellous 👏 . 🚢  🇮🇹  when tests are passed.

xavijam:
Bump package.json version before merge please (https://github.com/CartoDB/cartodb/blob/master/CONTRIBUTING.md#submitting-contributions).

alonsogarciapablo:
Ole!

PR #7374, opened 2016-05-11
Get geometry from analysis query schema

### LINKED ISSUE #7334: Attach last node sqlSchemaModel to the proper layer definition model
cc @viddo @javisantana @matallo

viddo:
This was based on [the existing code in tabledata.js](https://github.com/CartoDB/cartodb/blob/831b3505ae8759c7a39555284d96cef96e5d6523/lib/assets/javascripts/cartodb/models/tabledata.js#L771-L824), but for the new concepts added on the node-definition's query-schema model, since it's closely related to the schema/data available on the node's query.

```js
querySchemaModel.on('change:status', handleStateChange)
querySchemaModel.fetch()

// … once query-schema is fetched and where you need the geometry:

querySchemaModel.get('status') // => should be `fetched` when getting geom:
var geom = querySchemaModel.getGeometry() // => geometry value object
if (geom) {
  geom.getPostgresType() // => 'ST_MultiPoint'
  geom.getSimpleType() // => 'point'
}
```

Note that the geometry might not be available, it depends on the state of the analysis but also on there being data containing a geom for the given query, so even when finished there's no guarantee there is any.

This also resolves #7334 added a shortcut to access the current analysis-definition-node that the layer currently holds, but keep in mind that this object will change as the analysis changes (e.g. adding/removing nodes), so we should only listen to events on this object in views while editing something:

```js
layerDefModel.getAnalysisDefinitionNodeModel() // returns the "head", or undefined if the layer has no source (e.g. a basemap layer).
```

@xavijam please review


lib/assets/javascripts/cartodb3/components/modals/add-analysis/add-analysis-view.js
```diff
@@ -73,16 +63,29 @@ module.exports = cdb.core.View.extend({
...
-    this._$body().append(view.render().el);
-    this.addView(view);
-  },
-
   _$body: function () {
     return this.$('.js-body');
   },
 
+  _onQuerySchemaChange: function () {
+    if (this._isFetching()) return;
```
xavijam: Could we "expose" this methods outside? They will be super helpful.
isFetching
hasFailed
…

viddo: since usages will have to use the different states anyway, why not just the value of `m.get('status')`?

lib/assets/javascripts/cartodb3/data/analysis-definition-node-source-model.js
```diff
@@ -13,44 +11,10 @@ module.exports = AnalysisDefinitionNodeModel.extend({
...
-        if (geometryType) {
-          cb(null, geometryType);
-        } else {
-          cb('no geometry value');
-        }
-      })
-      .fail(function (err) {
-        cb(err || 'something failed');
-      });
+    this.querySchemaModel.set({
```
xavijam: Heavy cleaning :heart:

lib/assets/javascripts/cartodb3/data/layer-definition-model.js
```diff
@@ -115,6 +116,16 @@ module.exports = cdb.core.Model.extend({
...
   createNewAnalysisNode: function (nodeAttrs) {
     return this.collection.createNewAnalysisNode(this, nodeAttrs);
+  },
+
+  getAnalysisDefinitionNodeModel: function () {
+    return this.collection.findAnalysisDefinitionNodeModel(this.get('source'));
+  },
+
+  _onSourceChanged: function () {
+    if (this.infowindowModel) {
```
xavijam: Nice, I'd do the same with style model 🍔

lib/assets/javascripts/cartodb3/data/query-schema-model.js
```diff
@@ -1,10 +1,10 @@
-var $ = require('jquery');
 var _ = require('underscore');
 var cdb = require('cartodb.js');
 var Backbone = require('backbone');
+var geometry = require('./geometry');
 
 var MAX_GET_LENGTH = 1024;
-var WRAP_SQL_TEMPLATE = 'select * from (<%= sql %>) __wrapped limit 0';
+var WRAP_SQL_TEMPLATE = 'select * from (<%= sql %>) __wrapped limit <%= rows_sample_size %>';
```
xavijam: Q, for you @javisantana too. Does it make sense to add in this query 'the_geom is not null'?

xavijam: _Already answered, not possible to add that because if a query is applied and it is not requesting `the_geom` column, it will fail.

lib/assets/javascripts/cartodb3/data/query-schema-model.js
```diff
@@ -13,7 +13,8 @@ module.exports = cdb.core.Model.extend({
 
   defaults: {
     query: '',
-    status: 'unavailable' //, unfetched, fetching, fetched
+    status: 'unavailable', //, unfetched, fetching, fetched
+    may_have_rows: false // this flag indicate if there _may_ be data available on the given query (but there's no guarantee)
```
xavijam: Finally we go with this solution?

lib/assets/javascripts/cartodb3/editor/widgets/widgets-form/widgets-form-content-data-view.js
```diff
@@ -29,7 +30,8 @@ module.exports = cdb.core.View.extend({
 
   _initBinds: function () {
     this._widgetDefinitionModel.on('change:type', this._renderForm, this);
-    this._querySchemaModel.fetch().then(this.render.bind(this));
+    this._querySchemaModel.on('change:status', this.render, this);
+    this.add_related_model(this._querySchemaModel);
```
xavijam: Coul you add another add_related_model for _widgetDefinitionModel?

lib/assets/javascripts/cartodb3/components/modals/add-analysis/add-analysis-view.js
xavijam: :+1:

lib/assets/javascripts/cartodb3/data/query-schema-model.js
viddo: It's not needed to know/be checked externally, but I need to know it internally to decide whether to reset status or not. If you have a concrete proposal on how to remove it I'm all ears!

xavijam:
_Dale zapato_ => 🚢 🇮🇹

xavijam:
Ready to go?

viddo:
Since there are only frontend code changes for the new editor should be, but I'm still waiting for the frontend tests to finish… once they pass I'll merge

PR #7384, opened 2016-05-12
Data Observatory general integration

iriberri:
**Requires the migration https://github.com/CartoDB/cartodb/pull/7385 to be deployed in advance**

 * Metrics for observatory general: overquota, number of uses, remanining quota and so on
 * Quota for observatory general service
 * Soft limit for observatory general
 * Tests for the observatory general part


rafatower:
:+1:

ethervoid:
👍

Issue #7342, opened 2016-05-10
Layer management


alonsogarciapablo:
- [x] Adding new layers:
  - [x] Add an empty layer CartoDB instead of cloning the one at the top.
  - [x] Make sure layer is added below the layer with labels.
- [x] Deleting an existing layer:
  - [x] Display "Delete layer..." in contextual menu.
  - [x] Do not display "Delete layer..." if there's only one "CartoDB" or "Torque" layer.
  - [x] Clicking "delete layer" deletes the layer:
    - [x] Layer is removed from the map (CartoDB.js).
    - [x] Layer deletion is persisted in the DB.
    - [x] Analysis nodes owned by the layer are deleted.
  - [x] Confirm deletion via Infobox.
- [x] Moving layers:
  - [x] Layers can be drag & dropped.
  - [x] After a layer is moved, the map is refreshed.
  - [x] After a layer is moved, new order of layers is persisted in the DB.
  - [x] Before saving the layers, create analysis definitions in the DB.
  - [x] If Context Menu is visible, hide it right when dragging starts.
  - [x] "CartoDB" and "Torque" layers can't be moved below the basemap.
  - [x] "CartoDB" and "Torque" layers can't be moved above the layer with labels.
  - [ ] ~~Animated layers should be on top~~ (see https://github.com/CartoDB/cartodb/issues/8975)
- [x] Other stuff:
  - [x] Expanding and collapsing layers
  - [x] Showing / hidding layers
  - [x] Apply a new color to the icon of new layers. (dupe of #7436 ?)
  - [x] Display labels layer differently (currently says Basemap).
  - [x] Basemap icon (preview).
  - [x] Renaming layers
  - [ ] ~~Indicate geometry output on layer avatar~~ (see https://github.com/CartoDB/cartodb/issues/7490)

viddo:
Saw this issue after created #7479, perhaps you'd want to add to "Adding new layers" to keep track of it too?

viddo:
Another thing I encounter, this one related to "moving layers", the order of the layers is not matching between editor and the map:

```js
> layerDefinitionsCollection.pluck('table_name')
["madrid_oct_2015_listings_with_closest_metro", "paradas_metro_madrid", "barrios", undefined]
> layerDefinitionsCollection.pluck('order')
[1, 1, 1, 0]

> vis._layersCollection.pluck('layer_name')
[...]

PR #7387, opened 2016-05-12
Add new layers in the right place

alonsogarciapablo:
Fixes first part of #7342:

- [x] Add an empty CartoDB layer instead of cloning the one at the top.
- [x] Make sure layer is added below the torque layer/layer with labels.

@xavijam would you please take a look? Thx


xavijam:
👍  after tests say everything is 🍏

alonsogarciapablo:
Retest this please

PR #7390, opened 2016-05-12
Disable tests for organization user mover

gfiorav:

PR #7395, opened 2016-05-13
[user-mover] fix update_metadata issues w/ orgs

santisaez:
This PR makes `ImportJob` class compatible with organization migrations, this feature was broken after adding `update_metadata` method in [PR#7310](https://github.com/CartoDB/cartodb/pull/7310), changes:

- there are 2 new methods in charge of updating specific user or organization metadata
- the generic `update_metadata` method now updates user, organization or both metadata accordingly
- each `ImportJob` instantiation is stored to allow calling `update_metadata` after consistency-checks
- `update_metadata_*` expect a non-referenced `target_dbhost` to support rollback in the future

**IMPORTANT**

-  this PR also enables back again some user-mover tests that were disabled on [PR#7390](https://github.com/CartoDB/cartodb/pull/7390)
- https://github.com/CartoDB/cartodb-platform/pull/1913 must be merged after merging this PR
- make "if not nil" comparison more clear, related: https://github.com/CartoDB/cartodb/pull/7395/commits/1d300e07cc60e55440504b8a054e0d0fb5561e8e

@azamorano @luisbosque please review, thanks!

### DIFF
--- a/services/user-mover/import_user.rb
+++ b/services/user-mover/import_user.rb
@@ -207,11 +207,11 @@ module CartoDB
         # We first import the owner. If schemas are not split, this will also import the whole org database
         @logger.info("Importing org owner #{@owner_id}..")
         @import_job_admin_instance = ImportJob.new(file: @path + "user_#{@owner_id}.json",
-                      mode: @options[:mode],
-                      host: @target_dbhost,
-                      target_org: @pack_config['organization']['name'],
-                      logger: @logger, data: @options[:data], metadata: @options[:metadata],
-                      update_metadata: @options[:update_metadata])
+                                                   mode: @options[:mode],
+                                                   host: @target_dbhost,
+                                                   target_org: @pack_config['organization']['name'],
+                                                   logger: @logger, data: @options[:data], metadata: @options[:metadata],
+                                                   update_metadata: @options[:update_metadata])
         @import_job_admin_instance.run!
 
         # Fix permissions and metadata settings for owner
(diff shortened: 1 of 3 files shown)


services/user-mover/import_user.rb
```diff
@@ -576,6 +586,19 @@ def update_metadata
         end
       end
 
+      def update_metadata(target_dbhost)
+        if !@pack_config['organization'].nil?
```
azamorano: if not nil could be replaced by a standard if @pack_config['organization'] which is clearer

santisaez: @azamorano sure, in my opinion it's also confusing but I just did a copy from:

https://github.com/CartoDB/cartodb/blob/759239a9d60eccce7467455c004c05745889587b/services/user-mover/import_user.rb#L57

I'm going to replace with a more clearer expression in both places.

santisaez:
I get the following error when using this code from the "migrator-worker" tool:

>PG::Error: ERROR:  cannot execute ALTER ROLE in a read-only transaction

It happens when calling [`set_database_search_path`](https://github.com/CartoDB/cartodb/blob/02695b7f1bbdb3ae075c30cf16ec6862777840a8/app/models/user/db_service.rb#L228-L235) from [`configure_database`](https://github.com/CartoDB/cartodb/blob/02695b7f1bbdb3ae075c30cf16ec6862777840a8/services/user-mover/import_user.rb#L564), it works fine when using it the first time to apply on the organization admin but it fails when processing the first user, the reason is that `database_host` points to the source database and it's in read-only mode.

santisaez:
@zenitraM now it works fine with the migrator tool, I have performed several tests migrating normal users, enterprise users and organizations, etc. thanks!! :)

@azamorano can you review again this? thanks!

services/user-mover/import_user.rb
```diff
@@ -51,13 +53,17 @@ def initialize(options)
       end
 
       def run!
-        if !@pack_config['organization'].nil?
+        if @pack_config['organization']
```
azamorano: You could use the method organization_import just defined below for clarity

Issue #6929, opened 2016-03-21
Update analysis on nodes deletion


viddo:
Was postponed from /pushed out of scope of #6827

Removal of node(s) should remove dependent nodes, layers, and widgets, as well as update/destroy analysis definitions accordingly.

viddo:
Resolved by #7420 , closing

PR #7420, opened 2016-05-18
[WIP] Layer and analysis deletion

alonsogarciapablo:
Part of #7342 (layer deletion) and #6929 (analysis deletion).

layers

edit: In a nutshell:
- Deleting layers deletes its "owned" analysis nodes (nodes that having matching letter)
- Deleting nodes will delete dependent nodes, widgets, and layers
- Deleting nodes will update analysis (either change existing analysis-definition or delete duplicate analysis)
- Due to synchronization issues the concerned models on sync event listeners only the last save call will be adhered to (any ongoing request is aborted), migrated from the old code.

cc: @viddo


lib/assets/javascripts/cartodb3/editor/layers/layer-view.js
```diff
@@ -139,6 +137,38 @@ module.exports = cdb.core.View.extend({
...
+      {
+        label: 'Delete layer…',
+        val: 'delete-layer',
+        destructive: true
+      }
+    ]);
+
+    var menuView = new ContextMenuView({
+      collection: menuItems,
+      triggerElement: this.$el.find('.js-show-menu')[0],
```
viddo: Keeping references to DOM elements are typical causes memory leaks, IMHO it should be avoided if possible. If not, at least make sure the reference is removed when the view is removed (i.e. clean).

lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
```diff
@@ -0,0 +1,125 @@
...
+
+  _onEscapePressed: function (ev) {
+    if (ev.which === $.ui.keyCode.ESCAPE) {
+      this.hide();
+    }
+  },
+
+  _onDocumentElementClicked: function (e) {
+    var clickedElement = e.target;
+    if (this.options.triggerElement === clickedElement) {
```
viddo: Not entirely sure this will work consistently, I remember having trouble with this in the past due to varying behavior in different browsers. Also, if the triggerElement contains more than a text node the target might differ depending on where the user actually clicks.

lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    if (!opts.triggerElement) {
+      throw new Error('triggerElement option is required');
+    }
+
+    this.options = _.extend({}, this.options, opts);
+
+    this.model = new cdb.core.Model({
+      query: '',
```
viddo: What is this used for?

lib/assets/javascripts/cartodb3/editor/layers/layer-view.js
```diff
@@ -139,6 +137,38 @@ module.exports = cdb.core.View.extend({
...
+      }
+    }, this);
+
+    this.$el.append(menuView.render().el);
+    this.addView(menuView);
+  },
+
+  _hideContextMenu: function () {
+    if (this._menuView) {
+      this._menuView.remove();
```
viddo: I'd change this to use the default/conventional `.clean`, to make sure the view is properly cleaned up (override the default clean to remove the custom document event listeners)

lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
```diff
@@ -0,0 +1,125 @@
...
+  },
+
+  _initBinds: function () {
+    this.model.bind('change:visible', function (mdl, isVisible) {
+      this.render();
+    }, this);
+
+    this.collection.bind('change:selected', function (menuItem) {
+      this.hide();
+    }, this);
```
viddo: You can make this conciser, i.e.:
```js
this.model.on('change:visible', this.render, this);
this.collection.on('change:selected', this.hide, this);
```

Btw, I believe it's clearer to use the [default `on`/`off` methods instead of `bind`/`unbind` alias](http://backbonejs.org/#Events) to make a clearer separation from the native JS [`Function.prototype.bind`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind) method.

lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
```diff
@@ -0,0 +1,125 @@
...
+          this.$el.css(attr, val);
+        }
+      }, this);
+    }
+
+    if (this.isVisible()) {
+      this.$el.show();
+    } else {
+      this.$el.hide();
+    }
```
viddo: Could be conciser, `this.$el.toggle(this.isVisible);`

viddo:
@alonsogarciapablo Apart from my comments and unless you have any concerns regarding the changes I did related to deleting nodes and if there's any other known(?) issue, it's 👍  for me. 

BTW, updated the PR description to give an overview of the changes the PR contain.

lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
alonsogarciapablo: Fixed [here](https://github.com/CartoDB/cartodb/commit/23479068b888084722b818b46319860abdcc770b).

lib/assets/javascripts/cartodb3/editor/layers/layer-view.js
alonsogarciapablo: That method is actually not being used... 😲 Removing it!

lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
alonsogarciapablo: :+1:

lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
alonsogarciapablo: Ouh yeah!

lib/assets/javascripts/cartodb3/editor/style/style-converter.js
```diff
@@ -331,7 +331,8 @@ function generateStyle (style, geometryType) {
   if (style.type === 'none') {
     return {
       cartoCSS: GENERIC_STYLE,
-      sql: null
+      sql: null,
+      layerType: 'CartoDB'
```
viddo: @javisantana talked with @xavijam who directed me to you; Without this addition the layer's type was set to `undefined` upon adding an analysis node, so the layer could not be saved.

javisantana: I changed it here https://github.com/CartoDB/cartodb/pull/7428/files#diff-d43b3ab860e2013a71ff53dd2c8b92b7R343

xavijam:
Guys, I haven't checked the changes, but after seeing the GIF, we shouldn't provide the option to delete the last data layer available in the map.

alonsogarciapablo:
Good catch @xavijam. I tracked that in #7342. We can fix it on a separate PR.

viddo:
No backend changes and frontend tests passing, so merging

PR #7425, opened 2016-05-18
Fix color picker

piensaenpixel:

color-picker-ok


xavijam:
Let's fix rest of the things of the fill component and then we will review it ok? :)
Also, check tests 💃

piensaenpixel:
@xavijam fixed  the rest of dropdowns 💃

image

image

image

xavijam:
:+1:

PR #7430, opened 2016-05-18
Scroll view

### LINKED ISSUE #7407: Implement scroll (+ shadows) view for new content views
screen shot 2016-05-17 at 09 48 11

cc @piensaenpixel

nobuti:
This PR closes #7407.

This is still work in progress, but it's working. 

scroll

There are some things to fix to work better:

- [x] Right now, the component fills the remain height, and works for now. But we need to make some changes in the css in order to work with flexbox. It will allow to react to other views like the warnings or errors.
- [x] Because the reason above, we need to remove the resize binding, and fixed height logic.

Also, the approach taken here is to extend the base Backbone.View, in order to avoid bringing too much functionality. I'm not sure about it, because `cdb.core.View` has done a lot of work behind the scenes to ensure they'll just work. So, I'm wondering if a mixin approach would be better:

```
// scroll.js
module.exports = {
  applyScroll: function () {},
  destroyScroll: function () {},
  otherStuff: function () {}
}

// any view composing scroll
var Scroll = require('../mixins/scroll');

module.exports = cdb.core.View.extend(
  _.extend({}, Scroll, {
    ...
    // (Methods and attributes here)
    ...
  })
);
```

Thoughts?

cc @viddo @javierarce @xavijam


lib/assets/javascripts/cartodb3/components/scroll/scroll.js
```diff
@@ -0,0 +1,130 @@
+var _ = require('underscore');
+var $ = require('jquery');
+var Backbone = require('backbone');
+var Ps = require('perfect-scrollbar');
+var template = require('./scroll.tpl');
+
+module.exports = Backbone.View.extend({
+  tagName: 'div',
+  className: 'ScrollView',
+  defaults: {
```
xavijam: Let's not use defaults, let's use options, something already supported and managed by the proper cdb.core.View, and we will get rid of the next thing I'm gonna point it :joy:

lib/assets/javascripts/cartodb3/components/scroll/scroll.js
```diff
@@ -0,0 +1,130 @@
...
+    offset: 20
+  },
+
+  initialize: function (opts) {
+    if (!opts.view) throw new Error('Inner view is required');
+    this._view = opts.view;
+    this._inDom = false;
+    this._viewOptions = _.extend({}, opts.viewOptions);
+    this.options = _.extend({}, this.defaults, this.options);
```
xavijam: Not necessary using options instead of defaults (defaults is for Models :) ).

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindow-view.js
```diff
@@ -16,15 +17,27 @@ module.exports = cdb.core.View.extend({
       label: self.options.toggleLabels.off,
       selected: true,
       createContentView: function () {
-        return new InfowindowView(self.options);
+        return new ScrollView({
+          view: InfowindowView,
```
xavijam: instead of passing the view and then the options, why not consider...:

```js
new ScrollView({
  createContentView: function () {
    return new InfowindowView({
      options: self.options
    })
  }
});
```

lib/assets/javascripts/cartodb3/components/scroll/scroll.js
```diff
@@ -0,0 +1,130 @@
...
+  destroyScroll: function () {
+    if (!this._wrapper) {
+      return;
+    }
+
+    this._unbindEvents();
+    Ps.destroy(this._wrapper.get(0));
+  },
+
+  render: function () {
```
xavijam: Opinion: I like to have render function always after initialize function.
Super picky man, I know :joy:

xavijam:
I'd wait to review the whole thing until we have the flexbox thing working properly.

lib/assets/javascripts/cartodb3/components/scroll/scroll.js
```diff
@@ -0,0 +1,128 @@
+var _ = require('underscore');
+var $ = require('jquery');
+var Backbone = require('backbone');
+var Ps = require('perfect-scrollbar');
+var template = require('./scroll.tpl');
+
+module.exports = Backbone.View.extend({
```
viddo: > …, the approach taken here is to extend the base Backbone.View, in order to avoid bringing too much functionality. I'm not sure about it, because cdb.core.View has done a lot of work behind the scenes to ensure they'll just work. So, I'm wondering if a mixin approach would be better

Why would a mixin approach be better?

The `cdb.core.View` contains logic to manage the lifecycle of the view (and its subviews), which you're currently missing and would have to implement manually, e.g. the created content view is never cleaned up with current implementation.

viddo: Btw, we have the convention of suffix files by their type (-view.js, -model.js etc.) to easier find/navigate the source code (i.e. in this case rename the file to `scroll-view.js`)

lib/assets/javascripts/cartodb3/components/scroll/scroll.js
```diff
@@ -0,0 +1,128 @@
...
+
+  initialize: function (opts) {
+    if (!opts.createContentView) throw new Error('A factory createContentView function is required');
+    this.options = opts || {};
+    this._inDom = false;
+    this._offset = 20;
+  },
+
+  render: function () {
+    var view = this.options.createContentView.call(null);
```
viddo: …continuing on [my prev comment](https://github.com/CartoDB/cartodb/pull/7430/files#r63852680), e.g. if using cdb.core.View as base view you could/should have the subview managed by doing
`this.addView(view)`, this makes sure the subview is cleaned and removed properly when this (parent) view is removed too.

viddo: The canonical example for render to work as expected using the `cdb.core.View` would be:

```js
render: function() {
  this.clearSubViews(); // cleans up and removes any previously added sub-views

  var view = var view = this.options.createContentView();
  this.addView(view);
  this.$('.js-content').append(view.render().el);

  //…

  return this;
}

File: lib/assets/javascripts/cartodb3/components/scroll/scroll.js [hunk 7]
```diff
@@ -0,0 +1,128 @@
...
+
+    this.destroyScroll();
+    if (!this._content) {
+      this.$el.html(template());
+      this._content = this.$('.js-content');
+      this._wrapper = this.$('.js-wrapper');
+      this._shadowTop = this.$('.js-top');
+      this._shadowBottom = this.$('.js-bottom');
```
viddo: To avoid keeping the DOM refs around you could just move these behind private getter (continuing on the example in my [prev comment](https://github.com/CartoDB/cartodb/pull/7430/files#r63854215)):

```js
  this._$content().append(view.render().el);

// …

// also prefixing it with "$" to make the returned type more clear
_$content: function () {
  return this.$('.js-content');
}
```

lib/assets/javascripts/cartodb3/components/scroll/scroll.js
nobuti: I like the mixin pattern because it this scroll is only a wrapper, and this way, you will add the scroll functionality maintaining all the base logic from `cdb.core.View`. But you are right, it should be better to extend `cdb.core.View` than `Backbone.View` base class.

lib/assets/javascripts/cartodb3/components/scroll/scroll.js
```diff
@@ -0,0 +1,128 @@
...
+  },
+
+  remove: function () {
+    this.destroyScroll();
+    Backbone.View.prototype.remove.apply(this, arguments);
+  },
+
+  clean: function () {
+    this.trigger('clean');
+    this.remove();
```
viddo: If going with the prev recommendations, this should call `cdb.core.View.prototype.clean`, to [clean up subviews etc.](https://github.com/CartoDB/cartodb.js/blob/e9d37f0b478fce24aba02d88f4a29067779a1319/src/core/view.js#L44-L70), `remove` should not be necessary.

lib/assets/javascripts/cartodb3/components/scroll/scroll.js
nobuti: I added a nullify pattern on destroy to clear the references, this way we loses all the advantage of a cached element.

nobuti:
@xavijam working with flexbox /cc @piensaenpixel

lib/assets/javascripts/cartodb3/components/scroll/scroll-view.js
```diff
@@ -0,0 +1,105 @@
...
+var template = require('./scroll.tpl');
+
+module.exports = cdb.core.View.extend({
+  tagName: 'div',
+  className: 'ScrollView',
+
+  initialize: function (opts) {
+    if (!opts.createContentView) throw new Error('A factory createContentView function is required');
+    this.options = opts || {};
+    this._inDom = false;
```
xavijam: Is `_inDom` being used?

In any case, I'd try to use static variables outside the class like:

var OFFSET = 20;
var INTERVAL = 100;
…

xavijam:
It looks pretty promising, what do you think @viddo?
BTW, it would be great @nobuti if you can apply the scroll thing to all working panes.
Widgets list and form, style form, etc, pretty please.

nobuti:
On it.

nobuti:
Done.

viddo:
👍

javierarce:
👍  Looks good to me too.

nobuti:
🚀

PR #7437, opened 2016-05-19
Pass element ID instead of the DOM element itself

alonsogarciapablo:
This PR changes the ContextMenu component so that is uses DOM ids instead of references to DOM elements, as suggested [in this comment](https://github.com/CartoDB/cartodb/pull/7420/files#r63733560).

@viddo better?

cc: @javisantana

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
+++ b/lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
@@ -25,8 +25,8 @@ module.exports = cdb.core.View.extend({
       throw new Error('collection option is required');
     }
 
-    if (!opts.triggerElement) {
-      throw new Error('triggerElement option is required');
+    if (!opts.triggerElementID) {
+      throw new Error('triggerElementID option is required');
     }
 
     this.options = _.extend({}, this.options, opts);
@@ -49,7 +49,8 @@ module.exports = cdb.core.View.extend({
 
   _onDocumentElementClicked: function (e) {
     var clickedElement = e.target;
-    if (this.options.triggerElement === clickedElement || $.contains(this.options.triggerElement, clickedElement)) {
+    var triggerElement = $('#' + this.options.triggerElementID)[0];
+    if (triggerElement === clickedElement || $.contains(triggerElement, clickedElement)) {
       this.toggle();
     } else {
       this.hide();
--- a/lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
+++ b/lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
@@ -37,8 +37,8 @@ module.exports = cdb.core.View.extend({
 
     this._initBinds();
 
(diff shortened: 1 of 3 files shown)


lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
```diff
@@ -49,16 +49,17 @@ module.exports = cdb.core.View.extend({
...
       this.hide();
     }
   },
 
-  remove: function () {
-    $(document).unbind('keydown', this._onEscapePressed.bind(this));
-    $(document).unbind('click', this._onDocumentElementClicked.bind(this));
+  clean: function () {
+    $(document).unbind('keydown', this._onEscapePressed);
+    $(document).unbind('click', this._onDocumentElementClicked);
```
viddo: Considering the default clear function is overriden I believe a `cdb.core.View.prototype.clean.apply(this, arguments);` call is missing here

viddo: Btw, are the functions given in the unbind calls actually matching the functions used in the initialize function? Otherwise there's the risk that the unbind doesn't work as expected. It's possible that wasn't correct previously either due to doing separate .bind(this) calls, but probably should double check this. 

One way to avoid this would be to use [_.bindAll](http://underscorejs.org/#bindAll) to bind the functions for the created object, i.e.: 

```js
initialize: function () {
  _.bindAll(this, '_onEscapePressed', `_onDocumentElementClicked`);`
  $(document).bind('keydown', this._onEscapePressed);
  $(document).bind('click', this._onDocumentElementClicked);
}

…

clear: function () { 
  $(document).unbind('keydown', this._onEscapePressed);
  $(document).unbind('click', this._onDocumentElementClicked);
```

viddo:
👍

lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
alonsogarciapablo: > Btw, are the functions given in the unbind calls actually matching the functions used in the initialize function? Otherwise there's the risk that the unbind doesn't work as expected. It's possible that wasn't correct previously either due to doing separate .bind(this) calls, but probably should double check this.

Yes, that's why I used `$.proxy()`. According to the [jQuery docs](https://api.jquery.com/jQuery.proxy/):

> This method is most useful for attaching event handlers to an element where the context is pointing back to a different object. Additionally, jQuery makes sure that even if you bind the function returned from jQuery.proxy() it will still unbind the correct function if passed the original.

Anyway, I like your suggestion, so I'll change it and also add a test to ensure it's working properly. Thanks!

alonsogarciapablo: Done in [24df9d081c96b74f0da76f7f9105e56df9e25518](https://github.com/CartoDB/cartodb/pull/7437/commits/24df9d081c96b74f0da76f7f9105e56df9e25518). Thanks @viddo!

PR #7440, opened 2016-05-20
Reduces grunt memory usage by avoiding forking the process

javitonino:
We had problems deploying #7178 because we were running out of memory. The added files increased the memory size of the Grunt process, just enough for it to run out of memory when calling an external process.

This appears to be due to how Linux spawns process, by first doing a `fork()` which duplicates the process (double the memory). I changed the Grunt task to execute the HTTP request to Fastly by using the internal HTTP client instead of calling `curl`. This fixes the problem.

Can you CR @xavijam?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -93,16 +93,28 @@
 
     grunt.registerTask('invalidate', "invalidate cache", function() {
       var done = this.async();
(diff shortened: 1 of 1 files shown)


xavijam:
👍  neno :)

xavijam:
By the way, WTF! :joy:

MariaCheca:
@javitonino FTW.

PR #7446, opened 2016-05-20
Add --quote-all-identifiers option to pg_dump during user export

zenitraM:
This is a workaround for a (currently on the moderation queue) bug in Postgres 9.5's pg_dump which does not treat "over" as a reserved keyword while dumping a 9.3 database.

### DIFF
--- a/services/user-mover/export_user.rb
+++ b/services/user-mover/export_user.rb
@@ -53,7 +53,7 @@ module CartoDB
           @database_host,
           CartoDB::DataMover::Config[:user_dbport],
           @database_name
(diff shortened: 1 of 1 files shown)


santisaez:
👍

zenitraM:
Deemed as not a bug on PostgreSQL:
 http://www.postgresql.org/message-id/CAKFQuwaB45GTsZ9JDERjYxsoDhs=uUzaONwi87hag+DJ-LHaMw@mail.gmail.com 
but the parameter seems to be safe in general and also useful to ensure dumps work between major versions so merging it

PR #7452, opened 2016-05-23
Fix opacity selection

javierarce:
The opacity of the `rgba` color that the user selected in the fill component is not reflected on the UI.

This PR fixes the issue passing a RGBA color (that is, a color with the opacity value) to the colorpicker component instead of a regular RGB.


javierarce:
CR: @xavijam

xavijam:
:+1:

PR #7459, opened 2016-05-23
Problems related with context-menu cleaning fixed

### LINKED ISSUE #7457: Uncaught TypeError: Cannot read property 'ownerDocument' of undefined (context-menu-view:55)
Saw this exception in the dev console in production env:

```
jquery:1430 Uncaught TypeError: Cannot read property 'ownerDocument' of undefined
b.contains @ jquery:1430
b.exports.f.core.View.extend._onDocumentElementClicked @ context-menu-view.js:55
_.event.dispatch @ jquery:4435q.handle @ jquery:4121
```

[content-menu-view:55](https://github.com/CartoDB/cartodb/blob/350f182802ad10d32cfce4b24a88b52e1fb621dd/lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js#L55)

xavijam:
Fixes #7457.

So, from my point of view, only external views should open the context-menu, not the proper view.
And also, document binds should not be enabled until the view is visible.
What do you think mate?

CR: @alonsogarciapablo

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
+++ b/lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
@@ -22,9 +22,9 @@ module.exports = cdb.core.View.extend({
   initialize: function (opts) {
     opts = opts || {};
     if (!opts.collection) throw new Error('collection option is required');
-    if (!opts.triggerElementID) throw new Error('element id is required');
+    if (!opts.targetElement) throw new Error('targetElement is required');
 
-    this._triggerElementID = opts.triggerElementID;
+    this._targetElement = opts.targetElement;
 
     this.model = new cdb.core.Model({
       visible: false
@@ -86,7 +86,7 @@ module.exports = cdb.core.View.extend({
 
   _onDocumentElementClicked: function (ev) {
     var $el = $(ev.target);
-    if ($el.closest(this.$el).length === 0 && $el.attr('id') !== this._triggerElementID) {
+    if ($el.closest(this.$el).length === 0 && (ev.target !== this._targetElement)) {
       this.hide();
     }
   },
--- a/lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
+++ b/lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
@@ -22,9 +22,9 @@ module.exports = cdb.core.View.extend({
   initialize: function (opts) {
     opts = opts || {};
     if (!opts.collection) throw new Error('collection option is required');
-    if (!opts.targetElement) throw new Error('targetElement is required');
+    if (!opts.triggerElementID) throw new Error('element id is required');
 
-    this._targetElement = opts.targetElement;
+    this._triggerElementID = opts.triggerElementID;
 
     this.model = new cdb.core.Model({
       visible: false
@@ -86,7 +86,7 @@ module.exports = cdb.core.View.extend({
 
   _onDocumentElementClicked: function (ev) {
     var $el = $(ev.target);
-    if ($el.closest(this.$el).length === 0 && (ev.target !== this._targetElement)) {
+    if ($el.closest(this.$el).length === 0 && $el.attr('id') !== this._triggerElementID) {
       this.hide();
     }
   },
--- a/lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
+++ b/lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
@@ -86,7 +86,7 @@ module.exports = cdb.core.View.extend({
 
   _onDocumentElementClicked: function (ev) {
     var $el = $(ev.target);
-    if ($el.closest(this.$el).length === 0 && $el.attr('id') !== this._triggerElementID) {
+    if ($el.closest(this.$el).length === 0 && $el.closest($('#' + this._triggerElementID)).length === 0) {
       this.hide();
     }
   },
--- a/lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
+++ b/lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
@@ -20,55 +20,43 @@ module.exports = cdb.core.View.extend({
   tagName: 'div',
 
   initialize: function (opts) {
-    _.bindAll(this, '_onEscapePressed', '_onDocumentElementClicked');
-
     opts = opts || {};
-    if (!opts.collection) {
-      throw new Error('collection option is required');
-    }
+    if (!opts.collection) throw new Error('collection option is required');
+    if (!opts.triggerElementID) throw new Error('element id is required');
 
-    if (!opts.triggerElementID) {
-      throw new Error('triggerElementID option is required');
-    }
-
-    this.options = _.extend({}, this.options, opts);
+    this._triggerElementID = opts.triggerElementID;
 
     this.model = new cdb.core.Model({
       visible: false
     });
 
+    this._onEscapePressed = this._onEscapePressed.bind(this);
+    this._onDocumentElementClicked = this._onDocumentElementClicked.bind(this);
+
     this._initBinds();
-
-    $(document).bind('keydown', this._onEscapePressed);
-    $(document).bind('click', this._onDocumentElementClicked);
-  },
-
-  _onEscapePressed: function (ev) {
-    if (ev.which === $.ui.keyCode.ESCAPE) {
-      this.hide();
-    }
-  },
-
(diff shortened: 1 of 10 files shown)


lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
```diff
@@ -90,6 +78,19 @@ module.exports = cdb.core.View.extend({
...
 
+  _onEscapePressed: function (ev) {
+    if (ev.which === $.ui.keyCode.ESCAPE) {
+      this.hide();
+    }
+  },
+
+  _onDocumentElementClicked: function (ev) {
+    var $el = $(ev.target);
+    if ($el.closest(this.$el).length === 0 && $el.attr('id') !== this._triggerElementID) {
```
xavijam: Not sure about triggerElementID reference, but it is useful to check if it was clicked over that reference.

lib/assets/javascripts/cartodb3/editor/layers/layers-view.js
```diff
@@ -48,7 +49,8 @@ module.exports = cdb.core.View.extend({
 
   _addLayerView: function (m) {
     var view;
-    if (m.get('order') === 0) {
+
+    if (m.get('type') === 'Tiled') {
```
xavijam: Could it do the trick?

lib/assets/javascripts/cartodb3/editor/map/map-view.js
```diff
@@ -82,6 +82,7 @@ module.exports = cdb.core.View.extend({
     });
 
     this.$el.append(stackLayoutView.render().$el);
+    this.addView(stackLayoutView);
```
xavijam: One of the original problems.

lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
```diff
@@ -20,55 +20,43 @@ module.exports = cdb.core.View.extend({
...
-
-    this.options = _.extend({}, this.options, opts);
+    this._triggerElementID = opts.triggerElementID;
 
     this.model = new cdb.core.Model({
       visible: false
     });
 
-    this._initBinds();
+    this._onEscapePressed = this._onEscapePressed.bind(this);
```
xavijam: This solves the problem that document bindings were not removed correctly.

xavijam: Already used in the `custom-carousel-view.js` by the way, thanks to @nobuti.

lib/assets/test/spec/cartodb3/components/context-menu/context-menu-view.spec.js
```diff
@@ -121,7 +126,33 @@ describe('components/context-menu/context-menu-view', function () {
     expect(this.view.$el.css('wadus')).toBeUndefined();
   });
 
+  describe('document binds when item is not visible', function () {
+    beforeEach(function () {
+      spyOn(this.view, 'hide');
+    });
+
+    it('should not check outside click', function () {
```
alonsogarciapablo: `should not try to hide the context menu when clicking somewhere else`?

lib/assets/test/spec/cartodb3/components/context-menu/context-menu-view.spec.js
```diff
@@ -121,7 +126,33 @@ describe('components/context-menu/context-menu-view', function () {
...
+      $('body').click();
+      expect(this.view.hide).not.toHaveBeenCalled();
+      this.$button.click();
+      $('body').click();
+      expect(this.view.hide).toHaveBeenCalled();
+    });
+
+    it('should not check ESC keypress', function () {
```
alonsogarciapablo: `should not try to hide the context menu when pressing ESCAPE`?

lib/assets/test/spec/cartodb3/components/context-menu/context-menu-view.spec.js
```diff
@@ -46,32 +51,32 @@ describe('components/context-menu/context-menu-view', function () {
     expect(this.view.$el.find('button[title="Label 2"]')).toBeDefined();
   });
 
-  it('should be toggled when clicking on the triggerElement', function () {
+  it('should be toggled when clicking on the button', function () {
```
alonsogarciapablo: This test and the following are now the same... I think it's important to ensure that both clicking on an element (previously `this.triggerElement`) and in some element inside of if (`this.button`) works as expected.

lib/assets/test/spec/cartodb3/components/context-menu/context-menu-view.spec.js
```diff
@@ -121,7 +126,33 @@ describe('components/context-menu/context-menu-view', function () {
...
 
+  describe('document binds when item is not visible', function () {
+    beforeEach(function () {
+      spyOn(this.view, 'hide');
+    });
+
+    it('should not check outside click', function () {
+      $('body').click();
+      expect(this.view.hide).not.toHaveBeenCalled();
+      this.$button.click();
```
alonsogarciapablo: Perhaps we could use `this.view.show()` here and in the following test instead of `this.button.click()` to make it more obvious that we're showing the view?

lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
alonsogarciapablo: $el.closest() will only work if the element that triggered the click (`$(ev.target)`) is a descendent of `this.$el`, That's not the case in the context menu that is displayed for layers. That's why I had to use the triggerElementID reference.

lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
```diff
@@ -20,55 +20,43 @@ module.exports = cdb.core.View.extend({
...
   },
 
-  _onEscapePressed: function (ev) {
-    if (ev.which === $.ui.keyCode.ESCAPE) {
-      this.hide();
-    }
+  _initBinds: function () {
+    this.model.on('change:visible', function (mdl, isVisible) {
+      this.render();
+      if (isVisible) {
```
alonsogarciapablo: :+1: Like it!

lib/assets/javascripts/cartodb3/editor/layers/layer-view.js
```diff
@@ -121,6 +122,10 @@ module.exports = cdb.core.View.extend({
     this._stackLayoutModel.nextStep(this.model, 'layers', selectedNode);
   },
 
+  _onOpenContextMenu: function () {
+    this._menuView.toggle();
```
alonsogarciapablo: 👍  Like it too! :)

lib/assets/javascripts/cartodb3/editor/layers/layers-view.js
alonsogarciapablo: I guess we can live with it for now. I'm not a big fan of having this kind of magic literals for layer types (I'd rather have a global LAYER_TYPES.TILED or something along those lines) but since it's a common pattern, let's go with this.

I'd rename `BasemapLayerView` to `TiledLayer` or something different since the layer with labels is not really a basemap... But, we can do this in the future.

alonsogarciapablo:
I like the approach 👏  👏  👏  (more efficient than what we had before)

lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
alonsogarciapablo: `_.bindAll(...)` should also work, right?

lib/assets/test/spec/cartodb3/components/context-menu/context-menu-view.spec.js
```diff
@@ -23,13 +25,19 @@ describe('components/context-menu/context-menu-view', function () {
...
     this.triggerElement.append(this.button);
+
     $('body').append(this.triggerElement);
 
     this.view = new ContextMenuView({
       collection: this.collection,
       triggerElementID: this.triggerElement[0].id
     });
 
+    this.triggerElement.on('click', function (ev) {
```
alonsogarciapablo: Do we need this?

xavijam: you mean the paco thing? Yeah, it is cool.

alonsogarciapablo:
👍  Thanks @xavijam!

xavijam:
Passing travis frontend tests, merging!

matallo:
🙌

PR #7465, opened 2016-05-23
Add layer style issues

xavijam:
Three things about this PR:

1. If a new layer is added, infowindow and tooltip layer models are not initialized, because they don't have data about those things. From now on, both models will be created from the beginning, checking that the layer has a table_name, as we do with the style model.
Also, if infowindow models don't have any data, they will not be parsed in the layer-definition-model toJSON. -> CR: @alonsogarciapablo + @matallo

2. And, when a new layer is added, style-manager was not binding new layer changes -> CR: @javisantana.

3. If a layer changes (after saving it) and it doesn't have any source reference, it will not try to create the analysis definition. -> CR: @viddo

### DIFF
--- a/lib/assets/javascripts/cartodb3/data/layer-definition-model.js
+++ b/lib/assets/javascripts/cartodb3/data/layer-definition-model.js
@@ -41,14 +41,14 @@ module.exports = cdb.core.Model.extend({
       attrs.sql = r.options.query;
     }
 
-    if (r.infowindow) {
+    if (r.options.table_name) {
       if (!this.infowindowModel) {
         this.infowindowModel = new InfowindowDefinitionModel(r.infowindow, {
           configModel: opts.configModel || this._configModel
         });
       }
     }
-    if (r.tooltip) {
+    if (r.options.table_name) {
       if (!this.tooltip) {
         this.tooltipModel = new InfowindowDefinitionModel(r.tooltip, {
           configModel: opts.configModel || this._configModel
@@ -108,11 +108,13 @@ module.exports = cdb.core.Model.extend({
       options: options
     };
 
-    if (this.infowindowModel) {
+    var infowindowData = this.infowindowModel && this.infowindowModel.toJSON();
+    if (!_.isEmpty(infowindowData)) {
       d.infowindow = this.infowindowModel.toJSON();
     }
 
-    if (this.tooltipModel) {
+    var tooltipData = this.tooltipModel && this.tooltipModel.toJSON();
+    if (!_.isEmpty(tooltipData)) {
       d.tooltip = this.tooltipModel.toJSON();
     }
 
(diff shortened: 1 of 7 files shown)


lib/assets/javascripts/cartodb3/data/layer-definition-model.js
```diff
@@ -41,14 +41,14 @@ module.exports = cdb.core.Model.extend({
       attrs.sql = r.options.query;
     }
 
-    if (r.infowindow) {
+    if (r.options.table_name) {
```
viddo: Considering `table_name` is a legacy attr that doesn't really fit conceptually with the new layers, why not provide an empty infowindow and tooltip object [when the layer is created instead](https://github.com/CartoDB/cartodb/blob/4c141afc4f5617115901170c805c7c7499b88598/lib/assets/javascripts/cartodb3/components/modals/add-layer/add-layer-model.js#L213-L222)?

xavijam: It could be another option, @alonsogarciapablo agree?

xavijam:
ping pong!

lib/assets/javascripts/cartodb3/editor/style/style-manager.js
```diff
@@ -12,19 +12,34 @@ function StyleManager (layerDefinitionsCollection) {
...
+
+    function _bind (layerDef) {
+      if (layerDef && layerDef.styleModel) {
+        layerDef.styleModel.bind('change', _generate(layerDef), this);
+      }
+    }
+
+    function _unbind (layerDef) {
+      if (layerDef.styleModel) {
+        layerDef.styleModel.unbind(null, null, this);
```
javisantana: I'd unbind what it binds and not everything

lib/assets/test/spec/cartodb3/editor/style/style-manager.spec.js
```diff
@@ -90,8 +90,25 @@ describe('editor/style/style-manager', function () {
     });
 
     it('it should track style changes on add', function () {
-      // this.layerDefModel._styleModel.set({ fill: { fixed: '#000', opacity: 0.4 } });
-      // expect(this.layerDefModel.get('cartocss')).toBe('#layer {\nmarker-fill: #000;\nmarker-fill-opacity: 0.4;\n}');
+      this.collection.add({
```
javisantana: 👍

lib/assets/javascripts/cartodb3/editor/style/style-manager.js
xavijam: Already changed, github is :trollface:  me

xavijam:
No more comments means merge ;).

PR #7477, opened 2016-05-24
Remove tmp node creation on layer dragndrop

viddo:
Will need a proper implementation to work as expected, removing until we have time to do it properly.


xavijam:
👍

PR #7487, opened 2016-05-24
infowindow template

matallo:
already sending the template, but not really sure how to proceed here, or at least if we should move around the templates (i.e. https://github.com/CartoDB/cartodb/blob/7418-infowindow-template/lib/assets/javascripts/cartodb3/data/infowindow-definition-model.js#L6-L29)

also we need the tooltips to start working

comparing the PR against the other one at least until it's merged

CR @alonsogarciapablo 

/cc @javisantana


lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindow-select-view.js
```diff
@@ -39,14 +40,128 @@ module.exports = cdb.core.View.extend({
...
     if (!this._checkValidTemplate()) {
       this._setNoneTemplate();
     }
 
     this._templatesCollection.bind('change:selected', function (mdl) {
       if (mdl.get('selected')) {
-        this._layerInfowindowModel.setTemplate(mdl.getValue());
+        var value = mdl.getValue();
+
+        this._layerInfowindowModel.setTemplate(value, this._compileTemplate(value));
```
alonsogarciapablo: CartoDB.js already takes care of compiling the template, etc. I don't think we need to do it here. IMO all we need to do here is read the template from the corresponding file (based on the name of the template) and set it as a string.

matallo: but I need to do this to persist the template in the backend for the embed later, right?

Issue #7485, opened 2016-05-24
Polygon datasets are rendered as points in new editor when using "by value" styling


AbelVM:
Steps:
* Get a dataset with polygons
* Make a map with the old editor
* Switch to the new editor, everything looks ok
* Go to the styling and change the fill from "solid" to "by value"
* Bang! Now you have points

Note: the styling menus in the new editor displays the options for points from start, even when displaying polygons the right way

cc @xavijam

xavijam:
We have found the reason. The query we use to retrieve or guess the geometry type is giving us some headaches. We have agreed to use the same query we have in the current editor (first 20 rows, order by cartodb_id).

@AbelVM @javisantana @viddo

PR #7495, opened 2016-05-24
Issue 7485

xavijam:
Fixes https://github.com/CartoDB/cartodb/issues/7485.

cc @rochoa @javisantana @AbelVM

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/style/style-converter.js
+++ b/lib/assets/javascripts/cartodb3/editor/style/style-converter.js
@@ -21,13 +21,13 @@ function makeColorRamp (props) {
     } else {
       // colorramp name
       c.push(props.range);
+      if (props.bins) {
+        c.push(props.bins);
+      }
     }
   }
   if (props.quantification) {
     c.push(props.quantification.toLowerCase());
-  }
-  if (props.bins) {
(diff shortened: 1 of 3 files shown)


lib/assets/javascripts/cartodb3/editor/style/style-converter.js
```diff
@@ -21,11 +21,11 @@ function makeColorRamp (props) {
     } else {
       // colorramp name
       c.push(props.range);
+      if (props.bins) {
```
xavijam: Only add bins when colorbrewer is used.

lib/assets/javascripts/cartodb3/editor/style/style-converter.js
```diff
@@ -37,9 +37,9 @@ function makeWidthRamp (props) {
   if (props.range) {
     c.push(props.range[0]);
     c.push(props.range[1]);
-  }
-  if (props.bins) {
-    c.push(props.bins);
+    if (props.bins) {
```
xavijam: If there is a range, and bins is set, we can use it.

javisantana:
merged into https://github.com/CartoDB/cartodb/pull/7505

PR #7505, opened 2016-05-24
Style heatmap

javisantana:
this enables heatmap style based on torque

there is one issue, aggregation specify the grid size but and when the heatmap is animated the resoltuion is also present in the animated section. We should remove one of them (probably the animated section one)

cc @xavijam


lib/assets/javascripts/cartodb3/editor/style/style-defaults/heatmap-style-defaults.js
```diff
@@ -10,7 +10,7 @@ module.exports = _.defaults({
           fixed: 35
         },
         'color': {
-          range: ['blue', 'cyan', 'lightgreen', 'yellow', 'orange', 'red']
+          fixed: 'white'
```
xavijam: Why have we changed that @javisantana?

javisantana: because this is not a color range, it's an image-filter and should be managed in a different way

javisantana: don't worry so much about that, we could thing about that later

javisantana: I'm fixing it

javisantana:
@xavijam fixed

2 small problems:

- stroke and blending should not be in the form
- the fill component allows to get more than one column but it's aggregated so it should work in the same way than rectangle and hexagon grid, with the `points_agg` column

xavijam:
Ok, doing those changes right now.

xavijam:
Changes applied @javisantana ^^^.

PR #7509, opened 2016-05-25
Fix new layers created from uploaded/imported datasets

### LINKED ISSUE #7479: Importing a dataset from data library fails
1. Click add layer
1. Open data library
1. Select a dataset (e.g. _Madrid districts_)
1. Await the importer to finish successfully

Expected result:
- New layer added is added at the top with an analysis source of the new dataset

Actual result:
- A duplicated layer of labels at the top, instead of the expected data layer

edit: [the create layer logic](https://github.com/CartoDB/cartodb/blob/842118537b6fbf8b7c0a9bf17fdef659685f2598/lib/assets/javascripts/cartodb3/components/modals/add-layer/add-layer-model.js#L197-L238) seems wrong:
- the (add-layer) view model logic should not need to know all those internal details/scenarios, probably should be contained in the layer-definitions-collection
- The layer types were already defined in [layer-types-and-kinds](https://github.com/CartoDB/cartodb/blob/842118537b6fbf8b7c0a9bf17fdef659685f2598/lib/assets/javascripts/cartodb3/data/layer-types-and-kinds.js), perhaps can be used to do matching instead of duplicating/dispersing them in more places?

alonsogarciapablo:
Fixes #7479.

@viddo would you please take a look? Thanks!

### DIFF
--- a/lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
+++ b/lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
@@ -4,17 +4,12 @@ var LayerDefinitionModel = require('./layer-definition-model');
 var layerLetters = require('./layer-letters');
 var nodeIds = require('./analysis-definition-node-ids.js');
 var camshaftReference = require('./camshaft-reference');
+var layerTypesAndKinds = require('./layer-types-and-kinds');
 
 /**
  * Collection of layer definitions
  */
 module.exports = Backbone.Collection.extend({
-
-  LAYER_TYPES: {
-    CARTODB: 'CartoDB',
-    TORQUE: 'torque',
-    TILED: 'Tiled'
-  },
 
   /**
    * Intended to be called from entry point, to make sure initial layers are taken into account
@@ -110,9 +105,9 @@ module.exports = Backbone.Collection.extend({
     var layerOnTop = this.first();
     if (layerOnTop) {
       order = layerOnTop.get('order');
-      if (layerOnTop.get('type') === this.LAYER_TYPES.CARTODB) {
+      if (layerTypesAndKinds.isCartoDBType(layerOnTop.get('type'))) {
         order += 1; // Layer is placed on top
-      } else if (layerOnTop.get('type') === this.LAYER_TYPES.TORQUE || layerOnTop.get('type') === this.LAYER_TYPES.TILED) {
+      } else if (layerTypesAndKinds.isTorqueType(layerOnTop.get('type')) || layerTypesAndKinds.isTiledType(layerOnTop.get('type'))) {
         layerOnTop.set('order', order + 1); // Layer that was on top is kept on top
       }
     }
--- a/lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
+++ b/lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
@@ -3,11 +3,18 @@ var Backbone = require('backbone');
 var LayerDefinitionModel = require('./layer-definition-model');
 var layerLetters = require('./layer-letters');
 var nodeIds = require('./analysis-definition-node-ids.js');
+var camshaftReference = require('./camshaft-reference');
 
 /**
  * Collection of layer definitions
  */
 module.exports = Backbone.Collection.extend({
+
+  LAYER_TYPES: {
+    CARTODB: 'CartoDB',
+    TORQUE: 'torque',
+    TILED: 'Tiled'
+  },
 
   /**
    * Intended to be called from entry point, to make sure initial layers are taken into account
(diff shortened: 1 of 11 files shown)


lib/assets/javascripts/cartodb3/editor.js
```diff
@@ -235,8 +235,8 @@ deepInsights.createDashboard('#dashboard', vizJSON, {
   });
 
   backgroundPollingModel.bind('importCompleted', function (importModel) {
```
alonsogarciapablo: @viddo not sure if we should leave this here or there is a better place for it (perhaps could be moved to the backgroundPollingModel since we will always want to create the layer when an import is completed?)... Also, we don't seem to be testing this file so this is not tested right now.

lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
```diff
@@ -3,12 +3,19 @@ var Backbone = require('backbone');
...
 
 /**
  * Collection of layer definitions
  */
 module.exports = Backbone.Collection.extend({
 
+  LAYER_TYPES: {
+    CARTODB: 'CartoDB',
+    TORQUE: 'torque',
+    TILED: 'Tiled'
```
viddo: I'm still no fan of having these types defined here, what do you think about removing this here and access them through  
something like this?
```js
var type = layerOnTop.get('type'))
layerTypesAndKinds.isOfTypeCartoDB(type); // => true
layerTypesAndKinds.isOfTypeTiled(type); // => false
```

lib/assets/javascripts/cartodb3/editor.js
viddo: Ouch, good catch! Wasn't even aware of this one. :( The snippets was intended as a dev tool to be used from the dev console, until we had the actual UI features migrated/implemented, never to be called from the app code. 👍  to have that fixed.

> not sure if we should leave this here or there is a better place for it (perhaps could be moved to the backgroundPollingModel since we will always want to create the layer when an import is completed?).

If that is true then I agree

> Also, we don't seem to be testing this file so this is not tested right now.

Considering it requires a full editor pre-setup it's hard to unit test since it would require a humongous bootsrap to work. Ideally this file should be smaller and any app logic should be contained/moved elsewhere (and tested ofc).

lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
```diff
@@ -97,6 +104,49 @@ module.exports = Backbone.Collection.extend({
     return r.layers;
   },
 
+  createCartoDBLayerFromTableName: function (tableName, options) {
```
viddo: why not just call it `createLayerForTable`? Not sure why the layer type have to be mentioned in the method name, seems more like an internal detail than anything else.

alonsogarciapablo: 👍

lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
alonsogarciapablo: Done! I went for `.isCartoDBType`, `.isTorqueType`...

lib/assets/javascripts/cartodb3/editor.js
alonsogarciapablo: I talked with @xavijam about this and we realised `backroundPollingModel` doesn't currently know about the `layerDefinitionsCollection`. Since this is most likely going to change with the new notification system I left a TODO and decided to leave the code there.

alonsogarciapablo:
@viddo thanks for your feedbak! Will merge once the tests are green unless you have some more comments.

lib/assets/javascripts/cartodb3/editor.js
viddo: > … we realised backroundPollingModel doesn't currently know about the layerDefinitionsCollection

Although that could be changed ofc, especially if _we will always want to create the layer when an import is completed_ is correct. I don't see any problem with having the background polling knowing about the layer-definitions-collection if it needs to create layers. I'd actually prefer that to implicit behavior using events in this case. That would also solve the problem of testability you pointed out.

> and backgroundPollingModel is a subclass of a subclass of common/background_polling/background_polling_model.js

From what I can see it doesn't subclass anything from common/, and if it were we would have to break that dependency because the new editor can't depend on old models (since they depend on cartodb.js)… AFAIK these models were ported to work with the new cartodb.js v4. Can you confirm?

> Since this is most likely going to change with the new notification system

Why should the notification system care about _how_ the layer is created when the import finished?

> I left a TODO and decided to leave the code there.

Please, don't leave `// TODO …` in the source, create an issue and reference the concerned code instead.

viddo:
👍  although I still think we should resolve [my doubts (on a now outdated code piece)](https://github.com/CartoDB/cartodb/pull/7509#discussion_r64549260) before merging.

lib/assets/javascripts/cartodb3/editor.js
alonsogarciapablo: You're totally right @viddo! (Thanks for your persistence!). I didn't get the hierarchy of objects right and I now see that the class for the backgroundPollingModel is under `cartodb3/data/editor-background-polling-model`. Moving the code related to layer creation there then and writing  some tests :+1:

alonsogarciapablo:
Hey @viddo I addressed your final feedback on [this commit](https://github.com/CartoDB/cartodb/pull/7509/commits/ccb555060a0d1c375aa732ada2c25dfbe0663d94). PTAL. Thanks!

viddo:
👍

PR #7512, opened 2016-05-25
infowindow: template

matallo:
already sending the template, but not really sure how to proceed next, also we need the tooltips to start working

the template appears properly both in the vizjson and the embed, BUT as we don't save it anywhere else at the time of the fill component it always restores to the same value `#35AAE5` unless we add a new attribute to the infowindow

CR @alonsogarciapablo

/cc @viddo


lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindow-select-view.js
```diff
@@ -39,14 +63,42 @@ module.exports = cdb.core.View.extend({
     noneTemplate.set('selected', true);
   },
 
+  _isValidURL: function (url) {
```
matallo: prepare for image headers

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindow-form-model.js
```diff
@@ -18,10 +21,16 @@ module.exports = cdb.core.Model.extend({
...
     var self = this;
 
+    // depending on the field we'll update the select model for later transformation
+    // or the infowindow model directly
     _.each(this.changed, function (val, key) {
       if (self._layerInfowindowModel.get(key)) {
         self._layerInfowindowModel.set(key, val);
       }
+
+      if (self._infowindowSelectModel.get(key)) {
```
viddo: perhaps `model.has(key)` is preferable for this kind of guard?

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindow-select-view.js
```diff
@@ -23,6 +26,27 @@ module.exports = cdb.core.View.extend({
...
+    });
+
+    this.model.bind('change', this._onChange, this);
+  },
+
+  _onChange: function () {
+    var self = this;
+
+    _.each(this.model.changed, function (val, key) {
+      if (self.model.get(key)) {
```
viddo: [ditto m.has(key)](https://github.com/CartoDB/cartodb/pull/7512/files#r64564140)

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindow-form-model.js
matallo: ooh, much better!

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindow-form-view.js
```diff
@@ -30,14 +32,15 @@ module.exports = cdb.core.View.extend({
     }
 
     this._formModel = new InfowindowFormModel({
-      width: this._layerInfowindowModel.get('width'),
+      width: this._layerInfowindowModel.get('width') || '',
       headerColor: {
         color: {
-          fixed: '#CCC'
+          fixed: '#35AAE5;'
```
viddo: This color is repeated thrice (at least, judging by this PR), I presume their usages are interrelated? If so, I'd suggest to define it once and reference that in the other places? Like `defaultColor`, since it seems to be some kind of default(?)/fallback color.
Also, the headerColor struct here looks suspiciously similar to the data in new `infowindowSelectModel`, could this be consolidated?

matallo: uhm, I'm a bit concerned about this, as preferably I want to store it so it appears selected in the fill component, IMO the best place for this would be the infowindow attributes, but I'm not really sure about the implications

viddo: Thx for clarifying your intent. What I meant was something on the lines of (without changing too much):

```js
// infowindow-select-view.js
var defaultHeaderColor = '#35AAE5';
this.model = new cdb.core.Model({
  defaultHeaderColor: defaultHeaderColor,
  headerColor: defaultHeaderColor,
  headerImg: null
});

…
return template.replace('background: '+ this.model.get('defaultHeaderColor'), 
```

```js
// infowindow-form-view.js
this._formModel = new InfowindowFormModel({
  headerColor: {
    color: {
      fixed:  this._infowindowSelectModel.get('defaultHeaderColor')
```

viddo: Btw, I see another potential problem related to the form model; since the form is always given these attrs wouldn't the color _always_ be set to the default color even if the infowindow already have a color set? 
In other places where we use the forms we initiate it with the model's attributes and utilize the defaults (or parse) to set default values when the model is created.

matallo: doing so with the width! `width: this._layerInfowindowModel.get('width') || '',` that was my doubt on how to do it when I don't have a color to pull from

I set it that way so at least I could have the fill working, then I'd tackle that problem, but I need to know how first

matallo: @viddo discussed with @javisantana and solved 😎

lib/assets/javascripts/cartodb3/data/layer-definition-model.js
```diff
@@ -90,12 +90,6 @@ module.exports = cdb.core.Model.extend({
 
     this.on('change:source change:sql', this._onPosibleLayerSchemaChanged, this);
 
-    if (this.infowindowModel) {
```
alonsogarciapablo: Why did we move this to `infowindow-view.js`? IMO this logic is more model-related stuff and it shouldn't be handle by a view.

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindow-form-view.js
```diff
@@ -30,10 +30,10 @@ module.exports = cdb.core.View.extend({
     }
 
     this._formModel = new InfowindowFormModel({
-      width: this._layerInfowindowModel.get('width'),
+      width: this._layerInfowindowModel.get('width') || '',
       headerColor: {
         color: {
-          fixed: '#CCC'
+          fixed: this._layerInfowindowModel.get('headerColor') ? this._layerInfowindowModel.get('headerColor').color.fixed : '#35AAE5;'
```
alonsogarciapablo: We could move this logic the model perhaps and do something like:

```
      ...
      {
        fixed: this._layerInfowindowModel.getHeaderColor();
      }
      ...
```

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindow-select-view.js
alonsogarciapablo: Are we going to need this?

lib/assets/javascripts/cartodb3/data/layer-definition-model.js
matallo: @javisantana suggested this 🤔

javisantana: @alonsogarciapablo if we keep that in the model (as it works in the current editor) we have a pretty big problem with race conditions. It happened in the editor2.0 and we have crazy hacks to avoid that.

as a general rule: `save` should only be raised by a user action not by other model change.

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindow-select-view.js
```diff
@@ -39,14 +49,46 @@ module.exports = cdb.core.View.extend({
...
+
+    if (template_name === 'infowindow_light_header_blue' && color) {
+      template = this._transformTemplate(template, color);
+    }
+
+    return template;
+  },
+
+  _setTemplate: function (template_name) {
+    this._layerInfowindowModel.setTemplate(template_name, this._getTemplate(template_name));
```
alonsogarciapablo: I'd change this a bit and move `_getTemplate` and `_transformTemplate` to the layerInfowindowModel (these methods are not using anything from this view). That way, we can do just:

```
this._layerInfowindowModel.setTemplate(template_name);
```

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindow-view.js
```diff
@@ -8,9 +9,23 @@ var ScrollView = require('../../../../components/scroll/scroll-view');
 
 module.exports = cdb.core.View.extend({
   initialize: function (opts) {
+    if (!opts.layerInfowindowModel) throw new Error('layerInfowindowModel is required');
```
alonsogarciapablo: I'd pass this as `opts.model`. This will set `this.model` automatically (so you will have to replace `this._layerInfowindowModel` by `this.model`). The main benefit is that everything will be automagically unbound from this model when this view gets cleaned up.

alonsogarciapablo:
I tried this and it's working like a charm 👍 . Good job!

I finished the CR: I left some comments and I'd personally like to see some more testing around this, but I'll now leave it up to @matallo to change it.

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindow-select-view.js
matallo: not at all, doing at cartodb.js 🔥

matallo:
🚀

PR #7513, opened 2016-05-25
Op545 test ldap rake task

javitonino:
- Extracts create/test ldap parameter parsing code to `ldap_configuration_from_environment`
- Improves test ldap task with more checks (searches, login) and returns a JSON for automatic testing
- Adds a small `reset_ldap_configuration` task to reset configuration/disabling LDAP

### DIFF
--- a/app/models/carto/ldap/configuration.rb
+++ b/app/models/carto/ldap/configuration.rb
@@ -45,16 +45,16 @@ class Carto::Ldap::Configuration < ActiveRecord::Base
   attr_readonly :user_id_field
 
   validates :organization, :host, :port, :connection_user, :connection_password, :user_id_field, :username_field,
-  :email_field, :user_object_class, :group_object_class, :presence => true
+            :email_field, :user_object_class, :group_object_class, presence: true
 
-  validates :ca_file, :length => { :minimum => 0, :allow_nil => true }
+  validates :ca_file, length: { minimum: 0, allow_nil: true }
 
-  validates :encryption, :inclusion => { :in => [ ENCRYPTION_SIMPLE_TLS, ENCRYPTION_START_TLS ], :allow_nil => true }
-  validates :ssl_version, :inclusion => { :in => [ ENCRYPTION_SSL_VERSION_TLSV1_1 ], :allow_nil => true }
+  validates :encryption,  inclusion: { in: [ENCRYPTION_SIMPLE_TLS, ENCRYPTION_START_TLS], allow_nil: true }
+  validates :ssl_version, inclusion: { in: [ENCRYPTION_SSL_VERSION_TLSV1_1], allow_nil: true }
   validate :domain_bases_not_empty
 
   def domain_bases_list
-    self.domain_bases.split(DOMAIN_BASES_SEPARATOR)
+    domain_bases.split(DOMAIN_BASES_SEPARATOR) if domain_bases
   end
 
   def domain_bases_list=(list)
@@ -121,8 +121,8 @@ class Carto::Ldap::Configuration < ActiveRecord::Base
   private
 
   def domain_bases_not_empty
-    errors.add(:domain_bases, "No domain bases set") if self.domain_bases.to_s.length == 0
-    errors.add(:domain_bases_list, "Domain bases list empty") if domain_bases_list.length < 1
+    errors.add(:domain_bases, "No domain bases set") unless domain_bases.present?
+    errors.add(:domain_bases_list, "Domain bases list empty") unless domain_bases_list.present?
   end
 
   def search_in_domain_bases(filter)
--- a/lib/tasks/ldap.rake
+++ b/lib/tasks/ldap.rake
@@ -1,40 +1,68 @@
 namespace :cartodb do
   namespace :ldap do
-    desc "Tests an LDAP connection"
-    task :test_ldap_connection, [] => :environment do |t, args|
+    desc "Tests an LDAP connection. Returns a summary of all tests ran, with success status and error messages."
+    task :test_ldap_connection, [:from_database] => :environment do |_t, args|
+      args.with_defaults(from_database: false)
 
-      raise "Missing HOST" if ENV['HOST'].blank?
-      host = ENV['HOST']
+      test_user     = ENV['TEST_USER']
+      test_password = ENV['TEST_PASSWORD']
+      ldap = args.from_database ? Carto::Ldap::Configuration.first : ldap_configuration_from_environment
 
-      raise "Missing PORT" if ENV['PORT'].blank?
-      port = ENV['PORT']
+      # Test configuration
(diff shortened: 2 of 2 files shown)


javitonino:
Please CR @juanignaciosl

javitonino:
Changing CR @gfiorav

app/models/carto/ldap/configuration.rb
```diff
@@ -121,8 +121,8 @@ def last_operation_result
   private
 
   def domain_bases_not_empty
-    errors.add(:domain_bases, "No domain bases set") if self.domain_bases.to_s.length == 0
-    errors.add(:domain_bases_list, "Domain bases list empty") if domain_bases_list.length < 1
+    errors.add(:domain_bases, "No domain bases set") unless domain_bases.present?
+    errors.add(:domain_bases_list, "Domain bases list empty") unless domain_bases_list.present?
```
gfiorav: super minor: maybe `.empty?` is more fitting for a list?

lib/tasks/ldap.rake
```diff
@@ -1,107 +1,161 @@
...
-      raise "Missing USER_OBJECT_CLASS" if ENV['USER_OBJECT_CLASS'].blank?
-      user_object_class = ENV['USER_OBJECT_CLASS']
+  private
+
+  def ldap_configuration_from_environment(organization)
+    # Mandatory: connection parameters
+    host = ENV['HOST']
+    port = ENV['PORT']
+    ssl_version = ENV['SSL_VERSION'].blank? ? nil : ENV['SSL_VERSION']
+    encryption = ENV['ENCRYPTION'].blank? ? nil : ENV['ENCRYPTION']
```
gfiorav: Still thinking this naming is to generic!

app/models/carto/ldap/configuration.rb
javitonino: Yup, I refactored it to `empty?` at first, but `domain_bases_list` can be nil. I was tempted to wrap this two validations in a `domain_bases.present?` check (presence is already validated), so this validations only triggered once the domain bases were set, but I was not convinced.

lib/tasks/ldap.rake
```diff
@@ -1,107 +1,161 @@
...
+      Carto::Ldap::Configuration.delete_all
+    end
+  end
 
-      raise "Missing USER_OBJECT_CLASS" if ENV['USER_OBJECT_CLASS'].blank?
-      user_object_class = ENV['USER_OBJECT_CLASS']
+  private
+
+  def ldap_configuration_from_environment(organization)
+    # Mandatory: connection parameters
```
gfiorav: Maybe we should raise if environment variables are not present... I guess it will fail further down the stack anyway

lib/tasks/ldap.rake
javitonino: I would have prefixed all four env variables above with LDAP_, as they are quite generic. Still not convinced on changing this, as it may break scripts. Will ask around to check if anyone knows whether this is the case,

lib/tasks/ldap.rake
```diff
@@ -1,107 +1,161 @@
...
+      ssl_version:          ssl_version,
+      connection_user:      connection_user,
+      connection_password:  connection_password,
+      user_id_field:        user_id_field,
+      username_field:       username_field,
+      email_field:          email_field,
+      domain_bases:         domain_bases,
+      user_object_class:    user_object_class,
+      group_object_class:   group_object_class
+    )
```
gfiorav: I'm assuming that `Carto::Ldap::Configuration` ignores params when passed as `nil` right? I mean, for sure if they're not specified, but when specified as `nil`?

lib/tasks/ldap.rake
javitonino: We raise in https://github.com/CartoDB/cartodb/pull/7513/files#diff-4d736b6864b01685bdc4803b98d0e142R91 for the creation task. I cannot raise here except for the first 4 parameters, because almost everything is optional for the test task.

lib/tasks/ldap.rake
javitonino: They are set as nil (same as if not passed), and validation fails. Same thing for empty strings (e.g: `HOST= PORT=389` fails validation on host).

lib/tasks/ldap.rake
```diff
@@ -1,107 +1,161 @@
...
+
+      REQUIRED_FIELDS = [:host, :port, :connection_user, :connection_password].freeze
+      if !config_result[:success] && config_result[:error].keys.any? { |k| REQUIRED_FIELDS.include?(k) }
+        result[:connection] = { success: false, error: 'Configuration not valid' }
+      end
 
-      if result[:success]
-        puts "OK"
+      # Test connection
+      result[:connection] = ldap.test_connection unless result[:connection]
```
gfiorav: What timeout does this have?

lib/tasks/ldap.rake
```diff
@@ -1,107 +1,161 @@
...
+      config_result = if ldap.valid?
+                        { success: true }
+                      else
+                        { success: false, error: ldap.errors.to_h }
+                      end
+
+      result = { config: config_result }
+
+      REQUIRED_FIELDS = [:host, :port, :connection_user, :connection_password].freeze
+      if !config_result[:success] && config_result[:error].keys.any? { |k| REQUIRED_FIELDS.include?(k) }
```
gfiorav: Can't see why you check for any error regarding `REQUIRED_FIELDS`, isn't _any_ error enough?

lib/tasks/ldap.rake
javitonino: Ruby's net module default timeouts: 60s.

app/models/carto/ldap/configuration.rb
gfiorav: 🌴

lib/tasks/ldap.rake
gfiorav: This should be in sync with the timeout for the form

lib/tasks/ldap.rake
gfiorav: 🌴

lib/tasks/ldap.rake
gfiorav: 🌴

lib/tasks/ldap.rake
gfiorav: Ok then 🌴

gfiorav:
📦 it!

lib/tasks/ldap.rake
javitonino: No, as long as we have this 4 fields we can do a connection test. Other fields are used for specific tests: user login, object searches etc, and are checked in each tests. This way, we can run the tests without writing a lot of unneeded parameters.

Basically, is a shortcut for writing less for manual testing of LDAP.

PR #7527, opened 2016-05-26
Disable import from external services when not configured

javitonino:
Disables import services (dropbox/gdrive & co) when they are not configure (e.g: offline). As a bonus, it disables the tab navigation when not needed, and hides it completely when no service is configured (only data file available).

screenshot_20160526_145507

screenshot_20160526_145536


javitonino:
Please CR @xavijam. I may have been overeager hiding the navigation & tabs. It can be easily reverted if needed, as it changes the design a bit.

javitonino:
Hold that review! Making some changes...

javitonino:
Now with more new editor compatibility. All ready for CR @xavijam

xavijam:
Looks good to me :)

PR #7543, opened 2016-05-27
7171 CartoDBfy tables before switching in Sync Tables

### LINKED ISSUE #7171: Cartodbfying process should happen before overwriting the final table on sync tables
Right now the sync tables are first overwritten and then cartodbfied, which causes the sync table to not have cartodb_id or proper columns for a small amount of time during the sync proess.

cc @CartoDB/dataservices

iriberri:
Fixes #7171 

## The problem
Currently, the synchronization workflow is as follows:
  1. The new table is imported, it is never registered
  2. The old and the new tables are switched. The old table is dropped.
  3. CartoDBfication, import cleanups (cartodb_id) and geom checks (multi/simple geoms) are performed

And this provokes several problems:
  * When the switch is performed, the new table is not CartoDBfied (no cartodb_id, no the_geom) so any maps that contain this table as a layer will crash. See #7171 for a specific example.
  * If there's an error in the CartoDBfication or the rest of the checks, the synchronization will leave a broken table. The old table has been dropped so there's no possible rollback. The maps will be broken.
  * Also, if there's an error in the CartoDBfication, when the Ghost Tables are triggered they will unregister the table, breaking completely the synchronization. This would remove any layer in which the disconnected table was involved too.

When working in the issue, I noticed that the current flow is working due to the fact that when the switch between tables is performed, the new table gets automatically registered (it reuses the register of the old table) so the operations in the Table model that require the Table object can be performed on them.

As a side-effect, as the new table was not CartoDBfied it would not be captured when the ghost tables were triggered.

## The solution
We had to move the CartoDBfy process (and the rest of the checks) before the switch is performed to ensure that when it happens, the new table will be completely compatible with the maps and the Editor.

By "moving" these processes before the switch, we no longer have a registered table to work with. This is the reason why we duplicated code from Table (`cartodbfy`, `set_the_geom_column!` and `import_cleanup`) and adapted it to work with unregistered tables. Basically in this approach I am getting the column information via PostgreSQL queries, instead of checking the schema of the Table.

This code duplication is intended to be only part of a first iteration. Some parts of these methods can be directly removed with a correct configuration of Ogr2Ogr or the original methods can moved outside of Table.rb to create some kind of utility that can be used for both registered and unregistered tables.

To be able to CartoDBfy the table before switching, we had to move the table to the user schema. Some internal code of the CartoDBfication performs queries on CDB_UserQuotaSize and other functions that are expected to live in the schema in which the table lives, and `cdb_importer`, which is the default schema in which new tables are created does not contain these.

We also had to work with Ghost Tables. As the table is CartoDBfied and lives in the user schema, it would be visible for the Ghost Tables task. We're adding a commit to ignore tables in the User DB that are named as our `importer` tables.

Now the process is as follows:

  1. The new table is imported
  2. It is moved to the user schema
  3. It is CartoDBfied + import_cleanup + geom fixes(cartodb_id, the_geom)
  4. It is switched by the old table of the user

In this scenario, if something goes wrong while manipulating the table (incorrect CartoDB ID, or multiple geometry types issues, which were reported by Jaak) an exception will be raised and the sync will stop before any switch is made.

## Testing

Output of some tests: https://gist.github.com/iriberri/42a1a44a47f9ee5e428901dcd4a125d5

Other similar branches I created just for test purposes (see gist above):
* `7171-cdbfy_in_syncs_b`
* `cdbfy_master_with_sleep`

### Timing checks
I got some timings for the query edited in the `ghost_tables_manager` for my own production user with 2564 tables in the User DB.

* Original query, no name checks:
  * Time: 240.853 ms
* Adapted query 1: `t.tablename !~ 'importer_\\w{32}'`
  * Time: 364.519 ms
* Adapted query 2 (the one currently commited): `t.tablename !~ '^importer_'`
  * Time: 277.863 ms
* Adapted query 3: `t.tablename !~ '^importer_\\w{32}'`
  * Time: 278.990 ms

====

Pending: 
* [x] drop table in failure scenario to not leave stale tables
* [x] test thoroughly for cartodbfycation exceptions and special geom types (multipoint, simple linestrings) to check postgres queries flow
* [x] get some numbers on ghost tables query times

### DIFF
--- a/app/models/synchronization/adapter.rb
+++ b/app/models/synchronization/adapter.rb
@@ -202,7 +202,7 @@ module CartoDB
       def import_cleanup(schema_name, table_name)
         qualified_table_name = "\"#{schema_name}\".#{table_name}"
 
-        user.transaction_with_timeout(statement_timeout: STATEMENT_TIMEOUT) do |user_database|
+        user.db_service.in_database_direct_connection(statement_timeout: STATEMENT_TIMEOUT) do |user_database|
           # When tables are created using ogr2ogr they are added a ogc_fid or gid primary key
           # In that case:
           #  - If cartodb_id already exists, remove ogc_fid
--- a/app/models/synchronization/adapter.rb
+++ b/app/models/synchronization/adapter.rb
@@ -149,6 +149,7 @@ module CartoDB
           WHERE attrelid = '#{qualified_table_name}'::regclass
           AND attname = '#{THE_GEOM}'
           AND a.atttypid = t.oid
+          AND a.attstattarget < 0
           LIMIT 1
         }].first
         return nil unless the_geom_data
@@ -213,6 +214,7 @@ module CartoDB
             WHERE attrelid = '#{qualified_table_name}'::regclass
             AND (a.attname = 'ogc_fid' OR a.attname = 'gid')
             AND a.atttypid = t.oid
+            AND a.attstattarget < 0
             LIMIT 1
           }].first
           aux_cartodb_id_column = aux_cartodb_id_column[:attname] unless aux_cartodb_id_column.nil?
--- a/app/models/synchronization/adapter.rb
+++ b/app/models/synchronization/adapter.rb
@@ -103,6 +103,7 @@ module CartoDB
                               exception: exception,
                               user: user,
                               table: table_name)
+        drop(table_name) if exists?(table_name)
         raise exception
       end
 
--- a/app/models/synchronization/adapter.rb
+++ b/app/models/synchronization/adapter.rb
@@ -42,6 +42,7 @@ module CartoDB
         puts exception.to_s
         puts exception.backtrace
         puts '=================='
+        raise exception
       end
 
       def overwrite(table_name, result)
--- a/app/models/synchronization/adapter.rb
+++ b/app/models/synchronization/adapter.rb
@@ -135,11 +135,12 @@ module CartoDB
         fix_oid(table_name)
       end
 
+      # From Table#get_the_geom_type!, adapted to unregistered tables
+      # returns type to run Table#get_the_geom_type! afterwards again, which
+      # saves the type in table metadata
       def fix_the_geom_type!(schema_name, table_name)
         qualified_table_name = "\"#{schema_name}\".#{table_name}"
 
-        # Partial duplication of Table#set_the_geom_column! to be able to
-        # operate with non-registered tables.
         type = nil
         the_geom_data = user.in_database[%Q{
           SELECT a.attname, t.typname
@@ -196,10 +197,10 @@ module CartoDB
         type
       end
 
+      # From Table#import_cleanup, with column schema checks adapted to unregistered tables
       def import_cleanup(schema_name, table_name)
         qualified_table_name = "\"#{schema_name}\".#{table_name}"
 
-        # From Table#import_cleanup, with column schema checks adapted to unregistered tables
         user.in_database_direct_connection(statement_timeout: STATEMENT_TIMEOUT) do |user_database|
           # When tables are created using ogr2ogr they are added a ogc_fid or gid primary key
           # In that case:
--- a/app/models/synchronization/adapter.rb
+++ b/app/models/synchronization/adapter.rb
@@ -202,7 +202,7 @@ module CartoDB
       def import_cleanup(schema_name, table_name)
         qualified_table_name = "\"#{schema_name}\".#{table_name}"
 
-        user.in_database_direct_connection(statement_timeout: STATEMENT_TIMEOUT) do |user_database|
+        user.transaction_with_timeout(statement_timeout: STATEMENT_TIMEOUT) do |user_database|
           # When tables are created using ogr2ogr they are added a ogc_fid or gid primary key
           # In that case:
           #  - If cartodb_id already exists, remove ogc_fid
--- a/app/models/synchronization/adapter.rb
+++ b/app/models/synchronization/adapter.rb
@@ -211,7 +211,7 @@ module CartoDB
             SELECT a.attname, t.typname
             FROM pg_attribute a, pg_type t
             WHERE attrelid = '#{qualified_table_name}'::regclass
-            AND a.attname = 'ogc_fid' OR a.attname = 'gid'
+            AND (a.attname = 'ogc_fid' OR a.attname = 'gid')
             AND a.atttypid = t.oid
             LIMIT 1
           }].first
--- a/app/models/synchronization/adapter.rb
+++ b/app/models/synchronization/adapter.rb
@@ -291,7 +291,7 @@ module CartoDB
 
       def move_to_schema(result, schema=DESTINATION_SCHEMA)
         # The new table to sync is moved to user schema to allow CartoDBfication.
-        # This temporal table should not be registered (check ghost_tables_manager.rb)
+        # This temporary table should not be registered (check ghost_tables_manager.rb)
         return self if schema == result.schema
         database.execute(%Q{
           ALTER TABLE "#{result.schema}"."#{result.table_name}"
--- a/app/models/synchronization/adapter.rb
+++ b/app/models/synchronization/adapter.rb
@@ -4,6 +4,8 @@ module CartoDB
   module Synchronization
     class Adapter
       DESTINATION_SCHEMA = 'public'
+      STATEMENT_TIMEOUT = 1.hour * 1000
+      THE_GEOM = 'the_geom'.freeze
 
       attr_accessor :table
 
@@ -27,8 +29,10 @@ module CartoDB
           end
           copy_privileges(user.database_schema, table_name, result.schema, result.table_name)
           index_statements = generate_index_statements(user.database_schema, table_name)
+          move_to_schema(result)
+          geo_type = cartodbfy(result.table_name)
           overwrite(table_name, result)
-          cartodbfy(table_name)
+          setup_table(table_name, geo_type)
           run_index_statements(index_statements)
         end
         self
@@ -42,7 +46,6 @@ module CartoDB
 
       def overwrite(table_name, result)
         return false unless runner.remote_data_updated?
-
         temporary_name = temporary_name_for(result.table_name)
 
         # The relation might (and probably will) already exist in the user public schema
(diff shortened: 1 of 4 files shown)


app/models/synchronization/adapter.rb
```diff
@@ -105,6 +135,53 @@ def cartodbfy(table_name)
...
+            user_database.run("SELECT DropGeometryColumn('#{owner.database_schema}', '#{self.name}','the_geom');")
+            user_database.run(%Q{ALTER TABLE #{qualified_table_name} RENAME COLUMN the_geom_simple TO the_geom;})
+          end
+        type = 'point'
+        end
+
+        #if the geometry is LINESTRING or POLYGON we convert it to MULTILINESTRING and MULTIPOLYGON resp.
+        if %w(linestring polygon).include?(type)
+          user.db_service.in_database_direct_connection({statement_timeout: STATEMENT_TIMEOUT}) do |user_database|
+            user_database.run("SELECT public.AddGeometryColumn('#{owner.database_schema}', '#{self.name}','the_geom_simple',4326, 'GEOMETRY', 2);")
```
iriberri: This `self.name` is wrong. Fix all occurrences.

app/models/synchronization/adapter.rb
```diff
@@ -105,6 +135,53 @@ def cartodbfy(table_name)
         update_cdb_tablemetadata(table_name)
       end
 
+      def get_the_geom_type(qualified_table_name)
```
iriberri: Oops.

iriberri:
#<Sequel::DatabaseError: PG::Error: ERROR:  CDB(_CDB_Has_Usable_Primary_ID: multiple primary keys for table "importer_928dc3b4242411e69a3d080027880ca6" are not allowed): ALTER TABLE importer_928dc3b4242411e69a3d080027880ca6 ADD CONSTRAINT cartodb_id_pk PRIMARY KEY (cartodb_id), ADD CONSTRAINT cartodb_id_integer CHECK (cartodb_id::integer >=0);
CONTEXT:  SQL statement "SELECT _CDB_Error(sql, Format('_CDB_Has_Usable_Primary_ID: %s', SQLERRM))"
PL/pgSQL function _cdb_has_usable_primary_id(regclass) line 60 at PERFORM
SQL statement "SELECT _CDB_Has_Usable_Primary_ID(reloid)"
PL/pgSQL function _cdb_rewrite_table(regclass,text) line 50 at SQL statement
SQL statement "SELECT _CDB_Rewrite_Table(reloid, destschema)"
PL/pgSQL function cdb_cartodbfytable(text,regclass) line 51 at PERFORM


```
                                                     Table "public.importer_928dc3b4242411e69a3d080027880ca6"
   Column    |       Type        |                                          Modifiers                                          | Storage  | Stats target | Description 
-------------+-------------------+---------------------------------------------------------------------------------------------+----------+--------------+-------------
 ogc_fid     | integer           | not null default nextval('importer_928dc3b4242411e69a3d080027880ca6_ogc_fid_seq'::regclass) | plain    |              | 
 cartodb_id  | integer           |                                                                                             | plain    |              | 
 the_geom    | geometry          |                                                                                             | main     |              | 
 name        | character varying |                                                                                             | extended |              | 
 description | character varying |                                                                                             | extended |              | 
Indexes:
    "importer_928dc3b4242411e69a3d080027880ca6_pkey" PRIMARY KEY, btree (ogc_fid)
Has OIDs: no
Options: autovacuum_enabled=true, toast.autovacuum_enabled=true


```

iriberri:
https://github.com/CartoDB/cartodb/blob/b9245329c03dfe52db025a2e49d4a87ace7325d6/app/models/table.rb#L301-L360

lib/carto/ghost_tables_manager.rb
```diff
@@ -138,6 +138,7 @@ def fetch_non_raster_cartodbfied_tables
                  count(column_name::text) cdb_columns_count
           FROM information_schema.columns c, pg_tables t, pg_trigger tg
           WHERE
+            t.tablename !~ 'importer_' AND
```
jgoizueta: I'd anchored it to the start: `/\Aimporter_/`

db/migrate/20160516103200_drop_visualization_exports_user_id_null_constraint.rb
```diff
@@ -0,0 +1,16 @@
+Sequel.migration do
```
iriberri: commit retrieved from master as the inexistent migration was breaking my deploy attempts

app/models/synchronization/adapter.rb
```diff
@@ -79,30 +83,186 @@ def fix_oid(table_name)
...
+          FROM pg_attribute a, pg_type t
+          WHERE attrelid = '#{qualified_table_name}'::regclass
+          AND attname = '#{THE_GEOM}'
+          AND a.atttypid = t.oid
+          LIMIT 1
+        }].first
+        return nil unless the_geom_data
+
+        if the_geom_data[:typname] != 'geometry'
+          user.in_database.rename_column(qualified_table_name, THE_GEOM, :the_geom_str)
```
jgoizueta: `CDB_ColumnType()` from the cartodb extension could be used here

app/models/synchronization/adapter.rb
```diff
@@ -79,30 +83,186 @@ def fix_oid(table_name)
...
+          # All normal fields casted to text
+          varchar_columns = user_database[%Q{
+            SELECT a.attname, t.typname
+            FROM pg_attribute a, pg_type t
+            WHERE attrelid = '#{qualified_table_name}'::regclass
+            AND typname = 'varchar'
+            AND a.atttypid = t.oid
+          }].all
+
+          varchar_columns.each do |column|
```
jgoizueta: Wouldn't a single ALTER TABLE to change all the columns be more efficient?

jgoizueta:
:+1:

app/models/synchronization/adapter.rb
```diff
@@ -27,8 +29,10 @@ def run(&tracker)
           end
           copy_privileges(user.database_schema, table_name, result.schema, result.table_name)
           index_statements = generate_index_statements(user.database_schema, table_name)
+          move_to_schema(result)
+          geo_type = cartodbfy(result.table_name)
```
rafatower: So cartodbfy is doing 2 things: actually cartodbfy'ing and returning a geometry type. The Single Responsibility Principle is against that.

https://en.wikipedia.org/wiki/Single_responsibility_principle
http://code.tutsplus.com/tutorials/solid-part-1-the-single-responsibility-principle--net-36074

app/models/synchronization/adapter.rb
```diff
@@ -79,30 +84,189 @@ def fix_oid(table_name)
       end
 
       def cartodbfy(table_name)
+        schema_name = user.database_schema
+        qualified_table_name = "\"#{schema_name}\".#{table_name}"
+
+        type = fix_the_geom_type!(schema_name, table_name)
```
rafatower: I'd move this up to the caller

app/models/synchronization/adapter.rb
iriberri: Very good point. Thanks! It's indeed doing 3 different things, `import_cleanup` is also editing the table. I'll take those away.

app/models/synchronization/adapter.rb
```diff
@@ -79,30 +84,189 @@ def fix_oid(table_name)
...
                               table: table_name)
       ensure
         fix_oid(table_name)
-        update_cdb_tablemetadata(table_name)
+      end
+
+      # From Table#get_the_geom_type!, adapted to unregistered tables
+      # returns type to run Table#get_the_geom_type! afterwards again, which
+      # saves the type in table metadata
+      def fix_the_geom_type!(schema_name, table_name)
```
rafatower: dammit we should have a ticket to do this in CDB_CartodbfyTable some day

app/models/synchronization/adapter.rb
```diff
@@ -79,30 +83,185 @@ def fix_oid(table_name)
...
+          varchar_columns = user_database[%Q{
+            SELECT a.attname, t.typname
+            FROM pg_attribute a, pg_type t
+            WHERE attrelid = '#{qualified_table_name}'::regclass
+            AND typname = 'varchar'
+            AND a.atttypid = t.oid
+          }].all
+
+          varchar_columns.each do |column|
+            user_database.run(%Q{ALTER TABLE #{qualified_table_name} ALTER COLUMN "#{column[:attname]}" TYPE text})
```
rafatower: I am more than curious about knowing why this decision was taken (converting to text)

rafatower: also noteworthy for CDB_CartodbfyTable

app/models/synchronization/adapter.rb
```diff
@@ -126,6 +290,8 @@ def checksum
       end
 
       def move_to_schema(result, schema=DESTINATION_SCHEMA)
+        # The new table to sync is moved to user schema to allow CartoDBfication.
+        # This temporal table should not be registered (check ghost_tables_manager.rb)
```
rafatower: s/temporal/temporary

app/models/synchronization/adapter.rb
iriberri: Introduced in 2012: https://github.com/CartoDB/cartodb/commit/1a986c985a7bc7d34364ce2a0ecc95ff12c74f8e

Related: https://github.com/CartoDB/cartodb/blob/b9245329c03dfe52db025a2e49d4a87ace7325d6/config/initializers/string.rb#L142

iriberri:
@rafatower thank you very much for the feedback! I just applied the `cartodbfy` changes here: https://github.com/CartoDB/cartodb/pull/7543/commits/16ab6ce28e6bc801048ae7946ed5b1eafc75dcfc

Won't squash the commits yet so that these changes are isolated.

rafatower:
:+1:

rafatower:
Adding @gfiorav to the party, especially regarding this: https://github.com/CartoDB/cartodb/pull/7543/files#diff-ed41caec536553c86246a629a466a888

See the comment above about performance.

As commented in person, we know a better approach is to use CDB_TableMetadata and mark the tables there upon CDB_CartodbfyTable (and it would be more performant, checking against catalog oids rather than doing string comparisons) but we evaluated it and this solution is far simpler and effective FTM.

lib/carto/ghost_tables_manager.rb
```diff
@@ -138,6 +138,7 @@ def fetch_non_raster_cartodbfied_tables
                  count(column_name::text) cdb_columns_count
           FROM information_schema.columns c, pg_tables t, pg_trigger tg
           WHERE
+            t.tablename !~ '^importer_' AND
```
gfiorav: 📦 ready for me. I don't think the effect will be perceptible. 

Please ping me when it's merged so I benchmark some known heavy users. Last refactor of GhostTablesManager reduced run time from 40 seconds (for 2500 datasets) to 3 seconds. I'm almost 100% sure that this little change won't increase that time but just in case.

Thanks for the ping!

PR #7552, opened 2016-05-27
Kickstart basemaps

matallo:
CR @javisantana 

as a comment, 👏 for the stackLayout, it's been so easy to start working on this new view

/cc @xavijam @viddo


matallo:
I recreated the carousel component as a mosaic (the little squares)

@xavijam would it be worth it consolidating in a single component? Also, I had to do one slight change as with the current design the mosaic item can be _not-selected_, _not-highlighted_ (i.e. I change the source, and hover with the mouse see attached)

screen shot 2016-05-30 at 19 28 14

@javisantana @viddo is the `basemaps` well placed in the files structure? I tried to follow the same patterns of other objects in the `show` view

right now it only works with tiled basemaps, next would be adding Plain basemaps, but if it's ok in general I'd add tests then ask for merge

thanks!

lib/assets/javascripts/cartodb3/components/custom-mosaic/custom-mosaic-collection.js
```diff
@@ -0,0 +1,50 @@
...
+      this.each(function (m) {
+        if (m.cid !== changedModel.cid && m.get('selected')) {
+          m.set('selected', false);
+        }
+      });
+    }
+  },
+
+  getSelected: function () {
+    return _.first(this.where({ selected: true }));
```
viddo: Just a detail, but could be done simpler by `return this.find(function(m) { return m.get('selected'); })`

viddo: Actually, `return this.findWhere({ selected: true })` for attr selection

lib/assets/javascripts/cartodb3/editor/layers/basemap-layer-view.js
```diff
@@ -24,5 +32,16 @@ module.exports = cdb.core.View.extend({
     }));
 
     return this;
+  },
+
+  _onEditBasemap: function (e) {
+    e.stopPropagation();
+
+    if (this._preventEditClick) {
```
viddo: I don't see this var being set anywhere, is it really necessary?

viddo:
Looks reasonable 👍

viddo:
Btw, I find the `custom-` prefix on files/components to be superfluous, being a component implies that it should be customizable already, custom would/should the exception.

lib/assets/javascripts/cartodb3/editor/layers/basemap-layer-view.js
matallo: oopsie, not really as it's used for _dragging_ and basemaps won't drag

lib/assets/javascripts/cartodb3/components/custom-mosaic/custom-mosaic-collection.js
matallo: roger

matallo: actually, this didn't work as expected (maybe I did something wrong), reverting to @xavijam version as in Carousel

matallo:
merging as soon as it gets 💚 

right now just updating the url for tilelayer basemaps, next is working with labels-on-top basemaps and replacing the whole layer

Issue #7329, opened 2016-05-09
styling available datasets


javisantana:
New styling options have aggregation capablities so a points table can be shown aggregated by zipcode, adm0 adm1 and so on.

In order to aggregate we need to have those datasets as FDW tables in user account.

We need at least the following tables (please, complete the list @saleiva @urbanphes )
- adm0
- adm1
- postal codes

I'd need to names and schema of those tables and also the bbox where they apply, like

``` js
[{
table: 'adm0',
bbox: [ [ xmin, ymin, xmax, ymax], ...] //list of bbox where the dataset can be used 
}]
```

saleiva:
These will come from the DO, right?

Ideally we will have all the boundaries that we have in the what's on next
on the DO.

2016-05-09 11:44 GMT+02:00 javi santana <notifications@github.com>:

> New styling options have aggregation capablities so a points table can be
> shown aggregated by zipcode, adm0 adm1 and so on.
>
> In order to aggregate we need to have those datasets as FDW tables in user
> account.
>
> We need at least the following tables (please, complete the list @saleiva
> <https://github.com/saleiva> @urbanphes <https://github.com/urbanphes> )
>
>    - adm0
>    - adm1
>    - postal codes
>
> I'd need to names and schema of those tables and also the bbox where they
> apply, like
>
> [{
> table: 'adm0,bbox: [ [ xmin, ymin, xmax, ymax], ...] //list of bbox where the dataset can be used }]
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/CartoDB/cartodb/issues/7329>
>



-- 
Sergio Alvarez Leiva

rafatower:
@saleiva  the tables mentioned by @javisantana are the ones used by the internal geocoder, not the boundaries used by DO.

javisantana:
I'd use tables from DO and not internal geocoder, at the end the internal geocoder should be using DO tables as a source

saleiva:
exactly

2016-05-09 12:18 GMT+02:00 javi santana <notifications@github.com>:

> I'd use tables from DO and not internal geocoder, at the end the internal
> geocoder should be using DO tables as a source
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/CartoDB/cartodb/issues/7329#issuecomment-217827844>
>



-- 
Sergio Alvarez Leiva

javisantana:
[...]

PR #7559, opened 2016-05-30
changed regions generator to support aggregation FDW datasets

javisantana:
add support for this #7329

@xavijam for the moment we only have countries and provinces


xavijam:
Good enough for the moment

PR #7567, opened 2016-05-30
Cartocss parser

nobuti:
This PR implements the css parser.

parsecss

cc @xavijam


javisantana:
it would awesome if you could add support for turbocarto functions :D

xavijam:
Yeah! please please please! ^^^

lib/assets/javascripts/cartodb3/editor/components/code-mirror/code-mirror-view.js
```diff
@@ -74,5 +86,53 @@ module.exports = cdb.core.View.extend({
   clean: function () {
```
xavijam: I know clean has not been touched, but could we move this method at the end of the class? I know, I'm a maniac :joy:.

lib/assets/javascripts/cartodb3/editor/components/code-mirror/code-mirror-view.js
```diff
@@ -74,5 +86,53 @@ module.exports = cdb.core.View.extend({
...
+    $line.find('.CodeMirror-linenumber').addClass('has-error');
+    this._lineWithErrors.push($line);
+  },
+
+  _renderError: function (error) {
+    this._getConsole().append(errorTemplate(error));
+  },
+
+  _renderNoError: function () {
+    this._getConsole().html('No errors');
```
xavijam: Let's translate this text please 👼

lib/assets/javascripts/cartodb3/editor/components/code-mirror/code-mirror-view.js
nobuti: Oh, I thought we follow the public first, then private convention. I'll fix it.

lib/assets/javascripts/cartodb3/editor/style/style-view.js
```diff
@@ -61,15 +62,18 @@ module.exports = cdb.core.View.extend({
...
-
-    this._cartocssModel.set('content', content);
-
-    this._layerDefinitionModel.save({
-      cartocss_custom: true,
-      cartocss: content
-    });
+    var parser = new ParserCSS(content);
+    var errors = parser.errors();
+    this._codemirrorModel.set('errors', parser.parseError(errors));
```
xavijam: 👍

xavijam:
Not sure if we have a better link for docs, but: https://docs.cartodb.com/cartodb-platform/cartocss/

lib/assets/javascripts/cartodb3/editor/components/code-mirror/code-mirror-view.js
xavijam: That's a not bad idea, it is true, but for clean method I like to have it at the end, I don't know :).

xavijam:
BTW, looks fantastic :)

lib/assets/javascripts/cartodb3/editor/components/code-mirror/code-mirror-error.tpl
```diff
@@ -0,0 +1,8 @@
+<ul class="CDB-InfoBox-footer CodeMirror-error">
+  <li class="CDB-InfoBox-footerItem CodeMirror-errorMessage u-lSpace--xl u-rSpace--xl">
+    Syntax error. Line <%= line %>: <span><%= message %></span>
```
xavijam: Do we need <%= %> or with <%- %> is enough?

package.json
```diff
@@ -38,6 +38,8 @@
     "brfs": "^1.4.3",
     "browserify": "13.0.0",
     "browserify-resolutions": "1.1.0",
+    "carto": "https://github.com/CartoDB/carto/archive/master.tar.gz",
```
xavijam: pretty weird dependency, is it ok?

package.json
```diff
@@ -38,6 +38,8 @@
...
+    "torque.js": "2.15.1",
```
xavijam: Adding torque as dependency, could we have any problem with the current editor torque?
cc @viddo @rochoa @javisantana

viddo: If the dependency is not matching [cartodb.js](https://github.com/CartoDB/cartodb.js/blob/v4/package.json#L38) it can lead to weird mismatches between the libraries (and increase the bundle size due to duplicate code in multiple versions), probably should set it up like how cartodb.js/deep-insights is setup.

Btw, these should probably be defined under dependencies, not dev-dependencies.

package.json
nobuti: https://github.com/CartoDB/cartodb.js/blob/v4/package.json#L40 declares it the same way

package.json
xavijam: Ok, if we move them to dependencies we should be safe.

viddo: No, that's not enough. While it's safe _for now_, it will fail eventually (e.g. when cartodb.js updates version) and we forget to update here. That's why I suggest to use some more relaxed versioning in both places (e.g. `^2.15`), just like we done previously for cdb+DI (before @nobuti joined). See [previous conversation about this topic](https://github.com/CartoDB/cartodb.js/pull/1199#issuecomment-197777889) for more details/examples.

xavijam:
retest this please

PR #7581, opened 2016-05-31
Layer dragging and dropping

alonsogarciapablo:
Part of https://github.com/CartoDB/cartodb/issues/7342.

drag droppin

@javierarce 👍  or 👎  ? thx!


lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
```diff
@@ -11,6 +12,69 @@ var layerTypesAndKinds = require('./layer-types-and-kinds');
...
+  },
+
+  _resetOrders: function () {
+    this.each(function (layerDefModel, index) {
+      layerDefModel.set('order', index, { silent: true });
+    });
+  },
+
+  save: function (options) {
+    var promises = this.chain()
```
alonsogarciapablo: @viddo FYI: this is how I'm making sure analysis definition nodes are created before the collection of layers is saved (now triggered after user re-orders some layers).

I'll open a separate issue to discuss if we should do this in bulk and if we should make this operation atomic.

viddo: 👌

app/assets/stylesheets/editor-3/layers-list.css.scss
```diff
@@ -14,6 +14,9 @@
   background: #FFF;
   position:relative;
 }
+.Editor-ListLayer-item--placeholder {
+  background: #F2F6F9;
```
javierarce: This color is almost white, right? Not sure if this is right or we could reduce a bit the opacity.

javierarce:
👍  (I added just a super minor comment, though)

app/assets/stylesheets/editor-3/layers-list.css.scss
alonsogarciapablo: I used a color picker to pick the color we have in the mockups. Not sure about this. What color would you recommend? Here's the actual color:

screen shot 2016-06-01 at 11 34 49

alonsogarciapablo: Asking @piensaenpixel about this.

javierarce: Oh, that color is for the placeholder! I thought it was for a different thing, sorry. Then I guess it's fine.

alonsogarciapablo: I changed a bit anyway in https://github.com/CartoDB/cartodb/pull/7581/commits/4fa353ae6541798a12db82b810ae5fc3329678da. I replaced color by existing variable and changed the color of the border to match the background color. Thanks @piensaenpixel!

alonsogarciapablo:
Thanks for the CR @javierarce! Merging after all tests are green.

PR #7582, opened 2016-05-31
Do not vendorize and include odyssey in cdb.js and embed.js

alonsogarciapablo:
I was seeing some differences locally in:

`vendor/assets/javascripts/cartodb.mod.odyssey.uncompressed.js`

I was going to just commit those changes but then I figured we can just get rid of oddysey in the editor since we're not using it AFAIK.

Can we @xavijam @viddo @javierarce @javisantana?


xavijam:
Odyssey is working in the current editor, but not sure if we have "enabled" that functionality to any external user.

javisantana:
"working"

Just remove everything related with odyssey, if you need to remove things from the current editor, just do it

PR #7586, opened 2016-05-31
Reverse order of layerDefs in layer-definitions-collection to match the order attribute

alonsogarciapablo:
Also, it now matches the order of the collection of layers in CartoDB.js.

cc: @matallo


matallo:
🚢 🇮🇹

PR #7578, opened 2016-05-31
7494 analysis form models

viddo:
Yes, a big PR because the issues above were all related to the problems of current implementation, so solved them in one go.

The main benefit, though, is that it will be easy for others (who don't know how things are wired together), to add new analyses 🎉 they just need to:

1. add a [new option entry](https://github.com/CartoDB/cartodb/blob/7494-analysis-form-models/lib/assets/javascripts/cartodb3/components/modals/add-analysis/default-options.js)
1. Implement a [form model(+view to go with it) representing the new type](https://github.com/CartoDB/cartodb/blob/7494-analysis-form-models/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-forms-collection.js#L6-L10)

More details below:

**Closes #7494 - Change analysis nodes creation/update to work with form models**
The layer-contenta-analyses-view now works with form models instead of directly interacting with the node definitions. This have the benefit of having invalid models w/ trouble, and representing editing state in a simpler way and not polluting the node models.

Also simplifies add-analysis modal, since it now just need to add metadata about an analysis. In theory we should be able to move this to the camshaft reference even, having all the the options generated from the metadata. This also solved the bug of showing the secondary source as an error when creating a new point-in-polygon analysis. 🎉 

Validation is now also done using the params restrictions defined in the [camshaft reference](https://github.com/CartoDB/camshaft/blob/master/reference/versions/0.10.0/reference.json)

**Closes #7517 - [analysis] trade areas not working with time**
The form converts the model's values to strings (`123`, `false`, etc.). Solved by doing conversion based on the [type definitions in the camshaft reference](https://github.com/CartoDB/camshaft/blob/master/reference/versions/0.10.0/reference.json#L13).

**Closes #7468 - Render analysis controls view at bottom**
Changed the layout structure, and could remove some unnecessary views. The hierarchy now is basically:

```
LayerContentAnalyses
  ScrollView
    ListView
      Workflow
      Form
  AnalysesControls (apply etc.)
```

Also some general boyscouting while at it, e.g. stack layout bugs, node internals, duplicated code (add-analysis modal creation) etc.

---

Some screens:

screen shot 2016-05-31 at 16 41 57
screen shot 2016-05-31 at 16 42 02
screen shot 2016-05-31 at 16 42 06
screen shot 2016-05-31 at 16 42 20

@xavijam sorry for such a large PR again, but can you review? 🙏 🙇 
cc @javierarce @alonsogarciapablo

PR #7590, opened 2016-05-31
Fixes for updating camshaft reference to v0.10.0

### LINKED ISSUE #7483: buffer analysis not enabled for lines data sources
1) go to a visualization with a layer with lines
2) add analysis

buffer analysis is not enabled.

Looks like a simple issue but it's not. Buffer analysis is going to be integrated with the trade areas but trade areas does not make any sense with lines, so what do we do?

Maybe just keep it disabled is better 

cc @saleiva @viddo

### LINKED ISSUE #7486: Change how analysis types are decided to eligible for add-analysis workflow
Currently we're using [the camshaft wrapper's `getValidInputGeometriesForType `](https://github.com/CartoDB/cartodb/blob/e8d709f0253f20af08004a83a5a008ee953813f3/lib/assets/javascripts/cartodb3/data/camshaft-reference.js#L74-L90) to check if an analysis can be created for a given source node. This was the quick'n'dirty implementation while @rochoa was on vacation, but we missed to keep track of it afterwards…

@rochoa How could/should we solve this properly? Could the expected geometry types be defined statically directly on the camshaft reference instead to avoid duplicating data (types)? Or will we need something more dynamic to be checked at runtime (e.g. geometry types to vary depending on params)?

Kind of related to #7483 where I might have to change the expected type for a buffer analysis for example…. cc @javisantana

viddo:
Various fixes related to updating geometry types being retrieved from camshaft reference

- [x] resolves #7486 - Get eligable geometry types from camshaft ref 
- [x] resolves #7483 - Line only work for buffer, so should not allow to change type to time in form

Notice that requesting to merge to the pending PR on analysis refactoring #7578
 
ping @xavijam 🙏


viddo:
@xavijam ready for review 💋 , deploying to ded03… 🚀

xavijam:
:+1:

PR #7597, opened 2016-06-01
Using the same query from current editor in order to obtain geometry type

### LINKED ISSUE #7485: Polygon datasets are rendered as points in new editor when using "by value" styling

xavijam:
Basically we are not getting the geometry type properly from several datasets in the new editor (reported by @AbelVM), but looks like old editor was doing it correctly so we have decided to follow the same thing in the new builder (:fingers_crossed:).

Fixes #7485.

CR: @viddo 

cc @AbelVM @javisantana

### DIFF
--- a/lib/assets/javascripts/cartodb3/data/query-schema-model.js
+++ b/lib/assets/javascripts/cartodb3/data/query-schema-model.js
@@ -5,7 +5,12 @@ var syncAbort = require('./backbone/sync-abort');
 var geometry = require('./geometry');
 
 var MAX_GET_LENGTH = 1024;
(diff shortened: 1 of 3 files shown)


viddo:
👍

xavijam:

PR #7598, opened 2016-06-01
Remove job table deletion notification

iriberri:
We added this notification to trace the deletion of the import temporary table to supervise that these weren't being stuck in the user db in case of an import error.

After a couple of months, I don't think this trace is useful anymore, so I'm removing it to alleviate the big amount of notifications we raise.


iriberri:
Hey @rafatower do you agree? :-)

rafatower:
I agree, go ahead with the trace deletion, please

PR #7600, opened 2016-06-01
fix popups on layer creation, close #7556

### LINKED ISSUE #7556: After adding a new layer, popup panel is not working
### STR

1. Create a map (with one layer) and go to editor 3.
2. Add a new layer.
3. Go to popup block.
4. Nothing is rendered, check chrome dev console.

cc @alonsogarciapablo

matallo:
CR @xavijam

### DIFF
--- a/lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
+++ b/lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
@@ -103,6 +103,7 @@ module.exports = Backbone.Collection.extend({
     options = options || {};
     var order = 0;
     var layerOnTop = this.last();
+
     if (layerOnTop) {
       order = layerOnTop.get('order');
       if (layerTypesAndKinds.isCartoDBType(layerOnTop.get('type'))) {
@@ -122,8 +123,6 @@ module.exports = Backbone.Collection.extend({
         tile_style: camshaftReference.getDefaultCartoCSSForType(),
         cartocss: camshaftReference.getDefaultCartoCSSForType(),
         style_version: '2.1.1',
-        tooltip: {},
-        infowindow: {},
         visible: true
       },
       tooltip: {},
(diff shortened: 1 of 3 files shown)


lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
```diff
@@ -125,7 +125,9 @@ module.exports = Backbone.Collection.extend({
         tooltip: {},
```
xavijam: Why don't we remove the options ones?
Also, could we add a tests please?

xavijam:
Checking what I've commented, adding a spec, and ready to go.

lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
matallo: totally true! they were misplaced since the beginning :/

tests added, merging as it goes 💚

lib/assets/test/spec/cartodb3/data/layer-definitions-collection.spec.js
```diff
@@ -28,6 +28,25 @@ describe('data/layer-definitions-collection', function () {
   });
 
   describe('.createLayerForTable', function () {
+    it('should create layer with infowindows and tooltips', function () {
+      var TiledLayer = new Backbone.Model({
```
xavijam: I feel we don't need those TiledLayer or CartoDBlayer, right?
With only the collection and using the function is enough.

matallo: it gave me a hard time on the layer creation because of an invalid sourceId, that being related to the letters, that being related to the order

https://github.com/CartoDB/cartodb/blob/7556-popups/lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js#L42-L50

this is the best solution I found to workaround that

xavijam: Hey, if we need that, let's add it within `beforeEach` + a comment, and 🎉 .

matallo: roger

xavijam: Ok, already checked with @matallo that rest of specs within that describe do other things in a different way. So 🍏 man!

matallo:
🚀

PR #7601, opened 2016-06-01
Infobox

### LINKED ISSUE #7551: Implement infobox
screen shot 2016-05-27 at 17 11 07
screen shot 2016-05-27 at 17 11 17
screen shot 2016-05-27 at 17 11 24
screen shot 2016-05-27 at 17 13 12

We should review new modals implementation in order to use it for this new element.

cc @viddo

nobuti:
This PR closes #7551.

The use case scenario should be this:

```
var states = [
  {
    state: 'ready',
    createContentView: function () {
      return Infobox.createInfo({
        title: 'Info',
        body: 'Lorem ipsum dolor sit amet.'
      });
    }
  },
  {
    state: 'error',
    createContentView: function () {
      return Infobox.createConfirm({
        title: 'Error',
        body: 'Lorem ipsum dolor sit amet.',
        confirmLabel: 'Confirm'
      });
    }
  }
];

var infoboxModel = new InfoboxModel({
  state: 'ready'
});

var infoboxView = new InfoboxView({
  infoboxModel: infoboxModel,
  infoboxCollection: new InfoboxCollection(states)
});
```

cc @xavijam @viddo


app/assets/stylesheets/editor-3/infobox.css.scss
```diff
@@ -0,0 +1,23 @@
...
+@import 'cdb-variables/sizes';
+
+.Infobox .CDB-InfoBox {
+  margin-right: -25px;
+  margin-left: -25px;
+  padding: 25px;
+  border-radius: 0;
+
+  &.is-default {
+    background: $cThirdBackground;
```
xavijam: shouldn't this background be in the main `Infobox` class? I haven't seen any part where we include `is-default` class.
Also, I'd migrate CDB-InfoBox styles from CartoAssets to Editor, renaming it with just Infobox, what do you think @piensaenpixel?

lib/assets/javascripts/cartodb3/components/infobox/infobox-button.tpl
```diff
@@ -0,0 +1,7 @@
+<% if (type === 'link') { %>
+  <p class="CDB-Text is-semibold CDB-Size-small"><a href="#" class="js-action"><%= label %></a></p>
```
xavijam: Using <%= %> just in case right?

xavijam:
I like the idea and I only miss some specs. So 👍  in my opinion.

xavijam:
Q: shouldn't it be a infoboxModel storing the state of the infobox?

xavijam:
Example:

```js
var infoboxModel = new InfoboxModel({
  state: 'idle'
});
var infoboxCollection = new InfoboxStatesCollection([
  {
    state: 'idle',
    title: 'Hola mundo',
    body: 'Lorem ipsum dolor sit amet',
    primaryButton: {
      type: 'link',
      label: 'Close',
      action: function () {
        infoboxModel.set('state', 'error');
      }
    }
  }, {
    state: 'error',
    type: 'error',
    title: 'Oops',
    body: 'Quam impedit ipsam a fuga voluptatum dolore, ea error.'
  }
]);

new Infobox({
  statesCollection: infoboxCollection,
  model: infoboxModel
});
```

app/assets/stylesheets/editor-3/infobox.css.scss
nobuti: The problem is `CDB-InfoBox` has the `background-color`. Maybe we need to update the `CDB-InfoBox` component in cartoassets. I'm trying to use `setElement` here to avoid the usual div wrapper, but I found some errors, that's why I left the `Infobox` class name for the wrapper for now.

nobuti:
Cleaner, on it.

app/assets/stylesheets/editor-3/infobox.css.scss
xavijam: The thing is, CDB-InfoBox has nothing to do with Deep-insights or CartoDB.js, so in my opinion, take all from CartoAssets (remove it from there of course), and adapt it as you wish.

viddo:
Just by reading the use-case scenario described above it's not clear to me how this would be used, could you elaborate a little? E.g. what's the intended lifecycle of the collection and view?

nobuti:
@viddo surething! An infobox shows information about an event/action usually by the user. But there will be events/actions that are asynchronous, like the deletion of a node, or even actions that they end up with an error. So we define a collection of states, where every state defines the information that an infobox is rendering in every moment.

viddo:
Also, I don't think it's a good idea to expose implementation details (actual collection/models/views), that should not be known to developer just using this component, I'd much rather see a semantic API which just do what I want w/o me having to know the internal details (also makes it easier to extend/change behavior in the future). 

Looking at the [mocks in the issue](https://github.com/CartoDB/cartodb/issues/7551):

```js
infoboxes.createInfo({
  title: 'Why a popup?',
  body: 'He's a queer man'
})
```

```js
var infobox = Infoboxes.createConfirm({
  title: 'CartoCSS applied',
  body: 'You just applied, …',
  cancelLabel: 'clear',
  confirmLabel: 'Continue'
});

infobox.on('cancel', function () {
  styleModel.clear();;
})
```

viddo:
> An infobox shows information about an event/action usually by the user. But there will be events/actions that are asynchronous, like the deletion of a node, or even actions that they end up with an error. So we define a collection of states, where every state defines the information that an infobox is rendering in every moment.

I'd recommend to not mix the events/actions handling with the infobox implementation, since they're separate concerns. The semantic API I suggested above is only for the infobox itself, the events/actions should be tackled separately.

viddo:
…that separate component would be using the infobox though, ofc. So I still think this is a good start on that way. :)

nobuti:
Awesome, thanks @viddo @xavijam, I'll work on that.

nobuti:
I've updated the code (and the PR description) to incorporate those ideas. Please code review it again guys.

lib/assets/javascripts/cartodb3/components/infobox/infobox-factory.js
```diff
@@ -0,0 +1,42 @@
+var InfoboxView = require('./infobox-item-view');
```
xavijam: 👌

lib/assets/javascripts/cartodb3/components/infobox/infobox-item-model.js
```diff
@@ -0,0 +1,12 @@
+var Backbone = require('backbone');
+
+/*
+ *  List item model
```
xavijam: `/* Infobox item model */`

xavijam:
Looks good to me, just one question, are we going to keep using `CDB-InfoBox` from CartoAssets?

lib/assets/javascripts/cartodb3/components/infobox/infobox-collection.js
```diff
@@ -0,0 +1,60 @@
...
+  getState: function (state) {
+    return _.first(this.where({state: state}));
+  },
+
+  getSelected: function () {
+    var selected;
+    selected = _.first(this.where({selected: true}));
+    if (!selected) {
+      selected = _.first(this.models);
+      selected.set({'selected': true});
```
viddo: Changing state doesn't belong in the getter, why not set a selected one when the collection is created/reset instead?

nobuti:
Fixed styles and default selected management in the collection.

xavijam:
ready to go then?!

nobuti:
Vamos! 🚀

PR #7609, opened 2016-06-01
Sync tables - Check if results exist before drop in rescue

iriberri:
If `result.nil?`, the code raises an exception that end ups in the rescue. The rescue was trying to drop the stale table, but `result` is nil, so you can't access to `result.table_name`.


rafatower:
:+1:

PR #7615, opened 2016-06-01
Switch fix oid to find appropriate table

iriberri:


rafatower:
:+1: 

don't take my thumb as a disclaimer :laughing:

Issue #7612, opened 2016-06-01
Dropping a layer causes layers to shake a bit


alonsogarciapablo:
This is not vey cool and I'm not sure why it's happening:

![layers-shaking](https://cloud.githubusercontent.com/assets/390398/15713965/f4b704e0-2818-11e6-9787-09990d4e159c.gif)

@javierarce do you have any idea of why this can be hapenning? would you please take a look? Not very important, but would be nice to fix it...

cc: @xavijam

xavijam:
Adding to editor 3 milestone, super minor ;).

PR #7617, opened 2016-06-01
Removed commas from transition arguments

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb/issues/7612.

layers-shaking

@javierarce can you please review?

BTW: We seem to be using `@include transition(property, duration, effect);` (arguments separated by commas) and I don't think that has the desired outcome (seems to be applying the transtion to `all` properties). I think we should change those to `@include transition(property duration effect);` (no commas) cc: @xavijam @viddo @javierarce @nobuti.

### DIFF
--- a/app/assets/stylesheets/editor-3/block-list.css.scss
+++ b/app/assets/stylesheets/editor-3/block-list.css.scss
@@ -7,8 +7,8 @@
 }
 .BlockList-item {
   @include display-flex();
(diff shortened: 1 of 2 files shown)


xavijam:
Finy be me, what do you think @piensaenpixel?

javierarce:
👍

PR #7623, opened 2016-06-02
Layer order bugfix [WIP]

### LINKED ISSUE #7621: Ordering layers is not working in this case
## STR

Making the example with this map: https://nycdoittpublicdata.cartodb.com/u/nycpublicdata/viz/be21c8c2-def6-11e5-b40b-0e787de82d45/public_map.

1. Export and import the map in your account.
2. Go to the map (current editor) and change the basemap (just for commodity) to one without labels.
3. Go to the new builder.
3. Open the dev chrome console, we will need it.
4. Add a widget for `nyc_subway_line`, e.g `cartodb_id` category.
5. Go back to layers list and move layer layer `C`, `nyc_subway_line` to the last position.
6. Check that tiles have not been reloaded.
7. Get back `C` layer to first position.
8. Tiles have been reloaded but widget, suddenly, is not working.

cc @alonsogarciapablo

xavijam:
Two things:

1. We were sending incorrect orders due to `resetOrders` happens after the `add` action who starts all the machinery.
2. When that new `add` event provokes a new createLayer (:ok_hand:), but in that method we remove the source, that was the reason why we were losing the analysis definition in the map instantiation.

What do you think @alonsogarciapablo?
Fixes #7621


lib/assets/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -183,7 +184,9 @@ var LAYER_TYPE_TO_LAYER_CREATE_METHOD;
 F.prototype._createLayer = function (m, opts) {
   opts = opts || {};
   var attrs = JSON.parse(JSON.stringify(m.attributes)); // deep clone;
-  delete attrs.source; // a new layer should not have an analyse to begin with, even if we have it in the context of editor
+  if (!opts || !opts.keepSource) {
+    delete attrs.source; // a new layer should not have an analyse to begin with, even if we have it in the context of editor
```
alonsogarciapablo: I'm wondering if could check if there's an existing analysisDefinition for attrs.source and, if that's the case, we would not need to `delete attrs.source` cause we can assume the analysis definition node exists in the context of CartoDB.js.

@xavijam @viddo what do you think?

viddo: Sounds reasonable 👍

xavijam: Well, there will be always an analysis for each layer (the source one).

alonsogarciapablo: Yes, there will always be an analysis node... What I'm suggesting is checking it there's an existing **analysisDefinition** whose `node_id` points to `attrs.source` in the analysisDefinitionsCollection...

xavijam: Ahhhhh, fair enough!

viddo: 👍  `this._analysisDefinitionsCollection.findByNodeId(m.get('source'))` should do :)

lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
```diff
@@ -47,9 +47,10 @@ module.exports = Backbone.Collection.extend({
 
     var movingLayer = this.at(from);
     this.remove(movingLayer, { silent: true });
-    this.add(movingLayer, { at: to, parse: false });
+    this.add(movingLayer, { at: to, parse: false, silent: true });
     this._resetOrders();
     this.save();
+    this.trigger('add', movingLayer, this, { at: to });
```
alonsogarciapablo: We should probably trigger this event after `this.save()` have succeeded, right?

lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
```diff
@@ -47,9 +47,10 @@ module.exports = Backbone.Collection.extend({
 
     var movingLayer = this.at(from);
     this.remove(movingLayer, { silent: true });
-    this.add(movingLayer, { at: to, parse: false });
+    this.add(movingLayer, { at: to, parse: false, silent: true });
```
viddo: I know this wasn't changed now, but I'm a bit skeptical about the parse: false. Why not maintain the model and just update the order attr? Not doing the parse could have unwanted side-effects (e.g. internal models not being setup/initialized as expected).

xavijam: If we agree with that, I'd open a new ticket about that issue. I don't want to mix things and make the result bigger.
But in any case, if the solution was reviewed, we should keep with the idea, shouldn't we?

lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
xavijam: :+1:

lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
viddo: > …, if the solution was reviewed, we should keep with the idea, shouldn't we?

Why? That something was reviewed previously doesn't mean that it's correct. 
But +1 to open new issue to look into this one.

xavijam: Totally agree, let's open a new ticket.

alonsogarciapablo:
:+1:

lib/assets/test/spec/cartodb3/data/layer-definitions-collection.spec.js
```diff
@@ -499,16 +499,22 @@ describe('data/layer-definitions-collection', function () {
...
+      };
       this.collection.on('add', onAddCallback);
       this.collection.on('remove', onRemoveCallback);
+      this.collection.on('layerMoved', onLayerMovedCallback);
 
       this.collection.moveLayer({ from: 3, to: 1 });
 
-      expect(onAddCallback).toHaveBeenCalled();
+      expect(onAddCallback).not.toHaveBeenCalled();
+      expect(onLayerMovedCallback).toHaveBeenCalled();
```
alonsogarciapablo: `expect(onLayerMovedCallback).toHaveBeenCalledWith(this.collection.at(x), 1);`?

lib/assets/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -189,8 +189,8 @@ F.prototype._onLayerDefinitionRemoved = function (m) {
 };
 
 F.prototype._onLayerDefinitionMoved = function (m, at, c) {
-  if (!m) throw new Error('moved layer model not provided');
-  if (!at) throw new Error('at position provided');
+  if (!m) throw new Error('moved layer model required');
```
alonsogarciapablo: I'd personally not include this checks here. Instead, I'd ensure that whoever triggers a "layerMoved" event passes the right arguments (as suggested [here](https://github.com/CartoDB/cartodb/pull/7623/commits/373a7082bf975330a236d0e42816dd85dbe9f8b0#r65712758))

lib/assets/test/spec/cartodb3/deep-insights-integrations.spec.js
```diff
@@ -286,4 +286,37 @@ describe('deep-insights-integrations', function () {
       });
     });
   });
+
+  describe('when a layer is moved', function () {
+    beforeEach(function () {
+      // this.analysisDefinitionsCollection.add({ node_id: 'a0' });
```
alonsogarciapablo: We have some commented out here. Just sayin'

alonsogarciapablo:
Left a couple of super-minor comments (feel free to ignore them)

 :+1: :+1: :+1:

PR #7632, opened 2016-06-02
Disable panel with infobox

### LINKED ISSUE #7548: Disable style form panel when a custom cartocss is applied
screen shot 2016-05-27 at 17 01 56

nobuti:
This PR closes #7548.

disable

Some work needs to be done:
- [x] ~~Some unrelated tests are failing. Investigating.~~
- [x] Implement confirm and cancel actions properly in terms of styles, form and undo manager.

cc @xavijam


xavijam:
One thing @nobuti, when a custom style is applied, it should change automatically to the other panel. That message + the "opacity" element should appear only when user switches to the styles panel.

app/assets/stylesheets/editor-3/options-bar.css.scss
```diff
@@ -3,10 +3,11 @@
 
 .Options-bar {
   @include align-items(center);
+  height: 75px;
```
piensaenpixel: why do we need this height instead work with padding?

nobuti: My mistake, but we need to tell it not to grow. I'll fix it.

xavijam: I feel it should be a min-height right? Because in the designs they appear with that size, no matter if content takes all the space.

xavijam: Scroll content?

piensaenpixel:
I think we dont need font-family - https://github.com/CartoDB/cartodb/pull/7632/commits/eb3ae324c216b940becc1de71a98a08605cae655#diff-a9f5c1e56ebfb81d1aa9abf7009ed399R10

piensaenpixel:
Maybe use cartoAssets typography? - https://github.com/CartoDB/cartodb/pull/7632/commits/eb3ae324c216b940becc1de71a98a08605cae655#diff-a9f5c1e56ebfb81d1aa9abf7009ed399L22

app/assets/stylesheets/editor-3/infobox.css.scss
```diff
@@ -34,7 +34,6 @@
 }
 
 .Infobox-button {
-  @include flex(1);
 
   &.Infobox-button--right {
     margin-left: auto;
```
piensaenpixel: I think we don't need this one

jvillarf:
I'm re-running Onpremises tests, but only for testing purposes. They were fine yesterday, so don't worry if you see any new message about them. 

Thanks!

nobuti:
thanks for the heads up @jvillarf !

nobuti:
Fixed styles and interaction. For now both infobox actions just close the notice.

lib/assets/javascripts/cartodb3/editor/style/style-view.js
```diff
@@ -119,25 +153,51 @@ module.exports = cdb.core.View.extend({
...
+
+    var infoboxSstates = [
+      {
+        state: 'confirm',
+        type: 'alert',
+        createContentView: function () {
+          return Infobox.createConfirmAndCancel({
+            type: 'alert',
+            title: 'CartoCSS Applied',
+            body: 'You just applied styles with CartCSS editor. You can continue or clear all styles.',
```
xavijam: Take into account internationalization, please :).

nobuti: Fixed.

lib/assets/javascripts/cartodb3/editor/components/overlay/overlay-view.js
```diff
@@ -0,0 +1,32 @@
...
+  initialize: function (opts) {
+    if (!opts.overlayModel) throw new Error('overlayModel is required');
+    this.overlayModel = opts.overlayModel;
+    this._initBinds();
+  },
+
+  render: function () {
+    var visible = !this.overlayModel.get('visible') ? 'none' : 'block';
+
+    this.$el.css({
```
xavijam: I'd try to not using css with code, with not adding a class that have those properties?

xavijam:
Q: there will be another use case of infobox for this panel, that is if the layer is "hidden", an infobox will appear saying (more or less) "Until layer is not visible, style won't change". Would it be easy to add several ways of creating different infoboxes for the same view/context?

nobuti:
We will need to add the new state for the infobox, and then setting the infobox model to the state's infobox will do the trick.

xavijam:
:+1:

xavijam:
go ahead man! looks great ;)

Issue #7338, opened 2016-05-09
Improve Geocoding Error Messages 


michellechandra:
@iriberri 

Users email us frequently in Support when they encounter a general georeferencing error message that has no additional details.

These tickets are usually about two common geocoding issues:

1) There isn't enough storage space in their account to complete the georeferencing and they need to free up space 

2) They failed to delete the cartodb_georef_status column prior to geocoding which cause the geocoding to fail.

Improve the general error message so users don't need to email Support for help and can continue to use the Product.

iriberri:
related https://github.com/CartoDB/cartodb/issues/5230

iriberri:
I'm going to close this issue, at least temporarily, as I closed my PR to solve this. With the Builder the scenario is different, so let's see what challenges we have in there.

PR #7636, opened 2016-06-02
Add new storage quota exception for geocoding and remove premature fail status

### LINKED ISSUE #5230: Add geocoding error message for errors caused by excess of quota
When performing a geocoding process, adding the geocoded information to the table might exceed the quota set up for the user. In these cases, we get internally errors like:

`Sequel::DatabaseError: PG::Error: ERROR: Quota exceeded by 1460KB`

but in the UI we're just showing a generic Geocoding Error.

What about trying to move this inner error to the UI? This will reduce support and will allow the users to understand what's happening.

cc @saleiva

iriberri:
Closes #5230 and helps with https://github.com/CartoDB/cartodb/issues/7338

While working on adding an exception for quota issues, I realised about a problem in the current flow.

Whenever an exception was raised, the status was automatically set to failed. The polling of the frontend was detecting the failed status and setting the geocoding as failed, but with an unknown error because the error is detected afterwards by `handle_geocoding_failure`. The current `handle_geocoding_failure` changes the status to failed (again), so I removed the initial failure setup, which is hiding the real exceptions.

The flow was something like:

* Editor starts geocoding process
* Geocoding fails
* Status is set to failure in the backend
* The Editor detects the new failure state and shows a modal about the error
* The backend keeps working on handling the failure and sets the error code accordingly, which is not shown to the user

### DIFF
--- a/services/table-geocoder/lib/internal_geocoder.rb
+++ b/services/table-geocoder/lib/internal_geocoder.rb
@@ -51,7 +51,6 @@ module CartoDB
         copy_results_to_table
         change_status('completed')
       rescue => e
-        change_status('failed')
         raise e
       ensure
         drop_temp_table
@@ -133,6 +132,10 @@ module CartoDB
             @query_generator.copy_results_to_table_query,
             @table_schema, @table_name
           )
+      rescue => e
+        if e.is_a?(Sequel::DatabaseError) && e.message =~ /.*quota exceeded.*/i
+          raise Carto::GeocoderErrors::OverStorageQuotaError
+        end
       end
 
       def drop_temp_table
--- a/services/table-geocoder/lib/exceptions.rb
+++ b/services/table-geocoder/lib/exceptions.rb
@@ -9,7 +9,7 @@ module Carto
     GEOCODER_TIMED_OUT_WHAT_ABOUT = %q{
       Your geocoding request timed out.
       Please <a href='mailto:support@cartob.com?subject=The geocoder timed out'>contact us</a>
-      and we'll try to fix it quickly.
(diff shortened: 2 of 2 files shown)


iriberri:
Hi @ethervoid, you were the last one working with this part of the code -- specially with the "set status to failure" line that I'm attempting to remove. Could you please take a look? :-)

services/table-geocoder/lib/internal_geocoder.rb
```diff
@@ -133,6 +132,10 @@ def copy_results_to_table
             @query_generator.copy_results_to_table_query,
             @table_schema, @table_name
           )
+      rescue => e
```
ethervoid: You have to mark the result as failed in any moment isn't it?

iriberri: But it already does here, no? https://github.com/CartoDB/cartodb/blob/master/app/models/geocoding.rb#L370

ethervoid: Hum, you're right

iriberri: So you like it? :P

ethervoid:
Sounds good to me but tests it thoroughly because the status part is a tricky one

iriberri:
retest this please

iriberri:
This is gonna need some frontend work, as the HTML in the exceptions is not being interpreted.

iriberri:
I'm gonna close this without deleting the branch for the moment, because I don't think is worthy to spend any more time in the old geocoding process right now.

PR #7638, opened 2016-06-02
Infowindow render fixed

### LINKED ISSUE #1117: Column named 'length' causes an error with infowindows
related to this issue

https://github.com/jashkenas/underscore/issues/1590

seen in this table

https://team.cartodb.com/u/andrew/tables/severe_wind_x/map

the infowindows fail. i try to select fields in the ui and it just spins.

MariaCheca:
Closes #1117

This is a fix for an underscore bug where every object with `length` as a property name on it is taken as an array. The for-in loop solves the problem.

CR @viddo please :)


viddo:
Thx for the well-written commit message 👍 

Please write a test case that pinpoints the problem, which should fail with the old code but pass with your changes. Considering it seems to be a very isolated issue it should not be too hard.

MariaCheca:
@viddo tests are done, up and running!

viddo:
👍

PR #7642, opened 2016-06-03
Adds category selector for the fill component

javierarce:
From the original PR

- Fixed failing specs
- Added translation for the quantification method list
- Removed linter/hound warnings
- Addressed rest of the comments

Pending stuff:

- Add selected state for the color ramps.
- Show the 'Others' category at the bottom of the category list.
- Select a default ramp when switching to the ramp view.
- Add specs for the category list


javierarce:
retest this please

PR #7663, opened 2016-06-03
remove google_maps feature flag check

gfiorav:
CR @javitonino 

brings changes from 3.13.x

cc/ @javisantana @rafatower

### DIFF
--- a/app/views/admin/client_applications/api_key.html.erb
+++ b/app/views/admin/client_applications/api_key.html.erb
@@ -76,7 +76,7 @@
           <label class="FormAccount-label">Google Maps</label>
         </div>
         <div class="FormAccount-rowData">
-          <input type="text" value="<%= current_user.organization.present? ? current_user.organization.google_maps_key : current_user.google_maps_key %>" class="FormAccount-input FormAccount-input--long is-disabled" readonly />
+          <input type="text" value="<%= current_user.organization.present? ? current_user.organization.google_maps_private_key : current_user.google_maps_private_key %>" class="FormAccount-input FormAccount-input--long is-disabled" readonly />
         </div>
         <div class="FormAccount-rowInfo">
           <% if !current_user.organization.present? %>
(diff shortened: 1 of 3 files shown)


javitonino:
The feature flag is used in the views (account settings page) to show/hide the Google API keys. With the new code, this will never be shown. Maybe change the check to use  `google_maps_geocoder_enabled?`?

gfiorav:
good point, removed them from the templates

the presenter used the geocoder check so it should be fine

javisantana:
good catch @javitonino

gfiorav:
Hey @javitonino thanks for looking into it with so much detail!

I have changed a few things that were wrongly displayed. Have a look and I'll merge! 

Thanks.

app/views/admin/client_applications/api_key.html.erb
```diff
@@ -76,7 +76,7 @@
           <label class="FormAccount-label">Google Maps</label>
         </div>
         <div class="FormAccount-rowData">
-          <input type="text" value="<%= current_user.organization.present? ? current_user.organization.google_maps_key : current_user.google_maps_key %>" class="FormAccount-input FormAccount-input--long is-disabled" readonly />
+          <input type="text" value="<%= current_user.organization.present? ? current_user.organization.google_maps_private_key : current_user.google_maps_private_key %>" class="FormAccount-input FormAccount-input--long is-disabled" readonly />
```
javitonino: Pre-existing bug?

javitonino:
Looks good for backend. Will there be matching changes to the JS?

app/views/admin/client_applications/api_key.html.erb
gfiorav: Yes! Decided to fix it as well

gfiorav:
Thanks for reviewing! I think the issues with Travis build is due to something else. I'll try to rebuild.

app/views/admin/client_applications/api_key.html.erb
javitonino: :palm_tree: :heart:

gfiorav:
🍏 Here as well, merging! Thnx @javitonino!

PR #7664, opened 2016-06-03
Adds shareable urls property to embed

fdansv:
This will activate updating the address bar with the current state of the map/dashboard.

@javisantana @xavijam


xavijam:
No need to update internal packages?

fdansv:
yes, done!

PR #7671, opened 2016-06-06
3211 remove oddyssey from vizjson

### LINKED ISSUE #3211: odyssey viz.json maps_api_template is not set to https when is requested with https
- create an odyssey
- request the viz.json using https

the master slide is good, the first slide is http instead of https

javitonino:
Closes #3211


javitonino:
CR please @gfiorav

gfiorav:
Looks good. Named maps is based on vizjson and have nearly no tests. I'm 99% confident it will not affect at all, but it might be worth a test.

📦 it when ready!

PR #7673, opened 2016-06-06
adds restricted options for ff rakes

gfiorav:
This adds an optional param to `add_feature_flag` to make it unrestricted or restricted:

```
rake cartodb:features:add_feature_flag[feature,restricted]  # add feature flag and optionally set restricted (default is true)
```

~~If the feature already exists, it will change the restricted if needed.~~
If the feature already exists, an error is raised.

It also adds another task only to change restricted of an existing feature flag (also nice to have):

```
rake cartodb:features:change_feature_restricted[feature,restricted]  # change feature flag to restricted or unrestricted
```

Please CR @javitonino 

cc/ @santisaez

### DIFF
--- a/lib/tasks/enable_feature.rake
+++ b/lib/tasks/enable_feature.rake
@@ -135,7 +135,7 @@ namespace :cartodb do
         ff.restricted = restricted
         ff.save
 
-        puts "[INFO]\tFeature flag '#{args[:feature]}' restricted set to '#{restricted}'"
+        puts "[INFO]\tFeature flag '#{args[:feature]}' restricted set to '#{ff.restricted}'"
       else
         raise "[ERROR]\tFeature '#{args[:feature]}' doesn't exist"
       end
--- a/lib/tasks/enable_feature.rake
+++ b/lib/tasks/enable_feature.rake
@@ -109,15 +109,40 @@ namespace :cartodb do
     # WARNING: For use only at development, opensource and custom installs.
     # Refer to https://github.com/CartoDB/cartodb-management/wiki/Feature-Flags
     desc "add feature flag"
-    task :add_feature_flag, [:feature] => :environment do |t, args|
+    task :add_feature_flag, [:feature, :restricted] => :environment do |_task, args|
+      restricted = args[:restricted] ? args[:restricted].casecmp('false') != 0 : true
 
-      ff = FeatureFlag[:name => args[:feature]]
-      if ff.nil?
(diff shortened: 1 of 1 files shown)


lib/tasks/enable_feature.rake
```diff
@@ -108,16 +108,41 @@ namespace :cartodb do
...
-      if ff.nil?
-        ff = FeatureFlag.new(name: args[:feature], restricted: true)
+      ff = FeatureFlag[name: args[:feature]]
+      if !ff
+        ff = FeatureFlag.new(name: args[:feature], restricted: restricted)
         ff.id = FeatureFlag.order(:id).last.id + 1
         ff.save
+
+        puts "[INFO]\tFeature flag '#{args[:feature]}' created and restricted set to '#{restricted}'"
+      elsif ff.restricted != restricted
```
javitonino: I wouldn't put this here, given that there is another rake task that does exactly this. But I guess there is a reason for this (provisioning scripts?).

lib/tasks/enable_feature.rake
```diff
@@ -108,16 +108,41 @@ namespace :cartodb do
...
+
+      ff = FeatureFlag[name: args[:feature]]
+      if ff
+        ff.restricted = restricted
+        ff.save
+
+        puts "[INFO]\tFeature flag '#{args[:feature]}' restricted set to '#{restricted}'"
       else
-        raise "[ERROR]  Feature '#{args[:feature]}' already exists"
+        raise "[ERROR]\tFeature '#{args[:feature]}' doesn't exists"
```
javitonino: Typo (when changing verb): `exists` -> `exist`

gfiorav: thanks, changing

lib/tasks/enable_feature.rake
gfiorav: Yeah, it's a bit ugly, but i guess it saves time.

javitonino:
I would like to know why there are two rakes for doing the same. Otherwise, :ship: it.

gfiorav:
Thanks @javitonino 

It's not that two rakes do the same, it's that if you try to add two times the same ff, it will check if the restricted requested and set are the same and if not it will change them. 

It's a bit ugly I admit. I'll change it.

gfiorav:
@javitonino have another quick look please!

javitonino:
:+1:

gfiorav:
Thanks @javitonino 

Merging as soon as tests are 🍏 , @santisaez

PR #7675, opened 2016-06-06
7652 fill default ramp

### LINKED ISSUE #7652: [Fill component] Select a default color ramp when switching to the ramp view

javierarce:
This PR picks a default color ramp when the user switches to the color ramp selector in the category fill. Also, when a user updates the number of bins, the ramp gets updated with the corresponding one based on the new chosen value.

Closes #7652


javierarce:
CR: @xavijam

xavijam:
:+1:

PR #7678, opened 2016-06-06
brings fixes from onprem branch for ff rake tasks

gfiorav:
Same stuff for `master` @javitonino !


javitonino:
Thumbs up! Maybe add a NEWS item so people know this exists? Not very important though.

PR #7679, opened 2016-06-06
Disables the opacity selection for categories

### LINKED ISSUE #7667: [Fill component] Disable alpha option for categories

javierarce:
Since category ramps are just a list of opaque colors, the color picker shouldn't give the option to change the opacity. This PR does that.

This PR also sets the default opacity to 1 in case it's not defined.

screen shot 2016-06-06 at 13 20 25

Closes #7667

### DIFF
--- a/app/assets/stylesheets/editor-3/color-picker.css.scss
+++ b/app/assets/stylesheets/editor-3/color-picker.css.scss
@@ -25,6 +25,11 @@
   padding: 3px 4px;
   text-align: center;
 }
+
+.ColorPicker-input.is-disabled {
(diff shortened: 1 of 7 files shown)


javierarce:
CR: @xavijam 
CR (design): @piensaenpixel

xavijam:
Oh, ok! it is true we don't provide that option in the current editor, not sure if we should disable again in the new one... what do you think @urbanphes ?

xavijam:
BTW, code looks good to me 👌

javierarce:
retest this please

PR #7682, opened 2016-06-06
Fixes rendering issue on the category component when changing attributes

### LINKED ISSUE #7680: [Fill component] Having a category column applied, when other is selected, previous one keeps being chosen
category-bux

javierarce:
This PR also fixes a bug that rendered and empty list of ramps when the bin number from the server was bigger than the biggest default one (7).

Fixes #7680


javierarce:
CR: @xavijam

PR #7691, opened 2016-06-06
7635 generate analysis options

### LINKED ISSUE #7635: Generate analysis options by camshaft reference
The [latest version of the camshaft reference](https://github.com/CartoDB/camshaft/blob/master/reference/versions/0.10.0/reference.json) is good enough to generate "automatic" (parameterized) analysis options, 

edit: for _development/testing_ purposes, i.e. not end-users, so changed scope below:

There's a few things that needs to be done for it to work though:

**Scope:**
- Add config flag to generate analysis from camshaft reference definitions instead of [default analysis options](https://github.com/CartoDB/cartodb/blob/3fb48ec8222b1d031a85fc565e0d56f1915559a2/lib/assets/javascripts/cartodb3/components/modals/add-analysis/default-analysis-options.js)
- The defaults should be treated as overrides if flag is enabled
- Generate form from analysis definition's params

@andrewxhill @javisantana opinions? anything I missed or overlooked?
cc @rochoa anything else I should consider?

viddo:
Resolves #7635 

If the feature flag `generate_analysis_options` is enabled for the user it will generate source options for the options not implemented yet. It's not intended for end-users but for internal testing/development, until the custom forms are implemented:

screen shot 2016-06-06 at 18 22 21

Selecting one these will generate a form view based on the params from the camshaft reference:
screen shot 2016-06-06 at 18 23 47

Note that while param types and required/optional state are defined in the reference, other restrictions (e.g. optional/required combinations) are not, so there's no guarantee that the analysis will work even if the form validation passes! 
I've implemented some "long hanging fruits" restrictions though, to make it easier to test these options e.g.:
- if a generated option only require a single source it's not allowed to be changed (the same limitation we have for all custom ones already).
- if a generated option only require a single source and have a column param it's presented as a columns-selector w/ columns available from that source, so one don't have to lookup/write the column manually

Also, added [tape](https://github.com/substack/tape) for easier/faster testing, not intended as a replacement but _complement_ in addition to jasmine, to easier unit test things, since it doesn't require any browserify bundling (or source map decoding), and thus we don't have to wait several seconds to get test results (I'll create a separate issue to elaborate on ideas/proposals). It's run on changed fils as part of `grunt dev` (and `grunt test` ofc), manually by:
```bash
$ npm test

> cartodb-ui@3.25.6 test /Users/viddo/src/cartodb
> tape $npm_package_config_tapeTests | tap-summary


------------------------------------ Tests -------------------------------------
✔ each category excluding generated categories [pass: 23, fail: 0, duration: 4ms]
✔ each category including generated categories [pass: 53, fail: 0, duration: 11ms]

----------------------------------- Summary ------------------------------------
duration: 15ms
assertions: 76
pass: 76
fail: 0
```

@javierarce can you review please?
cc @xavijam @nobuti @alonsogarciapablo FYI 

cc @rochoa @dgaubert you might also find this useful I guess


lib/assets/javascripts/cartodb3/components/modals/add-analysis/analysis-options.js
```diff
@@ -0,0 +1,84 @@
...
+    var implementedAnalyses = _.reduce(Object.keys(categories), function (memo, category) {
+      return memo.concat(
+        categories[category].analyses.map(function (analysis) {
+          return analysis.nodeAttrs.type;
+        })
+      );
+    }, []);
+
+    categories['generated'] = {
+      title: 'Generated analyses by the Camshaft reference v' + _.last(camshaftReference.versions),
```
javierarce: I understand that this is not meant for end-users, but shouldn't this string be localized?

viddo: What's the point of localize it considering it's intended to only be used internally?

javierarce: Oh, ok, ok, I thought maybe it would make sense for other developers. Never mind then.

javierarce:
:+1:

nobuti:
Awesome.

andrewxhill:
@rochoa @javisantana @stuartlynn 

would it make sense for this point to a develop branch on camshaft? that way we could test features even before they land in master

rochoa:
@viddo what do you think?

Hint: having a develop branch means you cannot use the npm published version of the reference.

viddo:
@andrewxhill, to "test new features" I think the best would be for you to install CartoDB locally (just like how we do it for development), and then do a "npm link" of the camshaft-reference so you can modify it as you wish, and do normal PR workflow to push changes upstream.

andrewxhill:
yeah, that's fine, also we have a hangup for most of what we are working on related to the deployment of crankshaft. so all good. thanks!

PR #7692, opened 2016-06-06
Only allow deleting a layer when there're more than one Torque or CartoDB layers

alonsogarciapablo:
- Detect global "mousedown" events (instead of "click") to determine when the context menu should be hidden. This way the context menu is hidden when user starts dragging a layer.
- Context menu is now created and destroyed when the toggle icon is clicked. This way we can decide if "Delete layer" should be included in the menu.
- I placeholder item to the context menu so that the menu has something when there's only one layer. 

deleting-layers

@viddo 👍  or 👎 ? thanks!


lib/assets/javascripts/cartodb3/data/layer-definition-model.js
```diff
@@ -9,7 +9,7 @@ var layerTypesAndKinds = require('./layer-types-and-kinds');
 var InfowindowModel = require('./infowindow-click-model');
 var TooltipModel = require('./infowindow-hover-model');
 
-var OWN_ATTR_NAMES = ['id', 'order', 'color', 'table_name_alias', 'infowindow', 'tooltip'];
+var OWN_ATTR_NAMES = ['id', 'type', 'order', 'color', 'table_name_alias', 'infowindow', 'tooltip'];
```
alonsogarciapablo: @viddo I had to add this `type` here. Otherwise, we're always relying on the `kind` attribute when parsing the model](https://github.com/CartoDB/cartodb/blob/47c3ab0754fcc252870e3dded1f9fa3037903221/lib/assets/javascripts/cartodb3/data/layer-definition-model.js#L39). Can you think of any side effects? Thx!

viddo: Possible kind/type mismatches I guess, which one should take precedence? 

When I implemented the model I decided to stick with `kind` as the one "mandating" (used in parse/toJSON) since its the property that's stored at the top-level of the API endpoint, where as type is just another one in the bag of `options`.  IMHO we should avoid having to deal with double values since it increases the risk of inconsistent state.

viddo: Btw, actually, `OWN_ATTR_NAMES` is intended for props stored at top-level, type is not one of them so _why_ did you have to add it? Why is it a problem only relying on kind when parsing the model? The type would be inferred from the kind anyway so it's not clear to me why this would be needed.

alonsogarciapablo: I think I changed this as I wrote [this test](https://github.com/CartoDB/cartodb/pull/7692/files#diff-2e0d495281e372d50122808b240aa26fR526).

It's pretty confusing right now, as the viz.json includes several `types` at different levels.

```
{
  ...
  "layers": [
    {
      "options": {
        "type": "Tiled",
        ...
      },
      "type": "tiled"
      ...
    },
    {
      "type": "layergroup",
      "options": {
        ...
        "layer_definition": {
          ...
          "layers": [
            {
              ...
              "type": "CartoDB",
              "options": {
                 ...
              }
              ...
            },
            {
              ...
              "type": "CartoDB",
              "options": {
                 ...
              }
              ...
            }
          ]
        },
        ...
      }
    },
    {
      "options": {
        "type": "Tiled",
        ...
      },
      "type": "tiled"
      ...
    }
  ],
  ...
}
```

I'll change my test to something like:

```
this.collection.reset([
  {
    options: { type: 'Tiled' }
  }, {
    options: { type: 'CartoDB' }
  }, { 
    options: { type: 'torque' }
  } , { 
    options: { type: 'WADUS' }
  }
]);
```

alonsogarciapablo: Actually, it looks like `options.type` is not present in the DB anymore, so will use `kind` in my test then:

```
irb(main):022:0> Layer.map{|layer| layer.kind }.uniq
(0.003248s) SELECT * FROM "layers"
=> ["tiled", "carto", "torque"]
irb(main):023:0> Layer.map{|layer| layer.options[:type] }.uniq
(0.003993s) SELECT * FROM "layers"
=> [nil]
```

alonsogarciapablo: Reverted this in https://github.com/CartoDB/cartodb/pull/7692/commits/3cf95cd712b9278a2eaf994cdf7275a011bf6348. Ready for review again. Thanks your comments!

viddo:
LGTM 👍

PR #7703, opened 2016-06-07
Replace cdb.core.{Model,View} usages

viddo:
While working on #7691 I discovered some things I wanted to highlight:

1. [`cdb.core.Model`](https://github.com/CartoDB/cartodb.js/blob/2a84dd87ed202ae188fd58865d9b8e96a4791d35/src/core/model.js) contains code that's not actually used anywhere for the new code, so might as well just use `Backbone.Model`.
2. [`cdb.core.View`](https://github.com/CartoDB/cartodb.js/blob/2a84dd87ed202ae188fd58865d9b8e96a4791d35/src/core/view.js) do contain logic that's also used here (but not all)

Previously (in current/old editor) models and views were created in the context of the editor and passed to cartodb.js, since that's no the case anymore there's no reason to depend on those objects anymore.

So, I ported cdb.core.View here. The main benefit of 1 and 2 is that we can make a "clean cut" with cartodb.js, not having to require on common models/views anymore. Less complexity, and will enable both cartodb.js and the editor to change and improve w/o having to worry about one another.

Also, since we now have the `lib/assets/node_modules` dir, through how require calls are looked up we could finally fix those nasty `require('../../../../` calls, that's why `var CoreView = require('browserify/core-view')` works in all contexts. See [browserify-handbook#how-node_modules-works](https://github.com/substack/browserify-handbook#how-node_modules-works) for more details.

@xavijam @javier @alonsogarciapablo @javisantana I'd like to hear your opinions on this.

It's deployed on _ded03_ if you want to have a look, things works just like before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,7 +53,7 @@ vendor/bundle/*
 vendor/cache/*
 .vagrant
 Vagrantfile
-node_modules
+/node_modules
 .grunt/*
 vendor/assets/javascripts/cartodb.*
 vendor/assets/stylesheets/cartodb.*
--- a/lib/assets/javascripts/cartodb/models/export_map_model.js
+++ b/lib/assets/javascripts/cartodb/models/export_map_model.js
@@ -1,3 +1,5 @@
+/* global cdb */
+
 cdb.admin.ExportMapModel = cdb.core.Model.extend({
   /*
    * Creates an export_visualization job and polls until it finishes.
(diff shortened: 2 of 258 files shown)


javisantana:
I agree with the changes, I don't think this was the momento to do them

javisantana:
said that, I'd merge ASAP to avoid merge problems

xavijam:
Agree with @javisantana in everything.

xavijam:
Adding @nobuti to the thread.

Issue #7644, opened 2016-06-03
infowindows: create "none" template


matallo:
instead of disabling all the fields then not showing the popup on click, we'll still need to show a template BUT one telling "no fields are selected > go activate them" similar to

![screen shot 2016-06-03 at 10 39 40](https://cloud.githubusercontent.com/assets/36676/15773596/15586e4a-2978-11e6-8d2e-602243be1465.png)

I expect to tackle this so not many changes are needed in cartodb.js or migrations, by the next approach:

- if no template is sent and no fields are selected, popup won't appear
- if a template is passed, still no fields are selected, popup with the template will appear

this would involve making some changes in cartodb.js, but still worth it to prevent destructive migrations, also this covers the default no fields, `""` template case

however, I don't know how can we solve this _just_ for the editor, what to do outside when the template is set?

/cc @piensaenpixel @urbanphes @alonsogarciapablo

urbanphes:
We need to emphasize the message, and opted for the warning color in the popup
![image](https://cloud.githubusercontent.com/assets/14546701/15818329/1c4d998e-2bdd-11e6-9b55-3121660235af.png)

matallo:
this is already implemented, reassigning to @piensaenpixel for the styles

PR #7706, opened 2016-06-07
infowindows: none fields selected

matallo:
prepare for https://github.com/CartoDB/cartodb/issues/7644

this would need changes in cartodb.js, still not sure about this as I don't know how we would show these *only* in the editor.

/cc @alonsogarciapablo


alonsogarciapablo:
I investigated this a bit... It turns out that infowindows won't work in CartoDB.js unless they have been set up with some fields (Maps API can't compute the interactivity grid unless some interactivity columns have been setup when instantiating the map).

I talked briefly about this with @javisantana and what we can do is. We can set enable the field for the "cartodb_id" column at the editor level when the "none" template is selected so that interactivity is enabled in CartoDB.

The current editor disables the interactivity of layers when any of the selected columns is not present in the SQL for that layer:

screen shot 2016-06-07 at 19 54 17

This is something we can do in a separate issue though.

matallo:
what about this? also, I'm _updating_ the infowindow when doing the link to cover the case in the initial load, as well as in the changes

matallo:
also, I think we should remove `js-infowindow`from the template so the width isn't affected by the slider (and adapt the infowindow to the visuals in the proposal, of course)

/cc @piensaenpixel ?

matallo:
bump @alonsogarciapablo! ready to go?

lib/assets/javascripts/cartodb3/infowindow-manager.js
```diff
@@ -1,10 +1,31 @@
+var fs = require('fs');
+var INFOWINDOW_NONE = fs.readFileSync(__dirname + '/mustache-templates/infowindows/infowindow_none.jst.mustache', 'utf8');
+
 function InfowindowLink (layerDef, visMap) {
   if (!layerDef.infowindowModel) return;
 
+  function updateInfowindow (attrs) {
```
alonsogarciapablo: I'd rename `attrs` to `infowindowModel` cause it's actually a model...

lib/assets/javascripts/cartodb3/infowindow-manager.js
```diff
@@ -1,10 +1,31 @@
...
+    var jsonAttrs = attrs.toJSON();
```
alonsogarciapablo: Just to make sure we don't also add the field to the infowindowModel (I think `jsonAttrs.fields` is a reference to `attrs.attributes.fields` as it's implemented right now), I'd use the `JSON.parse(JSON.stringify(infowindowModel.attributes))` trick here.

matallo: ooh, I see I see, rookie mistake, I remember now

lib/assets/javascripts/cartodb3/infowindow-manager.js
```diff
@@ -1,10 +1,31 @@
...
+        position: 0
+      }];
+
+      jsonAttrs.template = INFOWINDOW_NONE;
+    }
+
+    visMap.getLayerById(layerDef.id).infowindow.update(jsonAttrs);
+  }
+
   function onChange (attrs) {
```
alonsogarciapablo: I'd use `infowindowModel` instead of `attrs` here too.

lib/assets/javascripts/cartodb3/mustache-templates/infowindows/infowindow_none.jst.mustache
```diff
@@ -0,0 +1,20 @@
+<div class="CDB-infowindow CDB-infowindow--light js-infowindow">
+  <div class="CDB-infowindow-container">
```
alonsogarciapablo: Should we add the markup for the close button to this template too?

```
<div class="CDB-infowindow-close js-close"></div>
```

alonsogarciapablo:
I left some comments 😸 . Also, I'm missing some tests in this PR, but I'm sure you're aware of this, so I leave it up you to add them. Thanks!

matallo:
struggled with `deep-insights-integrations.spec.js` but finally added, what about now? 😄

lib/assets/test/spec/cartodb3/deep-insights-integrations.spec.js
```diff
@@ -317,4 +321,110 @@ describe('deep-insights-integrations', function () {
...
+    describe('when update infowindow template_name', function () {
+      beforeEach(function () {
+        this.layerDefinitionModel.infowindowModel.set({
+          'template_name': 'infowindow_light',
+          'template': '<div class="CDB-infowindow"></div>'
+        });
+      });
+
+      it('should set none template if infowindow has no fields', function () {
+        expect(this.layer.infowindow.get('template')).toContain('You haven’t selected any fields to be shown in the infowindow.');
```
alonsogarciapablo: I'd be nice to assert here that the `cartodb_id` [field is added](https://github.com/CartoDB/cartodb/pull/7706/files#diff-633e1f4a71fa5124fea6b52adc35bb92R15) when this happens...

matallo: indeed, doing it so, thanks for the review!

lib/assets/test/spec/cartodb3/deep-insights-integrations.spec.js
```diff
@@ -317,4 +321,110 @@ describe('deep-insights-integrations', function () {
...
+            position: 0
+          },
+          {
+            name: 'name',
+            title: true,
+            position: 1
+          }
+        ]);
+
+        expect(this.layer.infowindow.get('template')).toEqual('<div class="CDB-infowindow"></div>');
```
alonsogarciapablo: I'd also ensure that `cartodb_id` has not been added to the list of fields in this case...
```
        var newFields = [
          {
            name: 'description',
            title: true,
            position: 0
          },
          {
            name: 'name',
            title: true,
            position: 1
          }
        ];

        this.layerDefinitionModel.infowindowModel.set('fields', newFields);

        expect(this.layer.infowindow.get('template')).toEqual('<div class="CDB-infowindow"></div>');
        expect(_.isEqual(this.layer.infowindow.get('fields'), newFields)).toBe(true);
```

lib/assets/javascripts/cartodb3/infowindow-manager.js
```diff
@@ -1,10 +1,35 @@
...
-  function onChange (attrs) {
-    visMap.getLayerById(layerDef.id).infowindow.update(attrs.toJSON());
+  function updateInfowindow (infowindowModel) {
+    var attrs = JSON.parse(JSON.stringify(infowindowModel.attributes));
+
+    if (infowindowModel.fieldCount() === 0 &&
+        // 'table/views/infowindow_light' is the default template
+        infowindowModel.get('template_name') !== 'table/views/infowindow_light' &&
+        // '' is the template set when click None
+        infowindowModel.get('template_name') !== '') {
```
alonsogarciapablo: From what I understand, it seems that fields are set to `cartodb_id` when:

1. InfowindowModel doesn't have fields.
2. `template_name` is not empty and is not `'table/views/infowindow_light'`.

Doesn't this override fields when, for example, `template_name` is 'table/views/infowindow_dark'?

I think the default template_name is 'infowindow_light' and not ''table/views/infowindow_dark'...

I'd expect this condition to be more like:

```
// If no fields and template has been selected
if (infowindowModel.fieldCount() === 0 && !infowindowModel.get('template_name') {
  ...
}
```

As an idea, we could move that logic to the infowindowModel and simplify the conditional:

```
if (infowindowModel.isEnabled()) { // not sure isEnabled is the best name, but you get the idea...
  ...
}
```

matallo: totally agree with the `isEnabled` method (name pending), moving it there

re: the templates stuff

> I think the default template_name is 'infowindow_light' and not ''table/views/infowindow_dark'...

I suppose you mean 'table/views/infowindow_light', actually I catched this while writing the tests, the default infowindow is 'table/views/infowindow_light' indeed (at layer creation), I think it comes from backend, can you confirm @gfiorav?

so it won't show that infowindow if the user didn't touch anything (default template) or if he chose one and came back to none

gfiorav: That is correct: all carto layers are created with `light` infowindow by default.

Would you guy like to change this behaviour?

matallo:
feedback applied

lib/assets/javascripts/cartodb3/data/infowindow-definition-model.js
```diff
@@ -154,6 +154,13 @@ module.exports = Backbone.Model.extend({
     }
 
     this.set(attrs);
+  },
+
+  isEmptyTemplate: function () {
+    var isEmpty = this.fieldCount() === 0;
+    // '' is none
+    var hasTemplate = this.get('template_name') === '' || (typeof (this.TEMPLATES[this.get('template_name')]) !== 'undefined');
```
alonsogarciapablo:
```
return this.get('fields').length > 0 && this.TEMPLATES[this.get('template_name')];
```

alonsogarciapablo:
👍  👍  👍  Thanks @matallo!!!

matallo:
🚀

alonsogarciapablo:
:clap:

PR #7700, opened 2016-06-07
Adds loader for categories list

javierarce:
animation

CR: @xavijam 
CR: @piensaenpixel

PR #7710, opened 2016-06-07
change loaders

piensaenpixel:
@xavijam could you take a look? only change the new markup with the cartoassets


xavijam:
Aren't there any other places where we use the "old" loader?

piensaenpixel:
i only found this one :(

xavijam:
Ok, update npm-shrinkwrap with `npm run update-...`, change loader in https://github.com/CartoDB/cartodb/pull/7700 and ready to go!

xavijam:
@jvillarf failing on-premise tests, should we be worried?

jvillarf:
@xavijam it has failed while trying to download npm files from github... I've checked it and github is using an extra IP range that wasn't included in our firewall. I've already included it. 

Thanks a lot for reporting!

xavijam:
Go ahead!

Issue #7674, opened 2016-06-06
Create table view


xavijam:
### DESIGN (@urbanphes)

- [x] Loading state (when rows are being requested).
- [ ] Error state (when a rows request fails).
- [x] 0 rows result (query applied, page doesn't exist,...).
- [x] Pagination elements (please, buttons, no magic scrolls).
- [x] No rows in a dataset.
- [x] Notification bar with important info (query applied, export query result, clean query,...). -> Those info will be placed in the left panel.
- [x] Change column type (it appears as another dropdown within the column options).
- [x] Editor views.
  - [x] String edition.
  - [x] Number edition.
  - [x] Date edition.
  - [x] Boolean edition.

### IMPLEMENTATION (@xavijam)

- [x] Create main view.
- [x] Create table head (thead) view.
- [x] Create table body (tbody) view.
- [x] Start loading first rows.
- [x] Add column behaviour.
- [x] Add row behaviour.
- [x] Possibility to disable table.
- [x] Add possibility to sort by any column.
- [x] Implement way to select between map/dataset.
- [x] Row options.
  - [x] Add new one.
  - [x] Copy value.
  - [x] Close options when table is scrolled.
  - [x] ~~Edit value~~ -> Included in edition views.
  - [x] Highlight row and cell when options menu is showed.
  - [x] Delete (modal).
    - [x] Styles (@piensaenpixel): red button, red title,....
  - [x] Open options above for the last rows.
  - [x] Scroll when new row is added.
  - [x] ~~Disable row if it is going to be removed, if there is error, enable it. All ok, remove it.~~ To be done with delete modal dialog.
- [x] Review table problems under Safari and Firefox.
- [x] Disable copy for Safari browsers.
- [ ] Add sync info if exists.
- [ ] Disabled state.
- [ ] Rows pending tests.

---

- [x] Implement pagination.
  - [x] Loader state.
  - [x] Error state.
  - [x] ~~Go to page X~~.
  - [x] Style (@piensaenpixel).
  - [x] Pending tests.

---
- [ ] Result views:
  - [x] Loader state.
  - [x] Error state.
  - [ ] No page result.
  - [ ] 0 rows.
  - [ ] Style (@piensaenpixel).

---
- [x] Column options.
  - [x] Delete (modal).
  - [x] Rename (modal).
  - [x] Change type (modal).
  - [x] Add new one.
  - [x] Order by.
  - [x] Styles (@piensaenpixel).
  - [x] Pending tests.

---
[...]

PR #7719, opened 2016-06-07
Table implementation approach

xavijam:
Table track: https://github.com/CartoDB/cartodb/issues/7674

baby-steps-table

Basically the most important thing is, table view is generated based on an `analysisDefinitionNodeModel`, that's something we can't change if we want to visualize the results from an analysis node.

```js
new TableView({
  analysisDefinitionNodeModel: analysisDefinitionNodeModel,
  configModel: configModel
});
```

I decided to not touch `query-schema-model` class, and create mirrored collections for getting dataset/table results (within `query-schema-model` there are `rowsSampleCollection` and `columnsCollection`).If there any change in the `query-schema-model` (SQL is applied for example) and it is fetched again, mirrored colletions were reseted.
I'm not hundred percent sure about this decision, but it is true that `query-schema-model` should not know anything about table names, thing that for example column-model needs in order to create a new one ([code](https://github.com/CartoDB/cartodb/pull/7719/files#diff-3add5371a0b55405feb6e1cac00a9964R11)).

This is a first approach, don't take into account that there is no tests, views are uncompleted, and table can only be showed by a global method.

@CartoDB/frontend


lib/assets/javascripts/cartodb3/components/table/body/table-body-row-view.js
```diff
@@ -0,0 +1,51 @@
...
+    this._generateRowsHTML();
+    return this;
+  },
+
+  _initBinds: function () {
+    this.model.bind('remove', this.remove, this);
+  },
+
+  _generateRowsHTML: function () {
+    var geom = this._querySchemaModel.getGeometry();
```
xavijam: Create a view per each cell is useless because it will not provide anything, only bad performance.

lib/assets/javascripts/cartodb3/data/query-schema-model.js
```diff
@@ -47,7 +47,7 @@ module.exports = Backbone.Model.extend({
   },
 
   fetch: function (opts) {
-    if (!this.canFetch()) return;
+    if (!this.canFetch() && (!opts || !opts.force)) return;
```
xavijam: Just in order to fetch in any case (not a final decision).

lib/assets/javascripts/cartodb3/components/table/paginator/table-paginator-view.js
```diff
@@ -0,0 +1,35 @@
...
+
+  tagName: 'button',
+
+  events: {
+    'click': '_onClick'
+  },
+
+  initialize: function (opts) {
+    if (!opts.rowsCollection) throw new Error('rowsCollection is required');
+    if (!opts.tableViewModel) throw new Error('tableViewModel');
```
nobuti: required missing I guess

nobuti:
👏

lib/assets/javascripts/cartodb3/editor.js
```diff
@@ -223,4 +224,13 @@ deepInsights.createDashboard('#dashboard', vizJSON, {
   window.analysisDefinitionNodesCollection = analysisDefinitionNodesCollection;
   window.deepInsightsIntegrations = deepInsightsIntegrations;
   window.editorModel = editorModel;
+
+  window.addTableView = function () {
+    $('#dashboard').html(
+      new TableView({
```
viddo: I know the old "table view" was called table-view, which made sense then when the table (primary at least) represented a database table, but since this now represents the SQL schema+data I think another name would be more suitable, e.g. `QueryDataView` or something on those lines.

lib/assets/javascripts/cartodb3/editor.js
```diff
@@ -223,4 +224,13 @@ deepInsights.createDashboard('#dashboard', vizJSON, {
...
+        analysisDefinitionNodeModel: layerDefinitionsCollection.at(1).getAnalysisDefinitionNodeModel(),
```
viddo: Looking at the usages of this object, why not just pass `querySchemaModel` and `table_name` here as these are the real dependencies you want? Only require what you (really) need.

lib/assets/javascripts/cartodb3/components/table/body/table-body-row-view.js
```diff
@@ -0,0 +1,51 @@
...
+      var columnType = columnModel.get('type');
+
+      if (columnType !== 'geometry') {
+        html += '<td><div class="Table-cell">' + value + '</div></td>';
+      } else {
+        html += '<td>' + simpleGeometry + '</td>';
+      }
+    }, this);
+
+    this.el.innerHTML = html;
```
xavijam: Yeah, next commit solves that, stay tuned.

lib/assets/javascripts/cartodb3/editor.js
viddo: Looking further… table_name only seems to be required to decide whether a column can be added, so rather than passing that I'd say that this should be mapped here already: `new TableView({ read_only: !!node.get('table_name'), …`, OR even better, have that encapsulated in the columns/column-model code somehow so you just need to check `columnsCollection.canCreateColumn` or similar, the view should not need to know _how_ that's decided.

lib/assets/javascripts/cartodb3/editor.js
xavijam: 👍

viddo:
Other looks promising 👍

lib/assets/javascripts/cartodb3/components/table/paginator/table-paginator-view.js
xavijam: :+1:

lib/assets/javascripts/cartodb3/editor.js
xavijam: Taking that into account :+1:

lib/assets/javascripts/cartodb3/editor.js
xavijam: The thing is, it ends up being a Table component, but it is true that it doesn't belongs to a Table definition anymore... hum...

javierarce:
Not much more to say other than looks like a great start.

xavijam:
_"Dale zapato"_!

lib/assets/javascripts/cartodb3/editor/layers/change-view-buttons.tpl
```diff
@@ -0,0 +1,8 @@
+<div class="u-flex u-justifySpace u-bSpace--xl">
```
xavijam: This template and the next class are not permanent, they will be until we move that logic to the editor router.

alonsogarciapablo:
Vamos @xavijam! You got this! 😄

xavijam:
Still lot of things to do @alonsogarciapablo 😞 , but I'm on it! 💪

xavijam:
Merging this first part when frontend tests are green. This belongs to this ticket https://github.com/CartoDB/cartodb/issues/7674, and represents, more or less, the 45% percent of the feature.

nobuti:
💪  vamos!

PR #7722, opened 2016-06-08
Basemaps: cartodb, stamen, here, color

alonsogarciapablo:
cc: @matallo


matallo:
this is awesome

just spotted when going color => labels on top, the labels layer won't get the proper name, but I'll take it from there

thanks mate

matallo:
this PR will include basemaps management for:

- CartoDB basemaps (labels below and on top)
- Stamen basemaps (labels below and on top)
- background color

WIP but just wanted to point out a couple of points before moving on with the tests:

---

we need to support fixed opacity in current fill component, current baseLayer color is a plain hex string (i.e. `#FABADA`) thus we cannot pass the object (we can but it won't work) the fill component generates

```
{
fixed: '#FABADA',
opacity: 1
}
```

ideally I would hide the fields in the component, but that is more work for sure  `¯\_(ツ)_/¯`

right now i'm doing the trick of passing the fill an object with the color from the layer model https://github.com/CartoDB/cartodb/pull/7722/files#diff-808cbfa04ba6911427606fbad3ebb8dbR30

then taking the color from the object and passing it as the string
https://github.com/CartoDB/cartodb/pull/7722/files#diff-66b394241bd69d57c87c1c6cdfeb62e4R27

whatever the opacity is

/cc @javierarce 

---

I made some changes in the layer definition model to be able to pass the color in the options json when saving the layer
https://github.com/CartoDB/cartodb/pull/7722/files#diff-96ed1f8a6f25d43fa3d5758b58158c56R12

this is how the request should look (it was omiting the color before)
screen shot 2016-06-08 at 20 58 03

however, I made some tests break in my way, I should take a look to this with @nobuti 

---

I've spotted when doing changes in the baselayers very quick (change baselayer before labels on top finished shaving, have unintended effects (having more than one labels on top layer). In the current editor we prevent this by disabling the basemaps dropdown until everything is finally saved, shall I replicate the same behaviour @xavijam ?

---

when trying to open the fill component it will currently break unless you comment these lines
https://github.com/CartoDB/cartodb/blob/basemaps-without-category-on-top/lib/assets/javascripts/cartodb3/components/form-components/editors/fill/input-color/input-color-dialog-content.js#L10-L12

these opts are needed when generating the `ValueContentView` https://github.com/CartoDB/cartodb/blob/basemaps-without-category-on-top/lib/assets/javascripts/cartodb3/components/form-components/editors/fill/input-color/input-color-dialog-content.js#L110-L117

despite I used hidePanes 'value' when creating the form as I won't need it here

@javierarce can we take a look at this when you have a minute? ^^^

---

I'm creating the tests and asking for CR once I sort this out

/cc @xavijam @javisantana

nobuti:
@matallo I've fixed the tests.

matallo:
BOOM, thanks @nobuti

xavijam:
@matallo, about waiting until new basemap is set and save, fine by me, but you should talk with design people in order to check how to manage/show that state.

javisantana:
I miss some tests :tmcw:

:es: go on

PR #7753, opened 2016-06-09
Background importer

### LINKED ISSUE #7699: Migrate background-polling functionality to new Notifier
- [x] Stop using old background polling view and only use new Notifier stuff.
- [x] Keep using open-dialog-buttons when an import/geocoding has been completed, failed, etc.
- [x] There is a bug with the geocoding polling, it make a geocoding request each X seconds. It should not happen because geocoding polling should not be enabled by default within the editor.

cc @viddo @javierarce

nobuti:
This PR closes #7699.

It adds some improvements to the notifier:
- it accepts html code for info message.
- it can show both buttons at the same time.
- it links better event and actions for buttons.

I've found a bit difficult to build the message for the notification's message, there is some code repetition. Also all the old code needs to be deleted. I want to make sure that this is the line to follow.

cc @xavijam


xavijam:
@piensaenpixel, check CSS please!

app/assets/stylesheets/editor-3/notifier.css.scss
```diff
@@ -42,3 +42,9 @@
   @include flex(1);
   padding-right: $baseSize * 2;
 }
+
+.Notifier-actions {
+  & + & {
```
xavijam: I'm not super fan of that, sorry, I had to say 🎉  😬 .

nobuti:
Removed one `cdb.god`, only one left, but I think it's not implemented right now. Not sure if we should leave it as a reminder.

xavijam:
I'd prefer to remove it and create a ticket about, add possibility to drag a file in order to import it. How does it sound?

lib/assets/javascripts/cartodb3/components/importer-manager/warnings-details-view.js
```diff
@@ -0,0 +1,39 @@
+var CoreView = require('backbone/core-view');
```
xavijam: This is a good one, I didn't remember that.

lib/assets/javascripts/cartodb3/components/importer-manager/background-importer.js
```diff
@@ -0,0 +1,181 @@
...
+var ImportItemView = require('./background-import-item-view');
+var GeocodingItemView = require('./background-geocoding-item-view');
+var ImportLimitItemView = require('./background-import-limit-view');
+var ImportsModel = require('../../data/background-importer/imports-model');
+
+/**
+ *  Background polling manager
+ *
+ *  It will pool all polling operations that happens
+ *  in Cartodb, as in imports and geocodings
```
xavijam: "... in the platform..." ;)

xavijam:
Looks promising:

- 🔥 the old code.
- what do you propose to not repeat the same notification message?

nobuti:
Done.

xavijam:
"Dale zapato" when 🍏  :)

Issue #7747, opened 2016-06-09
Connectors Prototype


jgoizueta:
Implement MySQL connector (using fdw_odbc) so that we can create an import that generates a dataset from a query in a MySQL database with:

```
curl -v -H "Content-Type: application/json" \
     -d '{"odbc":"DRIVER=MySQL;SERVER=10.0.32.28;USER=user;PASSWORD=123456;DATABASE=mydb;SQL_QUERY=..."}' \
     "http://cdb.localhost.lan:3000/api/v1/imports/?api_key=xxxyyzzz"
````

**Comment by javisantana** [comment]
where is the sql query?

**Comment by jgoizueta** [comment]
good point, well, the actual test API call, due to some restrictions with the current odbc_fdw will be this:

```
curl -v -H "Content-Type: application/json" \
     -d '{"odbc":"dsn=mysql;username=jgoizueta;password=123;database=bigco;table=sales;sql_query='"'"'select * from sales'"'"'"}' \
     "https://http://cdb.localhost.lan:3000/api/v1/imports/?api_key=fff19a2d37518fc9d573919043da9310b70537d4"
```

**Comment by jgoizueta** [comment]
Once the simple import works, we'll enable connectors for syncs also:

```
curl -v -H "Content-Type: application/json" \
     -d '{"odbc":"dsn=mysql;username=jgoizueta;password=123;database=bigco;table=sales;sql_query=select * from sales", "interval":10000}' \
     "http://cdb.localhost.lan:3000/api/v1/synchronizations/?api_key=xxxyyyzzz"
```

**Comment by jgoizueta** [comment]
Note: the `sales` parameter defines the name of the local table imported and has no relation to the datasource which is solely defined by the query.

**Comment by jgoizueta** [comment]
Synchronizations are working now as well:

```
curl -v -H "Content-Type: application/json" \
    -d '{"odbc":"dsn=mysql;username=xxxx;password=yyyy;database=mydb;table=sales;sql_query=select * from sales", "interval":900}' \
    "http://cdb.localhost.lan:3000/api/v1/synchronizations/?api_key=xxxyyyzzz"
```

rafatower:
I'd say the prototype as such was completed and we need to have a different ticket with different scope to track progress. So I'm closing this and redefining the scope for the next milestone.

PR #7755, opened 2016-06-09
Prototype new kind of import: connector (using odbc_fdw)

jgoizueta:
#7747 

This is prototype code that requires odbc_fdw to be installed in the database server and DSNs defined
for the sources to be used. It also has the imported table schema fixed in connector.rb.

To make this work odbc_fdw must be installed in the database host, as well as odbc drivers
for the sources to be used:

Building and installing `odbc_fdw`:

```bash
# required odbc dev. libs:
apt-get install -y unixodbc-dev
# required postgresql dev. libs:
apt-get install -y postgresql-server-dev-9.5

git clone https://github.com/CartoDB/odbc_fdw.git
cd odbc_fdw
PATH=$(pg_config --bindir):$PATH make USE_PGXS=1
sudo PATH=$(pg_config --bindir):$PATH make USE_PGXS=1 install
```

Installing ODBC drivers:

```bash
# MySQL driver
sudo apt-get install libmyodbc
sudo odbcinst -i -d -f /usr/share/libmyodbc/odbcinst.ini

# PostgreSQL driver
sudo apt-get install odbc-postgresql

# SQLite driver
sudo apt-get install libsqliteodbc
```

This can be tested with:

```
curl -v -H "Content-Type: application/json" \
    -d '{"odbc":"driver=MySQL;host=xxx;username=yyyy;password=zzzz;database=mydb;table=sales;sql_query=select * from sales;columns=product_id int, total int", "interval":900}' \
    "http://cdb.localhost.lan:3000/api/v1/synchronizations/?api_key=xxxyyyzzz"
```


jgoizueta:
For the time being columns will be passed as part of the connection string:

```
 curl -v -H "Content-Type: application/json" \
     -d '{"odbc":"dsn=mysql;username=xxx;password=yyy;database=mydb;table=sales;sql_query=select sale_date, total from sales;columns=timestamp without time zone, total int", "interval":900}' \
     "http://cdb.localhost.lan:3000/api/v1/synchronizations/?api_key=xxxyyyzzz"
```

jgoizueta:
For connector DataImport & Synchronizations, the `service_name` column is set to "connector" and `service_item_id` to a connection string of the form `ODBC:opt1=val1;opt1=val2;...`.

These columns are otherwise used for the datasource type (public_url, gdrive, dropbox, ...) and the url.

We should probably add specific columns for connectors.

jgoizueta:
@rafatower please review

jgoizueta:
Hold on the review, I'm attending hound's complaints and I'll rebase when I'm finished

rafatower:
okay, just let me/us know :smiley:

jgoizueta:
OK @rafatower, this is ready for reviewing now

rafatower:
re: the installation notes in the PR description:
let's split this into 2 things and create a ticket for the work about generating a package:
- run dependencies (unix-odbc, postgres)
- build dependencies (unix-odbc-dev, whatever package has the postgres headers and the pgxs support files (included in the makefile)

We can also polish the README and get rid of some options.

(I quite disagree with the reference to the `/usr/local/pgsql91/`)

jgoizueta:
oops `/usr/local/pgsql91/` should be set as in the prior line (using pg_config)

app/controllers/api/json/imports_controller.rb
```diff
@@ -40,10 +40,13 @@ def create
 
         if params[:url].present?
           validate_url!(params.fetch(:url)) unless Rails.env.development? || Rails.env.test?
-          options.merge!(data_source: params.fetch(:url))
+          options[:data_source] = params.fetch(:url)
+        elsif params[:odbc].present?
+          options[:service_name]    = 'connector'
+          options[:service_item_id] = "ODBC:#{params[:odbc]}"
```
jgoizueta: `param1=value1;param2=value2;...`
With the allowed paramters `driver`, `host`, `port`, `username`, `password`, `database`, `sql_query`, `sql_count`, `columns`.

app/models/data_import.rb
```diff
@@ -671,6 +673,41 @@ def new_importer
     importer.nil? ? false : importer.success?
   end
 
+  def new_connector
```
rafatower: this method is running a lot of things other than instantiating a connector. A method called `new_whatever` should not do that much job IMHO.

rafatower: looks like a factory coupled with the processing to me

app/models/data_import.rb
```diff
@@ -671,6 +673,41 @@ def new_importer
...
+      connector, registrar, quota_checker, database, id,
+      overviews_creator,
+      destination_schema, public_user_roles
+    )
+    log.append 'Before importer run'
+    importer.run(tracker)
+    log.append 'After importer run'
+
+    store_results(importer, connector)
+    importer.nil? ? false : importer.success?
```
rafatower: it can't be nil at this point, can it?

lib/cartodb/import_error_codes.rb
```diff
@@ -139,6 +139,21 @@ module CartoDB
...
       source: ERROR_SOURCE_USER
     },
+    1500 => {
+      title: 'Connector error',
+      what_about: "The connector used to fetch the resourced failed.",
+      source: ERROR_SOURCE_USER
+    },
+    1501 => {
+      title: 'Connector invalid channel',
+      what_about: "Type of connector you were trying to use is invalid or unsupported",
```
rafatower: minor comment: "*The* type of connector..."

services/importer/lib/importer/connector.rb
```diff
@@ -0,0 +1,218 @@
...
+        @job.log "Creating Server"
+        execute_as_superuser create_server_command
+        @job.log "Creating Foreign Table"
+        execute_as_superuser create_foreign_table_command
+        @job.log "Copying Foreign Table"
+        execute "CREATE TABLE #{qualified_table_name} AS SELECT * FROM #{foreign_table_name};"
+      rescue => error
+        @job.log "Connector Error #{error}"
+        @results.push result_for(@job.schema, table_name, error)
+      else
```
rafatower: This is the first time I see this ruby lang construct, nice!

services/importer/lib/importer/connector.rb
```diff
@@ -0,0 +1,218 @@
...
+        @id = @job.id
+        @unique_suffix = @id.delete('-')
+        channel, conn_str = connector_source.split(':')
+        raise InvalidChannelError.new(channel, @user) unless channel.casecmp('odbc') == 0
+        @conn_str = conn_str
+        # TODO: parser params properly, make param names lowercase
+        @params = Hash[@conn_str.split(';').map { |p| p.split('=').map(&:strip) }]
+        # Temporary hack to allow equal signs in parameter values (e.g. sql_query):
+        # the equal sign should be codified as .eq.
+        @params = Hash[@params.map { |k, v| [k, v.gsub('.eq.', '=').gsub('.sc.', ';')] }]
```
rafatower: a bit hard to read...

rafatower:
I left more questions than comments in the code. Thanks for the good job!

services/importer/lib/importer/connector.rb
```diff
@@ -0,0 +1,218 @@
...
+      end
+
+      def run(tracker = nil)
+        @tracker = tracker
+        @job.log "Connector #{@conn_str}"
+        # TODO: deal with schemas, org users, etc.,
+        # TODO: logging with CartoDB::Logger
+        table_name = @job.table_name
+        qualified_table_name = %{"#{@job.schema}"."#{table_name}"}
+        execute_as_superuser "CREATE EXTENSION IF NOT EXISTS odbc_fdw;"
```
jgoizueta: OK, agreed, but was handy for prototyping.

services/importer/lib/importer/connector.rb
```diff
@@ -0,0 +1,218 @@
+# encoding: utf-8
+
+require_relative './exceptions'
+
+module CartoDB
+  module Importer2
+    class Connector
```
jgoizueta: Dealing with (fdw_odbc) connectors. It is the equivalent of Runner when the import uses a connector instead of some import file. Like the runner it has a `run` method, that, instead of downloading, unpacking, ogr2ogring... configures the FDW connection, copies the table over it...

app/models/data_import.rb
jgoizueta: Good point. I'll think it over.

app/models/data_import.rb
jgoizueta: Right, shouldn't check for nil here if we didn't in the previous `importer.run`

services/importer/lib/importer/connector.rb
```diff
@@ -0,0 +1,218 @@
...
+
+      # Temptative terminology
+      # * connector channel (ODBC, ...)
+      # * provider          (MySQL, ...)     [driver]
+      # * connection        (database/query) [specific data source]
+
+      class ConnectorError < StandardError
+        attr_reader :channel_name, :user_name
+
+        def initialize(message = 'General error', channel = nil, user = nil)
```
jgoizueta: Only ODBC, and this is now coupled to the parameters (driver, host, etc.) we accept. So basically it's fixed/hardcoded, but at the time I wrote that I thought we might want to leave space for other possibilities, such as other FDWs...

services/importer/lib/importer/connector.rb
jgoizueta: I admit it, Q&D. Should be replace by proper parsing, or by having separate parameters.

jgoizueta:
cc @rafatower the las commit adapts to the new API, where everything is passed as JSON parameters in a `connector` parameter:

```json
{ "connector": {
    "connection": { "driver":"MySQL", "host":"10.0.32.28" }, 
    "table": "sales",
    "sql_query": "SELECT * FROM sales",
  },
  "interval": 900 }
```

The ODBC attributes that will make up the connection string are passed within the `connector.connection` parameter, and are open so that driver-specific attributes can be used.

The valid parameters inside `connector` are:

*  `connection` ODBC connection attributes (driver, host, port, database, … including any required driver-specific attributes)
* `schema` name of the remote schema (connection database name is used as default)
* `table` name of the remote table to import (if no sql_query given) and local name of the imported result table
* `sql_query` (optional): SQL code to be executed remotely to produce the dataset to be imported. If missing, schema.table will be imported.
* `sql_count` (optional): SQL code to compute the number of rows of the dataset. This shouldn't be required since a `count` query is generated by odbc_fdw but could be necessary for some driver having non-standard SQL.
* `encoding` (optional): character encoding used by the external database; default is UTF-8. (this isn't implemented in odbc_fdw yet).

This is being tested in staging

jgoizueta:
@iriberri @ethervoid can you review this? 

Tested in staging, and those tests will succeed, believe me :fingers_crossed:

iriberri:
:eyes:

lib/cartodb/import_error_codes.rb
```diff
@@ -139,6 +139,21 @@ module CartoDB
...
+      source: ERROR_SOURCE_USER
+    },
+    1501 => {
+      title: 'Connectors disabled',
+      what_about: "The type of connector you were trying to use is not enabled",
+      source: ERROR_SOURCE_USER
+    },
+    1502 => {
+      title: 'Connector invalid parameters',
+      what_about: "The connector couldn't be configure because the parameters were not valid.",
```
iriberri: configured*

app/models/data_import.rb
```diff
@@ -391,7 +392,15 @@ def dispatch
...
+    if service_name == 'connector'
+      importer, runner    = new_importer_with_connector
+      datasource_provider = nil
+      manual_fields       = nil
+    else
+      importer, runner, datasource_provider, manual_fields = new_importer
+    end
+    execute_importer importer, runner, datasource_provider, manual_fields
   rescue => exception
     puts exception.to_s + exception.backtrace.join("\n")
```
ethervoid: why do we have a puts here? Maybe is time to log it

app/models/data_import.rb
```diff
@@ -625,11 +640,6 @@ def new_importer
     if had_errors
       importer = runner = datasource_provider = nil
     else
-      tracker       = lambda { |state|
```
ethervoid: Why did you delete the tracker?

app/models/data_import.rb
```diff
@@ -391,7 +392,15 @@ def dispatch
     self.state = STATE_UPLOADING
     return migrate_existing   if migrate_table.present?
     return from_table         if table_copy.present? || from_query.present?
-    new_importer
+
+    if service_name == 'connector'
+      importer, runner    = new_importer_with_connector
+      datasource_provider = nil
```
ethervoid: You can make the assignation as you do below...

app/models/data_import.rb
ethervoid: nevermind, you move it to new_importer method

lib/cartodb/import_error_codes.rb
jgoizueta: oops thanks

app/models/data_import.rb
jgoizueta: Yes, didn't pay attention that, but it's not nice

app/models/data_import.rb
jgoizueta: You mean returning datasource_provider and manual_fields from `new_importer_with_connector` (always as nil), right?

Mmm, I have a vague memory I did that initially 😆

services/importer/lib/importer/connector.rb
```diff
@@ -0,0 +1,251 @@
...
+          CREATE USER MAPPING FOR postgres SERVER #{server_name};
+        }
+      end
+
+      def create_foreign_table_command
+        options = table_params
+        if @columns.present?
+          options_list = options.map { |k, v| "#{k} '#{v}'" } * ",\n"
+          %{
+            CREATE FOREIGN TABLE #{foreign_table_name} (#{@columns * ','})
```
iriberri: What if this runs in an organization? Isn't the schema missing here?

jgoizueta: Mmm, we actually are not taking care of the schema where the foreign table is created. It works because the foreign table is temporary and has a unique name, but I guess we should be more careful and either use the user's schema or `cdb_importer`.

jgoizueta: As the change to specify the foreign tables' schema is not trivial (it's easy but must take care of all the places where the foreign table is referenced and this needs some testing) and I'm confident this will work as it is I'll open a separate ticket for it and leave it as is for the initial release.

ethervoid:
At first sight looks good, despite some minor comments and tests failing.

PR #7761, opened 2016-06-10
7752 read quota redis both formats

rafatower:
this is the good one @iriberri


services/dataservices-metrics/spec/unit/service_usage_metrics_spec.rb
```diff
@@ -30,9 +30,8 @@
     end
 
     it "does not request redis twice when there's no need" do
-      @redis_mock.zincrby('org:team:here_isolines:isolines_generated:201606', 3141592, '20')
+      @redis_mock.expects(:zscore).once.with('org:team:here_isolines:isolines_generated:201606', '20').returns(3141592)
```
iriberri: :+1:

iriberri:
Looks even better to me now! I totally missed that the spec was not included in the Makefile.

PR #7762, opened 2016-06-10
Fix some tests in parallel by removing hidden Rails dependency

javitonino:
These two tests did fail in the parallel execution and required a second try. This is because they depend on function on a Rails initializer, while Rails is not loaded for unit tests. Solved by stubbing.


javitonino:
Can you CR @rafatower

rafatower:
:+1

PR #7764, opened 2016-06-10
Saving widget definition when type is changed

### LINKED ISSUE #7756: Changing widget type doesn't fire widget api call / change not persisted
STR:

1. Create a map
2. Add a widget
3. Change the widget type

Nothing happens, embed (for example) is not updated. Looking at the network no call is fired to the widgets API.

xavijam:
Not sure if we should save it within the method that changes widget type and attributes or where I've placed it.

CR: @javierarce 
cc @viddo 

Fixes #7756


viddo:
> Not sure if we should save it within the method that changes widget type and attributes or where I've placed it.

I think where you placed it is fine.

javierarce:
IMHO I think is ok where you placed it. Code looks 👍

xavijam:
Thanks mates! Merging due to frontend tests are passing 💋

Issue #7436, opened 2016-05-19
Render a different color per layer letter representation


viddo:
e.g. ![screen shot 2016-05-24 at 17 09 08](https://cloud.githubusercontent.com/assets/978461/15509033/3f4c3cca-21d2-11e6-9ab7-65e6f6cf7b0a.png)

scope: 
- the color should be based on the most prominent color used output from a layer's styles… 

TBD:
- should it be allowed to have same color for different layers?

PR #7766, opened 2016-06-10
Colorful layer icons :)

alonsogarciapablo:
close https://github.com/CartoDB/cartodb/issues/7436.

I'm following the same approach that we're using to assign letters to layers in the code.
@viddo can you please take a look? Thx!

I "randomly" picked the following color palette and colors will be assigned in this order:

colors

@piensaenpixel @urbanphes ^^ (what do you think about these colors? can you guys suggest a better color pallete?)

Here's how it looks:

screen shot 2016-06-10 at 13 05 09


viddo:
Looks legit, but i think an even easier solution would be a function that
takes the letter representation as input and return a color, that way we
don't have to modify the layer model and collection code more than
necessary.

On Friday, June 10, 2016, Pablo Alonso <notifications@github.com> wrote:

> close #7436 <https://github.com/CartoDB/cartodb/issues/7436>.
>
> I'm following the same approach that we're using to assign letters to
> layers in the code.
> @viddo <https://github.com/viddo> can you please take a look? Thx!
>
> I "randomly" picked the following color palette and colors will be
> assigned in this order:
>
> [image: colors]
> <https://cloud.githubusercontent.com/assets/390398/15962588/69947e38-2f0c-11e6-8dbb-968c65e81aa9.jpg>
>
> @piensaenpixel <https://github.com/piensaenpixel> @urbanphes
> <https://github.com/urbanphes> ^^ (what do you think about these colors?
> can you guys suggest a better color pallete?)
>
> Here's how it looks:
>
> [image: screen shot 2016-06-10 at 13 05 09]
> <https://cloud.githubusercontent.com/assets/390398/15962523/04a15cee-2f0c-11e6-8b06-2df9409cbc22.png>
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/CartoDB/cartodb/pull/7766
> Commit Summary
>
>    - Assign colors to layer models when they are created if they don't
>    have one
>    - Replaced _.each by _.find and renamed argument
>    - Removed duplication
>
> File Changes
>
>    - *A* lib/assets/javascripts/cartodb3/data/layer-colors.js
>    <https://github.com/CartoDB/cartodb/pull/7766/files#diff-0> (29)
>    - *M* lib/assets/javascripts/cartodb3/data/layer-definition-model.js
>    <https://github.com/CartoDB/cartodb/pull/7766/files#diff-1> (5)
>    - *M*
>    lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
>    <https://github.com/CartoDB/cartodb/pull/7766/files#diff-2> (25)
>    - *M* lib/assets/javascripts/cartodb3/editor/layers/layer-view.js
>    <https://github.com/CartoDB/cartodb/pull/7766/files#diff-3> (3)
>    - *M* lib/assets/javascripts/cartodb3/editor/layers/layer.tpl
>    <https://github.com/CartoDB/cartodb/pull/7766/files#diff-4> (2)
>    - *A* lib/assets/test/spec/cartodb3/data/layer-colors.spec.js
>    <https://github.com/CartoDB/cartodb/pull/7766/files#diff-5> (35)
>    - *M*
>    lib/assets/test/spec/cartodb3/data/layer-definitions-collection.spec.js
>    <https://github.com/CartoDB/cartodb/pull/7766/files#diff-6> (34)
>
> Patch Links:
>
>    - https://github.com/CartoDB/cartodb/pull/7766.patch
>    - https://github.com/CartoDB/cartodb/pull/7766.diff
>
> —
> <https://github.com/CartoDB/cartodb/pull/7766>, or mute the thread
> <https://github.com/notifications/unsubscribe/AA7uHejgU-lKZdKwRfgGy1nfAUm8Iuejks5qKUrlgaJpZM4Iy3Wx>
> .
>


-- 

*Nicklas Gummesson*, software engineer
*CartoDB* — Map Your World's Data

Live sessions to learn CartoDB <http://cartodb.com/webinars/>

alonsogarciapablo:
@viddo I liked your suggested approach. It's done in https://github.com/CartoDB/cartodb/pull/7766/commits/1855648a487a6b52701fa2f752734a77a0cf8954. Thanks!

xavijam:
Could we add #F24440 as the first color please :).

lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
```diff
@@ -113,9 +114,10 @@ module.exports = Backbone.Collection.extend({
       _.omit(d, ['options']
     ));
 
-    if (attrs.order !== 0) {
-      // Only for non-basemap layers
+    // Only for non-basemap layers
+    if (layerTypesAndKinds.isCartoDBType(o.type) || layerTypesAndKinds.isTorqueType(o.type)) {
```
alonsogarciapablo: @viddo I'm a bit confused about `kind` vs `options.type`. I added this check here to make sure that `letter` and `color` attributes are only set for CartoDB and Torque layers (`attrs.order !== 0` is not enough since the layer with labels doesn't have an order of 0 and shouldn't get a letter/color).

Tests are passing now, but editor is not working cause some layers don't actually have a type. Here's what the global `layersData` variable contains right now:

```
[
  {
    "options": {
      ...
    },
    "kind": "tiled"
  },
  {
    "options": {
      ...
    },
    "kind": "carto",
  },
  {
    "options": {
      "type": "CartoDB",
      ...
    },
    "kind": "carto",
    ...
  },
  {
    "options": {
      "type": "Tiled",
      ...
    },
    "kind": "tiled",
    ...
  }
]
```

Do you think I should replace all usages of `layerTypesAndKinds.isSomethingType(m.get('type'))` in this file by `layerTypesAndKinds.isSomethingKind(m.get('kind'))`?

alonsogarciapablo: From what I see [here](https://github.com/CartoDB/cartodb/blob/9f61aa66fecd4ec0c57ec8c72738c5989072116f/lib/assets/javascripts/cartodb3/data/layer-definition-model.js#L39), `kind` seems to be what I should be using since we're parsing data coming from the server. From the comment I the previous link, I see `type` is what the frontend uses.

Perhaps it's a good time to create a factory that, given some layer data coming from the server, creates a layerDefinitionModel with the right attributes?

viddo: > I added this check here to make sure that letter and color attributes are only set for CartoDB and Torque layers (attrs.order !== 0 is not enough since the layer with labels doesn't have an order of 0 and shouldn't get a letter/color).

I changed this on Fri on the branch I'm currently working on 👉 https://github.com/CartoDB/cartodb/compare/7686-drag-drop-node-as-new-layer#diff-f2e4d12c6436600353eb955a6ee77062R116 because the zero-index check wasn't enough, the layers-on-top layer shouldn't have the letter representation either. Not sure if that helps your use-case(s)?

viddo: > From what I see here, kind seems to be what I should be using since we're parsing data coming from the server. From the comment I the previous link, I see type is what the frontend uses.

The thing is that kind is an actual column on the persisted layers and what is actually validated by backend where as `type` is just another-attr-in-the-options bag. I guess we still have to continue to use both to maintain backward compatibility, but `kind` should be the priority/mandating prop. As you can see in my branch I've updated some tests after I did my internal change.

> Perhaps it's a good time to create a factory that, given some layer data coming from the server, creates a layerDefinitionModel with the right attributes?

Not sure if it's worth, and backbone already provides the mechanisms to translate to/from a server-side API (`parse`/`toJSON`) which we already utilize. What value would we get from extracting/changing it to a factory?

alonsogarciapablo:
@viddo I cherry-picked [this commit](https://github.com/CartoDB/cartodb/pull/7766/commits/4fe3419ae831bc7558fe46e02bf5931c1c138a5d) and [fixed a bunch of tests](https://github.com/CartoDB/cartodb/pull/7766/commits/dd7ddfd7d918d1ea01813597a53a49c2b0cc476a). Ready for CR now/again! :)

viddo:
:shipit:

PR #7816, opened 2016-06-14
Be more lenient validating email addresses

javitonino:
This is only used in the invitations view. Old code did not allow email addresses with + or with the new longer TLDs. This allows pretty much anything. Worst case, the invitation email is wrong and the invitation is not received, so no harm done.

Pending package bump


javitonino:
Please CR @xavijam

xavijam:
Totally agree 👍

PR #7818, opened 2016-06-14
Do not reset organization_id on failed user deletion

### LINKED ISSUE #6363: Subuser deletion gets rid of organization_id although delete fails
https://github.com/CartoDB/cartodb-management/issues/4379

https://github.com/CartoDB/cartodb/blob/0418ccd6bc640b92291c7a035e97b721bdca50db/app/models/user.rb#L302

According to @juanignaciosl, lines 301 and 302 are a possible source of this problem, as we empty the `organization_id` inside a `begin .. rescue` that will continue in case of failure.

When this happens, destroying the user is raising a:

````
irb(main):008:0> u.destroy
CartoDB::BaseCartoDBError: Tried to delete an organization user without org id
````

javitonino:
Asides from a bit of code cleaning, this only removes `self.organization_id = nil` which didn't make much sense.

User deletion still works for organization users (both owner and subs) and normal users. 

Closes #6363

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -346,7 +346,8 @@ class User < Sequel::Model
     if has_organization
       db_service.drop_organization_user(organization_id, !@org_id_for_org_wipe.nil?) unless error_happened
     elsif ::User.where(database_name: database_name).count > 1
-      raise CartoDB::BaseCartoDBError.new('The user is not supposed to be in a organization but another user has the same database_name. Not dropping it')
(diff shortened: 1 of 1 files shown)


javitonino:
Please CR @gfiorav

gfiorav:
looks ok! I'd test it in staging first though.

📦

javitonino:
Tested all kind of user creation / deletion I could think of. :rocket:

PR #7847, opened 2016-06-14
fixed problem with updating a infowindow in a non-existing layer

javisantana:
cherry-pick from "el sueco"

cc @viddo


viddo:
👍  and frontend tests passing, so good to go I think

PR #7849, opened 2016-06-14
Adds CartoColor ramps

### LINKED ISSUE #7795: Make CartoColors present on the editor
cc/ @javierarce

javierarce:
Closes #7795 

This PR adds the `carto-color` package and updates the categories ramp to support them. If a user had previously selected an old ramp, it will appear as a custom one in the ramp list, next to the list of new ramps.

animation

CR: @xavijam 
CC: @saleiva


xavijam:
💃 🚢 🇮🇹

viddo:
The only dependency change I can see is the cartocolor and I doubt it should include that many changes, right?

What command was run to update the shrinkwrap?

viddo:
Having a look at the shrinkwrap file it looks like it includes `devDependencies` too which it shouldn't, `npm shrinkwrap` should fix it.

PR #7852, opened 2016-06-14
Fix missing geometry type exceptions

### LINKED ISSUE #7728: Error when trying to add an analysis
**STR:**

1. Create a new map from an existing dataset (I tried with points).
2. Click "Add analysis", select "Buffer" / "Trade areas" and click "Add analysis".

**Expected result:**

- Analysis is added, modal is closed.

**Actual result:**

- JS error in the console:
`Uncaught Error: TrackJS Caught: simpleGeometryTypeInput is required`
- Modal is not closed and analysis is not created.

NOTE: I observed that it's possible to add Buffer / Trade Areas analyses after having added a Point in Polygon analysis first.

cc: @viddo

viddo:
Fixes #7728 

It had two origins:
1. Race-condition when the query-schema-model was updated while the
modal was closing, the model listeners weren’t removed until the modal
animation finished. Changed to be removed straight away.
2. Option view required geometry type, even though it might not be
available (yet, or due to failing node results), so relaxed the check

@alonsogarciapablo could you try this branch locally and confirm that it solves your problem, please?

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/modals/modal-view.js
+++ b/lib/assets/javascripts/cartodb3/components/modals/modal-view.js
@@ -1,3 +1,4 @@
+var _ = require('underscore');
 var CoreView = require('backbone/core-view');
 var template = require('./modal.tpl');
 
@@ -59,6 +60,7 @@ module.exports = CoreView.extend({
 
   _onDestroy: function () {
     this.hide();
+    _.invoke(this._subviews, 'allOff');
     this._delayDueToAnimation(function () {
       this.clean();
     });
--- a/lib/assets/test/spec/cartodb3/components/modals/add-analysis/analysis-option-view.spec.js
+++ b/lib/assets/test/spec/cartodb3/components/modals/add-analysis/analysis-option-view.spec.js
@@ -12,52 +12,69 @@ describe('components/modals/add-analysis/analysis-option-view', function () {
         type: 'buffer'
       }
     });
-
-    this.view = new AnalysisOptionView({
-      model: this.model,
-      simpleGeometryTypeInput: 'point'
-    });
-    this.view.render();
   });
 
-  it('should render the info', function () {
-    expect(this.view.$el.html()).toContain('Buffer');
-    expect(this.view.$el.html()).toContain('describes the buffer');
-    expect(this.view.$el.html()).toContain('area of influence');
-  });
-
-  it('should not have any leaks', function () {
-    expect(this.view).toHaveNoLeaks();
-  });
-
-  describe('when selected', function () {
+  describe('when given a simple geometry type', function () {
     beforeEach(function () {
-      this.model.set('selected', true);
-    });
-
-    it('should highlight the item', function () {
-      expect(this.view.el.className).toContain('is-selected');
-    });
-  });
-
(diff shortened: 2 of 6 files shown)


javisantana:
\o/

alonsogarciapablo:
Checking!

lib/assets/javascripts/cartodb3/components/modals/modal-view.js
```diff
@@ -59,6 +60,7 @@ module.exports = CoreView.extend({
 
   _onDestroy: function () {
     this.hide();
+    _.invoke(this._subviews, 'allOff');
```
alonsogarciapablo: First time I see `_.invoke` 👍  and `allOff`. Where's `allOff` defined? I'm sure that there's a reason for doing it this way, but just in case, why are not we just using `this.clean` here (without the delay)? Thanks!

alonsogarciapablo:
👍  👍  👍  Code looks 👍  and it works 👍  locally.

lib/assets/test/spec/cartodb3/components/modals/add-analysis/analysis-option-view.spec.js
```diff
@@ -12,52 +12,69 @@ describe('components/modals/add-analysis/analysis-option-view', function () {
...
+        spyOn(this.model, 'acceptsGeometryTypeAsInput').and.returnValue(false);
+        this.view.render();
+      });
+
+      it('should disable the view', function () {
+        expect(this.view.el.className).toContain('is-disabled');
+      });
+
+      it('should show alternative desc', function () {
+        expect(this.view.$el.html()).toContain('disabled-option-desc');
```
alonsogarciapablo: I guess we could assert that text is "Your layer geometry is something and this analysis needs wadus"... Here and when the geometry is unknown.

lib/assets/javascripts/cartodb3/components/modals/modal-view.js
viddo: http://underscorejs.org/#invoke - _Calls the method named by methodName on each value in the list._

> …, why are not we just using this.clean here (without the delay)?

Because clean removes the view (and sub-views, and [more stuff](https://github.com/CartoDB/cartodb/blob/master/lib/assets/node_modules/backbone/core-view.js#L48-L69)) right away which removes the animation. So, I extracted the model unbinding code to a separate method to be able to use only that.

viddo: > Where's allOff defined? 

It's in core-view, for some reason the changes are surpressed by default when viewed on Github though so you need to click to see it. ¯\_(ツ)_/¯

alonsogarciapablo: Perfect. Thanks for the clarification! :shipit:

lib/assets/test/spec/cartodb3/components/modals/add-analysis/analysis-option-view.spec.js
viddo: In test environment we don't load the actual strings (so output is the token itself), and just like visual things they need to be reviewed manually, so IMHO it's not worth testing in such detail.

PR #7862, opened 2016-06-15
Save node explicitly on layerDef.save

viddo:
Enables to modify an individual node and save it too, needed for SQL editor

@nobuti something on these lines should do the trick, right?


nobuti:
❤️  thank you so much!

PR #7870, opened 2016-06-15
Fix fill form

javisantana:
cc @javierarce

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/form-components/editors/fill/input-number/input-number-content-view.js
+++ b/lib/assets/javascripts/cartodb3/components/form-components/editors/fill/input-number/input-number-content-view.js
@@ -32,7 +32,7 @@ module.exports = CoreView.extend({
     return [
       Math.max(this.options.min, fixed - percent * delta),
       Math.min(this.options.max, fixed + percent * delta)
-    ]
+    ];
   },
 
   _generateForms: function () {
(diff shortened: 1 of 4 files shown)


lib/assets/javascripts/cartodb3/components/form-components/editors/fill/input-number/input-number-content-view.js
```diff
@@ -25,18 +25,41 @@ module.exports = CoreView.extend({
...
+    // calculate the range values based on this
+    var range = this.model.get('range')
+    var min = +range[0];
+    var max = +range[1];
+    if (range && min === max) {
+      var r = this._calculateRangeFromFixed(min);
+      // set the range so changes are propagated
+      min = r[0];
+      max = r[1];
+      this.model.set('range', r);
```
javierarce: I think writing this setter like this won't trigger any change due to the way Backbone works with arrays (we should do something like this instead: 

```this.model.set('range', [min, max]);```

javierarce: The fact that this is probably not triggering any change, makes me think we don't really need to update the model at this point. Instead, we could wait for a change in the form, saving us an extra call.

javierarce:
Just one comment. Aside for that, 👍

lib/assets/javascripts/cartodb3/components/form-components/editors/fill/input-number/input-number-content-view.js
javisantana: Think that r is being generated in a function and that generates a new array.

If we don't set that the change is not performed and the user needs to change manually to see the change

javierarce: Oh, 👍  then

PR #7872, opened 2016-06-15
Adds K-means and weighted centroids analyses

javierarce:
This PR introduces the K-means & Weighted centroid analyses. Using them together is possible to replicate a cluster analysis working with point datasets.

The `moran` analysis for polygons will be handled in another PR.

_k-means_
screen shot 2016-06-15 at 19 12 18

_weighted centroid_
screen shot 2016-06-15 at 19 17 06

CR: @xavijam 
CC: @javisantana @rochoa


lib/assets/javascripts/cartodb3/components/modals/add-analysis/analysis-option-models/kmeans-analysis-option-model.js
```diff
@@ -0,0 +1,8 @@
+var AnalysisOptionModel = require('./analysis-option-model');
```
xavijam: If it does nothing, let's not create it.

xavijam:
Not much to say 👍, in any case, having both @viddo and you working on this, I feel you can review together and merge faster :).

lib/assets/javascripts/cartodb3/components/modals/add-analysis/analysis-option-models/kmeans-analysis-option-model.js
viddo: +1, a model is only needed if custom logic is needed before creating the form model, see [point-in-polygon](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/components/modals/add-analysis/analysis-option-models/point-in-polygon-analysis-option-model.js) for an example

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/base-analysis-form-model.js
```diff
@@ -100,6 +100,29 @@ module.exports = Backbone.Model.extend({
   _setSchema: function (schema) {
     this.schema = schema;
     this.trigger('changeSchema', this);
-  }
+  },
 
+  _getColumnsByType: function (columnType) {
```
viddo: Only logic shared for all form models should go in the base model, this would be better extracted to a separate object (just like [source-options-model](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-source-options-model.js))

Also, which node/query-schema to get columns from and what filtering is needed will vary, so should probably stay in the form context however, something on the lines of:

```js
// in some foo-form-model
var nodeDefModel = this._layerDefinitionModel.findAnalysisDefinitionNodeModel(this.get('source')); 
var columnOptions = new ColumnOptions(nodeDefModel);
columnOptions.filterByType('type') // => […]
columnOptions.all() // => […]
```

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/weighted-centroid-form-model.js
```diff
@@ -0,0 +1,73 @@
...
+      source: {
+        type: 'Select',
+        text: _t('editor.layers.analysis-form.source'),
+        options: [ this.get('source') ],
+        editorAttrs: { disabled: true }
+      },
+      weight_column: {
+        type: 'Select',
+        title: _t('editor.layers.analysis-form.weight-column'),
+        options: this._getColumnsByType('number')
```
viddo: Related to https://github.com/CartoDB/cartodb/pull/7872/files#r67297815, an issue I'm seeing is that's assumed that the columns are always available which is not true. Just like I think the object should be extracted the form model should handle intermediate states of the query-schema-model and indicate them in the schema.

For an example see how the [source-options are handled in point-in-polygon](https://github.com/CartoDB/cartodb/blob/18130b2fc3f39e636aec3fa13da6f404d677170f/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/point-in-polygon-form-model.js#L13)

viddo:
Overall 👍  commented on the details I think should be fixed

javierarce:
Thanks, a ton!

PR #7874, opened 2016-06-15
Downsample raster overviews

rafatower:
Downsample original raster before proceeding with the rest of the
process of overview generation. Unfortunately this cannot be done along
with the reprojection as gdalwarp command does not support the `-scale`
parameter of gdal_translate, and conversely gdal_translate does not
support reprojection.

What I mean by "downsampling" is scaling values from the original raster (Byte, UInt16, Int16, UInt32, Int32, Float32, Float64, CInt16, CInt32, CFloat32 and CFloat64) to fit into a Byte (which is what we support in the tiler).

This can be done better but for the moment I think it fits the bill.

@jgoizueta can you please review?

cc/ @javisantana


jgoizueta:
:+1:

PR #7878, opened 2016-06-16
7686 drag drop node as new layer

### LINKED ISSUE #7686: be able to use a layer as a source for other layer
This should work like:

- having layer (A)
- drag the node to the layers panel

A new layer (B) should be created with A as source.

To check everything is working a widget added to A0 should filter B as well

viddo:
Resolves #7686 

This allows to create analysis by dragging an analysis "out" from a layer to create a new one.

To understand the scope and why behind changes/fixes it's probably easiest to read through the commit messages in chronological order.

There appear to be more changes than there really are, it's because the [jquery-ui](npmjs.com/jquery-ui) package on NPM is actually [not endorsed officially](https://forum.jquery.com/topic/publishing-jquery-ui-version-1-11-2-to-npm), and even outdated (0.10.5). There were a [bugfix released in v0.11.4](https://bugs.jqueryui.com/ticket/6889) that I needed, so installed the official way (downloading package), put it in lib/assets/node_modules so the affected code changes were minimal anyway.

@alonsogarciapablo since you know the layers implementation you're probably the best candidate to review this. 🙇 

There's still some pending enhancements (nice-to-have's and semi-related bugs) I've encountered while working on this, but I'll create separate issues for those.

edit: code review issues:

:+1: Great job @viddo! 👏  👏  👏 

Reviewed each commit and everything looks good. Some parts are harder to understand than others, but  I understand the problem is hard. I left some ideas that can be addressed later.

---

edit: code review comments TBD:

- [x] ~~_Once you start dragging an analysis, you can't leave it where it was/cancel_~~ no easy solution, the user can move the dragged item to somewhere until the placeholder disappear to abort (e.g. the list header) 
- [x] _After creating a new layer from an analysis, if you delete de layer that had the analysis
  in the first place, the new layer is also deleted and the map won't have any layer_
  - not trivial to handle, created separte issue: #7926
- [x] _When dragging an analysis, I see the name of the letter and title is clipped._  
  - CSS issue, set a min-width on the dragged template.
- [x] _I also found that clicking on a layer title to drag & drop sometimes triggers the layer edition. But IMO I'd expect to be able to drag the layer by clicking anywhere (except on an analysis). Perhaps we can find a way to make this work?_
- [x] ~~_Clicking on an analysis + dragging the layer edits the analysis on mouse up._ ~~ not reproducible anymore


lib/assets/javascripts/cartodb3/editor/layers/layer-analyses-view.js
```diff
@@ -60,6 +60,10 @@ module.exports = CoreView.extend({
 
   _attachDraggableToNodeView: function (nodeView, nodeDefModel) {
     if (!this.model.isOwnerOfAnalysisNode(nodeDefModel)) return;
+
+    var source = nodeDefModel.getPrimarySource();
+    if (source && !this.model.isOwnerOfAnalysisNode(source) && this.model.get('source') === nodeDefModel.id) return;
```
alonsogarciapablo: Perhaps we could move this logic to a method in `layerDefinitionModel`? `this.model.isSomething(nodeDefModel)?`

viddo: 👍 
btw, comment on the PR instead of individual commits, so comments gets folded when fixed.

lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
```diff
@@ -154,48 +150,124 @@ module.exports = Backbone.Collection.extend({
...
-        options.error && options.error();
       }
     });
   },
 
+  /**
+   * A layer for an existing node have different side-effects depending on the context in which the node exists.
+   * E.g. let's assume a nodeId X is wanted to be created from a layer A:
+   */
+  createLayerForAnalysisNode: function (nodeId, newLayerOrder) {
```
alonsogarciapablo: This method is pretty hard to understand (for me at least) as is. I know the problem is not easy so I understand it's hard to make it understandable, but It'd be nice if we could somehow clarify a bit. Perhaps by extracting some smaller methods?

alonsogarciapablo: Another idea... what about moving `createLayerForAnalysisNode` and `createLayerForTable` to a separate class that only takes care of creating layerDefinitionModels?

alonsogarciapablo:
:+1:

Reviewed each commit and everything looks good. Some parts are harder to understand than others, but  I understand the problem is hard. I left some ideas that can be addressed later.

Also, some things I noticed while playing with it (can be addressed later too):

1. Once you start dragging an analysis, you can't leave it where it was/cancel:


2. After creating a new layer from an analysis, if you delete de layer that had the analysis
  in the first place, the new layer is also deleted and the map won't have any layer:


3. I also found that clicking on a layer title to drag & drop sometimes triggers the layer edition. But IMO I'd expect to be able to drag the layer by clicking anywhere (except on an analysis). Perhaps we can find a way to make this work? 

4. Clicking on an analysis + dragging the layer edits the analysis on mouse up.

5. When dragging an analysis, I see the name of the letter and title is clipped. Why is layer-analysis-draggable.tpl so small? (should have the same width  as rest of layers)

lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
viddo: Yes, it's true, I'll give it a try and see if I can clarify the different paths somehow.

viddo:
1. From what I can remember couldn't find a way to abort… I'll check again.
2. That's a "feature"; all depending analysis/layers are currently deleted when one of its source(s) is deleted. The solution would be to move the dependeant analysis to another layer, if possible… I think this needs more consideration and should be tackled separately.
3. True, I disabled it in this PR because I had trouble with it; I'll check if I can solve it
4. Have not been able to reproduce it, or rather only if _not_ dragging, but I'd say that's expected behavior. What are the STR exactly?
5. It shouldn't be; it started to appear after updating with master yesterday/today, probably some style change after updating with master. I'll review

lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
viddo: I'll look into this as part of #7881

PR #7785, opened 2016-06-13
Column graphics at add widget modal

fdansv:

PR #7879, opened 2016-06-16
Data stats

### LINKED ISSUE #7815: Integrate stats in data panel
After #7759, we need to draw the graphs in the data main view.

nobuti:
This PR closes #7815.
It depends on https://github.com/CartoDB/cartodb/pull/7785, probably it needs to be merged first. I spotted some issues with the specs.

stats

We are generating stats for every type for all non blacklisted columns. Because the stats not always can be drawn, we can't know how many graphs we are going to have at the beginning, so the 'show more' button has to wait until then.

I'm working on:
- [x] Add/remove widget functionality on check
- [x] Check if the stat already present in the widgets collection to render it as selected

cc @xavijam


lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-show-more.tpl
```diff
@@ -0,0 +1,3 @@
+<% if (isVisible) { %>
+  <a href="#" class="CDB-Text CDB-Size-small u-upperCase js-button"><%-number %> columns more</a>
```
xavijam: I'd use a `<button>` instead of a `<a>`.

xavijam:
Go go go go! :)

Issue #7707, opened 2016-06-07
NoMethodError: Error in sync cartodbfy


rafatower:
See R/B 17319:
```
NoMethodError: undefined method `save' for nil:NilClass

1 File "app/models/table/privacy_manager.rb", line 15 in apply_privacy_change
2 File "app/models/table.rb", line 453 in after_save
3 File "app/models/table/user_table.rb", line 202 in after_save
+ 15 non-project frames 
19 File "app/models/synchronization/adapter.rb", line 86 in fix_oid
20 File "app/models/synchronization/adapter.rb", line 65 in overwrite
21 File "app/models/synchronization/adapter.rb", line 36 in run
22 File "app/models/synchronization/member.rb", line 220 in run
23 File "lib/resque/synchronization_jobs.rb", line 13 in block in perform
24 File "lib/resque/base_job.rb", line 21 in call
25 File "lib/resque/base_job.rb", line 21 in run_action
26 File "lib/resque/synchronization_jobs.rb", line 10 in perform
+ 30 non-project frames 
```

rafatower:
@gfiorav taking into account the purpose of the original exception management code 750326cb

I don't know if we really should be concerned of a table without a map at that point. I don't think so.

gfiorav:
Please do remove it if you find it adequate 😄

PR #7885, opened 2016-06-16
Do not raise exception when there's no map #7707

rafatower:
Go back to previous behavior but refining it:
- capture specific exception
- check for @table.map and exception message

then log the exception for debugging purposes (in case the sync has no
map for any reason) but do not stop the whole process just because of
it.

@gfiorav and @iriberri can you please review? there's a bit more historic info in the issue and previous commits.

### DIFF
--- a/app/models/table/privacy_manager.rb
+++ b/app/models/table/privacy_manager.rb
@@ -11,40 +11,37 @@ module CartoDB
     end
 
     def apply_privacy_change(metadata_table, old_privacy, privacy_changed = false)
-      begin
+      @table.map.save
+
+      # Separate on purpose as if fails here no need to revert visualizations' privacy
+      revertable_privacy_change(metadata_table, old_privacy) do
+        # Map saving actually doesn't changes privacy, but to keep on same reverting logic
         @table.map.save
+        set_from_table_privacy(@table.privacy)
+      end
 
-        # Separate on purpose as if fails here no need to revert visualizations' privacy
-        revertable_privacy_change(metadata_table, old_privacy) do
-          # Map saving actually doesn't changes privacy, but to keep on same reverting logic
-          @table.map.save
-          set_from_table_privacy(@table.privacy)
-        end
+      revertable_privacy_change(metadata_table, old_privacy,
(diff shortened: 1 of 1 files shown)


app/models/table/privacy_manager.rb
```diff
@@ -28,14 +28,16 @@ def apply_privacy_change(metadata_table, old_privacy, privacy_changed = false)
...
-        CartoDB.notify_debug("#{exception.message} #{exception.backtrace}", {
-          table_id: @table.id,
-          table_name: @table.name,
-          user_id: @table.user_id,
-          data_import_id: @table.data_import_id
-          })
-
-        raise exception
+        if @table.map.nil? && exception.message =~ /undefined method `save' for nil:NilClass/
+          CartoDB.notify_debug("#{exception.message} #{exception.backtrace}", {
```
gfiorav: `CartoDB.notify_debug` is deprecated.

You can use:
```ruby
CartoDB::Logger.debug(message: 'Whatever you like to appear in Rollbar',
                                       exception: exception,
                                       table_id: @table.id,
                                       table_name: @table.name,
                                       user_id: @table.user_id,
                                       data_import_id: @table.data_import_id)
```

Also, like Hound is pointing out, no need for `{` `}` !

gfiorav:
I made a comment regarding the use of deprecated `CartoDB.notify_debug`. Once that is changed it's 👍 on my side.

app/models/table/privacy_manager.rb
```diff
@@ -155,7 +154,7 @@ def revert_to_previous_privacy(metadata_table, old_privacy, additional_revert_ta
         end
       end
 
-      if errors.length > 0
+      if !errors.emtpy?
```
iriberri: typo, `empty`  instead :)

rafatower:
@gfiorav thanks for the comments. We shall eradicate the old logger.

app/models/table/privacy_manager.rb
gfiorav: also, you can use 
```ruby
unless errors.empty?
  (...)
end
```

if there is no `else`.

rafatower: thanks!

PR #7893, opened 2016-06-16
Adds moran analysis

### LINKED ISSUE #6948: Add moran analysis
- [ ] Add necessary analysis for models.
- [ ] Create proper analyses form templates.
- [ ] Check how to provide list of available nodes for `moran` analysis.
- [ ] add to add-analysis modal; 
  - need to be able to know the source's schema, and validate & pick two valid columns, if not valid should not display it as an option

javierarce:
Closes #6948

screen shot 2016-06-17 at 11 01 01
screen shot 2016-06-17 at 11 00 45


lib/assets/javascripts/cartodb3/components/modals/add-analysis/analysis-options.js
```diff
@@ -51,6 +52,13 @@ module.exports = function (alsoGenerateFromCamshaftReference) {
           title: _t('components.modals.add-analysis.option-types.point-in-polygon.title'),
           desc: _t('components.modals.add-analysis.option-types.point-in-polygon.desc')
         }, {
+          Model: MoranAnalysisOptionModel,
```
viddo: Since just using the default analysis-option-model you don't have to set any model at all, you could just omit this prop and the collection would use the default model instead (same applies to k-means+weighted-centeroid)

javierarce: Thanks, @viddo, I'll change it.

javierarce:
CR: @viddo 
CC: @rochoa

viddo:
👌

javierarce:
@rochoa, @saleiva, @urbanphes: the designs only show to parameters for this analysis (`key column` & `confidence`)… do you think is a good idea to hide the extra ones and use some defaults or makes more sense to allow users to customize it fully?

ohasselblad:
1. Suggestions to change `confidence` to `cluster significance` (if it fits!), and just `significance` (with an informative hover over or little tip on what this means)
2. Instead of a slider of continuous values for `significance`, have four options:
    a. **Very high: 0.1%**
    b. **High: 1%**
    c. **Medium: 5%**
    d. **Low: 10%**
3. There may have been some miscommunication, but this algorithm works on points and polygons, so there is no need for labeling it as polygons

javisantana:
nice explanation @ohasselblad 

question about the output, what is interesting for the user to see? I more or less understand it but it going to be really hard to regular users

ohasselblad:
Valuable outputs are hotspots, coldspots, or outliers. These are found by filter by quads. I wrote some helper functions to make it easier to get these outputs: https://github.com/CartoDB/crankshaft/blob/01fc2c1dd1087e58679cc2728ee88a0df5702ea6/src/pg/sql/10_moran.sql#L73-L83

javisantana:
cc @javierarce (i renamed the analysis to be more precise)

javierarce:
👌

PR #7894, opened 2016-06-16
Fix reset errors.

### LINKED ISSUE #7869: If a previous query failed, error info is showed, perfect, but it is not cleaned when a new query is done and it was ok

nobuti:
This PR fixes #7869.

cc @xavijam


xavijam:
👍

PR #7901, opened 2016-06-17
Filter category analysis; Split by comma instead of ||

### LINKED ISSUE #7705: [autogeneration] reject filters not working
1) aggregate a filter analysis
2) select a column (with at least two unique values) and reject one of the values

Nothing is renderered but it should render at the non rejected properties.

Not sure if this a UI problem or a backend one

viddo:
Fixes #7705

> The client should not send an empty array for accept.

cc @rochoa


viddo:

PR #7903, opened 2016-06-17
Fix scroll when modal content is larger than viewport

### LINKED ISSUE #7630: Scroll doesn't work on small viewports
E.g. add-widgets modal, probably missing some `Dialog--*` class/modifier to have overflow: scroll prop…

viddo:
Fixes #7630 

Compensates for the fixed footer size

@xavijam ok?


xavijam:
:+1: it will make @ramiroaznar and @AbelVM happier.

viddo:

PR #7919, opened 2016-06-17
GTM: Better sync run decisions 

gfiorav:
Please CR @javitonino

### DIFF
--- a/lib/carto/ghost_tables_manager.rb
+++ b/lib/carto/ghost_tables_manager.rb
@@ -19,7 +19,7 @@ module Carto
     def link_ghost_tables
       return if user_tables_synced_with_db?
 
-      if safe_sync?
+      if should_run_synchronously?
         link_ghost_tables_synchronously
       else
         ::Resque.enqueue(::Resque::UserDBJobs::UserDBMaintenance::LinkGhostTables, @user_id)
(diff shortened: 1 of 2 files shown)


lib/carto/ghost_tables_manager.rb
```diff
@@ -41,11 +42,15 @@ def user_tables_synced_with_db?
...
+    # It's nice to run sync if any unsafe stale (dropped or renamed) tables will be shown to the user but we can't block
+    # the workers for more that 180 seconds
+    def safe_sync?
       cartodbfied_tables = fetch_cartodbfied_tables
 
-      find_dropped_tables(cartodbfied_tables).empty? && find_stale_tables(cartodbfied_tables).empty?
+      dropped_and_stale_table_count = find_dropped_tables(cartodbfied_tables).count +
+                                      find_stale_tables(cartodbfied_tables).count
+
+      dropped_and_stale_table_count < MAX_TABLES_FOR_SYNC_RUN
```
javitonino: This always run synchronously if there are 0 stale/dropped tables, doesn't it? Even if there are lots of new tables.

I think the check we want is to only run sync if "there are any stale tables" and "there are less than X tables to operate on (including linking new tables)"

gfiorav: yeah, mindless overlook, give me a sec

javitonino:
LGTM
:package: :it: after tests

gfiorav:
Thanks for the review and deploying this!

Hope it's smooth !

PR #7925, opened 2016-06-17
Fixed problem with pagination after removing a row

xavijam:


xavijam:
Merging, frontend assets passed

PR #7935, opened 2016-06-17
Do not edit users that already have a quota set

iriberri:
If there have been manual editions to the users affected by the Rake, they would be overwritten. By checking if the quota is 0 we avoid overwriting the quota to them, and otherwise we edit them.


iriberri:
Please @rafatower could you review?

rafatower:
:+1:

PR #7937, opened 2016-06-20
Infowindow html editor

### LINKED ISSUE #7851: html editor for infowindows is broken
1. create a visualization with a layer
1. go to infowindow and enable light style, select some fields
1. go to html editor, it does not show up, this is shown instead

screen shot 2016-06-14 at 7 15 20 pm

nobuti:
This PR fixes #7851.

Todo:
- [x] ~~Undo redo history~~
- [x] Tooltip customs templates
- [x] Render custom templates in the map

cc @xavijam


nobuti:
@matallo @alonsogarciapablo I need some help here in order to save and load the custom templates in the map. Basically I have this code https://github.com/CartoDB/cartodb/blob/c2c5d5bc03c2046e7b3ffc0a2948afc2667aa409/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindow-view.js#L47-L73, but I need some info about how custom templates are managed in cartodb.js.
Also, when a custom templates is applies, which carousel item should be selected? I guess "none", right?

matallo:
I suppose in that case the "html" sould be checked and show the custom template, then clicking on the toggle would show something similar to the current editor where you'd lose your changes

I'll check the integration tomorrow, if you don't mind :)

nobuti:
infowindow_custom

Ready for review.

cc @xavijam @matallo 

I've found some issues:
- It's not possible to add inline style to pseudo elements. Because of that, the user won't be able to style the tip's color, not sure if we should change that @piensaenpixel 
- The custom html includes _functionality classes_ like `js-whatever`. Should we keep them?

xavijam:
1. That's a good one, @piensaenpixel, would you mind to open a ticket for trying to solve that in the future?

xavijam:
2. Yes, we need them, if not the infowindow will not work. That reminds me that we need to add some needed info for the users, like: "if you don't add that class, infowindow will not work..."

xavijam:
@nobuti, could we add the button as the design below?. Clicking over it should reset the custom html and apply the simple one?

https://cloud.githubusercontent.com/assets/14546701/16274733/2a096360-38a7-11e6-8070-c6fff60fd1bb.png

And when clicking,

nobuti:
@xavijam what a cliffhanger...

xavijam:
what? :joy:

nobuti:
> And when clicking,

The intrigue is killing me!

xavijam:
That has been edited!!! check the comment now :joy:

lib/assets/javascripts/cartodb3/components/infobox/infobox-button.tpl
```diff
@@ -3,5 +3,7 @@
     <span class="CDB-Button-Text CDB-Text is-semibold CDB-Size-small"><%= label %></span>
   </button>
 <% } else { %>
-  <p class="CDB-Text is-semibold CDB-Size-small"><a href="#" class="js-action"><%= label %></a></p>
+  <button class="CDB-Button CDB-Button--small js-action" style="padding-left: 0; padding-right: 0;">
```
matallo: ugh, inline styles?

nobuti: I know, I know, I had pending one comment. I think these both should be buttons:

screen shot 2016-06-23 at 09 47 45

AFAIK, there is no button or helper to eliminate the horizontal padding, in order to stick the text on the left side. Also, being both buttons, the vertical alignment between them is easy peasy.

matallo:
🙌  I'd wait to merge until infowindows migrations are, though (some templates are renamed) https://github.com/CartoDB/cartodb/pull/8085

/cc @juanignaciosl

matallo:
```Uncaught Error: Content property for codemirror is required.```

**str**

- create new map
- add point and infowindow in old editor
- go to new builder and popup panel

**expected behaviour**

- infowindow light to be selected

**actual behaviour**

- none is selected
- when trying to go to html editor the previous error arises

nobuti:
@matallo The problem is in that case `template_name` is `table/views/infowindow_light`, I'm going to make a fix but not sure if the issue belongs to this. AFAIK the template name should be ìnfowindow_light`, right?

matallo:
after https://github.com/CartoDB/cartodb/pull/8085 migrations, looks 👌 (tested locally)

nobuti:
Last minute CR?

matallo:
BOOM

PR #7940, opened 2016-06-20
Fix the dragged-layer size issues

viddo:
drag

1. Make sure the handle is always visible by enforcing a fixed left-offset
2. Render a fixed min-width so an element which is absolute positioned maintains the size of an item in the list

@piensaenpixel can you review?
cc @alonsogarciapablo


piensaenpixel:
👍

viddo:
Frontend tests pass, so merging

alonsogarciapablo:
looking good!

PR #7953, opened 2016-06-20
Escape + and / symbols in Surrogate Keys on Varnish HTTP and Telnet invalidation - fixes #7868

### LINKED ISSUE #7868: Varnish invalidations triggers should escape special regex symbols
Steps to reproduce:
- have a table that ends with a surrogate key containing a `+` with an environmnet configured with 
- invalidate that table

The `+` on the [Varnish HTTP trigger](https://github.com/CartoDB/cartodb/blob/master/app/models/user/db_service.rb#L1311) is sent unescaped; causing the regex to incorrectly match.

We should make the script replace at least `+` by `\+` and any other base64 characters that match special regex chars.

zenitraM:
Fixes #7868.

I'm not entirely sure on how to make a spec for this, but I've tested manually (by crafting by brute force tables whose Surrogate-Key ends up containing + and /) and it works fine.

could you review @rochoa @luisico?   🙏😄

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,10 +5,14 @@
 This release introduces a new Resque queue: `user_dbs`. It is needed for operation on user databases, i.e: linking
 ghost tables, importing common data and automatic index creation.
 
+There is also a fix on Varnish invalidation triggers in user databases to fix some edge cases where tables would not be properly invalidated on Varnish (see https://github.com/CartoDB/cartodb/issues/7868).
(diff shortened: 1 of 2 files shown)


xavijam:
Could we close this @luisbosque '

zenitraM:
or, you know, merge it :D

zenitraM:
@luisico @rochoa Closing this as I'm cleaning up open things on my GH, though this still might be a bug! 😅

rochoa:
Who are you? 😂

xavijam:
💃

PR #7954, opened 2016-06-20
Removed min, max and avg operations from aggregation selector

### LINKED ISSUE #7858: Category widgets' aggregation functions
The builder is offering a lot of different aggregation functions for category widgets, but only SUM and COUNT are allowed

MariaCheca:
Fixes #7858

CR @xavijam


xavijam:
:+1: thanks lady

PR #7956, opened 2016-06-20
Add new /dataset route for dataset view

javitonino:
Re #7950 

Works with canonical visualization uuid, or table name.

CR @gfiorav 
cc/ @xavijam

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -47,14 +47,16 @@ CartoDB::Application.routes.draw do
 
   # Editor v3
   scope module: 'carto', path: '(/user/:user_domain)(/u/:user_domain)' do
-    namespace :builder do
+    namespace :builder, path: '/' do
       # Visualizations
-      resources :visualizations, only: :show, path: '/', constraints: { id: /[0-z\.\-]+/ } do
+      resources :visualizations, only: :show, path: '/builder', constraints: { id: /[0-z\.\-]+/ } do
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -47,14 +47,16 @@
 
   # Editor v3
   scope module: 'carto', path: '(/user/:user_domain)(/u/:user_domain)' do
-    namespace :builder do
+    namespace :builder, path: '/' do
```
gfiorav: I find it kind of confusing that `builder` looses the `/builder` path :/

I'd even prefer to have a separate `resources :builder` for this if there's no elegant workaround.

javitonino: The alternative is:
```
    namespace :builder, path: '/' do
      # Visualizations
      resources :visualizations, only: :show, path: '/builder', constraints: { id: /[0-z\.\-]+/ } do
        namespace :public, path: '/' do
          match 'embed', to: 'embeds#show', via: :get
          match 'embed_protected', to: 'embeds#show_protected', via: :post
        end
      end

      resources :datasets, path: '/dataset', only: :show, constraints: { id: /[0-z\.\-]+/ }
    end
```
After an offline talk, it seems like the original is more compact and reflects better that `/dataset` is related to `/builder`, so leaving it as it is.

gfiorav:
📦 it!

PR #7958, opened 2016-06-20
Fix redirect for without subpath

javitonino:
So /editor returns 404 instead of 500.

CR @gfiorav

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -58,7 +58,7 @@ CartoDB::Application.routes.draw do
     end
 
     namespace :editor do
-      match '(*path)', to: redirect { |params, request| CartoDB.base_url_from_request(request) + '/builder/' + params[:path] }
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -58,7 +58,9 @@
     end
 
     namespace :editor do
-      match '(*path)', to: redirect { |params, request| CartoDB.base_url_from_request(request) + '/builder/' + params[:path] }
+      match '(*path)', to: redirect { |params, request|
+        CartoDB.base_url_from_request(request) + '/builder/' + params[:path].to_s
```
gfiorav: paranoid but can't `params[:path]` be `nil` in some circumstance?

javitonino: That's exactly what happened before, and what `to_s` solves. `nil.to_s => ''` :)

gfiorav: 😮 

🌴

gfiorav:
📦 it!

Issue #7875, opened 2016-06-16
The visualization is not updated if you save the SQL custom when the map is new


nobuti:
Steps to reproduce:

- Create a new map
- Go to the SQL editor and change the default query to a custom one
- CMD + S (nothing happens but all it's saved).
- F5
- Repeat steps 2 and 3 and see that the map is updated to show the result properly

As @viddo says, the problem is most likely that when reloaded the analysis is given to `cdb` through the vizjson, so the node and query exists, while when “fresh” `cdb` discards the node/query changes. If you compare the `map?stat` requests there’s no analyses on a fresh map, until a non-source node is created
but once reloaded it includes the analysis.

xavijam:
In my opinion this is an important bug, because we can't apply new queries, and also, after a deep check, we are making more requests than needed.

xavijam:
cc @CartoDB/frontend

alonsogarciapablo:
Ok, here's what's happening... When the map is created, the map is created in CartoDB.js through [`deepInsights.createDashboard`](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/editor.js#L72), using the data that is present in `window.vizJSON` (right after the redirection). The vizjson contains the following information about the "data" layer:

```
{
  "id": "9f5359e7-e4b6-46b6-8eef-801bac7bfc70",
  "type": "CartoDB",
  "infowindow": { ... },
  "tooltip": { ... },
  "legend": null,
  "order": 1,
  "visible": true,
  "options": {
    "layer_name": "paradas_metro_madrid_1",
    "cartocss": "#paradas_metro_madrid_1 { ... }",
    "cartocss_version": "2.1.1",
    "interactivity": "cartodb_id",
    "sql": "select * from paradas_metro_madrid_1"
  }
}
```

As you can see, there's no `source` option for this layer, so CartoDB.js doesn't know that it's connected to the analysis node whose SQL we're changing. When the map is reloaded, the vizJSON does include the `source` option for this layer, and everything starts working fine...

alonsogarciapablo:
[...]

PR #7961, opened 2016-06-20
Ensure `source` attribute is updated when layerDefinitionModel changes and layer in CartoDB.js didn't have a `source`

alonsogarciapablo:
Fixes: https://github.com/CartoDB/cartodb/issues/7875.

This was the quickest way to fix ☝️ . It basically ensures that, when a layerDefinitionModel is saved, the `source` attribute is updated when it's present in an editor's model and not in the corresponding layer in CartoDB.js. 

It fixes the bug but only because the `sql_history` attribute is changed when the SQL of a layer is updated. This triggers the udpate of the CartoDB.js model, previously ensuring that the `source` is set. Bug fix is executed as a side-effect of something else, which makes it a bit fragile...

@viddo PTAL! Thanks!

cc: @nobuti


nobuti:
👍

viddo:
👍

PR #7964, opened 2016-06-20
Fixes column selection in the input-number-value form

### LINKED ISSUE #7962: Input value number should allow to select the column if is not set

javierarce:
This PR fixes a problem with the input number value form of the fill component, that prevented choosing the column when opening the pane.

Closes #7962


javierarce:
CR:  @javisantana

javisantana:
:es:

PR #7967, opened 2016-06-20
min-height: 0 to everything.

### LINKED ISSUE #7946: New editor data tab not scrolling in Firefox
The data tab doesn't display scroll in Firefox (works fine on chromium), instead, it overflows and shows both horizontal and vertical scrollbars:

screenshot_20160620_120307

No errors on JS console.

cc @xavijam

nobuti:
This PR fixes #7946.

Whenever you've got an element with overflow: [hidden|scroll|auto] inside of a flex item, you need to give its ancestor flex item min-width:0 (in a horizontal flex container) or min-height:0 (in a vertical flex container), to disable this min-sizing behavior, or else the flex item will refuse to shrink smaller than the child's min-content size. 

See https://bugzilla.mozilla.org/show_bug.cgi?id=1043520

cc @piensaenpixel @xavijam @javierarce @javitonino

### DIFF
--- a/app/assets/stylesheets/editor-3/editor.css.scss
+++ b/app/assets/stylesheets/editor-3/editor.css.scss
@@ -20,18 +20,20 @@
   @include display-flex();
   position: relative;
   height: 100%;
+  min-height: 450px;
 }
 .Editor-panel {
   @include transition(width, 300ms, ease-in-out);
   @include display-flex();
   @include flex-direction(column);
+  @include flex(1);
   position: absolute;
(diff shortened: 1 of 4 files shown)


xavijam:
I hope it does not hurt IE (:joy:).

nobuti:
Tested in IE11, I've spotted some problems, but not related with the scroll.

ie11

xavijam:
Yay! ok, thanks man! 🚢 :it:

PR #7989, opened 2016-06-21
updated camshaft

javisantana:
cc @viddo


viddo:
you'd need to run `npm shrinkwrap` after `npm update camshaft-reference`, to actually use the new version for deploys

javisantana:
@viddo thanks, updated, makes sense?

viddo:
Looks legit 👍

Issue #7988, opened 2016-06-21
Remove trailing / leading white space on admin name values 


rafatower:
@clhenrick commented on [Sun Feb 21 2016](https://github.com/CartoDB/data-services/issues/212)

Thought I'd make a quick note of this here: 

Just attempted an import of this dataset and `Mexico ` was the only name that didn't geocode as it had a trailing white space. Seems like it should be an easy fix to remove trailing (as well as leading) white space before string matching with the geocoded data? 

```
name,region
Canada,North America
Mexico ,North America
United States,North America
Greenland,North America
```

PR #7991, opened 2016-06-21
Trim when joining results of geocoding #7988

rafatower:
Turns out we were trimming strings before sending to the geocoder and
the geocoder was getting results for them, but when updating the
original table there was an inconsistency causing some results not to be
added.

@iriberri can you please review?


iriberri:
LGTM! Perhaps it would be also interesting to trim characters such as line feeds, because `trim` without any extra parameter just gets rid of spaces, but looks good enough to me. Otherwise instead of a trim perhaps we would have to use a regexp directly.

rafatower:
I'll leave that for another time. This was meant to be a quick fix that took me longer, investigation of root cause included.

iriberri:
:+1: cool!

PR #7997, opened 2016-06-21
Replace tape with npm-based jasmine

viddo:
Replaces the alternative test suite that was added as part of #7691 with [jasmine-npm](https://github.com/jasmine/jasmine-npm)

So still be able to run tests in CLI w/o all the browserify/source-map overhead, but using the same test API everyone is used to.

For now only migrated the single test there were, but technically we should be able to move all non-view/DOM-related tests to this runner instead and avoid the browserify-build/source-map issues in the test env.

Test run output probably looks familiar:
```bash
$ npm test

> cartodb-ui@3.27.1 test /Users/viddo/src/cartodb
> jasmine

Started
.............................................................


61 specs, 0 failures
Finished in 0.042 seconds
```

@xavijam 
cc @javierarce @nobuti @alonsogarciapablo


xavijam:
👍

PR #8012, opened 2016-06-21
Fix an error in rescue from visualization renaming

javitonino:
Found this by chance, the rescue is wrong here. Fixed this and took the change to update to the new logger. CR @gfiorav


gfiorav:
📦 it!

PR #8019, opened 2016-06-21
fixes rectangle aggregation

### LINKED ISSUE #7993: Aggregated value column is not generated when aggregating by squares, hexabins or admin region
Create a map with a point dataset
Aggregate the points with any option
Change COUNT to SUM(value)

The resulting column doesn't appear on the column dropdown

javisantana:
fixes #7993


javisantana:
cc @xavijam

xavijam:
:+1:

xavijam:
ready to be merged right?

PR #8020, opened 2016-06-21
Fix named map tests in parallel executor

javitonino:


javitonino:
CR @gfiorav

gfiorav:
thanks! 🌴 

📦 it!

PR #8021, opened 2016-06-21
Migrates old color ramps and old quantification methods 

### LINKED ISSUE #7902: Choropleth styles migration is not working anymore
## STR
- Create a map of points.
- Go to the editor and apply the choropleth wizard.
- Go to the new builder, and check style panel.
- Screenshot below:

screen shot 2016-06-17 at 10 19 38

The problem basically is that we set the range with the `wizards-properties` value, that only represents the color, we should "convert" it in the fill component @javierarce.
screen shot 2016-06-17 at 10 13 32

javierarce:
Closes #7902

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/form-components/editors/fill.js
+++ b/lib/assets/javascripts/cartodb3/components/form-components/editors/fill.js
@@ -15,6 +15,13 @@ var INPUT_TYPE_MAP = {
   'color': InputColor
 };
 
+var QUANTIFICATION_REF = {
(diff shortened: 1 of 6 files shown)


javierarce:
CR: @xavijam and/or/xor @javisantana

xavijam:
Looks good man! yay! thanks.

xavijam:
Ready then?

javierarce:
Yes, I just thought I could simplify it a bit before merging.

Issue #7854, opened 2016-06-15
change order of import guessing 


oriolbx:
Context:
When you import a file with city column and a country column to CartoDB, it looks like that the order for the import guessing is first guessing by IP address, then by country and last by city name.

When CartoDB founds a column to georeference during the import guessing process, it stops searching for more columns to guess and it georeference the data with the column that finds first.

In the case of a table with city and country columns, it will georeference the data with country names but it won't use the city column to georeference the data, because the order of the import guessing first is to check by country name and then by city.  

Steps:

1- If you have a dataset with two columns, a column of cities and a column of countries like this table:

![screen shot 2016-06-15 at 10 28 23](https://cloud.githubusercontent.com/assets/11177693/16072956/ed5b5b1c-32e3-11e6-9d20-011f3f3272ef.png)

2 -When you import the dataset to CartoDB with the guessing option enabled, CartoDB georeference only the polygons. 

![screen shot 2016-06-15 at 10 38 47](https://cloud.githubusercontent.com/assets/11177693/16073266/629f7ab0-32e5-11e6-9296-b1be77f6a44b.png)

The import log shows this:

```json
 Trying ip guessing... true: Trying country guessing... true: country_proportion(city) = 0.0 true: 
country_proportion(country) = 1.0 true: Found country column: country true: Creating the_geom column true: 
Geocoding countries... true: Creating the_geom column true: Geocoding finished
```

3 - If I delete the country column to the table and import it to CartoDB

![screen shot 2016-06-15 at 10 31 25](https://cloud.githubusercontent.com/assets/11177693/16073059/5db5c406-32e4-11e6-8c47-505ca63b30b2.png)

4- CartoDB georeference by city name:

![screen shot 2016-06-15 at 10 37 12](https://cloud.githubusercontent.com/assets/11177693/16073287/84d9eade-32e5-11e6-8471-54ef5eb0c067.png)

 The import logs show this:

```json
Trying ip guessing... true: Trying country guessing... true: country_proportion(city) = 0.0 true:
 Trying namedplaces guessing... true: country_proportion(city) = 0.0 true: 
Found namedplace column: {:column_name=>"city", :data_type=>"character varying"} true: Creating the_geom column true: Geocoding namedplaces... true: Creating the_geom column true: Geocoding finished 
```

rafatower:
[...]

PR #8026, opened 2016-06-21
Change order of content guessing #7854

rafatower:
As requested by support, change the order of namedplaces and country
content guessing during import.

@iriberri easy review :smile:


iriberri:
LGTM, and I also like Hound's suggestion :-) Gonna test it.

PR #8033, opened 2016-06-21
Css update jun2016

klibloomberg:
in this update
asc, desc buttons position
magnifying glass, 
loading image
fixed grey out text in table


Cartofante:

Cartofante:

Cartofante:

PR #8042, opened 2016-06-22
7881 explicit persist side effects

### LINKED ISSUE #7881: Only persist models/collections on explicit user-actions
There are still some occasions where models/collection are saved on implicit changes. I've experienced some race-conditions due to this while working on the drag-node-to-create-layer feature #7878. These are the places 

I've found these, so far, by having a quick check (there might be more):

- [ ] widget-definitions-collection; sync => layerDef.save
- [ ] layer-definitions-collection; sync => create/update analysis
- [ ] analysis-definitions; change => save analysis

This is probably related to what you noticed @xavijam, that there are some unnecessary requests being done on some occasions.

cc @javisantana @alonsogarciapablo

viddo:
Resolves #7881 

This replaces the persisting side-effects done through implicit events with a single place where to find user-interactions. Main benefits:
- clearer/explicit where and what side-effects there are
- easier to control destructive state mutations, error handling/recovery* is actually manageable
- can identify and separate code vs. user mutations (useful to avoid unnecessary POST/UPDATE calls)
- can remove dependencies and event listeners between concerned models and collections => easier test setup, less implementation details

I started out with analyses, and while I've tried to keep changes at minimal it turned out to be tricky since starting to pull the thread(s) it became clear that it's hard to do changes in isolation (due to dependencies between the different concepts), so decided to stop here to get initial feedback on the approach. I think I'll use this PR as temporary "master", and do sub-branches/PRs with the remaining changes:

- [x] analysis node deletions #8124
- [x] create layer (from table + analysis node) #8384
- [x] destroy layer #8474 
- [x] update layer + widgets => need to persist analysis #8522
- [x] smoke/regression testing

Out-of-scope:
- *error-handling/recovery is still unresolved, but that could be tackled separately once this refactoring is done.

@alonsogarciapablo can you review please?
cc @javisantana @xavijam @javierarce @nobuti FYI


lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -0,0 +1,94 @@
...
+      if (!analysisFormModel.isValid()) return;
+
+      var nodeDefModel = analysisDefinitionNodesCollection.get(analysisFormModel.id);
+
+      if (nodeDefModel) {
+        analysisFormModel.updateNodeDefinition(nodeDefModel);
+        var analysisDefinitionModel = analysisDefinitionsCollection.findAnalysisThatContainsNode(nodeDefModel);
+        analysisDefinitionModel.save();
+      } else {
+        analysisFormModel.createNodeDefinition(this); // delegates creation to the form
```
alonsogarciapablo: Why this indirection here? Makes this method harder to understand...

lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -0,0 +1,94 @@
+var camshaftReference = require('./camshaft-reference');
```
alonsogarciapablo: One of the things that might be a bit confusing about this is the fact that each method takes some arguments and they are not very "uniform" (eg: one method takes a `nodeAttrs` object while another takes an `analysisFormModel` object). Do you think we could somehow homogenise this? (for example: only pass layer, analysis or widget models + attr objects for updates + options).

lib/assets/javascripts/cartodb3/data/user-actions.js
viddo: I wanted to minimize the changes on the form models when I did the refactoring, and while the user-actions knows _what_ should be done it do not necessarily need to know _how_, different types have different use-cases.

In the case for creation it could be refactored in such a way that it doesn't need to delegate, but there are cases where the update is not uniform. The problem is more obvious in the case of widgets, e.g. [time-series](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/components/modals/add-widgets/time-series/time-series-option-model.js#L30-L39).

I'll see if I can make it clearer.

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
```diff
@@ -117,10 +119,9 @@ module.exports = CoreView.extend({
   },
 
   _saveSQL: function () {
-    var content = this._codemirrorModel.get('content');
-    this._nodeModel.set('query', content);
-    this._sqlModel.set('content', content);
-    this._layerDefinitionModel.save();
+    var query = this._codemirrorModel.get('content');
+    this._sqlModel.set('content', query);
+    this._userActions.changeSourceQuery(query, this._nodeModel, this._layerDefinitionModel);
```
alonsogarciapablo: Is `userActions#changeSourceQuery` defined? I can't find it in this diff.

lib/assets/javascripts/cartodb3/data/user-actions.js
alonsogarciapablo: Another random thought... Since all (most?) user actions will be related to a "resource" (layer, analysis or widget?), perhaps we could namespace this actions... Example:

```
this._userActions.layers.createFromDataset(dataset);
this._userActions.analyses.createOrUpdate(analysisAttrs);
this._userActions.widgets.createOrUpdate(widgetAttrs);
```

This is something we can always do in the future, but just leaving it here in case you think it can provide some value.

viddo: > Do you think we could somehow homogenise this? (for example: only pass layer, analysis or widget models + attr objects for updates + options).

> Do you think we could somehow homogenise this? (for example: only pass layer, analysis or widget models + attr objects for updates + options).

It's kind of related to [my other answer](https://github.com/CartoDB/cartodb/pull/8042/files/fc29b6b5579a468d416bee03eefb218016ad3d8a#r68194305) though, in some cases it's not as straight forward since side-effects are not always uniform. IMHO I prefer some differences to having massive if-else statements (that passing raw values around tend to cause). In some cases a model doesn't even exist yet. Not sure having homogenous method signature should be a goal in itself.

I can understand the sentiment, I'll investigate how to solve this.

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
viddo: Renamed the method but seems like it was missed here (should be `.updateAnalysisSourceQuery`), good catch!

alonsogarciapablo:
I left some comments/ideas. I think it'll make our lives easier overall, since all user actions with side-effects are in one place, so I'd vote for with moving forward with this 👍 . It would be great to know what others think about it too...

PR #8044, opened 2016-06-22
Show columns, not widgets.

nobuti:
This PR fixes the issue about duplicated columns in the data panel.

cc @javisantana


javisantana:
:es:

Issue #6889, opened 2016-03-15
Duplicating a multilayer map does not respect layer order


iriberri:
**STR**:

* Create a map with two layers, I used NY fire companies and NY fire divisions from Data Library
* The layers will have a certain order:
  ![image](https://cloud.githubusercontent.com/assets/1730320/13781613/36beee9e-eac4-11e5-8c99-ade81e89c412.png)
* Click on "duplicate map"
* The resulting map has the layers in the reverse order:
  ![image](https://cloud.githubusercontent.com/assets/1730320/13781641/4e1807b0-eac4-11e5-820d-855e9a15b4c3.png)

**Expected behavior**
The order of the layers in the map to be duplicated is respected in the output.

**Current behavior**
Layer 1 becomes layer 2 and viceversa. 

My data, in case it's useful:
user: `vmg`
map id: `ad1a9862-eabb-11e5-861e-0e31c9be1b51`
duplicated map id: `e30fa624-eabb-11e5-abf1-0ecfd53eb7d3`

xavijam:
I'll take a look, but it could be related with the backend code. In any case, it should be easy to fix :).

javitonino:
Most likely will be solved with changes in https://github.com/CartoDB/cartodb/pull/8045

xavijam:
ole!, working then?

javitonino:
I tested it and worked fine, so I'd say close it!

PR #8045, opened 2016-06-22
Copy visualizations using export/import

### LINKED ISSUE #7657: duplicate map of a map with analyses fails
**str**

- import a map with analyses (classic paradas de metro)
- open map
- duplicate map
- open duplicate map in builder

**expected behaviour**

- panels and map load

**actual behaviour**

- panels and map fail, find attached the developer tools trace

```
editor3.js:21147 Uncaught Error: layerAnalysisDefinitionNodeModel is required
common_editor3.js:62620 Uncaught TypeError: Cannot read property 'toJSON' of undefined
```

javitonino:
Fix #7657 #6889

Uses the export/import flow to duplicate maps, in order to support the latest features. Note that creating a map from a canonical visualization still uses `TableBlender` and `Map::Copier` which do not support analyses, etc. This should not be a problem as canonical visualizations can't have those.


javitonino:
Please CR @gfiorav

Makefile
```diff
@@ -46,7 +46,6 @@ WORKING_SPECS_1 = \
 	spec/models/map_spec.rb \
 	spec/models/map/copier_spec.rb \
 	spec/models/visualization/collection_spec.rb \
-	spec/models/visualization/copier_spec.rb \
```
gfiorav: There are no new specs for this?

javitonino: I rely on the import/export tests, which include exporting and reimporting maps, so that should be covered. I also kept the controller level tests.

lib/carto/named_maps/api.rb
```diff
@@ -1,7 +1,7 @@
 # encoding: utf-8
 
 require_relative '../http/client.rb'
-require_relative 'template.rb'
```
gfiorav: Thanks man

Makefile
gfiorav: 🌴

gfiorav:
📦 ready for me if you tested it in staging!

PR #8069, opened 2016-06-22
Table row options fixes

xavijam:
- Don't offer copy the value if Safari is present.
- Added Browser-detect helper class.
- Specs for table-paginator-view.
- Using new table-confirmation dialog.
- Fixed remove row confirmation styles button.


xavijam:

PR #8084, opened 2016-06-22
fix crash when codeEditor is undefined

mbektas:


mbektas:
@javisantana can you review and merge this one? issue started occurring once we removed Add layer button from sidebar.

PR #8089, opened 2016-06-23
Fix failing controller_view_test

xavijam:
Just in order to check what's going on with frontend tests.

cc @javitonino


viddo:
@xavijam seems like the fix did the trick

xavijam:
you are the best, let's go with it! merging.

PR #8094, opened 2016-06-23
added link by node column analysis

javisantana:
cc @javierarce


lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/column-options.js
```diff
@@ -19,6 +24,7 @@ module.exports = Backbone.Model.extend({
   },
 
   filterByType: function (columnType) {
+    if (!this._querySchemaModel) return [];
```
javierarce: I think that this change should be added to the `all` method too.

lib/assets/javascripts/cartodb3/components/modals/add-analysis/analysis-option-models/weight-analysis-option-model.js
```diff
@@ -0,0 +1,17 @@
+var AnalysisOptionModel = require('./analysis-option-model');
+
```
javierarce: I think this file is not needed. I'm not sure what is it :/

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/column-options.js
javisantana: it calls filterByType at the end, rigth?

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/filter-by-node-column.tpl
```diff
@@ -0,0 +1,20 @@
+<form>
+  <div class="Editor-HeaderInfo">
+    <div class="Editor-HeaderNumeration CDB-Text is-semibold u-rSpace--m">2</div>
+    <div class="Editor-HeaderInfo-inner CDB-Text" data-fields="source,filter_source">
+      <div class="Editor-HeaderInfo-title u-bSpace--m">
+        <h2 class="CDB-Text CDB-HeaderInfo-titleText CDB-Size-large"><%- _t('analyses.cluster') %></h2>
```
javierarce: `analyses.cluster` should be `analyses.filter-by-node-column` and then an entry should be added to `en.json`.

javierarce:
I added some comments. I'd also recommend to bump the frontend version in the packajge.json before merging. 

Besides those things, :+1:

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/column-options.js
javierarce: Oh, true! Nevermind then :)

PR #8118, opened 2016-06-23
Add map copier dependency

javitonino:
Before it was taken implicitly from the controllers, and I removed it with the visualization copier refactor.

CR @gfiorav 
cc @matallo (thanks for reporting)


matallo:
thanks mate

gfiorav:
looks 📦

PR #8124, opened 2016-06-23
Migrate node deletions to user-actions

viddo:
Milestone for #7881

This seem to have some unwanted consequences for the layer creations (tests failing at least), since the deletion is not handled implicitly anymore, ~~so might have to migrate those too as part of this~~ fixe by (temporarily) delete faulty analysis. Will be fixed in the next PR so don't want to waste time on that now, at least the layer creation can be migrated separately. 

@alonsogarciapablo can you review?


lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -88,7 +91,52 @@ module.exports = function (collections) {
...
+        // If deleted node was a head of a layer change the source to be its primary source instead
+        layerDefinitionsCollection.each(function (m) {
+          if (m.get('source') === nodeId) {
+            m.save({source: primarySourceNode.id});
+          }
+        });
+      }
+
+      var containsNode = function (m) {
+        return m.id !== nodeDefModel.id && m.containsNode(nodeDefModel);
```
alonsogarciapablo: `widgetModel.containsNode` and `layerDefinitionModel.containsNode` doesn't make a lot of sense naming wise IMO, but I like 👇 so 👍

viddo: for widget it's questionable yes, but for layer I believe it's accurate. :)

alonsogarciapablo:
👍

viddo:

PR #8127, opened 2016-06-23
link infowindows with cartodb_id

matallo:
CR @javisantana

### DIFF
--- a/lib/assets/javascripts/cartodb3/deep-insights-integration/link-layer-infowindow.js
+++ b/lib/assets/javascripts/cartodb3/deep-insights-integration/link-layer-infowindow.js
@@ -8,17 +8,18 @@ module.exports = function linkLayerInfowindow (layerDef, visMap) {
     var attrs = JSON.parse(JSON.stringify(infowindowModel.attributes));
 
     if (infowindowModel.isEmptyTemplate()) {
-      attrs.fields = [{
-        name: 'cartodb_id',
-        title: true,
-        position: 0
-      }];
-
-      attrs.template = INFOWINDOW_NONE;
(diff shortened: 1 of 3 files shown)


javisantana:
not sure if I'm the best to review this one since I don't know so much about the queryModel

lib/assets/javascripts/cartodb3/deep-insights-integration/link-layer-infowindow.js
```diff
@@ -8,17 +8,22 @@ module.exports = function linkLayerInfowindow (layerDef, visMap) {
...
-        title: true,
-        position: 0
-      }];
-
-      attrs.template = INFOWINDOW_NONE;
+      var cartdb_id = layerDef.getAnalysisDefinitionNodeModel().querySchemaModel.columnsCollection.find(function (m) {
+        return m.get('name') === 'cartodb_id';
+      });
+
+      if (cartdb_id) {
```
viddo: So, what would/should happen if `isEmptyTemplate` and there's no `cartodb_id` column in the results? This scenario could happen if the node results is not available yet for example.

matallo: it would apply the changes in the map infowindow from the attributes passed in the change event

viddo:
Looks legit, added a small comment re: a detail that this change introduce I wasn't sure about though.

Also, while I was explained offline why this change was necessary I'd prefer to have a test case that outlines and asserts the scenario.

matallo:
yeah, tests on their way, right now they are broken :/

matallo:
🚀

PR #8131, opened 2016-06-23
Revert "Copy visualizations using export/import"

javitonino:
Reverts CartoDB/cartodb#8045


gfiorav:
In Github we trust!

📦 it

I'd wait for 🍏 just in case. We're not in that much of a hurry.

Issue #8145, opened 2016-06-23
Basemap thumbnail on layer list is missing corner-radius


saleiva:
![screen shot 2016-06-23 at 21 19 55](https://cloud.githubusercontent.com/assets/132149/16316768/488c5c02-3988-11e6-9f50-42b33ea1872c.png)

PR #8147, opened 2016-06-23
fix border-radius #8145

piensaenpixel:
- Fix https://github.com/CartoDB/cartodb/issues/8145


javierarce:
👍

PR #8151, opened 2016-06-24
Fix cursos css editor after saving

### LINKED ISSUE #8086: Cursor position when saving CartoCSS
in the builder, when using the cartocss editor, every time you save, the cursor goes all the way back to the top of the cartocss sheet and does not stay in the current line 

cc @nobuti

nobuti:
This PR fixes #8086.

cc @xavijam @makella


xavijam:
I feel we will need to do the same with the sql and infowindow editors.

nobuti:
I've checked it out and none of them have this problem. In the case of the infowindow, the code mirror's model's content update happens when we are not editing. In the sql view, we only updates it on initialisation.

PR #8152, opened 2016-06-24
Added cancel button.

### LINKED ISSUE #7941: I miss "CLEAR QUERY/VIEW" button in the new sql editor
Like we do in the current one:

screen shot 2016-06-20 at 10 28 55

Don't you @nobuti @urbanphes @saleiva ?

nobuti:
This PR fixes #7941.

cancel-query

cc @xavijam 

The cancel button only appears if there is a history in the undo-redo manager. Not sure if once cancelled, we should reset the history as well.

I guess we need some kind of `CDB-Button--terciary` right @piensaenpixel?


xavijam:
`CANCEL` or `CLEAR`?

xavijam:
And the button should only appear when a custom query is applied. About the history, there is no problem, it someone clicks over it, history should add that last query.

xavijam:
LGTM man! the only thing is we have called the component undo-redo but it does more than that. But it is not a problem, so let's go! :)

PR #8158, opened 2016-06-24
Copy visualizations using export/import

### LINKED ISSUE #7657: duplicate map of a map with analyses fails

javitonino:
Fixes #7657

Now more :bomb:-proof than ever.


javitonino:
Please CR @gfiorav

app/controllers/api/json/visualizations_controller.rb
```diff
@@ -40,53 +40,30 @@ def create
...
-                additional_fields = {
-                  type:       params.fetch(:type, Visualization::Member::TYPE_DERIVED),
-                  parent_id:  params.fetch(:parent_id, nil)
-                }
-
-                @stats_aggregator.timing('copy') do
-                  Visualization::Copier.new(current_user, source, name_candidate)
-                                       .copy(copy_overlays, copy_layers, additional_fields)
+        source_id = params[:source_visualization_id]
+        vis = if source_id
```
gfiorav: This seems really prone to corner cases but I guess it's the easiest path. Let's make a mental note to create a duplicate api or something like that.

gfiorav:
If you tested it and it works, 📦 it!

PR #8160, opened 2016-06-24
add check for image presence

gfiorav:
Aims to fix https://rollbar.com/vizzuality/CartoDB/items/19172/ and https://rollbar.com/vizzuality/CartoDB/items/19173/

They have only occurred for a given client, viz and only once. But it doesn't hurt.

CR please @javitonino


javitonino:
:+1:

P.S: The antonym of `present?` is `blank?` which checks for nil.

PR #8168, opened 2016-06-24
Check if notifier collection is available before removing the notifier model

xavijam:
This is the case that the notification was included through a collection, isn't it?

CR: @nobuti


nobuti:
👍  but when do not we want that? We should use https://github.com/CartoDB/cartodb/blob/ba5c11e6f8d73404f36c7237d3f96244dad18f91/lib/assets/javascripts/cartodb3/editor/components/notifier/notifier.js#L50-L52 to add a notification, right?

xavijam:
better? :)

nobuti:
🚀

xavijam:

PR #8169, opened 2016-06-24
Finish import process correctly when overviews creation fails

### LINKED ISSUE #8129: Timeouts in overviews process are not rescued and might create tables not fully setup
* The canonical map of a table is created in the `after_create` fallback. See: https://github.com/CartoDB/cartodb/blob/73c9d1aa0d88d48ac65a34ce1e96f5f30a8d801b/app/models/table.rb#L404

* The table gets created inside the `register` function in the import process. See: https://github.com/CartoDB/cartodb/blob/73c9d1aa0d88d48ac65a34ce1e96f5f30a8d801b/app/models/table_registrar.rb#L11

* As an exception in the `create_overviews` method is not rescued, this piece of code (`store_results`) is never run: https://github.com/CartoDB/cartodb/blob/a58af3f553cc48380e5fee1d110567aa538e025b/app/models/data_import.rb and it seems that [this piece of code](https://github.com/CartoDB/cartodb/blob/a58af3f553cc48380e5fee1d110567aa538e025b/app/models/data_import.rb#L712) could be key for the correct map_id setup.

* This is the last method which is raising the exception, inside Data Import: https://github.com/CartoDB/cartodb/blob/master/app/models/data_import.rb#L390-L398

If you need to reproduce the issue, I have a table for it, just let me know!

jgoizueta:
Fixes #8129
Previously when overviews creation failed the imported datasets were left in the db in an inconsistent state.

Since overviews are not essential, we are now ignoring any error during this phase and the import is completed without them.


jgoizueta:
Mmm, the rollbar error is grouped with other `Sequel::DatabaseError: Import error` and is hard to detect 😞

jgoizueta:
@rafatower @iriberri please review this. It's been tested in staging.

any suggestion to make the rollbar error more noticeable?

iriberri:
LGTM! I think the logging is good enough for the purpose, perhaps I'd use a debug/warning instead of an error but as you see. :-)

rafatower:
Try RQL, sending an example through slack :smile: 

:+1:

Issue #8171, opened 2016-06-24
add analysis button target is too small


matallo:
only the "+" is clickable

![screen shot 2016-06-24 at 13 07 25](https://cloud.githubusercontent.com/assets/36676/16336306/9e9212ac-3a0c-11e6-8b1b-0f632b168403.png)

/cc @saleiva

PR #8175, opened 2016-06-24
fix button #8171

piensaenpixel:
@matallo could you take a look?

Fix https://github.com/CartoDB/cartodb/issues/8171

button


matallo:
👍

PR #8182, opened 2016-06-24
Save map center changes

### LINKED ISSUE #7484: Moving the map on the new editor doesn't persist the position after reloading
Enter to a map, move it, reload. The position is the original one instead of the new one

xavijam:
Fixes #7484 

Thanks to @alonsogarciapablo, we save the map state when any moveend event is sent from cartodb.js map \o/, thanks @alonsogarciapablo.

CR: @alonsogarciapablo 

Sounds good?


lib/assets/javascripts/cartodb3/data/map-definition-model.js
```diff
@@ -1,18 +1,64 @@
...
     if (!opts.layerDefinitionsCollection) throw new Error('layerDefinitionsCollection is required');
+    if (!opts.userModel) throw new Error('userModel is required');
+    if (!opts.vis) throw new Error('vis is required');
+
+    this._configModel = opts.configModel;
+    this._mapModel = opts.vis.map;
+    this._userModel = opts.userModel;
+    this._layerDefinitionsCollection = opts.layerDefinitionsCollection;
+
+    _.delay(this._initBinds.bind(this), DELAY_TIME);
```
xavijam: A bit of delay in order to not save the map every time the map is initialized, workaround for the moment, until we have what I've mentioned in the description.

xavijam:
Forget about my previous comment, adding zoomend event and we are ready! :)

lib/assets/javascripts/cartodb3/editor.js
```diff
@@ -226,5 +227,21 @@ deepInsights.createDashboard('#dashboard', vizJSON, {
 
   vis.centerMapToOrigin();
 
+  window.mapDefModel = new MapDefinitionModel(
```
alonsogarciapablo: As I've said before, I'd try to avoid the use of `parse: true` for this. I'd much rather do something like:

```
var mapAttrs = _.extend({
  id: visualizationData.map_id
}, _.pick(vizJSON, ['drag', 'id', 'keyboard', 'legends', 'scrollwheel']))

new MapDefinitionModel(mapAttrs, {
  configModel: configModel,
  ...
});
```

xavijam: It is not used in any other place, and having that, at least for me, is more clear (my opinion 😄 )

lib/assets/javascripts/cartodb3/data/map-definition-model.js
```diff
@@ -1,18 +1,64 @@
...
+    this._userModel = opts.userModel;
+    this._layerDefinitionsCollection = opts.layerDefinitionsCollection;
+
+    this._initBinds();
+  },
 
-    var baseUrl = opts.configModel.get('base_url');
-    this.urlRoot = baseUrl + '/api/v1/maps';
+  _initBinds: function () {
+    this._mapView._leafletMap && this._mapView._leafletMap.on('moveend', _.debounce(this.save, DEBOUNCE_TIME), this);
```
alonsogarciapablo: I don't think this is going to work after https://github.com/CartoDB/cartodb.js/pull/1313 😞 .

`cdb.createVis` now returns a [model](https://github.com/CartoDB/cartodb.js/blob/v4/src/vis/vis.js#L18) that knows nothing about views (mapView now belongs to [this view](https://github.com/CartoDB/cartodb.js/blob/v4/src/vis/vis-view.js#L22)...

alonsogarciapablo: The good news: https://github.com/CartoDB/cartodb.js/pull/1318

xavijam:
You are ❤️

Issue #7801, opened 2016-06-13
[POPUPS] Popup fields order gets changed after clicking a new field


saleiva:
Create a point map
Go to popup
Select a field
Move that field to the first position
Click on another field
>>> The order of fields get changed.

saleiva:
It didn't happen to me on hover, but happened to me on click. Not sure if this is a normal rule

matallo:
submitted the PR https://github.com/CartoDB/cartodb/pull/8192 and reported a bug in the current editor https://github.com/CartoDB/cartodb/issues/8191

saleiva:
should we close this @matallo ?

Issue #8191, opened 2016-06-24
editor infowidow and map infowindow are inconsistent


matallo:
**str**
- create map
- add point
- create several columns
- add infowindow
- select all fields
- reload
- deselect several fields
- reload
- select several fields from the former deselected

**expected behaviour**
- fields order in infowindow and panel to be the same

**actual behaviour**
- fields order in infowindow and panel are different

![screen shot 2016-06-24 at 21 41 42](https://cloud.githubusercontent.com/assets/36676/16348762/ce550ee8-3a54-11e6-98b9-292b4c8c5d52.png)

[Untitled map 50 (on 2016-06-24 at 19.41.59).carto.zip](https://github.com/CartoDB/cartodb/files/332480/Untitled.map.50.on.2016-06-24.at.19.41.59.carto.zip)

they will get randomnly ordered in the infowindow (sometimes with proper order, some others not), as the pos stored in the fields array will be `Number.MAX_VALUE` for both at the time of save, thus it can't be properly sorted

ramiroaznar:
@matallo should we close this issue ?

matallo:
no work has been done in this, so i wouldn't close it

PR #8192, opened 2016-06-24
[popups] prevent reorder after field click

matallo:
close https://github.com/CartoDB/cartodb/issues/7801

CR @javisantana 

I've spotted bugs in current infowindows that drove me crazy while trying to reproduce and fix the issue, I've already created the corresponding issue https://github.com/CartoDB/cartodb/issues/8191

~~this PR will make the linter fail, but I don't really know how to ignore the `==` for that line~~not anymore

---

ps: I've spent way too much time tackling this, I'd say fix the reorder bug and move on for the sake of consistency with old editor, the way field positions are stored and then sent to the actual map infowindows needs to be rethought :(


saleiva:
👍

javisantana:
:es:

Issue #8174, opened 2016-06-24
SQL <-> DATA switch state is not persisted when changing sections or going back to the layer list and back to the layer again


saleiva:
If you are in the SQL editor and you go to any other section or to the layer list, Once you enter again to the layer, it should be shown by default under the DATA tab.

PR #8196, opened 2016-06-25
No data available

### LINKED ISSUE #8155: If there is no data available, we should show the "loading" state + infobox with a message
"There is no data available to be displayed" for example.
In any case it is weird we can't show any data about "european countries", cc @fdansv.

nobuti:
This PR fixes #8155, #8174, #8049.

cc @xavijam


xavijam:
I miss a screenshot 😬 , but 🚢 :it:

nobuti:
data-columns

xavijam:
👏

saleiva:
👍

PR #8209, opened 2016-06-27
Exclude the_geom that are null

### LINKED ISSUE #8181: [analysis] geometry type population should skip null geometries
When you try to add an analysis a query to the SQL API is performed in order to know the geometry type, the query might return null geometries.

We should filter `where the_geom is not null` so we get some geometry types and avoid getting nulls.

P.S.: My guesstimation about this is: this is happening since we started adding on the fly indexes for the widgets, making the returning order to change. What do you think about it? 😈 .

viddo:
Fixes #8181 

@rochoa OK?
cc @xavijam @saleiva

### DIFF
--- a/lib/assets/javascripts/cartodb3/data/query-schema-model.js
+++ b/lib/assets/javascripts/cartodb3/data/query-schema-model.js
@@ -5,7 +5,7 @@ var createGeometry = require('../value-objects/geometry');
 var QueryRowsCollection = require('./query-rows-collection');
 
 var MAX_GET_LENGTH = 1024;
-var WRAP_SQL_TEMPLATE = 'select * from (<%= sql %>) __wrapped';
+var WRAP_SQL_TEMPLATE = 'select * from (<%= sql %>) __wrapped where the_geom is not null';
 var PARAMS = {
   sort_order: 'asc',
   rows_per_page: 40,


rochoa:
Quick question: what do we use that query for? Apart from getting the geometry type, I mean. Do we use it for the table view?

viddo:
> Quick question: what do we use that query for? Apart from getting the geometry type, I mean. Do we use it for the table view?

To get the schema (columns) of the query, which is needed in some cases for the analysis.
What goes for the table I'm not sure, I think it's better @xavijam answers that.

rochoa:
I was asking because if we are using that query in other places we will constraint the results to only non-null geometries, and we don't what that behaviour.

Furthermore if we are using that query just for retrieving the geometry type I would `LIMIT 1`.

lib/assets/javascripts/cartodb3/data/query-schema-model.js
```diff
@@ -5,7 +5,7 @@ var createGeometry = require('../value-objects/geometry');
 var QueryRowsCollection = require('./query-rows-collection');
 
 var MAX_GET_LENGTH = 1024;
-var WRAP_SQL_TEMPLATE = 'select * from (<%= sql %>) __wrapped';
+var WRAP_SQL_TEMPLATE = 'select * from (<%= sql %>) __wrapped where the_geom is not null';
```
xavijam: I feel we discussed @javisantana, you (@viddo) and me about not adding this due to a side effect... do you remember why @javisantana?

viddo:
> I was asking because if we are using that query in other places we will constraint the results to only non-null geometries, and we don't what that behaviour.

That's the _internal_ query of that model though used to communicate with the SQL API, other usages depend on the `query` attr which are unaffected. Also confirmed with @xavijam that this model is not used in the table view.

> Furthermore if we are using that query just for retrieving the geometry type I would LIMIT 1.

Currently we limit the result through a query param, `rows_per_page: 40`, and we also fallback on `the_geom_webmercator` if there are no `the_geom` so this sample (larger than a single row) is still motivated, unless we want to change this?

lib/assets/javascripts/cartodb3/data/query-schema-model.js
viddo: From what I can recall we (re)moved the `LIMIT 1` and get a sample (40 rows) instead, assuming that at if the dataset have geoms they should appear in this sample at least (basically what I answered [in the PR below](https://github.com/CartoDB/cartodb/pull/8209#issuecomment-228751361)).

Although I just realized one possible limitation, this change is only valid if `the_geom` column is always present for the query source; AFAIK it is always present for cartodb-tables, but  does that old up for analyses too @rochoa ?

viddo:
I'll take this out of the milestone since it's apparently not as important anymore, feel free to bump it whenever you ready to talk about it.

viddo:
I'll close this for now, until we'll take a decision on this.

PR #8232, opened 2016-06-27
Don't sync query changes for source nodes

viddo:
The query is always fixed and set from editor, so no need to listen to changes from analysis nodes in cartodb.js. So, with this a layer with a source node (i.e. no analysis) should allow modifying the styles w/o risking the geometry to be unknown.

@xavijam please review
cc @AbelVM


xavijam:
Super important fix, thanks a lot man! 💋  🚢 :it:

viddo:
Frontend test passing, so merging

PR #8240, opened 2016-06-27
Fixes several issues with the filter analysis

javierarce:
This PR:

- [x] Shows 3rd form only when selecting a column above:  27c2453
- [x] Sets show option activated by default: 9243469
- [x] Adds new definition + new localization: 737aaa4
- [x] Shows select/text depending on the number of results + Fixes label names: bd50f39


javierarce:
CR: @viddo

viddo:
Looks legit 👍

javierarce:
Merging, since the build passed before and the last commit just bumps the package version.

Issue #8115, opened 2016-06-23
change font color in edit layer


piensaenpixel:
![image](https://cloud.githubusercontent.com/assets/1214201/16306467/1ad91478-395e-11e6-980f-08fffd87ab6f.png)

PR #8242, opened 2016-06-27
fix change color while editing

piensaenpixel:

Fixed - https://github.com/CartoDB/cartodb/issues/8115

editing


nobuti:
The fix is 💯  but the gif is 👎

piensaenpixel:
why the gif is 👎 :(((?

PR #8247, opened 2016-06-27
Do not force column selection on analyses generated forms

rochoa:
This avoid issues when the column is a new column that will be
created by the analysis itself, AKA data augmentation.

CR @viddo 🙏


viddo:
👍

PR #8252, opened 2016-06-27
Autostyle invalidation on layer style change

### LINKED ISSUE #8050: [WIDGETS] Unselect auto-style option when having auto-style activated and changing the style of the layer
Create a map,
Add a widget
Style by that widget
Change the style of the layer.

It works as expected but the button on the widget keeps checked

fdansv:
Fixes #8050 

cc @xavijam


lib/assets/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -98,6 +101,19 @@ F.prototype._onWidgetDefinitionSynced = function (m) {
...
+  var self = this;
+  var currentLayerId = model.get('id');
+  var widgetsToInvalidate = this.widgetDefinitionsCollection.where({'layer_id': currentLayerId});
+  widgetsToInvalidate.forEach(function (widget) {
+    var widgetModel = self._diDashboard.getWidget(widget.id);
+    if (widgetModel) {
+      widgetModel.cancelAutoStyle();
+    }
+  });
+  self._diDashboard._dashboard.vis._layersCollection.where({'id': currentLayerId})[0];
```
xavijam: here you don't need self.

xavijam:
I'd add a spec.

lib/assets/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -98,6 +101,19 @@ F.prototype._onWidgetDefinitionSynced = function (m) {
...
 };
 
+F.prototype._onLayerChangeCheckWidgetAutoStyling = function (model, style) {
+  var self = this;
+  var currentLayerId = model.get('id');
+  var widgetsToInvalidate = this.widgetDefinitionsCollection.where({'layer_id': currentLayerId});
+  widgetsToInvalidate.forEach(function (widget) {
+    var widgetModel = self._diDashboard.getWidget(widget.id);
+    if (widgetModel) {
+      widgetModel.cancelAutoStyle();
```
viddo: Why cancel instead of avoiding changing/updating in the first place?

fdansv: Because we need to cancel autostyling where, within the editor, the style has been changed. This makes sure that widgets associated to the layer whose style we're editing get their autostyling disabled.

viddo: I understand _what_ it does, but I was asking _why_ not handle that as a side-effect internally in DI instead? I.e. listen to layer changes on style and reset the autostyling if it's changed. 

Is there some use-case where the styles might change but the autostyle state should still be set (even if it's not auto-styled anymore)?

Btw, wouldn't this method call break for widgets that don't have auto-styling? From [what I can see](https://github.com/CartoDB/deep-insights.js/search?utf8=%E2%9C%93&q=cancelAutostyle&type=Code) it's only category and histogram that have that behavior.

fdansv: > Is there some use-case where the styles might change but the autostyle state should still be set (even if it's not auto-styled anymore)?

Yes, autostyling can be activated for more than one widget for a single layer (widths + colours), and widgets change directly the `cartocss` property of the layers, so we would have to handle widgets interfering with each other on autostyle.

lib/assets/test/spec/cartodb3/deep-insights-integrations.spec.js
```diff
@@ -406,6 +406,34 @@ describe('deep-insights-integrations', function () {
...
+        title: 'test',
+        layer_id: 'integration-test',
+        column: 'col',
+        source: 'a0'
+      });
+      model.trigger('sync', model);
+      var widgetModel = this.dashboard.getWidgets()[0];
+      widgetModel.set('autoStyle', true);
+      layerDefinitionModel.set('cartocss', 'differentCartocss');
+      document.body.removeChild(document.body.children[document.body.children.length - 1]);
```
xavijam: ?

xavijam: Do we need it?

viddo: I figure this is related to "fixing" the [tests that failed before this commit](https://travis-ci.org/CartoDB/cartodb/builds/140759078), the thing is that this is just sweeping the problem under the rug, the proper fix would be to investigate why the DOM is not cleaned up; most likely some view is missing to `this.addView(subViewContainingTooltip)`

viddo: Why merging before resolving this? This can potentially lead to memory leaks and problems later. 😞

fdansv: Sorry, I merged not seeing this thread of comments, probably because they were made before I pushed the last commit, so this is an outdated diff. The solution I pushed [wasn't that different](https://github.com/CartoDB/cartodb/pull/8252/files#diff-595367dadfc7a09d99fe5cc28ebc5905R432), but it is to circunvent what @viddo mentioned. I'll report it in deep insights and fix the test afterwards.

PR #8254, opened 2016-06-27
7641 cancel export on modal close

### LINKED ISSUE #7639: Failed export with bad config continues to poll status
1. set a invalid S3 entry in the `app_config.yml` file, e.g.:
```yml
    s3:
      access_key_id:
      secret_access_key:
      bucket_name:
      url_ttl:
      async_long_uploads: false
```
2. open a map and try to export it

**Expected result:**

- The polling to stop after the first request having state `failed`.
- Some more useful information about invalid config

**Actual result:**

- An error message (👍 ), but the export polling don't stop (👎 ) 
- errors found on `select entries from logs order by created_at desc limit 1;` on `carto_db_development` shows:
```
 2016-06-03 07:32:39 UTC: Exporting: #<NoMethodError: undefined method `[]' for nil:NilClass>                +
 /home/ubuntu/www/production.cartodb.com/current/app/models/carto/visualization_export.rb:45:in `run_export!'+
 /home/ubuntu/www/production.cartodb.com/current/lib/resque/exporter_jobs.rb:11:                             +
``` which if I understood it correctly is due to the invalid config entry for the S3 key. A guard there and some more useful output for devop would be useful?

### LINKED ISSUE #7641: Close export modal spams continue to poll export indefinitely
1. open a map and try to export it*
1. Close the modal right away, before the export is done

**Expected result:**

- Cancel/discard the export and any request(s) in flight
- Stop polling

**Actual result:**

- Polling continues and more intensively, it seem to do like 1-2 requests/second

_* this was after fixing the config as reported in #7639_

gfiorav:
Fixes #7641 
Fixes #7639

Please CR @xavijam I feel you won't agree on the trigger class approach


xavijam:
Change package version.

PR #8263, opened 2016-06-28
Table edition

xavijam:
Basically:

- Fixed problem with notifier in the table-edit-cell.
- Enable possibility to edit any value in the table view.

#### Pending
- [ ] Geometry edition.
- [ ] Check date styles.
- [ ] Table body class is super big.
- [ ] Table body tests.
- [ ] Table edition tests.


xavijam:
Passing frontend tests, removing conflicts and merging, a rush!

PR #8266, opened 2016-06-28
Show all analysis nodes in the source options

### LINKED ISSUE #8265: no analysis nodes on select forms
steps:

- add a layer with some analysis
- add a new analysis that needs to select a different dataset (like intersection). Analysis nodes are not shown in the select and they should be there

viddo:
Fixes #8265 

At the time of implementation it was only allowed to select layers' head nodes as input sources, this changes the options to include _all_ nodes (except for the self). 
Also fixed the bug where source options weren’t shown in the options if the query-schema was already fetched. 

E.g.

screen shot 2016-06-28 at 11 30 59 screen shot 2016-06-28 at 11 30 53

@javisantana @saleiva  Right now only the input source is filtered out (A2 in the example above), but we could potentially have trouble if we don't filter out nodes that creates cycles (e.g. a node in layer A, which has a node that points to a node in B…), but we can solve that separately…

@xavijam review, please

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -21,8 +21,6 @@
       throw grunt.util.error(env +' file is missing! See '+ env +'.sample for how it should look like');
     }
 
-    env.tape_tests = 'lib/assets/test/tape/**/*-test.js';
-
     var aws = {};
     if (grunt.file.exists('./lib/build/grunt-aws.json')) {
       aws = grunt.file.readJSON('./lib/build/grunt-aws.json');
--- a/lib/assets/javascripts/cartodb3/data/query-schema-model.js
+++ b/lib/assets/javascripts/cartodb3/data/query-schema-model.js
@@ -44,8 +44,12 @@ module.exports = Backbone.Model.extend({
     return this._configModel.getSqlApiUrl();
   },
 
+  isFetched: function () {
+    return this.get('status') === 'fetched';
+  },
+
   canFetch: function () {
-    return this.get('query') && !this._isFetched();
+    return this.get('query') && !this.isFetched();
   },
 
   fetch: function (opts) {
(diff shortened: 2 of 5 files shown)


xavijam:
:+1:

saleiva:
Is this still showing your of the datasets like before?

javisantana:
@saleiva take into account the analysis it's showing is link to layer (pending a PR to change the name) and that's only meant to nodes but it should show the datasets as well

saleiva:
Ok, thanks!

2016-06-28 11:46 GMT+02:00 javi santana <notifications@github.com>:

> @saleiva <https://github.com/saleiva> take into account the analysis it's
> showing is link to layer (pending a PR to change the name) and that's only
> meant to nodes but it should show the datasets as well
>
> —
> <https://github.com/CartoDB/cartodb/pull/8266#issuecomment-229002849>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAIENQOiGdXHiUBSyxtChwHfVI-6GLqdks5qQO2SgaJpZM4I_6dr>
> .
>



-- 
Sergio Alvarez Leiva

viddo:
@saleiva @javisantana yes, by _default_ datasets are also included, but for this particular analysis is [filtering out the datasets](https://github.com/CartoDB/cartodb/blame/90dc886a0404f90c309423e49c05a0a99af8b7ed/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/filter-by-node-column.js#L83) (I guess that was done on purpose?)

javisantana:
yes @viddo, I did it on purpose

viddo:
👍  for intersection you'll get both for example: screen shot 2016-06-28 at 11 51 51

viddo:
Merging since frontend tests already passed

PR #8271, opened 2016-06-28
Do not count raster tables with the_geom twice in ghost tables

### LINKED ISSUE #7942: Ghost tables runs unnecessarily sometimes
It seems it is related to tables being detected as both raster and non-raster.

javitonino:
Fixes #7942


javitonino:
CR @gfiorav

gfiorav:
Thanks! Please do test this in staging, it can be rally tricky! 

📦 it when 🍏 and staging proof!

PR #8274, opened 2016-06-28
Reset analysis payload once it has been used

### LINKED ISSUE #7826: After applying an analysis, it offers you to create it instead of apply changes
## STR

1. Create a map with a layer of points (layer A).
2. Go to the builder.
3. Apply a simple style to the layer A.
4. Add a buffer analysis over that layer A.
5. Apply a radius 40 meters.
6. Go to style again, style should be reseted.
7. Apply a simple one and change the color.
8. Go to the CartoCSS panel and change the `polygon-fill` to `red`, for example, apply changes.
9. Go back to analyses tab and check that the analysis already applied is offering `Create` instead of  `Apply`.

screen shot 2016-06-14 at 12 43 55

Reported by @saleiva

viddo:
Fixes #7826 

Navigating between the tabs caused the payload from the creation to be reused (due to the lifecycle closure of the parent render), causing the analysis view to create a new form model for the payload, instead of creating it from the persisted node of the layer as expected.

cc @xavijam not much to review, but oh well

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/layers/layer-content-view.js
+++ b/lib/assets/javascripts/cartodb3/editor/layers/layer-content-view.js
@@ -105,6 +105,7 @@ module.exports = CoreView.extend({
         } else {
           selectedNodeId = self.layerDefinitionModel.get('source');
         }
+        analysisPayload = null; // remove payload once we have used it, to not have it being invoked again when switching tabs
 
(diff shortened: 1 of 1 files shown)


xavijam:
package and ready to go

viddo:
Frontend tests passed so merging

Issue #8281, opened 2016-06-28
[popups] when load popup on none, it shows form


matallo:
it should show form if popup is none, however it works if I change tab or if I change template

![screen shot 2016-06-28 at 13 54 26](https://cloud.githubusercontent.com/assets/36676/16414450/ed386d10-3d37-11e6-9694-c25dd1615861.png)

PR #8282, opened 2016-06-28
[popups] close several minor issues

matallo:
#8272
#8273
#8275
#8281

CR @xavijam


lib/assets/javascripts/cartodb3/data/infowindow-definition-model.js
```diff
@@ -148,17 +148,14 @@ module.exports = Backbone.Model.extend({
   },
 
   setTemplate: function (templateName) {
-    var template = this.TEMPLATES[templateName];
+    var template = (typeof (this.TEMPLATES[templateName]) === 'undefined') ? '' : this.TEMPLATES[templateName];
```
xavijam: I really don't like typeof checks.
Also, below, we don't add template or template_type if template is not defined...

matallo: if it's not defined it would just change the template name, remaining the template, then the `isCustomTemplate` check won't work, we currently do like that for infowindows

https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/data/infowindow-click-model.js#L43-L61

xavijam:
Merge as soon as possible please :)

matallo:
🚀

PR #8284, opened 2016-06-28
Apply default style when a new analysis has been added and it is ready

xavijam:
This is a rudimentary option to set the style when a new analysis has finished. Bad things:

- We have to wait until the analysis is ready.
- Then we need to get the geometry in order to apply a proper style, so query-schema-model should be fetched.
- And finally, map will be reloaded, yes or yes, when a new analysis is ready, so we need to make another map instantiation with that new style.

**This is a first approach**, what do you think?

@javisantana @viddo @saleiva @alonsogarciapablo

### DIFF
--- a/lib/assets/javascripts/cartodb3/deep-insights-integrations.js
+++ b/lib/assets/javascripts/cartodb3/deep-insights-integrations.js
@@ -74,10 +74,13 @@ F.prototype._onAnalysisAdded = function (nodeDefModel) {
           if (nodeDefModel.querySchemaModel.get('status') === 'fetched') {
             var geom = nodeDefModel.querySchemaModel.getGeometry();
             var simpleType = geom && geom.getSimpleType();
+            var layer = _.first(this._layerDefinitionsCollection.where({ source: node.get('id') }));
 
-            setTimeout(function () {
-              this._layerDefinitionsCollection.at(1).styleModel.setDefaultPropertiesByType('simple', simpleType);
-            }.bind(this), 1000);
+            if (layer && simpleType) {
+              setTimeout(function () {
+                layer.styleModel.setDefaultPropertiesByType('simple', simpleType);
+              }, 1000);
+            }
             nodeDefModel.querySchemaModel.unbind('change:status', onChangeStatus, this);
           }
         };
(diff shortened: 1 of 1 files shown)


lib/assets/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -54,6 +56,48 @@ F.prototype._analyseDefinition = function (m) {
...
+        nodeDefModel.querySchemaModel.set({
+          query: node.get('query'),
+          may_have_rows: true
+        });
+
+        var onChangeStatus = function () {
+          if (nodeDefModel.querySchemaModel.get('status') === 'fetched') {
+            var geom = nodeDefModel.querySchemaModel.getGeometry();
+            var simpleType = geom && geom.getSimpleType();
+            var layer = _.first(this._layerDefinitionsCollection.where({ source: node.get('id') }));
```
viddo: `var layerDef = this._layerDefinitionsCollection.findOwnerOfAnalysisNode(node);`

lib/assets/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -54,6 +56,48 @@ F.prototype._analyseDefinition = function (m) {
   this._analyseDefinitionNode(nodeDefModel);
 };
 
+F.prototype._onAnalysisAdded = function (nodeDefModel) {
```
viddo: IMHO this should only be done once, and should probably be contained in [`_tryToSetupDefinitionNodeSync`](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/deep-insights-integrations.js#L65)

lib/assets/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -54,6 +56,48 @@ F.prototype._analyseDefinition = function (m) {
...
+        var onChangeStatus = function () {
+          if (nodeDefModel.querySchemaModel.get('status') === 'fetched') {
+            var geom = nodeDefModel.querySchemaModel.getGeometry();
+            var simpleType = geom && geom.getSimpleType();
+            var layer = _.first(this._layerDefinitionsCollection.where({ source: node.get('id') }));
+
+            if (layer && simpleType) {
+              setTimeout(function () {
+                layer.styleModel.setDefaultPropertiesByType('simple', simpleType);
+              }, 1000);
```
viddo: I really don't think we should use setTimeout

viddo: Actually, we should not have side-effects that affect the persisting data models, I think this would solve itself when the styles are updated in a more appropriate place. AFAIK, the styles should only need to be changed when analysis are changed for a layer right? I think this would be fairly easy to solve once #7881 is done.

xavijam: Ok, let's wait for that before "tackling" this problem.

PR #8291, opened 2016-06-28
8286 forbid editing derived visualizations builder

javitonino:
Also, no raster editing in the dataset view (keeps current behaviour).


javitonino:
Please CR @gfiorav

gfiorav:
looking good! 📦 it!

PR #8297, opened 2016-06-28
POC/WIP moving side-effects to each corresponding model

viddo:
@alonsogarciapablo addressing some of the concerns from previous PR (and offline conversation), it's a POC so have only migrated the code quickly to discuss some different points. 
If we finally agree this PR is not the wrong road to take I'll just close it.

The primary goal to solve in #7881 is to get rid of the indirect CRUD operations done in listener callbacks, secondary to know have the things easily maintainable. While it can be convenient to have side-effects in one place (like we have with the deep-insights-integration) I'm not sure if that should be a goal in itself.

That said, on specific things from prev PR:

> (re: `updateOrCreateAnalysis`) [Why this indirection here? Makes this method harder to understand...](https://github.com/CartoDB/cartodb/pull/8042#discussion_r68192171)

Like I responded in the thread there, it ~~is~~ was necessary due use-cases have different side-effects, by keeping the ~~`updateOrCreateAnalysis`~~ `analysisForm.save` logic internal (in the form model) the indirection is no longer there. A bonus consequence is that `userActions.userActions.createAnalysisNode` could be removed.

> (re: userActions methods' params signatures) [One of the things that might be a bit confusing about this is the fact that each method takes some arguments and they are not very "uniform" (eg: one method takes a nodeAttrs object while another takes an analysisFormModel object). Do you think we could somehow homogenise this? (for example: only pass layer, analysis or widget models + attr objects for updates + options).](https://github.com/CartoDB/cartodb/pull/8042#discussion_r68194177)

> the good thing about your current approach is being able to have everything in the same place

I agreed with this sentiment, but the only thing in common between the various methods are really only the collections, so there is really no benefit of having the logic in the same place afterall. The important thing is to have the side-effects in the same place however (which is resolved by this).

This is resolved, by having the model encapsulating its own state and only require the collections they need to do the necessary side-effects. For now I kept userActions to not have to pass around the individual collections in all views (although some are already passed around), but what goes for me could even 🔥 it (to remove indirection).

Thoughts?

edit: addressing more concerns from offline conversation

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/modals/add-widgets/category/category-option-model.js
+++ b/lib/assets/javascripts/cartodb3/components/modals/add-widgets/category/category-option-model.js
@@ -5,7 +5,7 @@ module.exports = WidgetOptionModel.extend({
 
   defaults: _.defaults({type: 'category'}, WidgetOptionModel.defaults),
 
-  createUpdateOrSimilar: function (widgetDefinitionsCollection) {
+  _createUpdateOrSimilar: function (widgetDefinitionsCollection) {
     var columnName = this.columnName();
 
     var attrs = {
--- a/lib/assets/javascripts/cartodb3/components/modals/add-widgets/widget-option-model.js
+++ b/lib/assets/javascripts/cartodb3/components/modals/add-widgets/widget-option-model.js
@@ -19,8 +19,27 @@ module.exports = Backbone.Model.extend({
     return this._selectedItem().columnModel.get('name');
   },
 
-  createUpdateOrSimilar: function () {
-    throw new Error('createUpdateOrSimilar should be implemented by child');
+  save: function (attrs, opts) {
+    if (!opts.analysisDefinitionsCollection) throw new Error('analysisDefinitionsCollection is required');
+    if (!opts.widgetDefinitionsCollection) throw new Error('analysisDefinitionsCollection is required');
+
+    var nodeDefModel = this.analysisDefinitionNodeModel();
+
+    // Might not always have a node-definition, e.g. time-series none-option
+    if (nodeDefModel) {
+      var analysisDefinitionModel = opts.analysisDefinitionsCollection.findAnalysisThatContainsNode(nodeDefModel);
+
+      if (!analysisDefinitionModel) {
+        opts.analysisDefinitionsCollection.create({analysis_definition: nodeDefModel.toJSON()});
+        this.layerDefinitionModel().save();
+      }
+    }
+
+    this._createUpdateOrSimilar(opts.widgetDefinitionsCollection);
+  },
+
+  _createUpdateOrSimilar: function () {
+    throw new Error('_createUpdateOrSimilar should be implemented by child');
   },
 
   _selectedItem: function () {
--- a/lib/assets/javascripts/cartodb3/data/user-actions.js
+++ b/lib/assets/javascripts/cartodb3/data/user-actions.js
@@ -1,8 +1,7 @@
-var _ = require('underscore');
-var camshaftReference = require('./camshaft-reference');
-
 /**
- * Coordinate side-effects done on explicit interactions.
+ * Adapter for CRUD operations that require various collections,
+ * to not have to pass them all around various view hierarchies.
+ *
  * @param {Object} collections
  * @param {Object} collections.analysisDefinitionsCollection
  * @param {Object} collections.analysisDefinitionNodesCollection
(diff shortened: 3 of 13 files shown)


lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -22,121 +21,37 @@ module.exports = function (collections) {
...
-        analysisDefinitionModel.save({node_id: nodeDefModel.id});
-      } else {
-        analysisDefinitionsCollection.create({analysis_definition: nodeDefModel.toJSON()});
-      }
-
-      layerDefmodel.save({
-        cartocss: camshaftReference.getDefaultCartoCSSForType(nodeDefModel.get('type')),
-        source: nodeDefModel.id
+    saveAnalysis: function (analysisFormModel) {
+      analysisFormModel.save(null, {
```
alonsogarciapablo: We have access to `analysisDefinitionsCollection` and `analysisDefinitionNodesCollection` here because `user-actions.js` knows about all of these collections... If we got rid of `user-actions.js`, in this case for example, I guess `layers/layer-content-views/analyses/analysis-controls-view.js` would need to have references to `analysisDefinitionsCollection` and `analysisDefinitionNodesCollection`, right? And these references might only be needed because view triggers and action with side-effects on these collections... My point here, what are the benefits of this approach, vs something like:

```
var AnalysisFormModel = Backbone.Model.extend({
  initialize: function (attrs, deps) {
    this._dep1 = deps.dep1;
    this._dep2 = deps.dep2.
  },
  save: function () {
     this._dep1.aSideEffect();
     this._dep2.anotherSideEffect()
  }
};
```

On the other hand, we *could* have the problem of having to pass all or most collections because one of the side-effects requires more collections, like:

```
save: function (null, deps) {
  deps.dep1.save(null, _.pick(deps, 'dep4', 'dep5'));
  deps.dep2.doSomething(_.pick(deps, 'dep6'));
}
```

:point_up save only needs dep1 and dep2, but since dep1 and dep2 require dep4, dep5 and dep6, whoever calls .save, would need to have references to all of them...

lib/assets/javascripts/cartodb3/components/modals/add-widgets/category/category-option-model.js
```diff
@@ -5,7 +5,7 @@ module.exports = WidgetOptionModel.extend({
 
   defaults: _.defaults({type: 'category'}, WidgetOptionModel.defaults),
 
-  createUpdateOrSimilar: function (widgetDefinitionsCollection) {
+  _createUpdateOrSimilar: function (widgetDefinitionsCollection) {
```
alonsogarciapablo: Not very important here but... is this `_createOrUpdate` or `_findOrCreateOrUpdate`?

lib/assets/javascripts/cartodb3/components/modals/add-widgets/widget-option-model.js
```diff
@@ -19,8 +19,27 @@ module.exports = Backbone.Model.extend({
     return this._selectedItem().columnModel.get('name');
   },
 
-  createUpdateOrSimilar: function () {
-    throw new Error('createUpdateOrSimilar should be implemented by child');
+  save: function (attrs, opts) {
```
alonsogarciapablo: Without knowing exactly what this method does I see we're calling ` this._createUpdateOrSimilar(opts.widgetDefinitionsCollection);` which it's not just "saving" the model, so perhaps `save` could be missleading.

alonsogarciapablo: Without knowing exactly what this method does I see we're calling ` this._createUpdateOrSimilar(opts.widgetDefinitionsCollection);` which it's not just "saving" the model, so perhaps `save` could be misleading.

lib/assets/javascripts/cartodb3/components/modals/add-widgets/category/category-option-model.js
viddo: No, depends on the widget type, see https://github.com/CartoDB/cartodb/search?utf8=%E2%9C%93&q=createUpdateOrSimilar&type=Code and in particular [time-series one](https://github.com/CartoDB/cartodb/blob/f1d77864afb98fb6f989a99f0da09054f265e182/lib/assets/javascripts/cartodb3/components/modals/add-widgets/time-series/time-series-option-model.js). Like the others I'm thinking that even renaming it to `save` is probably a better name here too

lib/assets/javascripts/cartodb3/components/modals/add-widgets/widget-option-model.js
viddo: true, related to your other question and why it was named createUpdateOrSimilar… if you have any suggestion to a better name I'm all ears. :)

lib/assets/javascripts/cartodb3/data/user-actions.js
viddo: Yeah, I was thinking about that one too. It would encapsulate the dependencies better ofc, with the downside that it requires collections upfront which might be a PITA in itself (especially when setting up use-cases in the test env). But

I don't think the amount of deps will be a problem, all side-effects basically are around layers, analysis and nodes.

viddo: > I guess layers/layer-content-views/analyses/analysis-controls-view.js would need to have references to analysisDefinitionsCollection and analysisDefinitionNodesCollection, right? 

Not necessarily, we could use the technique I've used for [add-analysis](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/editor/editor-map-view.js#L15), let view events bubble up and handle them higher up the view hierarchy, i.e.

```
SomeParentView <-- have access collections, listen to 'click .js-destroy-analysis'
  SubView
     SubView
        SubView
           LayerAnalysesView <-- e.g. <a class="js-destroy-analysis" data-node-id="123", …
```

PR #8321, opened 2016-06-29
Moved notifier to general components

xavijam:
In order to be used in other places, like dataset view.

CR: @nobuti.


nobuti:
💯

xavijam:

PR #8324, opened 2016-06-29
Add migration to remove arcgis column

javitonino:


javitonino:
CR @gfiorav More data-loss inducing migrations = more fun

gfiorav:
📦 it when 🍏

PR #8331, opened 2016-06-29
fixed acum

### LINKED ISSUE #7796: Accum Torque doesn't work
Create a point map, click on animated, select Accum. It doesn't accumulates the data

javisantana:
fixes #7796


javisantana:
cc @xavijam

xavijam:
:clap:

PR #8334, opened 2016-06-29
Allows number columns to be used in timeseries widgets

fdansv:
time

@javisantana


viddo:
> Haha awesome, it's turning the cartodb_id numbers into dates.

IIRC It was only implemented to work with dates (as the name indicates). There is the [`column_type = 'number'`](https://github.com/CartoDB/deep-insights.js/blob/master/src/widgets-service.js#L140) which is passed to the dataview which I think you can use to indicate that the column is a number and not a date though, perhaps is of help?

javisantana:
merging this since we need to move forward and see what problems we have with number columns as dates inside the whole thing

PR #8341, opened 2016-06-29
Unsets the quantification when switching to the category list

### LINKED ISSUE #8078: Switching from ramps to categories doesn't remove the quantification
In categories the quantification makes no sense. It should be removed.

javierarce:
Closes #8078 

CR: @xavijam


xavijam:
Not much to say 👍  :es:

PR #8346, opened 2016-06-30
Remove uniqueId.

### LINKED ISSUE #8338: Notifier's last item cannot be close when having more than 1 at the same time
The error it says is
4tracker.js:4 Uncaught Error: TrackJS Caught: Cannot read property 'remove' of undefined

STR
Rename two layers very fast, and close one of the notifications.
When trying to close the other one it will fail.

nobuti:
Fixes #8338. 

It doesn't work well with concurrency apparently. Models had the same id.

cc @xavijam


xavijam:
🚢 🇮🇹

PR #8384, opened 2016-06-30
7881 explicit persist side effects layers

viddo:
Milestone for #7881 

This contains the migration of layer creation (from table and analysis node), from the layer-definitions-collection to the user-actions. Also includes the layer move/reordering since that's required for the layer-creation too.

Also tried to make the two big use-cases layer-creation from an analysis node (head or non-head) clearer by having clearly distinct code paths.

@alonsogarciapablo CR, please?


lib/assets/test/jasmine/cartodb3/data/user-actions.spec.js
```diff
@@ -868,6 +868,16 @@ describe('cartodb3/data/user-actions', function () {
       });
     });
 
+    // Node is NOT a head of a node, e.g. given nodeId is 'a3' this would create a new layer B which takes over the
```
alonsogarciapablo: given nodeId is 'a2' ?

lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -220,82 +220,130 @@ module.exports = function (collections) {
...
-      analysisDefinitionNodesCollection.each(function (m) {
-        m.changeSourceIds(nodeId, newNodeDefModel.id);
-      });
+        newLayerDefModel.save(null, {
+          success: function () {
+            layerDefinitionsCollection.save(); // for orders to be up-to-date
+          }
+        });
+      } else {
+        // Node is NOT a head of a node, e.g. given nodeId is 'a3' this would create a new layer B which takes over the
```
alonsogarciapablo: given nodeId is 'a2'

lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -220,82 +220,130 @@ module.exports = function (collections) {
...
-          order: newOrderForLayerWithNodeAsHead,
-          letter: tmpLetter,
-          color: layerColors.getColorForLetter(tmpLetter),
-          source: nodeDefModel.clone(tmpLetter + '1').id
-        });
 
-        // Remove layerWithNodeAsHead ("layer A") temporarily, to move the layers to expected positions
-        layerDefinitionsCollection.remove(layerWithNodeAsHead, {silent: true}); // silent to avoid unwanted side-effects; re-added again later;
+      var prevLayer = layerDefinitionsCollection.findWhere({source: nodeId});
+      if (prevLayer) {
```
alonsogarciapablo: ASCII art definitely makes it easier to understand 👍, but method is kind of long... Perhaps we could extract branches of this if/else to a couple of smaller methods?

alonsogarciapablo:
👍 Left some minor comments, but I think this is great! Great job!

lib/assets/javascripts/cartodb3/data/user-actions.js
viddo: Totally agree, and yes a lot of things could probably be folded into smaller methods that actually fit better in collections/models. But considering current time contraints I'd say this can wait for later, to finish the remaining stuff (layer deletion etc.)

PR #8387, opened 2016-06-30
Allows to specify the count operation without a column

### LINKED ISSUE #8235: Remove column from the aggregation form of the intersection filter when the operation is count

javierarce:
Closes #8235

This PR replaces the `aggregate_column` and ``aggregate_function` fields with a single field that allows selecting both things.

CR: @xavijam 

animation

PS: Despite the fact that the user is no longer required to specify a column, due to https://github.com/CartoDB/camshaft/issues/102 we still need to send a string (which is why we are sending `cartodb_id` behind curtains).


javierarce:
cc: @saleiva

xavijam:
Yay, operator for the win \o/.
🚢 :it:

saleiva:
Vamos!

PR #8393, opened 2016-06-30
Copy table_name_alias for new layer created.

### LINKED ISSUE #8360: source layer name reverts after dragging an analysis
### STR
- Add a layer `A`
- Rename layer `A` to something like 'Metro Stations'
- Add Area of Influence analysis
- Drag the analysis node `A1` to create a new layer `B`
- The new layer `B` gets the name of the source layer `A`. The source layer reverts to its original name

### Expected result

The source layer `A` should keep its name. 
The new layer `B` should get a default name related to the analysis it comes from: `area_of_influence` or something like that

nobuti:
This PR fixes #8360.

alias-drag

cc @xavijam @viddo

### DIFF
--- a/app/assets/stylesheets/editor-3/inline-editor.css.scss
+++ b/app/assets/stylesheets/editor-3/inline-editor.css.scss
@@ -12,7 +12,6 @@
   position: absolute;
   top: -1px;
   left: 0;
-  // width: calc(100% - 16px);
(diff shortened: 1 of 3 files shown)


nobuti:
CR 🙏

xavijam:
Not much to say :es:

PR #8398, opened 2016-07-01
Extract analyses definitions to a common object

### LINKED ISSUE #8317: Name shown in the layer list for some analysis is wrong
- [ ] Link to layer analysis.
- [ ] Cluster.

STR, add analysis to the layer, and go to the layer list.

viddo:
Resolves #8317 

Extracted from views and avoid some duplication. 
For now only the title and form model, later we could consolidate the analysis options here too, to have them in defined in one place.

@javierarce ok?
cc @saleiva


javierarce:
👍 🎈

viddo:
Was merged to #8407, so will be released with that enhancement instead, pending translation I'll do pending translation fixes in another PR later if necessary.

PR #8411, opened 2016-07-02
Node dataset component

### LINKED ISSUE #7775: Show previous analysis id+name on input form field for an analysis
All "input" fields in the analyses forms currently only display the id (e.g. "a0"), it should display the input with more visual cues, like this:

screen shot 2016-06-13 at 12 40 46

Should be implemented in such a way that's it's reusable for all the different kind of forms that have a source.

In the case it's a secondary source it should not be disabled and slightly different styling too:
screen shot 2016-06-13 at 12 43 23

### LINKED ISSUE #7832: Dropdown list when selecting layers on analysis, should also show layers on the map, and those should be the first ones

xavijam:
Basically it does this thing:

screen shot 2016-07-02 at 11 35 04

We have added more information in the node within the `analysis-source-options-model`, and we could display it in a different way in a new component called `NodeDataset`. This new component can display nicely sources or datasets. If we want to show the selected node as it appears above, we have created a function called `getSourceOption` (in the `base-analysis-form-model`) that returns all the needed info to display it nicely.

Example: 
```js
…
var schema = {
      source: {
        type: 'NodeDataset',
        title: _t('editor.layers.analysis-form.source'),
        options: this.getSourceOption(),
        editorAttrs: {
          disabled: true
        }
      },
    …
…
```

Also, I've changed some small things in several analysis, please take a look at length. :kiss:

It is likely I've missed any analysis that needs this component, so please, take a look.
@viddo @nobuti @javierarce @javisantana 

Fixes #7775.
Fixes #7832.


lib/assets/javascripts/cartodb3/components/form-components/editors/node-dataset/node-dataset-selected.tpl
```diff
@@ -0,0 +1,10 @@
+<% if (typeof type != 'undefined' && type === 'node') { %>
```
xavijam: If desired info is not provided, no problem, it will work as it used to do.

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/base-analysis-form-model.js
```diff
@@ -31,6 +31,15 @@ module.exports = Backbone.Model.extend({
     return undefined; // override to return custom template data
   },
 
+  getSourceOption: function () {
```
xavijam: Already commented in the description, but just in case.

lib/assets/javascripts/cartodb3/components/form-components/editors/node-dataset/node-dataset-selected.tpl
nobuti: `type === 'node'` already implies `typeof type != 'undefined'`, doesn't it?

xavijam: Sometimes node could be not provided, and type could be dataset. I feel it was working 🆗 , but probably it doesn't make much sense :S.

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/base-analysis-form-model.js
viddo: I think it would be preferred to keep things that belong together closer to each other. Also, from the look of it, the _primary_ source is always displayed in the same way? With this in mind, I think we could make the intention more clear and tidy/DRY things up a little while at it, by encapsulating the whole schema item, e.g.

```js
// base-analysis-form-model:
_primarySourceSchemaItem: function (customLabel) {
  return {
    type: 'NodeDataset',
    title: customLabel || _t('editor.layers.analysis-form.source'), // not sure customLabel is even needed, but for the sake of the example I'll include it
    options: [{
      val: this.get('source'),
      type: 'node',
      layerName: this._layerDefinitionModel.getName(),
      color: this._layerDefinitionModel.get('color')
    }],
    editorAttrs: {disabled: true}
}

// usage, e.g. area-of-influence-form-model.js above:
 source: this._primarySourceSchemaItem(),
 type: {
   type: 'Text',
   label: …
```

xavijam: :+1:

viddo: For point-in-polygon I see that we need more custom options, so updated prev example a little bit (the `options: …` part)

viddo:
A little comment on an improvement, otherwise 👌

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/base-analysis-form-model.js
xavijam: What will `_getSourceOption` provide?

viddo: With the updated example, the same as you have now, I simply suggest to encapsulate the default definition for a source item in the schema (i.e. add a new method `_primarySourceSchemaItem`). 😄

xavijam:
Ok, suggestions applied + a gift:

screen shot 2016-07-02 at 13 43 17

Loading state! 🎉

xavijam:
Merging!

viddo:
Neat!

nobuti:
Osom

javisantana:
:raur:

PR #8412, opened 2016-07-02
Map options

### LINKED ISSUE #8202: Moran cluster -> Moran's I Cluster
description text should be.
"Find areas which have similar values and identify outliers "

### LINKED ISSUE #8302: Add map options
- [x] Change map name.
- [x] Delete map (+ confirmation modal).
- [x] Map metadata modal.
- [x] Map options in the "settings panel" (image below).

screen shot 2016-06-28 at 19 28 37

nobuti:
This PR fixes #8302.

map-options

Map options view is kind of placeholder because we need a better design. Also, we have to make clear which options we are going to include.

cc @xavijam


xavijam:
If save fails, what would happen? How would the error state be?

xavijam:
LGTM! :es:, only the question I've asked above.

nobuti:
If save fails, it shows an error window, but nothing about the error itself. I'm going to parse it and show as part as the message.

nobuti:
screen shot 2016-07-02 at 12 59 19

xavijam:
Put that cockroach in red :joy:, '--negative'.

saleiva:
hey @urbanphes can you check this? I'm not sure if we were planning to keep this in a modal window (I think so) but I'm not 100% sure. In any case, for the announcement date we will go with this. Thanks @nobuti! Great job!

PR #8413, opened 2016-07-02
Fix color.

### LINKED ISSUE #8405: Input text element font color in dark background
When on a dark background (like writing custom SQL) input text field have a white font color, making it unreadable

### STR
1 - Open a map
2 - Click on a layer and go to 'DATA'
3 - Click on the toggle at the bottom to activate the SQL editor, the background should turn to black
4 - Try and rename the layer: The font color is white, as the background is, so there is no way to see what we are writing **(1)**  unless we make a selection **(2)**

**(1)** screen shot 2016-07-01 at 16 17 32

**(2)** screen shot 2016-07-01 at 16 17 41

### what's expected
the font color should be dark

nobuti:
This PR fixes #8405.

cc @xavijam @piensaenpixel


xavijam:
:+1:

PR #8416, opened 2016-07-02
Start using visualization modal URLs

xavijam:
After trying to fix https://github.com/CartoDB/cartodb/issues/8116, I've realized that shared datasets (in an organization) can't be used in the builder, we should take a look after this week. This could be a workaround for now.

cc @CartoDB/frontend.

### DIFF
--- a/lib/assets/javascripts/cartodb3/data/vis-definition-model.js
+++ b/lib/assets/javascripts/cartodb3/data/vis-definition-model.js
@@ -34,6 +34,11 @@ module.exports = Backbone.Model.extend({
     return baseUrl + '/builder/' + this.id + '/embed';
   },
 
+  builderURL: function () {
+    var baseUrl = this._configModel.get('base_url');
+    return baseUrl + '/builder/' + this.id + '/';
+  },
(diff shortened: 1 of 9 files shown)


lib/assets/javascripts/cartodb3/data/vis-definition-model.js
```diff
@@ -34,6 +34,11 @@ module.exports = Backbone.Model.extend({
     return baseUrl + '/builder/' + this.id + '/embed';
   },
 
+  builderURL: function () {
```
nobuti: 👍  we must use this in the share modal.

xavijam:
@jvillarf on-premises tests failing, should we be worried?

jvillarf:
Will take a look :)

jvillarf:
@xavijam again, there was a connectivity issue with registry.npmjs.org, the IP changed... :(

I've added the new IP to the firewall, so you might be able to re-test it.

Javi.

xavijam:
Merging until we start using better URL management.

PR #8418, opened 2016-07-02
Clear styles on continue when custom css is applied

### LINKED ISSUE #7823: [STYLES] Clear and Continue do the same action
cc/ @urbanphes 

screen shot 2016-06-14 at 12 41 02

I believe those options are not the ones proposed on the designs.

nobuti:
This PR fixes #7823.

reset

cc @xavijam


lib/assets/javascripts/cartodb3/editor/style/style-view.js
```diff
@@ -90,15 +90,22 @@ module.exports = CoreView.extend({
...
 
   _cancelApplyStyles: function () {
-    // Pending to update the map.
+    var geom = this._querySchemaModel.getGeometry();
+    var geometryType = geom ? geom.getSimpleType() : 'point';
+
+    this._styleModel.setDefaultPropertiesByType('simple', geometryType);
     this._infoboxModel.set({state: ''});
     this._overlayModel.set({visible: false});
+    this.render();
```
nobuti: I spent a while thinking how to avoid this, only to update the carousel. Observing changes in the `_styleModel`'s type is not an option, because the carousel changes the type.

saleiva:
BUT in that case, CLEAR should go to the last state, not to the absolute default...

xavijam:
:es:

PR #8425, opened 2016-07-03
Fixes autostyling using an older version of the original style

fdansv:
Before:
bef
After:
aft

@javisantana cc @makella


javisantana:
:es:

saleiva:
Ole!

PR #8432, opened 2016-07-04
Apply infobox state when layer hidden.

### LINKED ISSUE #7895: When layer is hidden, we have to show an infobox with message and the opaque view
But before we need this PR merged :).

nobuti:
This PR fixes #7895.

infobox

cc @xavijam


xavijam:
LGTM 🇪🇸  😃

nobuti:
Applied hide/show option in the context menu but now I'm not sure if we need some kind of tip in the UI to show the user the layer is hidden, I feel only the map won't be enough @xavijam @saleiva @urbanphes 

screen shot 2016-07-04 at 11 11 35

xavijam:
With infobox message should be enough.

saleiva:
Yep, merge!

PR #8474, opened 2016-07-04
7881 explicit persist side effects layer deletion

viddo:
Milestone for #8042 , migrating the layer deletion logic to the user-actions.

There are still some (minor) things to fix, but I'd appreciate some feedback on the solution meanwhile.

Pending:
- [x] Check `layer.canBeDeletedByUser` on views to show/hide delete-layer menu item
- [x] Use `userAction.deleteLayer` instead of doing `layer.destroy` directly
- [x] ~~for the sake of simplicity, and my sanity, all layers have a separate analysis. It's not the most optimal solution (unnecessary/duplicated persistance of nodes), but I'd prefer to improve that in isolation later.~~
- [x] keep one analysis per layer, updated methods to take this into account

@alonsogarciapablo review, please?


lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -174,37 +169,27 @@ module.exports = function (collections) {
...
-          layerOnTop.set('order', orderLimit + 1); // Layer that was on top is kept on top
-        }
-
-        // If wanted order is still on top settle on the layer
-        if (isNaN(order)) {
-          order = orderLimit;
-        } else if (order > orderLimit) {
-          order = orderLimit;
+      if (layerOnTop && !layerTypesAndKinds.isCartoDBType(layerOnTop.get('type'))) {
+        layerOnTop.set('order', layerOnTop.get('order') + 1); // persisted on collection.save in success callback below
```
alonsogarciapablo: Right after the layer is added to the collection `order` of all layers will be reset, so I think we could skip this (`at` is all that matters).

lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -174,37 +169,27 @@ module.exports = function (collections) {
...
+        id: nodeIds.next(newLetter),
         tableName: tableName
       });
       analysisDefinitionsCollection.create({analysis_definition: nodeDefModel.toJSON()});
 
       var attrs = createDefaultCartoDBAttrs();
-      attrs.letter = letter;
-      attrs.order = order || 1; // start on top of basemap as fallback
+      attrs.letter = newLetter;
+      attrs.order = at;
```
alonsogarciapablo: I think we could skip this too (see the above comment :point_up:)

lib/assets/javascripts/cartodb3/data/layer-definition-model.js
```diff
@@ -150,8 +149,12 @@ module.exports = Backbone.Model.extend({
     );
   },
 
+  canBeDeletedByUser: function () {
```
alonsogarciapablo: `...ByUser` suggests to me that this has something to do with user permissions... `canBeDeleted`?

viddo: > ...ByUser suggests to me that this has something to do with user…

Yes, that's the intention! Although not _permissions_ (as in feature flags or such) per se, but more the current-state-of-things. It's only intended to be checked for UI actions, internally other layers (e.g. layer-on-top etc.) may be allowed to be deleted, so wanted to make that clearer.

```js
canBeDeletedByUser: function () {
  return this.collection.getNumberOfDataLayers() > 1 && this.isDataLayer() && !this._isAllDataLayersDependingOnAnyAnalysisOfThisLayer();
 +  },
```

lib/assets/javascripts/cartodb3/data/user-actions.js
viddo: No, because until the layer is saved then two layers would have the same order value, by which the collection is sorted, so the new layer could appear on top (even if it isn't).

viddo:

PR #8482, opened 2016-07-05
Fix name.

### LINKED ISSUE #8458: Name of widgets when adding them from the DATA tab is incorrect
We shouldn't prepend the type of widget like we do when adding the widgets from the widgets modal.

nobuti:
This PR fixes #8458.

widgets-name

cc @saleiva


saleiva:
Good good!

PR #8486, opened 2016-07-05
Add MultiSelect component

javierarce:
screen shot 2016-07-05 at 09 57 00

```
      columns: {
        type: 'MultiSelect',
        title: 'what',
        options: this._columnOptions.filterByType(['string', 'number']),
        editorAttrs: {
          showSearch: true
        }
      }
```

CR: @xavijam


saleiva:
OMG! :)

lib/assets/javascripts/cartodb3/components/custom-list/custom-list-collection.js
```diff
@@ -10,7 +10,6 @@ var _ = require('underscore');
  */
```
xavijam: I see so much logic, why not create an extended custom-list-collection file?
multiple-choice-list-collection?

lib/assets/javascripts/cartodb3/components/custom-list/custom-list-item-view.js
```diff
@@ -4,6 +4,7 @@ var _ = require('underscore');
 module.exports = CoreView.extend({
```
xavijam: Same for this one, I'll create an extended item.

lib/assets/javascripts/cartodb3/components/custom-list/custom-list-item-with-checkbox.tpl
```diff
@@ -0,0 +1,17 @@
+<button type="button" class="CDB-ListDecoration-itemLink u-ellipsis
+  <% if (isSelected) { %> is-selected <% } %>
+  <% if (isDestructive) { %>
+  u-errorTextColor
```
xavijam: Don't we miss a <% } %>. It is a bit difficult to read.

xavijam:
It looks great man! But in my opinion, I'd separate the custom-list implementation from the stuff you have built, because the custom-list thing was built like something super simple, easy customizable with the items. With this new logic, my feeling is we should create an extended component.

javierarce:
I thought our current policy was duplicate over extend, but it's a simple change.

xavijam:
Yeah, totally, but IMHO this element is getting more complexity than he should have.

lib/assets/javascripts/cartodb3/components/custom-list/custom-list-item-with-checkbox.tpl
javierarce: It should be fine. It's exactly what we have in custom-list-item.tpl plus the checkbox input.

xavijam:
❤️ 💋 😍

javierarce:
Ok, refactored. Thanks!

xavijam:
ready to merge! :)

javierarce:
Thanks!

PR #8489, opened 2016-07-05
Log error instead of trowing it

alonsogarciapablo:
Throwing the error was preventing the UI from being rendered.

Also, I bumped internal deps as part of this PR.

@xavijam easy peasy... 😄


xavijam:
🎉

PR #8495, opened 2016-07-05
Sync dataset info (dataset and builder)

xavijam:
screen shot 2016-07-05 at 12 03 19
screen shot 2016-07-05 at 12 03 26
screen shot 2016-07-05 at 12 05 38

- Shows sync info in the builder if layer source is synced.
- Display sync info and options in the dataset view if so.
- Let change sync options in dataset and builder.

cc @piensaenpixel @nobuti


app/assets/stylesheets/editor-3/sync-info.css.scss
```diff
@@ -0,0 +1,57 @@
...
+  @include align-items(center);
+}
+.SyncInfo--separator {
+  position: relative;
+  margin-bottom: 24px;
+  padding-bottom: 18px;
+
+  &::after {
+    position: absolute;
+    right: -24px;
```
xavijam: supper unhappy with negative margins, but impossible right now avoiding this in the current builder.

PR #8532, opened 2016-07-06
Move test from CLI to browser-based test suite

viddo:
Due to the implicit require of cartodb.js that makes it fail in the non-browser env.

cc @javisantana


viddo:
Travis fails due to npm timetouts, our own CI worked though, so merging

PR #8538, opened 2016-07-06
Don't make explicit requests

xavijam:
- Imports and geocodings requests disabled cc @nobuti.
- Don't make table fetch within analysis definition node source model cc @viddo.

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor.js
+++ b/lib/assets/javascripts/cartodb3/editor.js
@@ -148,8 +148,9 @@ deepInsights.createDashboard('#dashboard', vizJSON, {
   });
 
   var backgroundPollingModel = new BackgroundPollingModel({
-    showGeocodingDatasetURLButton: true,
-    importsPolling: true
+    showGeocodingDatasetURLButton: false,
+    geocodingsPolling: false,
+    importsPolling: false
   }, {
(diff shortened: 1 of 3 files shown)


viddo:
👍

lib/assets/javascripts/cartodb3/editor.js
```diff
@@ -148,8 +148,9 @@ deepInsights.createDashboard('#dashboard', vizJSON, {
   });
 
   var backgroundPollingModel = new BackgroundPollingModel({
-    showGeocodingDatasetURLButton: true,
-    importsPolling: true
+    showGeocodingDatasetURLButton: false,
+    geocodingsPolling: false,
```
nobuti: I think these properties are the default for the model https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/data/background-importer/background-polling-model.js#L17-L18 but it's ok to be explicit

xavijam:
Merging due to our frontend tests are passing.

lib/assets/javascripts/cartodb3/editor.js
xavijam: Yep, let's leave them there.

PR #8549, opened 2016-07-06
Migrated export modal and adding export-data option in builder

xavijam:
Basically using the same export-modal we had, just replacing the new modals and done.

screen shot 2016-07-06 at 11 48 10
screen shot 2016-07-06 at 11 48 15
screen shot 2016-07-06 at 11 48 23
screen shot 2016-07-06 at 11 48 49

cc @nobuti


xavijam:
Hadsfaslkdjfalñsdfj, I forgot to translate modal template strings, on it.

xavijam:
Frontend tests passed, merging!

PR #8565, opened 2016-07-06
Update npm deps

javitonino:
cc @xavijam


javitonino:
Please @javierarce have a look at those tests, after updating npm dependencies (including cartocolors)

PR #8573, opened 2016-07-06
Rename and reorganize analyses

andrewxhill:
working PR to organize a bit the way analyses are displayed in the builder


javierarce:
👍

javierarce:
Frontend green, so merging.

PR #8575, opened 2016-07-06
Fixes format params for the merge analysis

### LINKED ISSUE #8572: Merge analysis should send an empty array of columns in case none of them are selected
Otherwise the analysis will select all the datase columns.

javierarce:
This PR fixes two params that weren't being correctly sent:

- `source_gometry` now correctly indicates if the user wants to use the left or right source in the analysis.
- `left_source_columns` and `right_source_columns` has a default empty value (`[]`) to indicate that the user doesn't want to include any column from those sources (not sending anything would mean that the user wants to include every column).

Closes #8572


javierarce:
CR: @xavijam 🙏

xavijam:
:es:

PR #8576, opened 2016-07-06
Don't allow to delete node with ref source

viddo:
E.g. the last node of a layer that depends on another layer, since
otherwise it would be left w/o any data input and would fail

@xavijam CR


lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-view.js
```diff
@@ -13,28 +13,29 @@ module.exports = CoreView.extend({
     if (!opts.analysis) throw new Error('analysis is required');
     if (!opts.analysisFormsCollection) throw new Error('analysisFormsCollection is required');
     if (!opts.viewModel) throw new Error('viewModel is required');
-    if (!opts.layerId) throw new Error('layerId is required');
+    if (!opts.layerDefinitionModel) throw new Error('layerId is required');
```
xavijam: layerDefinitionModel is required

xavijam:
:+1:

viddo:
Frontend tests passing, merging

PR #8594, opened 2016-07-06
Notifier events

nobuti:
This PR implements notifications for:

- Query applied (check)
- Error in query (error)
- CartoCSS applied (check 10s)
- Error in CartoCSS (error)
- Layer added (check 10s)
- Layer deleted (check 10s)
- Importing file (progress)
- Widgets

cc @xavijam


xavijam:
:+1:

PR #8596, opened 2016-07-06
When an analysis is added/removed, table is refreshed

### LINKED ISSUE #8106: Need to refresh to show the data on the table view after doing an analysis
Do an intersection analysis between points and polygons for example,
Click on table view - Data will be the source data
Refresh
Data will be the correct data.

### LINKED ISSUE #8437: After removing all analysis you need to refresh to remove the "read-only-mode" from the table view

xavijam:
Basically, table view will be "re-initialized" each time the layer source has changed, and if the query-schema-model status or query has changed, table will load new data.

Fixes #8437.
Fixes #8106.

CR: @nobuti


xavijam:
Frontend tests passing, merging! 💃

nobuti:
👍

PR #8620, opened 2016-07-07
Update label.

### LINKED ISSUE #8183: change behavior style (data panel)
If you click on the style's link, should go to the style panel (not to the widget).
Also some things:
- Maybe we could add another button to customize the widget. If we add this one, the arrow back should be back to data panel

nobuti:
This PR fixes #8183.

The label is updated, the style part will be discussed in a future issue.

cc @xavijam


xavijam:
:+1:

PR #8627, opened 2016-07-07
Editor modal

xavijam:
cc @noguerol @urbanphes


xavijam:
And good bye avatar :(

Issue #8643, opened 2016-07-07
"Duplicate query" sending wrong params for import


rafatower:
Aside from the weird name to create a dataset from a query;

![image](https://cloud.githubusercontent.com/assets/1914276/16655869/9baa3ee8-445c-11e6-94b7-da5b6cac30e5.png)

It is sending the wrong params to the import API, resulting in a failed import. The correct params are:
```js
{
  "table_name": "your_new_table_name",
  "sql": "SELECT expression"
}
```

rafatower:
related to: https://github.com/CartoDB/cartodb/issues/8640

Issue #8640, opened 2016-07-07
Duplicate dataset fails because new query parameter is sent in request


iriberri:
~~It looks that we are using the name for the new duplicated table in the SQL query to duplicate a dataset, instead of the name of the current user.~~~

Investigating more, it seems that we changed the type of import. I'm seeing old duplicates using `external_table` while now this is using  `query`.

We're now adding a `query` parameter which **should not be there** or the data_import will treat is not as a copy table. See below old and new requests comparison.

This is causing import errors per each table duplication because the table doesn't exist (as the new name instead of the name of the table that is going to be duplicated is used).

**STR**:

* Go to a table
* Click on duplicate dataset
* You get an import error:

  ![image](https://cloud.githubusercontent.com/assets/1730320/16653394/d4d57004-4450-11e6-8ae5-2f4b8e0f4a20.png)

In another account outside team, **duplicating a table is creating an import with type `external_table` while in team the type is `query`**.

**Log**:

````
{"item":{"data":{"data_source":"","*******data_type":"query"********,"table_name":"twitter_catalansvote9n","state":"failure","success":false,"logger":"xxxxxx","updated_at":"2016-07-07T12:40:50+00:00","created_at":"2016-07-07T12:40:47+00:00","error_code":8003,"queue_id":null,"tables_created_count":null,"table_names":null,"append":false,"migrate_table":null,"table_copy":"twitter_catalansvote9n_copy","from_query":null,"synchronization_id":null,"id":xxxxxxxx,"table_id":null,"user_id":"xxxxxxxx","service_name":"public_url","service_item_id":null,"stats":"{}","type_guessing":true,"quoted_fields_guessing":true,"content_guessing":false,"server":"xxxxxxx","host":null,"resque_ppid":1513,"upload_host":xxxxxxx,"create_visualization":false,"visualization_id":null,"user_defined_limits":"{\"twitter_credits_limit\":0}","import_extra_options":null,"original_url":"","privacy":null,"cartodbfy_time":0.0,"http_response_code":null,"rejected_layers":null,"runner_warnings":null},"log":"2016-07-07 12:40:50 UTC: Running on server xxxxxxxx with PID: 26897
2016-07-07 12:40:50 UTC: from_table()
2016-07-07 12:40:50 UTC: import_from_query()
2016-07-07 12:40:50 UTC: Going to set merge error with code 8003
[...]

PR #8664, opened 2016-07-07
Fixing dataset duplication

xavijam:
Fixes https://github.com/CartoDB/cartodb/issues/8643.
Fixes https://github.com/CartoDB/cartodb/issues/8640.

Thanks @iriberri and @rafatower.


rafatower:
:flushed: I would've fixed it myself if I knew where

:+1: 

Thanks a lot! :smile:

Issue #8586, opened 2016-07-06
Zero transparency for point datasets


jsanz:
Steps to reproduce:

- Create a new map and add a points dataset
- Style with a stroke color and width enough
- Style fill with any color and move the transparency slider to zero
- The transparency is not applied
- Move to a bigger value like 0.01 and the transparency will be applied

![2016-07-06_19 40 54_selection](https://cloud.githubusercontent.com/assets/188264/16628464/2b3787ba-43b2-11e6-8eab-2dab2baa3d19.png)

![2016-07-06_19 41 13_selection](https://cloud.githubusercontent.com/assets/188264/16628473/308fcb3c-43b2-11e6-9927-67431b03d37b.png)

This only happens with points, polygon layers work as expected.

xavijam:
For me it is working :S, the only thing is that the color-picker, once you select 0 opacity, if you open it again, it resets it to 1. Already fixed in https://github.com/CartoDB/cartodb/issues/8586

jsanz:
OK I'll check again whet it's deployed. The problem was that setting 0 removed the CartoCSS line so I guess when you come back it sets again to 1 as the default value.

Thanks!

PR #8703, opened 2016-07-08
If opacity was 0, color picker set it as 1 when it was initialized

xavijam:
Fixes https://github.com/CartoDB/cartodb/issues/8586.

CR: @javierarce


javierarce:
👍

PR #8705, opened 2016-07-08
fix align buttons

piensaenpixel:
- Fix the align buttons in add layer and add widget modal.

image


xavijam:
:+1:

PR #8713, opened 2016-07-08
[WIP] Default styling

javisantana:


saleiva:
VAMOS NENO!

javisantana:
cc @makella I've created a default-cartography.json file so you can edit defaults. I'll add more there

lib/assets/javascripts/cartodb3/editor/style/style-converter.js
```diff
@@ -2,7 +2,31 @@ var _ = require('underscore');
 
 var CONFIG = {
   HEATMAP_IMAGE: 'url(http://s3.amazonaws.com/com.cartodb.assets.static/alphamarker.png)',
-  GENERIC_STYLE: '// points\n [mapnik-geometry-type=point] {\n    marker-fill: #FF6600;\n    marker-opacity: 1;\n    marker-width: 12;\n    marker-line-color: white;\n    marker-line-width: 3;\n    marker-line-opacity: 0.9;\n    marker-placement: point;\n    marker-type: ellipse;marker-allow-overlap: true;\n  }\n\n //lines\n [mapnik-geometry-type=linestring] {\n    line-color: #FF6600; \n    line-width: 2; \n    line-opacity: 0.7;\n  }\n\n //polygons\n [mapnik-geometry-type=polygon] {\n    polygon-fill:#FF6600;\n    polygon-opacity: 0.7;\n    line-opacity:1;\n    line-color: #FFFFFF;\n   }',
+  GENERIC_STYLE: [
+    '[mapnik-geometry-type=point] {',
```
alonsogarciapablo: Also, why do we have to define these default styles in two different places instead of having them in a single place? Are there any differences between the two?

makella:
ohhh great... thank you.

what do i do now?

lib/assets/javascripts/cartodb3/editor/style/style-converter.js
javisantana: this should not be used anymore, let me remove it

PR #8715, opened 2016-07-08
use here for trade areas

javisantana:
cc @saleiva
CR: @xavijam


xavijam:
That's it?

saleiva:
Please, let's check the error

javisantana:
that's from the frontend side, the error depends on backend

PR #8736, opened 2016-07-11
Shows Intact & Dissolve options when doing a buffer analysis

### LINKED ISSUE #7502: [ANALYSIS] Intact & Dissolve should be also present when doing a buffer and not only an area of influence
Attaching screenshot. That boundaries options should also appear when doing by distance and not only by time

screen shot 2016-05-24 at 21 03 05

cc/ @xavijam

javierarce:
Closes #7502


javierarce:
CR: @viddo

viddo:
👍

viddo:
Have some failing tests though

javierarce:
Thanks! Yes, the failing tests are those that you fixed some minutes ago. I just merged with master.

Issue #8695, opened 2016-07-08
Fix styles issue on Widget Style panel


saleiva:
![screen shot 2016-07-08 at 12 16 06](https://cloud.githubusercontent.com/assets/132149/16684527/cbfe71e6-4505-11e6-978c-bbb7d474dc45.png)

Issue #8436, opened 2016-07-04
Weird bullet on the size styling tool for labels


saleiva:
![screen shot 2016-07-04 at 11 54 41](https://cloud.githubusercontent.com/assets/132149/16557099/2af6bf94-41de-11e6-99e5-79c969cc5e51.png)

PR #8740, opened 2016-07-11
8695 fix form aligment

piensaenpixel:

Fix https://github.com/CartoDB/cartodb/issues/8695

it's the only solution I found for fix this weird align in all the cases

image

image

Also fix this issue

Fix https://github.com/CartoDB/cartodb/issues/8436

image


javierarce:
👍

PR #8726, opened 2016-07-11
Dataset metadata

nobuti:
This PR fixes #8687 and #8494.

metadata-dataset

cc @xavijam

PR #8741, opened 2016-07-11
CartoDB.css bug

xavijam:
#### History
- All of this started having `cartodb.css` file to be committed each time any developer had run `grunt` in their machine.
- That was provoked by a change we made [here](https://github.com/CartoDB/cartodb/blob/55cc37d62899d6b54057ddfaaac6bd60dcfc971d/vendor/assets/stylesheets/cartodb.css), rebranding the file, but this file is auto-generated using CDB submodule each time grunt is run, so that was the reason why it always appeared like edited.
- We removed the file by mistake (my bad for not putting both eyes in that change) [here](https://github.com/CartoDB/cartodb/pull/8726/files) in order to not have that file to commit each time.
- It was ok, because each time grunt is run, it tries to generate that file from the CDB submodule, but it fails:

```js
Running "cdb" task
cartodb.js not updated (due to cd lib/assets/javascripts/cdb; npm install
grunt-contrib-watch@0.5.3 node_modules/grunt-contrib-watch…
```
Due to that, it gets the last file it is present in vendor/stylesheets/cartodb.css in order to generate the proper stylesheet for the editor.
- And without that file, old maps were not rendering properly our custom elements.

#### Solution
- Get back the needed `cartodb.css` file, without the rebranding, that should happen in CartoDB.js#develop (cc @alonsogarciapablo), and when it is done, they will be visible in the editor `cartodb.css` file.
- We should fix the problem of the `cartodb.css` file generation in the grunt task.

cc @CartoDB/frontend @javitonino @rochoa

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,6 +56,7 @@ Vagrantfile
 /node_modules
 .grunt/*
 vendor/assets/javascripts/cartodb.*
+vendor/assets/stylesheets/cartodb.*
 public/test_support.csv
 public/test_guess_country.csv
 doc/manual/build
--- a/.gitignore
+++ b/.gitignore
@@ -56,7 +56,6 @@ Vagrantfile
 /node_modules
 .grunt/*
 vendor/assets/javascripts/cartodb.*
-vendor/assets/stylesheets/cartodb.*
 public/test_support.csv
 public/test_guess_country.csv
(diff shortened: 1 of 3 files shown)


xavijam:
Merging, frontend tests passing

PR #8746, opened 2016-07-11
Duplicate action.

### LINKED ISSUE #8451: Add 'duplicate map' option in the map context menu
There is an example of the behaviour in the dataset view, duplicating it.

nobuti:
This PR fixes #8451.

duplicate

cc @xavijam.


xavijam:
:es:

PR #8748, opened 2016-07-11
Fix creation of mobile apps #8747

### LINKED ISSUE #8747: Cannot create mobile apps in org accounts

javitonino:
Fixes #8747

Without passing the user, the URL generated is for the user (not the org), causing a redirection which fails.


javitonino:
CR @gfiorav

gfiorav:
You _could_ use the new syntax notation, but OK

📦 when ready!

javitonino:
:rocket: :negative_squared_cross_mark: Thanks!

PR #8749, opened 2016-07-11
8571 add query schema model cachebust

### LINKED ISSUE #8571: Wrong style when changing geometry type on an analysis
STR

- Create an intersection analysis between polygons and points, selecting the Filter option.
- This will render a map with points.
- Edit the analysis parameters to make an aggregation analysis.
- This will render a map with polygons.
- Edit the analysis parameters (again) to make a filter analysis.
- This will render a map with points.
- Edit the analysis parameters (again) to make an aggregation analysis.
- This will render a map with points - ERRR!

viddo:
Fixes #8571 with a tmp solution until this can be fixed on the maps API.  Requests will have an additional query string param that will change value for every request, e.g. `?cachebust=1468243056998&…`

edit: also set default style as _points_ if could not find any geometry

cc @javisantana @rochoa

### DIFF
--- a/lib/assets/javascripts/cartodb3/data/query-schema-model.js
+++ b/lib/assets/javascripts/cartodb3/data/query-schema-model.js
@@ -61,6 +61,7 @@ module.exports = Backbone.Model.extend({
     opts.data = _.extend(
       opts.data || {},
       {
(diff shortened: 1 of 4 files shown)


javisantana:
:es:

PR #8752, opened 2016-07-11
Replaces label title in the georeference form

javierarce:
Replaces 'Column' label with 'Type'. 

It also creates a "measure-by" entry in the localization file that was needed.

### DIFF
--- a/lib/assets/locale/en.json
+++ b/lib/assets/locale/en.json
@@ -1147,7 +1147,7 @@
         "tune-analysis": "Tune your analysis",
         "tune-centroid": "Tune your centroid",
         "tune-clusters": "Tune your clusters",
-        "type": "Measure by",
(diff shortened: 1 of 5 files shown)


javierarce:
CR: @xavijam 
CC: @urbanphes

lib/assets/locale/en.json
```diff
@@ -1147,7 +1147,7 @@
         "tune-analysis": "Tune your analysis",
         "tune-centroid": "Tune your centroid",
         "tune-clusters": "Tune your clusters",
-        "type": "Measure by",
+        "measure-by": "Measure by",
```
javierarce: In case you were wondering, there was already a "type" entry on the file.

xavijam:
:+1:

PR #8754, opened 2016-07-11
Add used org quota to superadmin api

javitonino:


javitonino:
Please CR @gfiorav. This is added to avoid querying each user for its quota from superadmin, when we only need the organization total.

gfiorav:
📦 !

PR #8755, opened 2016-07-11
Filter existing visualizations to avoid visualization_id collide

ethervoid:
We should previously filter the visualizations to avoid visualization_id collision problem

### DIFF
--- a/lib/tasks/explore_api.rake
+++ b/lib/tasks/explore_api.rake
@@ -272,35 +272,34 @@ namespace :cartodb do
     end
 
     def insert_new_visualizations_at_user(most_recent_created_date, most_recent_updated_date)
-
       puts "INSERTING NEW CREATED"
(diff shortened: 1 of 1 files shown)


ethervoid:
@iriberri CR this please

lib/tasks/explore_api.rake
```diff
@@ -272,35 +272,34 @@ namespace :cartodb do
     end
 
     def insert_new_visualizations_at_user(most_recent_created_date, most_recent_updated_date)
```
iriberri: Why are `most_recent_created_date` and `most_recent_updated_date` defined if you're only using `most_recent_created_date` inside the function? Is this a typo?

ethervoid: This was this way since the beginning but i has more sense to use the update

PR #8760, opened 2016-07-11
Added geometry type when query schema is fetched.

### LINKED ISSUE #7490: Indicate geometry output on layer avatar representation
screen shot 2016-05-24 at 17 01 47

Based on the simple type (points, polygons, lines).
It's available through the querySchemaModel, but need to take into account that the geometry type changes over time and might not be available (e.g. if analysis fails for whatever reason).

### LINKED ISSUE #8827: Start storing geometry type for analysis nodes
So for example, if a new node is added, when the related query-schema-model is fetched, we should take the geometry type and store it in the node model.
We have to have clear that this is something that could not be present always.

It's motivated to be able to render the geometry icon of a layer, w/o having to query the schema/rows every time the browser is reloaded, see #7490 

cc @CartoDB/frontend because it is kind of important addition.

nobuti:
Fixes #7490
Fixes #8827 

geometry-type

cc @xavijam @viddo


lib/assets/javascripts/cartodb3/editor/layers/layer-views/data-layer-view.js
```diff
@@ -92,6 +117,26 @@ module.exports = CoreView.extend({
     return this;
   },
 
+  _getGeometryTemplate: function (geometry) {
+    var template;
+
+    switch (geometry) {
+      case 'line':
+        template = geometryLinesTemplate;
```
xavijam: return whatever in each line and done no?

xavijam:
This is great but I'm not sure if we can avoid those petitions from the beginning. Imagine if we have a map with 10 layers, doing a request per each one only for guessing their geometries. I don't know, it shouldn't be so harmful but in any case pinging @viddo and @javisantana. What do you think guys?

lib/assets/javascripts/cartodb3/editor/layers/layer-views/data-layer-view.js
```diff
@@ -39,23 +43,41 @@ module.exports = CoreView.extend({
...
+    this._querySchemaModel = nodeDefModel.querySchemaModel;
+
     this.viewModel = new Backbone.Model();
 
     this.listenTo(this.model, 'change', this.render);
     this.listenToOnce(this.model, 'destroy', this._onDestroy);
     this.listenTo(this.viewModel, 'change:collapsed', this.render);
+    this._querySchemaModel.on('change:status', this.render, this);
+
+    if (!this._querySchemaModel.isFetched()) {
```
viddo: `canFetch()` is probably better (checks the isFetched internally)

viddo:
> This is great but I'm not sure if we can avoid those petitions from the beginning. Imagine if we have a map with 10 layers, doing a request per each one only for guessing their geometries. I don't know, it shouldn't be so harmful but in any case pinging @viddo and @javisantana. What do you think guys?

I think we should find some way to memoize/save the geometry, so we don't have to request it again every time the user enters the builder. Don't have an easy answer to where/how I think it should be solved right now, though…

xavijam:
Yes, I agree, we need to store the geometry type in the nodes. If we do so, we would avoid those unnecessary requests from the beginning. Let's discuss in the ticket https://github.com/CartoDB/cartodb/issues/8827.

saleiva:
oooooh

xavijam:
It will be done @saleiva, but before that we need to implement https://github.com/CartoDB/cartodb/issues/8827.

viddo:
I'll fix #8827 as part of this PR

viddo:
Due to the async nature of the query-schema-model, and since a page load of the Builder shouldn't provoke any state-changes to the viz, it will be persisted _eventually_ (when node/layer is persisted), and restored when the query-schema-model is created.
@xavijam can you review?

xavijam:
yay! :es:

Issue #8616, opened 2016-07-07
Underline help tooltips in forms


javierarce:
The ? should have a little underline to distinguish it from the label text.

![screen shot 2016-07-11 at 12 13 45](https://cloud.githubusercontent.com/assets/132146/16727367/0cb7718a-4761-11e6-99b9-5b28a1f91809.png)

PR #8761, opened 2016-07-11
fix align with help icon

piensaenpixel:
Fix - https://github.com/CartoDB/cartodb/issues/8616

image


javierarce:
👍

PR #8782, opened 2016-07-12
Torque fixes 3

fdansv:
Sorry this has taken so long to come out for such a tiny changeset :(

* Animated layers will stop getting their styles screwed up when there is a reply from the tiler.
* Timeseries widget isn't regenerated with each attribute, only when there has been a change of animated state or a change in the column attribute.
* Turbo carto now works, as we are sending `value` instead of the column itself.
* Minimum number of steps is now 1. Being able to set them to zero made torque crash.

cc @javisantana


fdansv:
I think some merge went wrong at some point, let me fix those changes

xavijam:
This PR is outdated, closing, @fdansv 💋 .

Issue #8229, opened 2016-06-27
Alignement details


saleiva:
![screen shot 2016-06-27 at 15 20 03](https://cloud.githubusercontent.com/assets/132149/16381014/ad4fa334-3c7a-11e6-90f7-b34381b1e584.png)

PR #8787, opened 2016-07-12
8229 fix aligments

piensaenpixel:
Fix https://github.com/CartoDB/cartodb/issues/8229
cc/ @urbanphes 

image

Also put 'Add widget' button in the right position
image


javierarce:
👍

Issue #8589, opened 2016-07-06
MapView Toggle Button Hides CartoDB Logo on Basemap


michellechandra:
Steps to reproduce:

1) Create a map using new editor
2) Select a data layer to view options for that layer (data / style options for the layer)
3) MapView toggle button hides cartodb basemap logo on map

![screen shot 2016-07-06 at 11 34 42](https://cloud.githubusercontent.com/assets/5423288/16630097/a9e3e95e-436d-11e6-811d-1176e856e194.png)

xavijam:
Yeah, it is on purpose, there is no other place to positionate it for the moment, it is a ticket for @urbanphes.

xavijam:
It has to be, because if not logo in the map options could not have any sense.

saleiva:
Well, I agree is not the best solution, but until we find a better solution
for it I'd rather prefer this than looking like an error

2016-07-06 20:56 GMT+02:00 Javier Álvarez Medina <notifications@github.com>:

> It has to be, because if not logo in the map options could not have any
> sense.
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/CartoDB/cartodb/issues/8589#issuecomment-230870296>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAIENZzALzJzCyPeU6_2F5piGb5_-D1bks5qS_pngaJpZM4JGYEU>
> .
>



-- 
Sergio Alvarez Leiva

michellechandra:
It appeared as though an error to me when I saw it, especially as the position of the toggle view button changes when you select Table View (it is positioned to the right side in Table View and in the middle in Map View). It should stay in the same place ideally.

urbanphes:
When you are on (Map or table) active element we will change this state with fill color without opacity. The color will be #2E3C43.
![image](https://cloud.githubusercontent.com/assets/14546701/16723855/e13cb094-474f-11e6-9eb7-aaf719f3b12d.png)

<br><br>
We need review the icons
![image](https://cloud.githubusercontent.com/assets/14546701/16723868/f3e66456-474f-11e6-9020-38904310e15a.png)

PR #8792, opened 2016-07-12
change background color and fix svg

piensaenpixel:
Fix - https://github.com/CartoDB/cartodb/issues/8589
image

cc/ @urbanphes


javierarce:
👍

PR #8795, opened 2016-07-12
Use userActions when widget is added from data view

xavijam:
We were not using userActions within data-view panel and if you create a new map with one layer, and try to add a widget from that panel, map instantiation fails due to node is not saved before saving the widget.

CR: @viddo 

cc @saleiva @nobuti


viddo:
👍

xavijam:
Tests failing already fixed in local, urgent, merging.

PR #8796, opened 2016-07-12
Don't try to create layer on drag'n'drop analysis node when max_layers limit is reached

### LINKED ISSUE #8758: Don't try to create layer on drag'n'drop analysis node when max_layers limit is reached
Related to #8712, since the limit is finally not removed, we need to handle this when an analysis node is dragged'n'dropped to avoid things breaking badly.

Until there's a better solution - I'll simply abort if the user reached the max_layers quota and show some error message in the notifier.

 cc @urbanphes @saleiva

viddo:
Resolves #8758

@xavijam ok?

### DIFF
--- a/lib/assets/javascripts/cartodb3/data/user-actions.js
+++ b/lib/assets/javascripts/cartodb3/data/user-actions.js
@@ -7,22 +7,25 @@ var layerColors = require('./layer-colors');
 
 /**
  * Coordinate side-effects done on explicit interactions.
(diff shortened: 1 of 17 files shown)


viddo:
Also noticed that the messages in the notifier gets badly wrapped, created an issue for that 👉  #8798

xavijam:
:+1:

xavijam:
UserModel everywhere!

PR #8801, opened 2016-07-12
add template_type in vizjson presenter

matallo:
close template compile issue in embeds described in https://github.com/CartoDB/cartodb/issues/8558

rather than change the default `template_type` in cartodb.js for the new embeds (which for some reason was underscore, instead of moustache, as in old embeds) we are now exposing the template_type (updated at the time it's changed in the builder) for namedmaps

CR @gfiorav (thanks mate) /cc @alonsogarciapablo

### DIFF
--- a/app/controllers/carto/api/vizjson3_presenter.rb
+++ b/app/controllers/carto/api/vizjson3_presenter.rb
@@ -447,7 +447,7 @@ module Carto
       ).freeze
 
       INFOWINDOW_AND_TOOLTIP_KEYS = %w(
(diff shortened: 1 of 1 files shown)


gfiorav:
Have you checked that the default case works fine? i.e, when you don't specify a template type?

gfiorav:
If you also checked old embeds, than it's 📦 after CI is 🍏 

Also, *don't forget to purge varnish cache* please :)

matallo:
sure I did! thanks for the thorough check 😅

alonsogarciapablo:
:clap:

PR #8803, opened 2016-07-12
Allows to specify a custom value in the select views

### LINKED ISSUE #8759: Custom value is missing from the dropdown and needed for example in geocoding analysis
screen shot 2016-07-11 at 17 40 36

javierarce:
See description of the ticket #8759

animation

Here's how to use it:

```js
      country: {
        type: 'Select',
        options: this._columnOptions.filterByType('string'),
        title: _t('editor.layers.analysis-form.country'),
        help: _t('editor.layers.analysis-form.country-help'),
        validators: this._getFieldValidator('country'),
        editorAttrs: {
          showSearch: true,
          allowFreeTextInput: true
        }
      },
```

CR: @xavijam


xavijam:
If we show the "value" added before, I would display it with something especial like `"Italy"`, what do you think?

saleiva:
I'd just Add quotes " ". so we keep readability

saleiva:
BTW @javierarce this looks so goodo

xavijam:
:es:, just the quotes thing.

javierarce:
Like this, right?

screen shot 2016-07-12 at 17 45 20

xavijam:
Yeah!

Issue #8784, opened 2016-07-12
Layer not showing on top after dragging an analysis-node layer below it


ramiroaznar:
## Metro Madrid use case - step 4

1-Add AOI analysis to Layer 1.
2-Drag AOI analysis node (Layer 2) below Layer 1.

![1](https://cloud.githubusercontent.com/assets/5215798/16761187/625ca7f4-481f-11e6-9070-408a75778abd.png)

The expected view should be Layer 1 on top of Layer 2. This is solved after refreshing the browser:

![2](https://cloud.githubusercontent.com/assets/5215798/16761277/b952f1da-481f-11e6-9db6-2ff662e9f6ac.png)

xavijam:
The problem is hitting again, could you (any of both @viddo or @alonsogarciapablo) take care of it?

ramiroaznar:
Not even refreshing works, because the point layer that is on top in the last screenshot is from the analysis node.

viddo:
> The problem is hitting again, …

Just to confirm what problem we're talking about:
@ramiroaznar, from the screenshots it looks like the layers appear in correct order in the layers list on the left, correct? Assuming that's the case:

It looks like it's the same problem as described in issue #8245 and supposedly fixed in PR #8292. @alonsogarciapablo could there be something that changed related to this in cartodb.js since the last fix that provokes this from happening again?

ramiroaznar:
@viddo it is the same problem, but I think it is not solved. Check the following screenshots:

The Layer 1 (black points) should be displayed on top of Layer 2 (orange points and buffers):
![3](https://cloud.githubusercontent.com/assets/5215798/16762270/1cc7618e-4824-11e6-92ea-cb7bd18bb06b.png)

Because when I disable the Layer 2 view, Layer 1 is shown:
![5](https://cloud.githubusercontent.com/assets/5215798/16762274/202e87da-4824-11e6-9dd8-af7e260a1971.png)

alonsogarciapablo:
> @alonsogarciapablo could there be something that changed related to this in cartodb.js since the last fix that provokes this from happening again?

We haven't changed anything lately that could have broken this as far as I know.

saleiva:
Hey, that happened to me also a few days ago (before releasing). Hard to replicate since it doesn't not happen always, but looks something related with the cache @rochoa

alonsogarciapablo:
[...]

PR #8810, opened 2016-07-12
Insert cdb.js layers at the right position

alonsogarciapablo:
Closes https://github.com/CartoDB/cartodb/issues/8784.

Use the actual position of the layerDefModel in the layerDefinitionsCollection instead of using the `order` attribute to determine where the layer should be inserted.

@viddo can you please review? Thx!

### DIFF
--- a/lib/assets/javascripts/cartodb3/data/user-actions.js
+++ b/lib/assets/javascripts/cartodb3/data/user-actions.js
@@ -310,7 +310,6 @@ module.exports = function (params) {
         tableName = prevLayer.get('table_name');
         tableNameAlias = prevLayer.get('table_name_alias');
         attrs = createDefaultCartoDBAttrs();
-        attrs.order = prevOrder;
         attrs.options = _.extend({}, attrs.options, {
           sql: 'SELECT * FROM ' + tableName,
           table_name: tableName,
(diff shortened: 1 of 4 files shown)


lib/assets/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -288,26 +288,15 @@ F.prototype._onLayerDefinitionRemoved = function (m) {
...
 
 var LAYER_TYPE_TO_LAYER_CREATE_METHOD;
-F.prototype._createLayer = function (layerDefModel, opts) {
-  opts = _.extend({
-    at: layerDefModel.get('order')
-  }, opts);
+F.prototype._createLayer = function (layerDefModel) {
   var attrs = JSON.parse(JSON.stringify(layerDefModel.attributes)); // deep clone;
 
-  if (attrs.source) {
```
alonsogarciapablo: I moved this below to check if there's a "create" method for this type of layer first.

alonsogarciapablo:
Retest this please.

lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -310,7 +310,6 @@ module.exports = function (params) {
         tableName = prevLayer.get('table_name');
         tableNameAlias = prevLayer.get('table_name_alias');
         attrs = createDefaultCartoDBAttrs();
-        attrs.order = prevOrder;
```
viddo: why only remove this one? if it's not needed I'd say 🔥  from the other place too

viddo:
👍

lib/assets/javascripts/cartodb3/data/user-actions.js
alonsogarciapablo: Agree, there are at least two other places in this same file where we're setting this. I'll 🔥 them! thx

PR #8811, opened 2016-07-12
Modification queries ready

### LINKED ISSUE #8093: UPDATE queries not working on SQL panel

xavijam:
update-queries

So, if a user applies a modification query, it will be applied and then SQL editor will put default query again, same behaviour than previous editor.

Fixes #8093.

CR: @nobuti.
cc @viddo @alonsogarciapablo @javisantana


lib/assets/locale/en.json
```diff
@@ -1226,6 +1226,9 @@
         "save": "CMD + S to apply your query."
       },
       "notifier": {
+        "sql-alter-loading": "Applying modification query...",
```
xavijam: Not sure which is the best way to name those type of queries, any idea @ohasselblad or @jsanz?

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
```diff
@@ -162,6 +200,23 @@ module.exports = CoreView.extend({
...
+    this._sqlModel.set('content', originalQuery);
+    this._userActions.saveAnalysisSourceQuery(originalQuery, this._nodeModel, this._layerDefinitionModel);
+    this._querySchemaModel.set('query_errors', []);
+    this._clearSQLModel.set({ visible: false });
+    // Force reset schema
+    this._querySchemaModel.set({
+      query: originalQuery,
+      status: 'unfetched'
+    });
+    this._querySchemaModel.fetch();
```
xavijam: We had problems in this case within the builder, so we needed to force the map instantiation and the query-schema-model fetch.

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
```diff
@@ -162,6 +200,23 @@ module.exports = CoreView.extend({
...
+    // Force map instantiation
+    this._vis.instantiateMap();
```
alonsogarciapablo: So far, we've made a pretty good job at keeping references to CartoDB.js models in [deep-insights-integrations.js](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/deep-insights-integrations.js).  @xavijam do you think we could somehow move this to that file (eg: triggering an event here). If we start passing/using internal cdb.js models around, we might loose control of what's going on...

lib/assets/javascripts/cartodb3/data/query-schema-model.js
```diff
@@ -58,19 +56,30 @@ module.exports = Backbone.Model.extend({
     this.set('status', 'fetching');
 
     opts = opts || {};
+    var data = {
```
viddo: I think it would be better to leave this model untouched, it has a clear purpose (analysis results, read-only queries and to get metadata out of that query), vs. the new use-case that is so different (destructive and about modifying the underlying data). Having a separate model would also give you the benefit of having a clear place to implement that "SQLUtils" logic.

I'm thinking something on the lines of:
```js
// e.g. data/query-model.js
var queryModel = new SQL('DELETE FROM table WHERE cartodb_id=2');
queryModel.altersData() // => true 
queryModel.execute({…}); // just to have a more semantic name, can call fetch internally => POST whatever
```

lib/assets/javascripts/cartodb3/dataset/dataset-options/dataset-options-view.js
```diff
@@ -172,17 +175,52 @@ module.exports = CoreView.extend({
...
+      var alterQueryModel = new QuerySchemaModel({
+        query: query
+      }, {
+        configModel: this._configModel
+      });
 
-    this._querySchemaModel.fetch({
-      success: self._saveSQL.bind(self)
-    });
+      this._sqlModel.set('content', query);
```
nobuti: This adds an additional step in the sql history, the alter query and then the default one on success. For me it's a bit odd, that an operation ends up changing the pattern applied.

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
xavijam: I'm ok with that, but in any case I'll need to pass the deep-insights model to data-view.

xavijam: About the event... Isn't it better to just create a method in deep-insights-integration where you can call the instantiation map?

lib/assets/locale/en.json
ohasselblad: What about just saying 'Modifying table...' or 'Updating table...'

xavijam: Because it could be a deletion or a creation...

ohasselblad: 'Modifying table' still works for insert, update, or delete

xavijam: :+1:

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
alonsogarciapablo: +1 to something like ☝️

lib/assets/javascripts/cartodb3/data/query-schema-model.js
xavijam: Auto-answering: yes, it exists and we are still using it in the builder, so perfect.

xavijam:
Suggestions applied! What do you think?

viddo:
:shipit:

lib/assets/locale/en.json
```diff
@@ -1226,6 +1226,9 @@
         "save": "CMD + S to apply your query."
       },
       "notifier": {
+        "sql-alter-loading": "Modifying table...",
+        "sql-alter-error": "Modifying failed",
+        "sql-alter-success": "Modifying applied",
```
ohasselblad: For the last two, 'Table modification failed' and 'Table modification applied' sound better to me

xavijam: We try to not mention table anymore. Not sure if we should add it here.

PR #8812, opened 2016-07-12
Better wrap and break.

### LINKED ISSUE #8798: Notifier messages gets badly wrapped
screen shot 2016-07-12 at 11 02 09
screen shot 2016-07-12 at 14 15 22

Appears on both Chrome and Firefox

nobuti:
This PR fixes #8798.

I made some tests and the issue has no optimal solution that works everywhere, because there are too much inconsistencies between vendors, even the hyphens are inconsistent within the same browser (Firefox doesn't hyphen words with underscores) . IMO this is the best approach: http://codepen.io/nobuti/pen/kXxyxO/

screen shot 2016-07-12 at 17 37 57

Even applying this, we can't assure no weirdos will appear.

cc @piensaenpixel

### DIFF
--- a/app/assets/stylesheets/editor-3/notifier.css.scss
+++ b/app/assets/stylesheets/editor-3/notifier.css.scss
@@ -3,9 +3,12 @@
 @import 'cdb-variables/sizes';
 
 .Notifier {
+  @include css3-prefix(hyphens, auto);
   margin: 0 -25px;
   padding: 0 $baseSize * 3;
   border-top: 1px solid $cSecondaryLine;
+  word-wrap: break-word;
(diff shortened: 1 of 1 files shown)


piensaenpixel:
But we removed all the name's layers so... i don't know if we need now this one, what do you think?

nobuti:
At least we need to get rid of the breaks all thing. The second image from the issue's description doesn't have the layer name.

piensaenpixel:
In the dashboard we are using line-clamp https://css-tricks.com/line-clampin/ i don't know if it's better, but if you see this is ok go ahead :) we can try it.

PR #8823, opened 2016-07-13
8231 fix dragndrop of source node

### LINKED ISSUE #8231: Add the option to create layer from source without analysis
If I want to create a new layer from the node B0 **without adding a new analysis**

selection_008

I drag'n'drop the node, but then, I get a new layer pointing to a new node C0 instead of my original B0!!

selection_009

But, If I add an analysis, and then create the layer

selection_010

cc @xavijam

### LINKED ISSUE #8696: When an analysis has been not applied we should show a button to cancel
Just in the same place we show the DELETE option when it's being applied. 

screen shot 2016-07-08 at 12 18 02

viddo:
Closes #8231 
Fixes #8696 - Show cancel button if new node is not yet persisted.
draganddelete

This enables some new user-cases to work, e.g.:
```
// New layer pointing at source of another layer:
 _______       _______   ______
| A    |      | A    |  | B    | 
|      |      |      |  |      |
| [A0] |  =>  | [A0] |  | [A0] |
|______|      |______|  |______|
```

Also cleaned up remaining code related to `point-in-polygon`, after talking with @rochoa since it's basically deprecated by intersection type, and already removed as an alternative in the UI (and will be finally removed from camshaft-reference for v1), and it helped simplify the analysis-option-model which I needed to touch to solve the primary issue.

@alonsogarciapablo can you review please?
cc @AbelVM @javisantana @saleiva FYI


lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -259,77 +259,101 @@ module.exports = function (params) {
...
-          sql: 'SELECT * FROM ' + tableName,
-          table_name: tableName,
-          table_name_alias: tableNameAlias,
-          letter: prevLetter,
-          color: layerColors.getColorForLetter(prevLetter),
-          source: nodeDefModel.getPrimarySource().id
-        });
-        var newLayerDefModel = layerDefinitionsCollection.add(attrs, {at: prevOrder});
-        analysisDefinitionsCollection.saveAnalysisForLayer(newLayerDefModel);
+        if (nodeDefModel.hasPrimarySource()) {
```
alonsogarciapablo: I think I've suggested this before, but I think extracting some smaller methods for each of the execution paths would make this method easier to understand:

```
if (prevLayer) {
  if (nodeDefModel.hasPrimarySource() {
    this._createLayerForHeadAnalysisNode(...)
  } else {
    this._createLayerForSourceAnalysisNode(...)
  }
} else {
    this._createLayerForIntermediateAnalysisNode(...)
}
```

Not sure the naming is right, but you get the idea...

lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -259,77 +259,101 @@ module.exports = function (params) {
...
+          // Create a new layer which simply points to that source
+          //   _______       _______   ______
+          //  | A    |      | A    |  | B    |
+          //  |      |      |      |  |      |
+          //  | [A0] |  =>  | [A0] |  | [A0] |
+          //  |______|      |______|  |______|
+          tableName = prevLayer.get('table_name');
+          tableNameAlias = prevLayer.get('table_name_alias');
+          attrs = createDefaultCartoDBAttrs();
+          attrs.order = newPosition;
```
alonsogarciapablo: Remember we don't need to set the order explicitly (https://github.com/CartoDB/cartodb/pull/8810/files)

lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -259,77 +259,101 @@ module.exports = function (params) {
...
+          attrs.options = _.extend({}, attrs.options, {
```
alonsogarciapablo: This chunk of code is very similar to [this one on line #310](https://github.com/CartoDB/cartodb/pull/8823/commits/01ca13851960501a1276cffb0ed1c44f23db0623#diff-3aa163d112e234b7529ca92fa7aaf665R310). Do you think it could be worth abstracting it? Perhaps "merge" it with `createDefaultCartoDBAttrs`? Perhaps some other methods could use it...

viddo: Yes, I had it on my radar too. Agree that it could be extracted, will do it after I fixed the other pending stuff in this PR

lib/assets/javascripts/cartodb3/data/user-actions.js
viddo: this was (already) removed in a later commit 🙆  (fixed in the merge w/ master)

lib/assets/javascripts/cartodb3/data/user-actions.js
viddo: Yes, that's a good suggestion. Now that the different use-cases are all implemented I think it's more motivated.

lib/assets/javascripts/cartodb3/data/analysis-definitions-collection.js
```diff
@@ -64,10 +64,16 @@ module.exports = Backbone.Collection.extend({
...
+    if (layerDefModel.isOwnerOfAnalysisNode(nodeDefModel)) {
+      if (analysisDefModel) {
+        analysisDefModel.set('node_id', nodeDefModel.id);
+      } else {
+        analysisDefModel = this.add({analysis_definition: nodeDefModel.toJSON()});
+      }
     } else {
-      analysisDefModel = this.add({analysis_definition: nodeDefModel.toJSON()});
+      if (analysisDefModel) {
+        analysisDefModel.destroy(); // layer's node is owned by another layer, so the existing analysis should be deleted
```
alonsogarciapablo: If the analysisDefModel is owned by another layer and we destroy it... Wouldn't that layer point to a non existing analysisDefModel?

alonsogarciapablo:
Done with the CR. Left some comments/questions and I didn't see anything else that caught my attention. Again, I have to admit that some parts are a bit hard to understand/follow given the complexity of this.

lib/assets/javascripts/cartodb3/data/analysis-definitions-collection.js
viddo: No, let's see if I can explain with an example, let's say we have two layers:
```
 _______   ______
| A    |  | B    | 
|      |  |      |
|      |  | [B1] |
| [A0] |  | [A0] |
|______|  |______|
```

…each layer with its corresponding analysis persisted (i.e. `A0` and `B1`).

Now, when deleting the node `B1` the source of the layer would be `A0`, so when saving the analysis for the layer `var analysisDefModel = this.findAnalysisForLayer(layerDefModel);` would return the analysis for layer B (`B1`), but since the new source of the layer (`A0`) is not owned by that layer it means it's persisted by another one - so it should just be deleted.

Does that answer your question? Do you have any proposal how to clarify this in the code?

lib/assets/javascripts/cartodb3/data/user-actions.js
viddo: Let's continue work/conversation in #8846

viddo:

PR #8825, opened 2016-07-13
Specifies schema name in tablestats query

fdansv:
@rochoa after asking around, apparently we don't have/use the user's schema name in the editor ui, so I opted to get the schema name according to whether the user is within an organisation or not.


lib/assets/javascripts/cartodb3/components/modals/add-widgets/tablestats.js
```diff
@@ -35,7 +36,8 @@ TableStats.prototype = {
       var schemaModel = new QuerySchemaModel(null, {
         configModel: this.configModel
       });
-      schemaModel.set('query', 'with a as (select reltuples from pg_class where relname = \'' + table + '\'), b as (select * from pg_stats where tablename = \'' + table + '\') select * from a,b');
+      var schema = userModel.isInsideOrg() ? userModel.username : 'public';
```
rochoa: Why don't we put a method in userModel to retrieve the schema? So if in the future we improve userModel by getting the real schema from somewhere we will get that improvement here for free.

fdansv: ✅

lib/assets/javascripts/cartodb3/components/modals/add-widgets/tablestats.js
```diff
@@ -22,6 +22,7 @@ TableStats.prototype = {
   },
 
   _getPgStats: function (table, callback) {
+    var userModel = window.userModel;
```
javisantana: I don't think we should use a global variable, this should be pass as a param when creating the object

fdansv: Just changed it so that it's passed through

PR #8828, opened 2016-07-13
Fixes opacity for points

### LINKED ISSUE #8817: Alpha value in the builder styles does not work
If you set Alpha to 0 that won't work, but 0.01 works. See below.
alpha

javierarce:
It looks like [we are not sending the opacity when the value is zero](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/editor/style/style-converter.js#L96) like we are doing [for polygons](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/editor/style/style-converter.js#L118).

CR: @javisantana 

Closes #8817


javisantana:
:es:

PR #8829, opened 2016-07-13
Add header metadata to new embeds

### LINKED ISSUE #8813: Add map preview image to new builder embeds
Basically we are not adding the preview of the map in the new embeds.
Could you take care of it @javitonino?

javitonino:
Fixes #8813


xavijam:
Looks nice! :)

javitonino:
Please CR @gfiorav. I moved all the code to use @vizjson so this benefits from mapcaps<sup>TM</sup>

gfiorav:
Very railish indeed 😄 

_Benefits_ is a strange word to use here 😆  but let's take care of that when it finally comes.

📦  when 🍏

Issue #8368, opened 2016-06-30
Infowindows or pop-ups


fernando-carto:
Small detail (at all critical): now infowindows are called pop-ups, right?
It's little misleading when you see "infowindow" mentioned in this message
![captura de pantalla 2016-06-30 a las 13 22 51](https://cloud.githubusercontent.com/assets/19906145/16486615/2f024560-3ec6-11e6-9667-7e8ff2602d34.png)

saleiva:
Well spotted! Thanks!

PR #8830, opened 2016-07-13
popups: change literal (and move to en.json)

matallo:
close https://github.com/CartoDB/cartodb/issues/8368

CR @javierarce


javierarce:
👍

Issue #7985, opened 2016-06-21
Notifier missing events


saleiva:
Above you can find a first batch of events that should be happening on the Notifier:

- [ ] Appliying query (progress indicator)
- [x] Query applied (check)
- [x] Error in query (error)
- [x] CartoCSS applied (check 10s)
- [x] Error in CartoCSS (error)
- [x] Layer added (check 10s)
- [x] Layer deleted (check 10s)
- [x] Importing file (progress)
- [x] First time add some widget (loader)

- [x] Map instantiation error.
- [x] Analysis enqued
- [x] Analysis in progress (progress)
- [x] Analysis finished
- [x] Analysis error (error)
- [x] TurboCartoCSS error. <- These are displayed in the cartocss editor

@urbanphes missing anyone else?

cc/ @xavijam

xavijam:
Commenting:

- Applying query: not possible to know progress right now, and I don't know if we provide progress with batch API, for the future (@dgaubert @rochoa).
- Query applied: 🆗 
- Error in query: 🆗 
- Analysis in progress: we already have the analysis loader, do we need anything else?

rochoa:
> Applying query: not possible to know progress right now, and I don't know if we provide progress with batch API, for the future (@dgaubert @rochoa).
How is that related to Batch API?

xavijam:
1. In case we decide to apply queries with batch api from the builder in the future (no idea if makes sense or not).
2. Yeah yeah, but we have those errors over the nodes in the UI, not sure if we need to show them in the notifier.

dgaubert:
From Batch API side, it isn't possible to return a progress status, at least, at query level. We can know, how many queries are done and how many are left for any job, but we don't know the cost of each query.

saleiva:
status feedback is never too much, let's add it also to the analysis
progress @xavijam

2016-07-04 11:45 GMT+02:00 Daniel <notifications@github.com>:

> From Batch API side, it isn't possible to return a progress status, at
> least, at query level. We can know, how many queries are done and how many
> are left for any job, but we don't know the cost of each query.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/CartoDB/cartodb/issues/7985#issuecomment-230250441>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAIENQtVZM4ThLYFUghlRBIwxviebqkTks5qSNYxgaJpZM4I6k2v>
> .
>



[...]

PR #8848, opened 2016-07-13
Display a notification when map instantiation fails

alonsogarciapablo:
Part of https://github.com/CartoDB/cartodb/issues/7985. This implements notifications for "map instantiation error".

Basically, when there's an unknown error, the following notification is displayed:

screen shot 2016-07-13 at 18 08 15

Right now, the notification is not closable, but I can change this. The notification will be automatically closed when the next request to the Maps API succeeds. (cc: @saleiva, makes sense? how do you like the copy?)

@nobuti can you CR please? Thx!

### DIFF
--- a/lib/assets/javascripts/cartodb3/vis-notifications.js
+++ b/lib/assets/javascripts/cartodb3/vis-notifications.js
@@ -1,0 +1,41 @@
+var Notifier = require('./components/notifier/notifier');
+
+var NOTIFICATION_ID = 'visNotification';
+
+var VisNotifications = {
+
+  track: function (vis) {
+    vis.on('change:error', this._toggleVisErrorNotification, this);
+    this._toggleVisErrorNotification(vis);
+  },
+
+  _toggleVisErrorNotification: function (vis) {
+    var error = vis.get('error');
+    if (error) {
+      this._showVisErrorNotification();
+    } else {
+      this._hideVisErrorNotification();
+    }
+  },
+
+  _showVisErrorNotification: function () {
+    Notifier.addNotification(this._getNotificationAttrs());
+  },
+
+  _hideVisErrorNotification: function () {
+    if (Notifier.getNotification(NOTIFICATION_ID)) {
+      Notifier.removeNotification(NOTIFICATION_ID);
+    }
+  },
+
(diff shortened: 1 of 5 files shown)


lib/assets/locale/en.json
```diff
@@ -1474,6 +1474,9 @@
     }
   },
   "notifications": {
+    "vis": {
+      "failed": "This map couldn't be rendered. Please try to refresh the page and contact us if the error persists."
```
nobuti: Not sure if we need some styles here, @piensaenpixel?

lib/assets/javascripts/cartodb3/vis-notifications.js
```diff
@@ -0,0 +1,41 @@
...
+    var error = vis.get('error');
+    if (error) {
+      this._showVisErrorNotification();
+    } else {
+      this._hideVisErrorNotification();
+    }
+  },
+
+  _showVisErrorNotification: function () {
+    Notifier.addNotification(this._getNotificationAttrs());
```
nobuti: Should we need to check if the notification is already open?

nobuti:
Just 2 doubts, the rest is 💯

lib/assets/javascripts/cartodb3/vis-notifications.js
alonsogarciapablo: Notification has an ID so the Notifier won't create more than one of these notifications 😄

piensaenpixel:
I think we should add "This map couldn't be rendered" in bold and red.

lib/assets/javascripts/cartodb3/vis-notifications.js
```diff
@@ -32,7 +34,10 @@ var VisNotifications = {
     return {
       id: NOTIFICATION_ID,
       status: 'error',
-      info: _t('notifications.vis.failed'),
+      info: NOTIFICATION_TEMPLATE({
```
alonsogarciapablo: cc: @xavijam

xavijam: 👍

alonsogarciapablo:
screen shot 2016-07-14 at 12 26 34

PR #8850, opened 2016-07-13
Parse error correctly if it's an object.

### LINKED ISSUE #8842: Error not being reported on the UI
The SQL API gives the following error but the UI doesn't reflect anything when doing
`UPDATE table_name SET column_name = round(column_name, 2)`

No function matches the given name and argument types. You might need to add explicit type casts.

nobuti:
This PR fixes #8842.

If the query alters data, the error comes from an ajax request directly, not from query schema model, so we need to handle it properly.

cc @xavijam


xavijam:
Thanks for taking care man! 
I'd do it in the previous step instead of adding things in the error parser.

xavijam:
:kiss:

PR #8862, opened 2016-07-14
7258 rewrite visualization queries on import

### LINKED ISSUE #7258: Rewrite visualization queries on import table rename
If you import a full visualization with existing layers import works but the new visualization uses the existing tables instead of the new ones. For example, importing twice a visualization using a table `X` creates two visualizations using both `X`.

If during import table names are not kept, visualization metadata should rewrite the queries.

javitonino:
Re #7258 

Rewrites table names on import, both in the layer and analyses options. As this can sometimes fail, it tries running the query (with an explain) to ensure it is valid, before actually performing the change.


javitonino:
Please CR @gfiorav

app/services/carto/visualizations_export_persistence_service.rb
```diff
@@ -6,15 +6,16 @@ module Carto
...
+    def save_import(user, visualization, renamed_tables: {})
       apply_user_limits(user, visualization)
       ActiveRecord::Base.transaction do
         visualization.id = random_uuid
         visualization.user = user
 
         ensure_unique_name(user, visualization)
 
-        visualization.layers.map { |layer| fix_layer_user_information(layer, user) }
+        visualization.layers.map { |layer| fix_layer_user_information(layer, user, renamed_tables) }
```
gfiorav: why `.map`?

javitonino: No idea. I didn't notice it when adding a line below. Will change to each as we are ignoring the results.

app/services/carto/visualizations_export_persistence_service.rb
```diff
@@ -101,19 +102,59 @@ def apply_user_limits(user, visualization)
...
     end
 
-    def fix_layer_user_information(layer, new_user)
+    def fix_layer_user_information(layer, new_user, renamed_tables)
       new_username = new_user.username
 
       options = layer.options
       if options.has_key?(:user_name)
-        old_username = options[:user_name]
+        @old_username = options[:user_name]
```
gfiorav: I'd avoid the global!

app/services/carto/visualizations_export_persistence_service.rb
```diff
@@ -101,19 +102,59 @@ def apply_user_limits(user, visualization)
...
         options[:user_name] = new_username
+      end
+
+      if options.has_key?(:table_name)
+        old_table_name = options[:table_name]
+        options[:table_name] = renamed_tables.fetch(old_table_name, old_table_name)
```
gfiorav: `fetch(old_table_name, old_table_name)`

If you coalesce with `nil` in case `old_table_name` is `nil`: why not `renamed_tables[old_table_name]`?

app/services/carto/visualizations_export_persistence_service.rb
```diff
@@ -101,19 +102,59 @@ def apply_user_limits(user, visualization)
...
 
-        # query_history is not modified as a safety measure for cases where this naive replacement doesn't work
-        query = options[:query]
+    def fix_analysis_node_queries(node, new_user, renamed_tables)
+      options = node.options
 
-        options[:query] = rewrite_query_for_new_user(query, old_username, new_user) if query.present?
+      if options && options.has_key?(:table_name)
+        old_table_name = options[:table_name]
+        options[:table_name] = renamed_tables.fetch(old_table_name, old_table_name)
```
gfiorav: Same stuff here

app/services/carto/visualizations_export_persistence_service.rb
javitonino: Yeah, it's very ugly. However I cannot think of a better way to get the old username (the one in the visualization export).

app/services/carto/visualizations_export_persistence_service.rb
javitonino: `old_table_name` cannot be nil here (it checks in the if that the key exists). This avoids modifying a table name that has not been imported (and thus, it does not appear in renamed_tables).

app/services/carto/visualizations_export_persistence_service.rb
gfiorav: Probably it should be passed around... up to you!

app/services/carto/visualizations_export_persistence_service.rb
gfiorav: `renamed_tables` is not a hash?

javitonino: Yes, it is. It maps original table names to the actual ones (after the necessary renaming in case of duplicates). So, say renamed_tables is empty (because this was a metadata-only .carto):

```
renamed_tables[old_table_name] -> nil
renamed_tables.fetch(old_table_name, old_table_name) -> old_table_name
```

I want the second behaviour, so this code doesn't change anything if no renames are specified.

app/services/carto/visualizations_export_persistence_service.rb
javitonino: Yeah, but then I'd have to return this from this methods, which feels kinda weird. Also, as this is called in a loop, I would have to do something like:
`old_username = visualization.layers.map { |layer| fix_layer_user_information(layer, user, renamed_tables) }.compact.last`
which feels even worse to me.

Thinking about including the original owner username in the export...

app/services/carto/visualizations_export_persistence_service.rb
gfiorav: alrighty then!

Not too fan of fetch!

gfiorav:
📦 it when you think it's ready!

app/services/carto/visualizations_export_persistence_service.rb
gfiorav: I don't really see it that bad, honestly 😁

javitonino:
Rewrote the importer without that ugly global, and used a field in the export JSON instead, which makes way more sense. Can you take a second look?

gfiorav:
Looks nicer 👍 

📦 it!

PR #8863, opened 2016-07-14
fix title and large labels

piensaenpixel:
Now when the label is very long the text is hidden, talking with @urbanphes we decided add and ellipsis and a title 

labels


javierarce:
:D 👍

PR #8868, opened 2016-07-14
Added max, min, avg operators to aggregation category

### LINKED ISSUE #7957: Add AVG, MAX, and MIN for aggregation in the category widgets
Only when this ticket is closed:
https://github.com/CartoDB/Windshaft-cartodb/issues/513

MariaCheca:
Fixes #7957 

screen shot 2016-07-14 at 15 20 59

CR @xavijam


xavijam:
:+1:, take into account you are not bumping the package.json version. It is not a problem but it is likely if you deploy your changes you won't see them immediately.

MariaCheca:
You're right, I'll update that :) Thanks!

PR #8873, opened 2016-07-14
Add widget source id validation

gfiorav:
This adds validation so that `source_id` has to be a proper node `natural_id`.

There are 441 legacy widgets without them. They're fixed by visiting the dashboard.


gfiorav:
This got left behind, I'll retake on it on the future and with updated analysis logic.

PR #8875, opened 2016-07-14
Changed text "CartoDB logo" => "CARTO logo"

MariaCheca:
This text appears in the builder map settings view as a checkbox to
select to show or not our logo in the map.

screen shot 2016-07-14 at 16 17 21

CR @xavijam


xavijam:
:+1:

PR #8897, opened 2016-07-15
Dataset view issues

### LINKED ISSUE #6017: Make read only in sync'd tables more explicit
When working with a sync table, it is not obvious that the data type cannot be changed like in other tables. When clicking on the data type, the column editing information appears instead. 

It'd be great to have a little info hover that says 'data types cannot be changed for tables that are sync'd' or something like that.

ref #6015

### LINKED ISSUE #8490: Animated maps: Trails limits
ANIMATED Maps: Trails is set to be between 0 and 10, and in many cases it is great to be able to add another value numerically in the box (like 20-30) for a nicer visual effect. Twitter uses this a lot. 

I propose you take out that limitation, as was the behavior in the Editor.

BTW, when applying trails after the limit error message appears, with a new value between 0 and 10, I get no visible effect when applying a different number there, and the CartoCSS code is not changing from the default Torque code.

### LINKED ISSUE #8658: Adding more than one row to a table only updates the pagination tooltip value once
**STR**:

* Create a new empty dataset
* Click on "add row" several times:

  image

**Current behavior**:
  The "pagination tooltip" updates to "0 to 1" after 5 new added rows. It requires a page refresh for it to show "0 to 5".

**Expected behavior**:
  Each row added to the table would update the pagination tooltip value.

After a refresh this is the result:

  image

### LINKED ISSUE #8668: the_geom and the_geom_webmercator appear as possible values for labels
**STR**

* In a map, try to add labels to one of your layers
* `the_geom` and `the_geom_webmercator` can be chosen among the rest of columns in the table (although nothing seems to be rendered)

  image

**Expected behavior**: 

 `the_geom` and `the_geom_webmercator` should not be options to be used as a label in a map.

### LINKED ISSUE #8743: When data is not georeferenced PREVIEW option in the dataset view should be disabled

### LINKED ISSUE #8860: Change background when you are on table view
Now the background color is #F2F6F9<br>
image

<br><br>
... and should be **#2E3C43**<br>
image

xavijam:
It fixes several problems reported in the table-view element and in the dataset page.

- Adding dark background for table view (only showed in the builder) - fixes #8860.
- Change table-paginator when a new row is added or an old row is removed - fixes #8658.
- Start table-paginator count from 1, not from 0 - commented by @urbanphes.
- Set `is-disabled` class properly when any change happens related with the table-view.
- Make read only state more explicit - fixes #6017.
- Don't display `preview` button in dataset page if there is no geometry and `the_geom_webmercator` is not present - fixes #8743.

Extra balls:
- Changed animated-trails limit to 30 - fixes  #8490.
- Don't let the user choose `the_geom` or `the_geom_webmercator` columns or date type columns in the label attribute, within the styles - fixes #8668.
- Updated dependencies.
- Fixed master frontend tests.

CR: @nobuti (I encourage you to follow commit by commit, each fix is done in a commit).

### DIFF
--- a/lib/assets/javascripts/cartodb3/dataset/dataset-options/dataset-actions-view.js
+++ b/lib/assets/javascripts/cartodb3/dataset/dataset-options/dataset-actions-view.js
@@ -4,7 +4,8 @@ var template = require('./dataset-actions.tpl');
 
 var REQUIRED_OPTS = [
   'mapAction',
-  'previewAction'
+  'previewAction',
+  'querySchemaModel'
 ];
 
 module.exports = CoreView.extend({
@@ -20,11 +21,20 @@ module.exports = CoreView.extend({
       if (!opts[item]) throw new Error(item + ' is required');
       this['_' + item] = opts[item];
     }, this);
+
+    this._querySchemaModel.bind('change:status', this.render, this);
+    this.add_related_model(this._querySchemaModel);
   },
 
   render: function () {
     this.clearSubViews();
-    this.$el.html(template());
+    var geom = this._querySchemaModel.getGeometry();
+    var geomWebmercator = this._querySchemaModel.columnsCollection.findWhere({ name: 'the_geom_webmercator' });
+    this.$el.html(
+      template({
+        hasGeometry: geom && geomWebmercator
+      })
+    );
     return this;
   },
 
--- a/lib/assets/javascripts/cartodb3/editor/style/style-form/style-form-components-dictionary.js
+++ b/lib/assets/javascripts/cartodb3/editor/style/style-form/style-form-components-dictionary.js
@@ -278,10 +278,7 @@ module.exports = {
     var filterFunction = function (item) {
       var columnName = item.get('name');
       var columnType = item.get('type');
-      if (columnName !== 'the_geom' && columnName !== 'the_geom_webmercator' && columnType !== 'date') {
-        return true;
-      }
-      return false;
+      return (columnName !== 'the_geom' && columnName !== 'the_geom_webmercator' && columnType !== 'date');
     };
     return generateSelectByStyleType('labels-attribute', querySchemaModel, filterFunction, styleType);
   },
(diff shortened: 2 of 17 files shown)


lib/assets/javascripts/cartodb3/editor/style/style-form/style-form-components-dictionary.js
```diff
@@ -275,7 +275,15 @@ module.exports = {
   },
 
   'labels-attribute': function (querySchemaModel, styleType) {
-    return generateSelectByStyleType('labels-attribute', querySchemaModel, null, styleType);
+    var filterFunction = function (item) {
+      var columnName = item.get('name');
+      var columnType = item.get('type');
+      if (columnName !== 'the_geom' && columnName !== 'the_geom_webmercator' && columnType !== 'date') {
```
nobuti: return (columnName !== 'the_geom' && columnName !== 'the_geom_webmercator' && columnType !== 'date');

lib/assets/javascripts/cartodb3/dataset/dataset-options/dataset-actions-view.js
```diff
@@ -20,11 +21,20 @@ module.exports = CoreView.extend({
       if (!opts[item]) throw new Error(item + ' is required');
```
nobuti: Not strictly related with this PR, I recommend to apply strict triple equal comparison with `undefined`. Sometimes we want boolean options, and it will fail for a required option.

xavijam: Yeah, totally aggre.

nobuti:
LGTM. I'm going to assume the `provider` thing is going to work well, because I lack of background here to understand what it does mean.

xavijam:
@nobuti: we removed here provider yesterday from area-of-influence model, and there were serveral tests checking that those attirbutes exist.

xavijam:
Merging due to frontend tests are passing with Travis and we have frontend tests broken in master 💋

PR #8910, opened 2016-07-15
Remove some S3 dead code

### LINKED ISSUE #8833: Check if S3Uploader and data repository are dead code
The classes `S3Uploader` and the module `DataRepository::Filesystem::S3` seem to be dead code. Double check and remove if possible.

javitonino:
Closes #8833


javitonino:
Tested to death with all file uploads I could think of: importer, map exporter, marker images, avatars. Nothing broke, so I say let's :fire: it. Please CR @gfiorav

gfiorav:
Well done! 🔥  it!

PR #8913, opened 2016-07-15
Fixing problem with fill-component in Firefox

### LINKED ISSUE #8905: Cannot edit custom color palettes in Firefox
Clicking on a color doesn't do anything:

.

xavijam:
Firefox doesn't understand what `event.toElement` mean (using the older and beloved `event.target`) and the index was always giving -1, that was the reason of the problem.

Fixes #8905.

CR: @javierarce 
cc @javitonino


javierarce:
👍

xavijam:
:kiss:

PR #8916, opened 2016-07-15
Removed trackjs from application_builder layout

xavijam:
cc @javitonino


javitonino:
This only affected the `/builder` and `/dataset` views, so perfect :) :package: it as soon as :green_apple:

xavijam:
Backend tests passed, merging.

PR #8917, opened 2016-07-15
Fix font in invalid format message when importing dataset

javitonino:
Before:
screenshot_20160715_164110

After:
screenshot_20160715_165625

Alternative font size:
screenshot_20160715_165707

This PR comes with the small size (middle screenshot) because it wrapped better, but easy to change it.

--

Also fixes or separator in marker image upload.

Before:
screenshot_20160715_170702

After:
screenshot_20160715_170719


javitonino:
Some frontend :tophat: fixes while testing S3 uploads. Please CR @xavijam

xavijam:
👍

PR #8921, opened 2016-07-15
8712 remove max layers

### LINKED ISSUE #8712: Remove layer limit for builder
Currently set to 4. This shouldn't be a limiting factor anymore. We'll just set the limit to the max always (10).

cc/ @saleiva @xavijam

gfiorav:
Fixes #8712 

Central PR: https://github.com/CartoDB/cartodb-central/pull/1325


xavijam:
We will need to take care of this when @gfiorav is out, @juanignaciosl and @javitonino.

PR #8929, opened 2016-07-18
Refresh map when widgets order changes

### LINKED ISSUE #8892: The widgets lost the order and there is no way to sort them out

cc @saleiva

nobuti:
This PR fixes #8892.

When moving a widget, we force a new map triggering the movedLayer event of the associated layer of the widget, if anyone.

I think this is the easy way to achieve it the goal, but not completely sure about the approach @xavijam @viddo @alonsogarciapablo

### DIFF
--- a/lib/assets/javascripts/cartodb3/data/user-actions.js
+++ b/lib/assets/javascripts/cartodb3/data/user-actions.js
@@ -432,6 +432,14 @@ module.exports = function (params) {
       });
     },
 
+    moveWidget: function (w) {
+      var layerId = w.get('layer_id');
+      var movingLayer = layerDefinitionsCollection.get(layerId);
+      if (movingLayer) {
+        layerDefinitionsCollection.trigger('layerMoved', movingLayer);
+      }
+    },
+
     deleteLayer: function (id) {
       var layerToDelete = layerDefinitionsCollection.get(id);
       if (!layerToDelete.canBeDeletedByUser()) return;
--- a/lib/assets/javascripts/cartodb3/editor/widgets/widgets-view.js
+++ b/lib/assets/javascripts/cartodb3/editor/widgets/widgets-view.js
@@ -58,7 +58,16 @@ module.exports = CoreView.extend({
     }
   },
 
-  _onSortableFinish: function () {
+  _onSortableFinish: function (event, ui) {
+    var widgetDefModel;
+    var $draggedWidgetElement = ui.item;
+    var widgetId = $draggedWidgetElement.data('model-cid');
+
+    if (widgetId) {
+      widgetDefModel = this._widgetDefinitionsCollection.get(widgetId);
+      this._userActions.moveWidget(widgetDefModel);
+    }
+
     var self = this;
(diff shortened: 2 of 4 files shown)


lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -432,6 +432,14 @@ module.exports = function (params) {
       });
     },
 
+    moveWidget: function (w) {
```
xavijam: I'd do more explicit the parameter name, widgetModel?

lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -432,6 +432,14 @@ module.exports = function (params) {
...
+      var layerId = w.get('layer_id');
+      var movingLayer = layerDefinitionsCollection.get(layerId);
+      if (movingLayer) {
+        layerDefinitionsCollection.trigger('layerMoved', movingLayer);
```
xavijam: Why do we send an event that mentions layer when we have touched widgets?
If we need a new map instantiation when a widget is ordered, we can listen widget changes here:
https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/deep-insights-integrations.js#L45

And then, when order changes finishes (maybe a debounce is needed), force a map instantiation with:

https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/deep-insights-integrations.js#L384

What do you think @alonsogarciapablo ?

alonsogarciapablo:
After reading issue and description... I still don't know what the problem is here 😕 ... @nobuti could you please elaborate on what's wrong now?

lib/assets/javascripts/cartodb3/data/user-actions.js
nobuti: Basically because I didn't know how to force a new map instantiation 😁 . Also, as an addition to your proposal, I'd add that we should follow the same approach as we have with layers, getting rid of order and sort in favour of collection's order to render the sidebar view, I think this is the only way we use the `order` parameter (in deep-insights)

nobuti:
@alonsogarciapablo being on the widgets tab pane, when the user drag and drop a widget to reorder them, the map should respond to that action. I guess in some cases the map should be different according to the order of the widgets.

xavijam:
Jum... but the widget order should not affect the map, right?

alonsogarciapablo:
Thanks for the clarification @nobuti. I know understand what's going on here... The Publish button is not re-enabled when widgets are re-ordered because the map is not affected at the cdb.js level and the `reload` event is not being triggered by the `vis` object (which we're using [here](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/deep-insights-integrations.js#L52)).

The truth is... the map wouldn't need to be refreshed in this circumstance, because re-ordering widgets doesn't have any effect on the map/widgets data.

nobuti:
Closing this for useless.

PR #8930, opened 2016-07-18
Update locale

### LINKED ISSUE #8926: When an analysis is not waiting but is running loading message should say "Analysis XX running..."
Right now it just says Analysis waiting

It's important that we differentiate when it's running and when it's enqued/waiting on the UI.

nobuti:
This PR fixes #8926.

cc @saleiva


xavijam:
:es:

PR #8932, opened 2016-07-18
Update GIS StackExchange tag names to CARTO

iriberri:
We updated the tag in GIS SE and made `cartodb` and `carto` synonyms.

http://gis.stackexchange.com/questions/tagged/carto

cc @xavijam


xavijam:
You are pure ❤️ , :es:

iriberri:
The quote fix is looking good in staging @xavijam:

image

xavijam:
🚢 🇮🇹

PR #8948, opened 2016-07-18
Listen to vis in order to change status of notifications.

### LINKED ISSUE #7985: Notifier missing events

nobuti:
This PR fixes #7985.

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/notifier/notifier-item-view.js
+++ b/lib/assets/javascripts/cartodb3/components/notifier/notifier-item-view.js
@@ -19,6 +19,7 @@ module.exports = CoreView.extend({
     if (!opts.editorModel) throw new Error('editorModel is required');
 
     this._editorModel = opts.editorModel;
+    this._vis = opts.vis;
     this.template = this.options.template || Template;
     this._notifierModel = opts.notifierModel;
     this.$el.attr('id', this._notifierModel.get('id'));
@@ -47,6 +48,12 @@ module.exports = CoreView.extend({
 
     this.listenTo(this._editorModel, 'change:edition', this._changeStyle);
     this.add_related_model(this._editorModel);
+
+    // In dataset we don't have vis
+    if (this._vis !== undefined) {
+      this._vis.on('reload', this._triggerVisReload, this);
+      this.add_related_model(this._vis);
+    }
   },
 
   _initViews: function () {
@@ -105,6 +112,14 @@ module.exports = CoreView.extend({
     var status = this._notifierModel.getStatus();
     var action = this._notifierModel.getAction();
     this._notifierModel.trigger('notification:action', action || status);
+  },
+
+  _triggerVisReload: function () {
+    this._notifierModel.trigger('vis:reload');
+  },
+
+  _triggerVisError: function () {
+    this._notifierModel.trigger('vis:error', this._vis.get('error'));
   },
 
   _changeStyle: function () {
--- a/lib/assets/javascripts/cartodb3/components/notifier/notifier-item-view.js
+++ b/lib/assets/javascripts/cartodb3/components/notifier/notifier-item-view.js
@@ -19,7 +19,7 @@ module.exports = CoreView.extend({
     if (!opts.editorModel) throw new Error('editorModel is required');
 
     this._editorModel = opts.editorModel;
-    this._vis = opts.vis;
+    this._visDefinitionModel = opts.visDefinitionModel;
     this.template = this.options.template || Template;
     this._notifierModel = opts.notifierModel;
     this.$el.attr('id', this._notifierModel.get('id'));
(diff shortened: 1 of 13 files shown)


lib/assets/javascripts/cartodb3/components/notifier/notifier-item-view.js
```diff
@@ -19,6 +19,7 @@ module.exports = CoreView.extend({
     if (!opts.editorModel) throw new Error('editorModel is required');
 
     this._editorModel = opts.editorModel;
+    this._vis = opts.vis;
```
alonsogarciapablo: I'm "fighting" against passing the `vis` object around... So far, we've tried to encapsulate the integration between cdb.js/deep-insights.js and the builder in [this file](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/deep-insights-integrations.js) and I think it's been working great... I'm afraid passing `vis` object around and listening to events might make things unmaintainable....

alonsogarciapablo: But I can see how, in this scenario, it's hard to work with the notification given that:

1. notification is displayed when a user action is performed.
2. notification should closed when something happened inside cartodb.js.

nobuti: I can proxy the `vis` events into a new pubsub object, but still we need to pass this object around, I don't know how to avoid this. Ideas?

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
```diff
@@ -189,16 +193,25 @@ module.exports = CoreView.extend({
...
-      closable: true,
-      delay: 10000
+    this._sqlNotification = Notifier.addNotification({
+      id: PREFIX_NOTIFICATION_ID + this.cid,
+      status: 'loading',
+      info: _t('editor.data.notifier.sql-applying'),
+      closable: false
     });
 
+    this._sqlNotification.once('vis:reload', this._updateVisNotification, this);
```
alonsogarciapablo: One think I don't get to understand is why the notificiationItemView has to act as a proxy for this event too... Why not inject the visModel here now that it triggers the `vis:reload` event?

nobuti: Mainly for two reasons:
- I was thinking on to implement the 'autoCloseOnVisReload' as attribute. We have a similar case in the dataset view that  it could be fit.
- The cost of injecting the visModel in this view is higher in terms of bringing the model through a bunch of files.

alonsogarciapablo: Fair points.

lib/assets/javascripts/cartodb3/components/notifier/notifier-item-view.js
```diff
@@ -47,6 +48,12 @@ module.exports = CoreView.extend({
 
     this.listenTo(this._editorModel, 'change:edition', this._changeStyle);
     this.add_related_model(this._editorModel);
+
+    // In dataset we don't have vis
```
alonsogarciapablo: Why does `notifier-item-view` have to do this "re-triggering"?

Given that the `Notifier.init` gets the `visModel` as an option, perhaps `Notifier.addNotification` could ensure that `visModel` is injected into every `notifier-model`... Here's some pseudo-code...

```
// notifier.js

...

addNotification: function (attrs) {
   collection.add(attrs, {
     visModel: this._visModel
   }
}),

...

```

```
// notifier-collection.js

...

model: function (attrs, opts) {
   return new NotifierModel(attrs, {
     visModel: opts.visModel
   });
}

...

```

```
// notifier-model.js

initialize: function () {

  ...

  this._retriggerEvent(this._visModel, 'vis:reload');
  this._retriggerEvent(this._visModel, 'vis:error');

  ...
},

_retriggerEvent: function (model, event) {
  model.on(event, function () {
    this.trigger('event', arguments);
  }, this);
},

...
```

What do you think?

lib/assets/javascripts/cartodb3/components/notifier/notifier-view.js
```diff
@@ -10,6 +10,7 @@ module.exports = CoreView.extend({
     if (!opts.editorModel) throw new Error('editorModel is mandatory');
 
     this._editorModel = opts.editorModel;
+    this._visDefinitionModel = opts.visDefinitionModel;
```
alonsogarciapablo: We probably don't need this anymore, right?

alonsogarciapablo:
👍 (although I'd add a couple of tests to `cartodb3/editor/layers/layer-content-view/data/data-view.spec.js`)

PR #8969, opened 2016-07-19
Fixes related with SQL panel

### LINKED ISSUE #8928: update statements with alias fail
PostgreSQL allows for tablenames to be aliased in `UPDATE` statements, but Carto complains about it. Docs on it here: https://www.postgresql.org/docs/9.5/static/sql-update.html

The following used to work:

```
UPDATE awesome_table As at
SET the_geom = ...
WHERE at.awesome_count > 10
```

But now fails: "Syntax error at or near As"

### LINKED ISSUE #8966: Table view does not update after clicking on "CLEAR" (SQL)
### STEPS

0- Create a map* from a dataset. Go to table view from a layer. Enable the SQL panel.
1- Apply a simple query.

query

2- Click on "CLEAR". It should display the whole dataset (in fact, it shows `select * from table` on the SQL command), but it continues showing the data from the previous query.

clear

You need to click on "APPLY" to get the expected result. 

*This behaviour does not apply to the Dataset view. The Dataset view works fine.

xavijam:
- update statements with alias fail (Fixes #8928): We were not taking into account \r or \n in the sql-utils parser. CR: @javisantana 
- Table view does not update after clicking on "CLEAR" (Fixes #8966): Missing changing status of the query-schema-model as we do in the dataset view. CR: @nobuti

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
+++ b/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
@@ -152,6 +152,7 @@ module.exports = CoreView.extend({
     } else {
       // Parser errors SQL here and saving
       this._querySchemaModel.set({
+        status: 'unfetched',
         query: query
       });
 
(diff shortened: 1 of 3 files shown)


nobuti:
👍

xavijam:
Hey @nobuti, last commit really fixes the problem with the "clear query" behaviour, sorry.

xavijam:
Frontend tests passing, merging 💋

PR #8976, opened 2016-07-19
Fixing problem adding a shared datasets in a map within the builder

### LINKED ISSUE #8954: Check problems creating maps with shared datasets in the builder
- Backend problem creating the viz-json definition, where it is not including the owner user name before the name of the table (https://github.com/CartoDB/cartodb/issues/8974).
- Frontend problems related with table name, not adding the needed ownerUserName before the table name.

xavijam:
So, if a user add a shared dataset, it will work in the new builder.
But if the user creates a new map with any shared dataset, map will not work due to a bug in the vizJSON generation (https://github.com/CartoDB/cartodb/issues/8974).
In any case, fixing this problem, and the other is being managed by @juanignaciosl. 

Fixes #8954
CR: @javierarce

### DIFF
--- a/lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
+++ b/lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
@@ -81,16 +81,22 @@ module.exports = Backbone.Collection.extend({
 
       // Create source attr if it does not already exist
       var sourceId = nodeIds.next(o.letter);
-      var tableNameWithOwner = self._configModel.get('user_name') !== o.user_name ? (o.user_name + '.' + o.table_name) : o.table_name;
 
       if (o.table_name && (!o.source || o.source === sourceId)) {
+        o.table_name = o.table_name && o.table_name.replace(/"/g, '');
+        var tableName = o.table_name;
+
+        if (o.table_name.search(/\./i) === -1) {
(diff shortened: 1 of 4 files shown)


lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
```diff
@@ -84,13 +83,20 @@ module.exports = Backbone.Collection.extend({
       var sourceId = nodeIds.next(o.letter);
 
       if (o.table_name && (!o.source || o.source === sourceId)) {
+        o.table_name = o.table_name && o.table_name.replace(/"/g, '');
+        var tableName = o.table_name;
+
+        if (o.table_name.search(/\./i) === -1) {
+          tableName = o.user_name && self._configModel.get('user_name') !== o.user_name ? (o.user_name + '.' + o.table_name) : tableName;
```
javierarce: I would consider to improve the legibility of this oneliner:
```
if (o.table_name.search(/\./i) === -1 && o.user_name && self._configModel.get('user_name') !== o.user_name) {
   tableName =  o.user_name + '.' + o.table_name;
}
```

Even though the condition is also long, I find it's easier to read than the one with the ternary operator.

javierarce:
Other than my suggestion, 👍  and 💃

xavijam:
:kiss:

xavijam:
Frontend tests passed, merging 💋 .

PR #8995, opened 2016-07-19
reverse ramps

javisantana:
cc @xavijam


xavijam:
:es: why do we want this?

javisantana:
@xavijam the colors are not properly sorted

PR #9008, opened 2016-07-20
add check for visualization_id

gfiorav:
Previous behaviour checked for `visualization_id` presence. I thought it was overkill. It's not. There are a lot of `data_imports` with no `visualization_id` (only data imports?).

Anyway, this restores the previous behaviour and stops Rollbar verbosity.

CR please @javitonino


javitonino:
Looks :package: ready to me!

PR #9009, opened 2016-07-20
Define "carto" dependency as git branch to freeze it at shrinkwrap

rochoa:
Using a URL like `https://github.com/CartoDB/carto/archive/master.tar.gz` means it will get downloaded each time `npm install` happens, even if `carto` dependency is frozen at `npm-shrinkwrap.json` file it will redownload whatever is on `master` branch.

However using a `"username/repo#branch"` (or `git://...`)  pattern allows npm to freeze the commit at `npm-shrinkwrap.json` so you get the same version afterwards.

Just to let you know: we also have `cartocolor` development dependency declared as we had `carto` dependency. But I'm not that familiar with the usage of that dev dependency, so please, consider reviewing that situation.

cc @xavijam @javisantana

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1087,8 +1087,8 @@
     },
     "carto": {
       "version": "0.15.1-cdb1",
-      "from": "https://github.com/CartoDB/carto/archive/master.tar.gz",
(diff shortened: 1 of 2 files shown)


xavijam:
:es:

xavijam:
The question is, why did we decide that?

PR #9014, opened 2016-07-20
8963 add exceeded quota event

### LINKED ISSUE #8963: [Metrics] Add new segment event 'Exceeded quota'
Track a new event 'Exceeded quota' into Segment.com every time a new dataset import fails due to not enough available space in disk:

The event will be identified by user id (box id) and include the following attributes:
- username: (user.username)
- email: (user.email)
- plan: (user.account_type)
- organization: (user.organization_user? ? user.organization.name : nil)
- event_origin: 'Editor'
- creation_time: (Time.now.utc)
- import_size: (size of dataset to be imported)

@gfiorav Is it possible to get the last attribute?

gfiorav:
Fixes #8963 

Please CR @javitonino


app/controllers/api/json/tables_controller.rb
```diff
@@ -46,6 +46,7 @@ def create
                         }, 400)
         end
       rescue CartoDB::QuotaExceeded
+        Carto::Tracking::Events::ExceededQuota.new(current_user).report
```
javitonino: I think this will never trigger, as we don't have users with table quotas now (checked).

lib/carto/tracking/events.rb
```diff
@@ -94,13 +91,7 @@ def properties(visualization, views)
 
       class VistedPrivatePage < TrackingEvent
```
javitonino: Typo, if you feel like changing it, do it, but not important.

gfiorav: Nice catch!

app/controllers/api/json/tables_controller.rb
gfiorav: Yeah, totally. Maybe I should remove it? My fear is that it should be shadowing other types of exceded quotas. I'd stick with it.

javitonino: Yeah, not a big deal if it is not called, and maybe we will add the table limit some day (unlikely, but you never know)

javitonino:
Ship it when :green_apple: !

gfiorav:
Fixed some typos! have another 👀 please!

gfiorav:
Fixed some typos! have another 👀 please!

cc/ @jmartinter

javitonino:
LGTM!

gfiorav:
@javitonino : we found another place were the raise is done and also noted that the exceeded quota can be computed from there. Please have another look 👼

app/models/data_import.rb
```diff
@@ -209,6 +215,10 @@ def run_import!
     handle_failure(invalid_cartodb_id_exception)
     self
   rescue => exception
+    if exception.is_a?(CartoDB::Importer2::StorageQuotaExceededError)
+      Carto::Tracking::Events::ExceededQuota.new(current_user).report
```
javitonino: You may be double tracking in some instances. One track where the exception is raised and another one here. Have you tested this, is it an actual issue?

javitonino:
I left a comment. :ship: ready after considering that.

gfiorav:
Ok @javitonino @jmartinter 

I've tested this last version and it works for:
+ Import
+ URL
+ Datasource:
 - Dropbox
 - Google Drive
 - Box

I've basically extracted the quota checks that were performed in various places (downloader, datasource_downloader, runner) and put it in a helper so they all consume the same method (this is were i've taken the opportunity to storage quota exceeded as well).

For that part, we will need a data services person to have a look! @jgoizueta, care to do the honours? :)

Be advised, this will probably be deployed the first week of August when I come back from vacation.

Regarding that last topic: @javitonino if you see it fit and feel confutable with this code, deploy it if you want to once we all agree it's good enough 😁 .

Thanks to all of you!

jmartinter:
Tested some cases and seems ok for me. 
Anyway, let's wait for Guido to return in order to deploy it.

app/models/data_import.rb
```diff
@@ -809,23 +815,30 @@ def get_downloader(datasource_provider)
...
-      log.append "File will be downloaded from #{downloader.url}"
+      metadata_url = metadata[:url]
+      resource_url = (metadata_url.present? && datasource_provider.providers_download_url?) ? metadata_url : data_source
+
+      log.append "File will be downloaded from #{resource_url}"
+
+      http_options = { http_timeout: ::DataImport.http_timeout_for(current_user) }
+      options = {
+        importer_config: Cartodb.config[:importer],
+        user_id: current_user.id
```
jgoizueta: What's the reason of adding this `user_id` option?

app/models/data_import.rb
```diff
@@ -809,23 +815,30 @@ def get_downloader(datasource_provider)
...
-                                       }, log
-      )
-    end
 
-    downloader
+      http_timeout = ::DataImport.http_timeout_for(current_user)
+      options = {
+        http_timeout: http_timeout,
+        importer_config: Cartodb.config[:importer],
+        user_id: current_user.id
```
jgoizueta: The new field is consistently added :slightly_smiling_face:

app/models/synchronization/member.rb
```diff
@@ -320,37 +320,38 @@ def get_downloader
...
+          http_options = {
+            http_timeout: DataImport.http_timeout_for(user),
+            etag: etag,
+            last_modified: modified_at,
+            checksum: checksum,
+            verify_ssl_cert: false
+          }
+          options = {
+            importer_config: Cartodb.config[:importer],
+            user_id: user.id
```
jgoizueta: OK, forget my comments about `user_id`, I got how it has been added to the Downloader :smile:

services/importer/lib/importer/datasource_downloader.rb
```diff
@@ -107,6 +109,9 @@ def set_downloaded_source_file(available_quota_in_bytes=nil)
             @http_response_code = @datasource.get_http_response_code if @datasource.providers_download_url?
           rescue => exception
             if exception.message =~ /quota/i
+              user_id = @options[:user_id]
+              report_over_quota(user_id) if user_id
```
jgoizueta: Only one very small concern here, that I leave up to your consideration:

What if we call `raise_if_over_storage_quota(requested_quota: 1, available_quota: 0, user_id: user_id)` here to let all the storage quota exceptions logic (i.e. the particular exception class , `StorageQuotaExceededError` which is raised and reporting events) be handled by the `QuotaCheckHelpers` module?

jgoizueta:
I took a look and left only a minor comment, up to you what to do about it. And I see some tests are broken.

All in all, good job! :+1:

services/importer/lib/importer/datasource_downloader.rb
gfiorav: You're right! the line below, which raises `StorageQuotaExceededError` is not needed anymore.

jgoizueta: 👍

gfiorav:
It seems some test code relied on no-quota set users. Instead of changing the test behaviour, I simply respected the previous decision of not checking if `available_quota_in_bytes` is `nil`. This is up for revision of course, but the time and scope is not now or this one.

I'll ping for a small CR when 🍏 

Thanks for the comments @javitonino and @jgoizueta

app/models/data_import.rb
```diff
@@ -809,23 +815,30 @@ def get_downloader(datasource_provider)
...
+      metadata_url = metadata[:url]
+      resource_url = (metadata_url.present? && datasource_provider.providers_download_url?) ? metadata_url : data_source
+
+      log.append "File will be downloaded from #{resource_url}"
+
+      http_options = { http_timeout: ::DataImport.http_timeout_for(current_user) }
+      options = {
+        importer_config: Cartodb.config[:importer],
+        user_id: current_user.id
+      }
```
javitonino: You can extract `http_options` and `options` outside the if, else. It is the same in both branches.

javitonino:
:+1:

PR #9020, opened 2016-07-20
Activates widget avg/max/min within the editor

fdansv:
Fixes https://github.com/CartoDB/deep-insights.js/issues/415

@xavijam @piensaenpixel


xavijam:
This will only work in the first iteration, but if a new widget is added it will not work. In my opinion this should not be there, should be in the widget creation.

lib/assets/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -47,10 +47,10 @@ var F = function (opts) {
   opts.widgetDefinitionsCollection.on('add remove reset', this._invalidateSize, this);
 
   this.widgetDefinitionsCollection = opts.widgetDefinitionsCollection;
-  this.activateWidgetsStats();
 
   this._analysisDefinitionsCollection.each(this._analyseDefinition, this);
-  this._vis().on('reload', this._recordChange, this);
+  this._vis().on('reload', this._visReload, this);
+  this._vis().on('change:error', this._visErrorChange, this);
```
xavijam: What is that? Is related with this PR?

fdansv: no, that's a bad merge I just added by accident, reverting

xavijam:
This is already working. 💋

PR #9050, opened 2016-07-21
Move and freeze actual dependencies from devDependencies

rochoa:
It's very critical to keep dependencies list tidy. That means any dependency `require`d at `lib/assets/javascripts/cartodb3/` should be declared within package.json's `dependencies` and not within `devDependencies`.

Not declaring them within `dependencies` means that they are not frozen at `npm-shrinkwrap` so the dependencies we use to develop, test locally, test in PR, and deploy might be different. So we can end with a different version in production that the want we testing in the whole release process.

That's even more critical if we declare the dependencies as use any minor release (`^`), or use any bugfix release (`~`). That's why I'm also using some fixed versions for libraries we don't control, e.g. `codemirror`, `bootstrap-colorpicker`, or `clipboard`.


xavijam:
cc @CartoDB/frontend ^^^ QUITE IMPORTANT.

xavijam:
🇪🇸

alonsogarciapablo:
:+1:

rochoa:
This introduces a small problem: different backbone versions. backbone-undo requires 1.2.1, where we define 1.2.3.

rochoa:
I updated the dependency to a more recent version that has backbone dependency updated.

PR #9052, opened 2016-07-21
added checks to run grunt

javisantana:
sad to add this but needed to avoid problems 

cc @CartoDB/frontend @rochoa


xavijam:
🙈  🇪🇸 , I'd ping in #builder channel after merge.

alonsogarciapablo:
nice!

rochoa:
Grunt will fail if you don't have node@0.10.x or npm@2.14.x, and it will also fail if for some reason we introduce different versions in npm-shrinkwrap for some specific dependencies

javisantana:
\o/

matallo:
and it failed 😂

for those like me who had the stack put together with duct tape, here's what I did to successfully fix the local versions for node & npm

- remove global packages and re-install system node & npm
 https://gist.github.com/DanHerbert/9520689

- reinstall nvm
https://github.com/creationix/nvm/issues/298
https://github.com/creationix/nvm

- install node via nvm fixed to `v0.10.26`
https://github.com/creationix/nvm

- install npm fixed to `2.14.16`
http://stackoverflow.com/questions/9755841/how-can-i-change-the-version-of-npm-using-nvm

PR #9074, opened 2016-07-21
Only show columns with weight > 0 in analysis

talos:
This should solve issues with measures like "aggregate travel time to work", that should be hidden, from appearing in the Builder Analysis.

cc @rambo @saleiva


saleiva:
cc/ @javisantana

javisantana:
sounds good to me

javisantana:
(i'll fix the PR, no worries)

PR #9084, opened 2016-07-22
Implement tooltip.

### LINKED ISSUE #8918: Set error tooltip when a node has failed
screen shot 2016-07-18 at 10 12 45

There could be an error in a node within a layer, so error notification is great, but is disappear after some seconds, so it would be great to have the tooltip providing the error at any time.

nobuti:
This PR fixes #8918.

analysis-tooltip-error

It changes styles and markup to be able to interact with the red dot. And also it applies tipsy thing to the view. For the message, we mimic the notification.

cc @xavijam @viddo


saleiva:
Now we just need better errors :D

viddo:
👍

PR #9093, opened 2016-07-22
changing default point outline from yellow to white

makella:
@saleiva @javisantana 

Please look and merge if it is OK!!

Just please get the yellow outline off of everything.

I don't know why point outline goes to other geometries, but here, I changed the basic default.


xavijam:
Code looks good to me.

PR #9094, opened 2016-07-23
Size options for custom list.

### LINKED ISSUE #9077: Make dropdown a bit bigger so Delete fits without scrolling
screen shot 2016-07-22 at 09 52 58

nobuti:
This PR fixes #9077.

I found this issue tricky because it could affect a lot of elements in the UI. For instance, a lot of form components use this custom list internally. 

The custom max height of the list view should depend on the inner item's height, and how many items we want to show without scrolling. So I'm introducing a size option for the custom list, that it handles how many items are visible at first sight. This is customizable for every custom view.

custom-list-size

In the screencast above you can see the context menu has a custom size of 5, to fix the issue. And I set the default size to 3, because TBH, those huge form controls bugged me up and I want to save some vertical space. In any case, now we can show some of them with different sizes.

Thoughts @xavijam @piensaenpixel @saleiva?

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
+++ b/lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
@@ -76,7 +76,7 @@ module.exports = CoreView.extend({
     this.$el.toggle(this.isVisible());
 
     $('body').append(this.el);
-
+    this._setListMaxSize();
     this.$el.css('top', this.options.position.y + this.options.offset.y);
     this.$el.css('left', this.options.position.x + this.options.offset.x - this.$el.outerWidth());
 
@@ -97,15 +97,23 @@ module.exports = CoreView.extend({
   },
 
   _renderList: function () {
-    var listView = new CustomListView({
+    this._listView = new CustomListView({
       model: this.model,
       collection: this.collection,
       typeLabel: '',
       ItemView: this.options.itemView,
-      itemTemplate: this.options.itemTemplate
+      itemTemplate: this.options.itemTemplate,
+      size: 5
     });
-    this.$el.append(listView.render().el);
-    this.addView(listView);
+    this.$el.append(this._listView.render().el);
+    this.addView(this._listView);
+  },
+
+  _setListMaxSize: function () {
+    setTimeout(function () {
+      this._listView.setMaxSize();
+      this.$el.css('visibility', 'visible');
+    }.bind(this), 0);
   },
 
   show: function () {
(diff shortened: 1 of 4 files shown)


lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
```diff
@@ -76,7 +76,7 @@ module.exports = CoreView.extend({
     this.$el.toggle(this.isVisible());
 
     $('body').append(this.el);
-
+    this._setListMaxSize();
```
nobuti: To calculate the height of one dom element, this element needs to be in the dom, otherwise you'll end up with 0 pixels. In order to avoid jumps, we need to make visible the element after the list item's height calculation. This is why the element containing the custom list is in charge of triggering the calculation.

lib/assets/javascripts/cartodb3/components/custom-list/custom-list-view.js
```diff
@@ -62,6 +67,13 @@ module.exports = CoreView.extend({
     return this;
   },
 
+  setMaxSize: function () {
+    var height = this.$('.js-listItem:first').outerHeight();
```
xavijam: if the item doesn't contain that class?. Should we add a default height in this case?

lib/assets/javascripts/cartodb3/components/custom-list/custom-view.js
```diff
@@ -93,13 +94,21 @@ module.exports = CoreView.extend({
...
     });
     this.$el.append(this._listView.render().el);
     this._listView.highlight();
     this.addView(this._listView);
   },
 
+  _setListMaxSize: function () {
+    setTimeout(function () {
+      this._listView.setMaxSize();
+      this.$el.css('visibility', 'visible');
```
xavijam: I'd like to avoid css changes directly, could we make use of classes?

lib/assets/javascripts/cartodb3/components/context-menu/context-menu-view.js
```diff
@@ -97,15 +97,23 @@ module.exports = CoreView.extend({
...
-    this.$el.append(listView.render().el);
-    this.addView(listView);
+    this.$el.append(this._listView.render().el);
+    this.addView(this._listView);
+  },
+
+  _setListMaxSize: function () {
+    setTimeout(function () {
+      this._listView.setMaxSize();
+      this.$el.css('visibility', 'visible');
```
xavijam: Same here ^^^, classes instead of css properties?

xavijam:
I really like the idea, just those two comments. I'm a bit afraid about the places where we are using the styles of the custom list and we could hide the list, could you double check where that is happening (it is likely we are not using it in that way so :+1:)?

saleiva:
ohh!

nobuti:
@xavijam I can't see any element using custom list styles outside the context menu or custom list view context. Do you remember any case in particular? @piensaenpixel does it ring your bell?

piensaenpixel:
For me it's chachi piruli

xavijam:
"pues dale niño"

PR #9119, opened 2016-07-26
8977 fix layer creation create missing analyses

### LINKED ISSUE #8977: If you create a map with one layer, and then you add a new layer, if you refresh, builder fails
### STR

- Create a map from the dashboard, with one layer, any dataset, it doesn't mind.
- Once you are in the builder, without doing anything else, add a new layer, no matter the dataset.
- Wait until it is added and you are sure that the layer is added correctly.
- Reload the page.
- Error appears saying that `a0` node doesn't exist.

Basically, the problem is when we add a new layer, all layers are saved, and if any of the previous one has not been saved before, it will store `source` attribute (for example a0) pointing to nowhere, because there is no analysis defined about that source.

cc @alonsogarciapablo @viddo

viddo:
Fixes #8977 

Make sure the analysis are persisted for all existing layers/nodes at the time of creating the new layer, since saving the collection will persist the source references.

@alonsogarciapablo can you review, please?

### DIFF
--- a/lib/assets/test/jasmine/cartodb3/data/user-actions.spec.js
+++ b/lib/assets/test/jasmine/cartodb3/data/user-actions.spec.js
@@ -906,7 +906,9 @@ describe('cartodb3/data/user-actions', function () {
 
     it('should create analysis for all layers that lack one', function () {
       this.layerDefinitionsCollection.add({kind: 'tiled'});
-      this.layerDefinitionsCollection.add({
+
+      // Layers w/ analysis
+      this.layerA = this.layerDefinitionsCollection.add({
         id: 'layerA',
         kind: 'carto',
         options: {
@@ -914,7 +916,16 @@ describe('cartodb3/data/user-actions', function () {
           source: 'a0'
         }
       });
-      this.layerDefinitionsCollection.add({
+      this.analysisDefinitionsCollection.add({
+        analysis_definition: {
+          id: 'a0',
+          type: 'source',
+          params: {}
+        }
+      });
+
+      // Layer w/o analysis
+      this.layerB = this.layerDefinitionsCollection.add({
         id: 'layerB',
         kind: 'carto',
         options: {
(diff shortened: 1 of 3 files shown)


lib/assets/test/jasmine/cartodb3/data/user-actions.spec.js
```diff
@@ -914,6 +903,33 @@ describe('cartodb3/data/user-actions', function () {
...
+          source: 'b0'
+        }
+      });
+
+      spyOn(this.analysisDefinitionsCollection, 'saveAnalysisForLayer');
+
+      this.userActions.createLayerForTable('tableName');
+
+      expect(this.analysisDefinitionsCollection.saveAnalysisForLayer).toHaveBeenCalled();
+      expect(this.analysisDefinitionsCollection.saveAnalysisForLayer.calls.count()).toEqual(2);
```
alonsogarciapablo: I guess we could use `.toHaveBeenCalledWith` to be more explicit:

```
      ...
      var layerA = this.layerDefinitionsCollection.add({
        id: 'layerA',
        kind: 'carto',
        options: {
          letter: 'a',
          source: 'a0'
        }
      });
      var layerB = this.layerDefinitionsCollection.add({
        id: 'layerB',
        kind: 'carto',
        options: {
          letter: 'b',
          source: 'b0'
        }
      });

      expect(this.analysisDefinitionCollection).toHaveBeenCalledWith(layerA, layerB);
      ...
```

alonsogarciapablo:
👍

lib/assets/test/jasmine/cartodb3/data/user-actions.spec.js
viddo: I'm not very fond of referencing the models in that call, because when the test fail the output is horrendously large. Nevertheless, made the test more assertive (asserting ids instead).

PR #9125, opened 2016-07-26
Fix notifications

piensaenpixel:
cc @saleiva


xavijam:
Fine by me, still pending to remove `<span>` from notifications.

PR #9130, opened 2016-07-26
Fit the area-of-influence interval inputs to distance/kind

### LINKED ISSUE #7501: [ANALYSIS] Area of Influence numeric inputs parameters improvement
edit: updated numbers based on new numbers given in #8166

**By Distance (buffer)**
~~-> m: min: 10, max: 1500, step: 1~~
~~-> km&mi: min: 1, max: 5, step: 0.5~~

new values:
- m (10-1000)
- km (0.5-100)
- mi (0.5-100)

**By Time (trade-area)**
-> walking: min: 60s, max: 3600, step: 1
-> driving: min: 60s, max: 14400, step: 10

On the future we should think how to change the unit when reaching high values

viddo:
Fixes #7501

Also make the value fit the range boundaries when changing the distance/kind, to avoid showing error feedback since the user have not really done anything wrong.

aos-inputs

@javierarce can you review?
cc @saleiva


javierarce:
👍

Issue #9041, opened 2016-07-21
On basemap section, the arrow isn't aligned


urbanphes:
On basemaps shows like this

![image](https://cloud.githubusercontent.com/assets/14546701/17016202/4baa016c-4f2e-11e6-97a4-d8680f0adf27.png)
<br><br>
... and should be

![image](https://cloud.githubusercontent.com/assets/14546701/17016251/8d6454ae-4f2e-11e6-8ff9-5f13868f9345.png)

PR #9138, opened 2016-07-27
fix arrow position

piensaenpixel:

image

Fix - https://github.com/CartoDB/cartodb/issues/9041


javierarce:
👍

Issue #9116, opened 2016-07-26
Removing analysis node or dragging a node outside a layer, makes the layer style unchangeable


saleiva:
This is because the aggregation method is set to NONE, instead to the default one.

Might be related to:
https://github.com/CartoDB/cartodb/issues/8827
https://github.com/CartoDB/cartodb/issues/9042

saleiva:
cc/ @javisantana

viddo:
Can't reproduce it, from what I know this should already be fixed by #9144 (#9146 might also have helped). Please reopen if still happen.

PR #9144, opened 2016-07-27
use default style when type is none

### LINKED ISSUE #9143: removing all the nodes apply the wrong style
- add a layer with an analysis (sampling)
- remove the node

the style applied is the wrong default one and it should be one generated from cartography.json

javisantana:
fixes #9143

There is a problem. This fixed the default style applied but does not the fact that "none" style is applied when you remove all the analysis in a layer

cc @viddo


viddo:
Probably solved #9116 too

viddo:
:checkered_flag:

Issue #9042, opened 2016-07-21
AOI buffers disappear after running a second analysis


ramiroaznar:
### Steps

I have encountered a couple of issues while replicating the 2nd demo @cartodave made in #the-retraining:

- Create a map with points dataset (`layer1`).
- Run AOI analysis in `layer1`.

![aoi1](https://cloud.githubusercontent.com/assets/5215798/17017453/a8e78098-4f33-11e6-9547-8f363892582a.png)

- Continue the workflow analysis in `layer1`. This time "Enrich from DO". I select `United States` as COUNTRY, `Transportation` as MEASUREMENT and `Commuters by Subway or Elevated` as SEGMENTS.

![aoi2](https://cloud.githubusercontent.com/assets/5215798/17017527/f6d4e62e-4f33-11e6-9a1b-92486ceb5ca8.png)

- The buffers disappear. In addition, the values from the DO are really low. According to @AbelVM it is because the values are normalized. But another explanation could be that the second analysis is using the point geometry instead of the buffer geometry to extract the DO data.

![commuters](https://cloud.githubusercontent.com/assets/5215798/17017633/644d2c52-4f34-11e6-9944-9cb3af422979.png)

saleiva:
as you see, the geometries are still polygons so this looks as a bug with the cache of layers and the style we apply.

saleiva:
Does this happen always?

ramiroaznar:
It always happens with "Enrich with DO" analysis.

saleiva:
@xavijam feel free to assign to @viddo with priority tho...

saleiva:
Happening to me as well when doing an "Intersect with second layer" analysis in a polygon layer with a previously analized points layer. As soon as you any property on the style form on the polygons layer, they get correctly rendered as polygons 

cc/ @javisantana

talos:
The really low column values were a result of numbers being calculated per sq m instead of per sq km.  This should be resolved now.

talos:
Also want to flag https://github.com/CartoDB/dataservices-api/pull/237 here, as once that's merged then we'll get different numbers again for polygon measures -- the absolute number instead of a rate per sq km.  This is the original intended behavior, but there were several layers of problems thus the back and forth.

javisantana:
ok, several things here:

- hey, I can't reproduce with my datasets, could you send the .carto so I can check?
- I think DO analysis should have an option to normalize by pop at UI level cc @saleiva

[...]

PR #9146, opened 2016-07-27
9042-set-none-styles-for-unknown-geometry

viddo:
Fallback solution for #9042 

ping @javisantana


javisantana:
:es:

viddo:

PR #9151, opened 2016-07-27
Only downsmaple rasters if the band type if more than Byte

### LINKED ISSUE #8709: Overviews should only be downsampled (aka interpolated) when needed
Raster overviews are generated on the original raster with the `-scale` option. See bb77480.

This is done so because overviews are focused on efficient rendering and the tiler supports 8bit images so adding or keeping more depth doesn't make sense.

But it should only be done when values cannot be held in 8bit ([0, 255] or [-127,128] intervals)

ethervoid:
Closes #8709 

The band types could be more than Byte and we don't support them in
the tiler so we need to make downsample but if the band is already a
Byte (could be 8BSI, 8BUI, 4BSI, 2BSI..) we don't need to make
downsample.

In case the band is for example a Int16 type but the max/min values are
[0,255] we downsample it to reduce the size of the column


ethervoid:
@rafatower could you review this please

rafatower:
LGTM

PR #9152, opened 2016-07-27
9148 fix thrown exceptions

viddo:
Potential fix for #9148

I'll add inline comments for feedback

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/data-observatory-measure-model.js
+++ b/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/data-observatory-measure-model.js
@@ -23,7 +23,9 @@ module.exports = BaseAnalysisFormModel.extend({
   },
 
   _fetchDOMetadata: function () {
-    this.doMetadata.setGeometryType(this.nodeDefModel.querySchemaModel.getGeometry().getSimpleType());
+    var geometry = this.nodeDefModel.querySchemaModel.getGeometry();
+    var simpleGeometryType = geometry && geometry.getSimpleType() || 'point';
+    this.doMetadata.setGeometryType(simpleGeometryType);
     this.listenTo(this.doMetadata, 'sync', this._setSchema);
     this.on('change:area', this._areaChanged, this);
     this.on('change:measurement', this._measurementChanged, this);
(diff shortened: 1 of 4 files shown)


lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/data-observatory-measure-model.js
```diff
@@ -23,7 +23,9 @@ module.exports = BaseAnalysisFormModel.extend({
   },
 
   _fetchDOMetadata: function () {
-    this.doMetadata.setGeometryType(this.nodeDefModel.querySchemaModel.getGeometry().getSimpleType());
+    var geometry = this.nodeDefModel.querySchemaModel.getGeometry();
+    var simpleGeometryType = geometry && geometry.getSimpleType() || 'point';
+    this.doMetadata.setGeometryType(simpleGeometryType);
```
viddo: @javisantana The deleted row was trying to get the simple type of a none-existing geometry object (it's not guaranteed to exist), so falling back on `point` as you mentioned offline.

javisantana: :+1:

lib/assets/javascripts/cartodb3/editor/style/style-defaults/simple-style-defaults.js
```diff
@@ -10,6 +10,7 @@ module.exports = _.defaults({
 
   generateAttributes: function (geometryType) {
     return _.extend(
+      {},
```
viddo: From the [stack trace I mentioned](https://github.com/CartoDB/cartodb/issues/9148#issuecomment-235641538) in the issue:
> `Uncaught TypeError: Cannot read property 'fixed' of undefined`, I happened to just hit this same exception (not conclusive that it's the same, but likely…):

> ```
> (anonymous function) (VM28973:6)
> i (underscore:1461)
> getDefaultCartoCSSForType (camshaft-reference.js:148)
> createAnalysisNode (user-actions.js:85)
> createNodeDefinition (base-analysis-f…m-model.js:111)
> saveAnalysis (user-actions.js:70)
> _onSaveClicked (analysis-controls-view.js:181)
> …
> ```

I have not been able to reproduce it consistently again, but looking at [the usages of the faulty object](https://github.com/CartoDB/cartodb/search?utf8=%E2%9C%93&q=default-cartography) there's only one place I can see where that object could be modified (3da295b), so change this extend call to work on a new object instead of the returned fill object.

@xavijam can you review this part?

xavijam: :+1:

PR #9156, opened 2016-07-28
9155 fix letter match

### LINKED ISSUE #9155: Cannot read property '0' of null bug
### STR

- Create a map with a dataset of polygons.
- Add a "Filter points in polygons" analysis.
- Select a `target layer`, any is fine.
- Don't apply the analysis.
- Go back to layers list.
- Try to add a new analysis, you can try with the same one.
- JS error should appear in the console: Cannot read property '0' of null

Seems like we are not removing target layers from analyses node collection.
We could check that and also be sure we take the proper letter from the nodes.

viddo:
Fixes #9155 

This bug only happens if there are secondary sources that have the format of `”other_user”.some_table_name”`, since the regex previously assumed the node ids to start with a letter.

@xavijam ping


viddo:

xavijam:
:es:

PR #9159, opened 2016-07-28
8805 fix layer content letter cue

### LINKED ISSUE #8805: Provide visual layer cue on layer-content view
Compare:
screen shot 2016-07-12 at 16 29 08 vs: screen shot 2016-07-12 at 16 29 08

When editing layer I quickly forget what layer I was actually editing, why not be consistent and provide the letter+color+head node id next to the layer name like we do in other places (e.g. layer dropdown)?

cc @saleiva

viddo:
Resolves #8805 

screen shot 2016-07-28 at 10 42 45

@piensaenpixel can you review please? Not sure if I've used the CSS styles appropriately.
cc @urbanphes


piensaenpixel:
🇪🇸

PR #9164, opened 2016-07-28
Widgets options on builder side

### LINKED ISSUE #8882: Add options to the widgets module
On the widget list:
- [x] Rename widget with double click

On the widget contextual menu:
- [x] Edit widget
- [x] Remove widget.

On the widget form:
- Header:
    - [x] Rename widget.
    - [x] Rename widet with double click.
    - [x] Delete widget.
- Form
    - [x] Remove title input.

nobuti:
This PR fixes #8882.

widgets-options

cc @xavijam


lib/assets/javascripts/cartodb3/mixins/has-context-menu.js
```diff
@@ -0,0 +1,63 @@
+var ContextMenuView = require('../components/context-menu/context-menu-view');
```
xavijam: I like the idea, but not sure if I like more the idea of creating a view for this purpose, not something that is extended.

xavijam: The typical:

```js
var menu = ContextMenu.init(collection);
menu.show(position);
menu.hide();
menu.clean();
view.addView(menu);
```

nobuti: I like the mixin pattern because the context is the own view, so you can take advantage of the backbone events hash and delegation, without adding a new object to the chain. Also, I wanted to delegate all the management to the mixin, completely, in your proposal the main view is responsible for show, hide and so on, right? 

But I see your concern, I guess we need a stronger system/method/whatever in order to apply the mixin pattern. So I've implemented a new view which is listening the trigger element to manage the dropdown https://github.com/CartoDB/cartodb/blob/70380d52bf3af89f3870d29443d0092d7f138f53/lib/assets/javascripts/cartodb3/components/context-menu-factory-view.js what do you think?

xavijam:
Yeah, in my honest opinion it is safer. It is true that we don't introduce anything new but sometimes old things keep working "prettily" 💃 .
🇪🇸  🌴

PR #9165, opened 2016-07-28
Dataset view issues

### LINKED ISSUE #8865: Cannot edit shared datasets in the builder
1. Share a dataset as RW with some colleague in the organization
 2. From the second user (non-owner) open the dataset in the builder

You should be able to edit the dataset, but you can't, all edition options are blocked.

cc @xavijam

### LINKED ISSUE #9161: An error pops up after dataset renaming, but the renaming succeed
### Steps

* Open a dataset.
* Rename the dataset. After clicking on "enter", you get an error:

error

* Then the error disappears, and the renaming was OK.

xavijam:
- Cannot edit shared datasets in the builder (fixes #8865).
- An error pops up after dataset renaming, but the renaming succeed (fixes #9161): we were fetching 2 times the query-schema-model. It was not necessary to fetch when it is renamed in the sql editor.

CR: @nobuti

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/table/table-view-model.js
+++ b/lib/assets/javascripts/cartodb3/components/table/table-view-model.js
@@ -58,8 +58,11 @@ module.exports = Backbone.Model.extend({
   },
 
   isCustomQueryApplied: function () {
+    var sanitizeName = function (str) {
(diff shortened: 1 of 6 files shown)


nobuti:
👍

PR #9170, opened 2016-07-28
Fixes AOI completion text

javierarce:
We were showing the description for the georeference instead of the one for AOI.

screen shot 2016-07-28 at 18 42 19


javierarce:
CR: @piensaenpixel

piensaenpixel:
🇪🇸

Issue #8552, opened 2016-07-06
Layer list should reference the different nodes with the name of the layer + the analyisis


saleiva:
[Layer name] source name

viddo:
Fixes by #9173

PR #9173, opened 2016-07-28
Render analysis short-title for select list

viddo:
#8552 i.e. render `[layer-name] analysis-short-title` in node selector for widgets and analysis forms.

screen shot 2016-07-28 at 18 15 53
screen shot 2016-07-28 at 19 00 13

@xavijam ok?
cc @saleiva


saleiva:
Pity is with long layer names... uhm... 
What about writing in the other way around? 

Node_name (Layer_name) ?

Easier to scan?

xavijam:
Fine by me whatever you decide. I'd try with:

[A0] AOI _paradas_metro_madrid_

Last part in light grey color. (my two cents)

viddo:
Sure,
screen shot 2016-07-29 at 08 41 17

+1 what @xavijam propose, but need to figure out how/where the text is actually rendered to structure it like that.

Btw, isn't it weird that we call the label "Select layer" when it's really an analysis node (that may be used by multiple layers for that matter)?

xavijam:
True, the thing is we don't talk about nodes in the new builder, and not sure if we could confuse users:

Select Input?

viddo:
Or "Select analysis", since that's how its described in the UI everywhere else.

Btw, better this way
screen shot 2016-07-29 at 09 13 13

xavijam:
But you will select an analylsis or another dataset...
Yeah, I like it a lot, it is likely we need to add a bit of space between node name and layer name.

saleiva:
@piensaenpixel what do you think?

viddo:
> But you will select an analylsis or another dataset...

Ok, scratch my proposal then, I'll leave it be for the moment.

viddo:
With a little more spacing:
screen shot 2016-07-29 at 11 14 31

viddo:
Some more screenshots, I think it looks a little clearer this way indeed:
screen shot 2016-07-29 at 12 02 14
screen shot 2016-07-29 at 12 02 28
also added a title attribute so one can see the full layer name if it does't fit:
screen shot 2016-07-29 at 12 03 04

piensaenpixel:
@viddo i need fix some css issues (margins and paddings) :)

viddo:
@piensaenpixel ok, feel free! let me know if you need anything from me 😘

piensaenpixel:
@viddo for me now is ok, so... when you want/can feel free

viddo:
Then I'd say let's :shipit: once the CI pass

viddo:

PR #9176, opened 2016-07-29
change loaders

piensaenpixel:

Change loaders in add layer and add widgets

add-layer

add-widget


javierarce:
👍

PR #9178, opened 2016-07-29
9168 fix delete layer with high node ids

### LINKED ISSUE #9168: Cannot delete layers or add widgets under certain circumstances
Not sure about the STR, but I have a .carto (remember to rename the .zip to .carto before)
[Untitled Map (on 2016-07-28 at 15.19.17).zip](https://github.com/CartoDB/cartodb/files/388719/Untitled.Map.on.2016-07-28.at.15.19.17.zip)

screen shot 2016-07-28 at 17 20 59

viddo:
Fixes #9168 
Fixes #9178

E.g. (from the reported issue) deleting layer C will now make its node to be folded under another layer as expected.

Before delete:
screen shot 2016-07-29 at 17 52 18 
After delete (`c2` moved to `b2`):

The problem was that layer B's node already had a rather high sequence number (starting at `[b4, b3]`), so when reorganizing the nodes under the new layer (B) and only having 3 nodes `[b4,b3,c2]` the algorithm wasn't prepared for this and wanted to rename them to `[b3,b2,b1]`, which caused node b3 to point at itself (thus the infinite loop later in the code path). 

Changed the code to take this into account and instead start from the highest sequence instead (thus naming nodes `[b4,b3,b2]` instead). Test case is using data from the extracted carto file.

@alonsogarciapablo can your review?
cc @saleiva @xavijam

### DIFF
--- a/lib/assets/javascripts/cartodb3/data/user-actions.js
+++ b/lib/assets/javascripts/cartodb3/data/user-actions.js
@@ -484,6 +484,13 @@ module.exports = function (params) {
 
         // Since will reassign ids from the start of list need to calculate which sequence id to start on
         var idSequence = newParentLinkedNodesList.length;
+        var firstSequence = nodeIds.sequence(newParentLinkedNodesList[0].id);
+
+        // If first node has a higher sequence start from that one to avoid issues on renaming nodes
+        if (idSequence <= firstSequence) {
+          idSequence = firstSequence;
+        }
+
         var lastNode = _.last(newParentLinkedNodesList);
         if (lastNode.get('type') !== 'source') {
           idSequence++; // to start on right id sequence, so last item in newParentLinkedNodesList gets assigned x1 as id, since its source will be a none-source node belong to another layer
--- a/lib/assets/test/jasmine/cartodb3/data/user-actions.spec.js
+++ b/lib/assets/test/jasmine/cartodb3/data/user-actions.spec.js
@@ -1893,6 +1893,7 @@ describe('cartodb3/data/user-actions', function () {
         expect(this.layerDefinitionsCollection.pluck('id')).toEqual(['A', 'B'], 'layer C should be deleted');
         expect(this.analysisDefinitionNodesCollection.pluck('id')).toEqual(['a0', 'a1', 'b2', 'b3', 'b4'], 'c2 should not be b2');
         expect(this.analysisDefinitionNodesCollection.get('b3').get('source')).toEqual('b2', 'b3 should point on the renamed node');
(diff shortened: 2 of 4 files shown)


lib/assets/test/jasmine/cartodb3/data/user-actions.spec.js
```diff
@@ -1798,6 +1798,104 @@ describe('cartodb3/data/user-actions', function () {
...
+
+        this.userActions.deleteLayer('C');
+      });
+
+      it('should fold node c2 under layer B', function () {
+        expect(this.layerDefinitionsCollection.pluck('id')).toEqual(['A', 'B'], 'layer C should be deleted');
+        expect(this.analysisDefinitionNodesCollection.pluck('id')).toEqual(['a0', 'a1', 'b2', 'b3', 'b4'], 'c2 should not be b2');
+        expect(this.analysisDefinitionNodesCollection.get('b3').get('source')).toEqual('b2', 'b3 should point on the renamed node');
+        expect(this.analysisDefinitionNodesCollection.get('b2').get('type')).toEqual('centroid', 'b2 should be the prev c2');
+      });
```
alonsogarciapablo: We could also check that `expect(this.analysisDefinitionNodesCollection.get('b2').get('source')).toEqual('a1', 'b2 should be the prev c2');`.

lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -484,6 +484,13 @@ module.exports = function (params) {
 
         // Since will reassign ids from the start of list need to calculate which sequence id to start on
         var idSequence = newParentLinkedNodesList.length;
+        var firstSequence = nodeIds.sequence(newParentLinkedNodesList[0].id);
+
+        // If first node has a higher sequence start from that one to avoid issues on renaming nodes
```
alonsogarciapablo: > // If first node has a higher sequence start from that one to avoid issues on renaming nodes

Why do this only "if first node has a higher sequence" and not always?

viddo: Because it should not always start from that sequence, it depends on how many nodes that needs to be moved. E.g. let's say there are two layers, layer `A` with nodes `[a3,a2,a1,a0]`, and layer `B` with nodes `[b1,a3]`. Deleting layer A in this case should fold the nodes from layer A under B and to do so it need to start not from the first node but from the calculate one (`idSequence`).

alonsogarciapablo:
Looks legit 😛 !!! :+1:

viddo:
Frontend tests passing so closing

Issue #9194, opened 2016-08-01
Basemaps are not loading correctly (tiles and labels) or disabling the zoom in control beyond level 10


arianaescobar:
OS: mac OS El Capitán 10.11.5
Browser: Chrome Version 51.0.2704.106 (64-bit)
CARTO Account used when this happened: arianaescobar 

## Issue

On the Basemaps picker, the basemaps are sometimes taking a lot of time to load, are loading incorrectly (not loading tiles or labels, or loading another basemaps label styles instead), or not loading at all. Also sometimes the basemaps are only letting me zoom in until zoom level 10. 

Tiles didn't load: 
![screen shot 2016-08-01 at 13 23 13](https://cloud.githubusercontent.com/assets/16247628/17292658/d33f6638-57eb-11e6-9412-0a119358fcb3.png)

Positron label styles on a Dark Matter basemap:
![screen shot 2016-08-01 at 13 26 47](https://cloud.githubusercontent.com/assets/16247628/17292659/d345553e-57eb-11e6-9dd6-f82874fdbe51.png)

Take a look on how the + icon on the zoom controls is disabled: 
![screen shot 2016-08-01 at 13 27 26](https://cloud.githubusercontent.com/assets/16247628/17292660/d34be1a6-57eb-11e6-9919-02ed738e2529.png)
![screen shot 2016-08-01 at 13 29 36](https://cloud.githubusercontent.com/assets/16247628/17292722/0fd5f030-57ec-11e6-95dd-371b8b5f0105.png)

## Steps taken when it happened 

1. I logged in into the my account
2. I opened one of my maps
3. I clicked on the Basemap layer and the basemaps picker appeared.
5. Then I clicked on differents basemaps and the errors began to occur. 

## Expected behaviour 

When I click on different basemaps than the one picked, it should load it and change to that one faster and correctly (tiles and labels). 

## Console errors

**tracker.js:26 Uncaught Error: TrackJS Caught: Cannot read property 'getName' of undefined**

## Reproduction 

I can reproduce it on every map in that account. I tried to reproduce it creating a new map but then I got this old bug #8879 when I opened the basemaps picker. 

If there's any other information or context I can give to help reproduce the issue, just let me know, I'll be happy to :)

xavijam:
the getName error is something I've experienced last days, we will take a look. About the zoom limitation, do we set that @matallo?

matallo:
checking the `getName` error, the zoom issue seems more rare as we do set the limit

PR #9195, opened 2016-08-01
basemaps: fix getname

matallo:
fix some of the cases in https://github.com/CartoDB/cartodb/issues/9194

checking the zoom after

CR @xavijam my bad after doing that last change you asked I didn't put enough attention when doing so


xavijam:
:+1:

PR #9196, opened 2016-08-01
Fix invalid analyses

viddo:
Make sure to delete orphaned analyses, that does not any longer contain any node.

@xavijam Not entirely sure about the root cause still since I have not been able to reproduce the reported errors, but nevertheless invalid analyses are not worth keeping around, so using the cleanup method to delete this ones too. Added a guard check for the containsNode call to return false for this case instead of throwing an exception.


xavijam:
:+1:

viddo:

PR #9237, opened 2016-08-02
Don't run the same applied query

### LINKED ISSUE #9184: "Applying query" message gets stuck after applying SELECT * query
### STEPS

* Add a `layer` to a map.
* Apply `SELECT * FROM layer`.
* The "Applying query" message gets stuck, as shown in the following screenshot:

applyingquery

### CONTEXT

* With some queries, the notification area shows `SQL query applied` and then disappear as expected. **This only happens sometimes**, and we couldn't find a clear pattern. Examples: 
  * queries with `WHERE` clauses
  * queries with `LIMIT n`

xavijam:
Seems like #9184 is due to the fact that the sql notification disappear when the vis is reloaded. Trying to run the same query that is applied, query is applied but event about `viz:reload` is not sent (I feel it makes sense, because nothing has changed).

Fixes #9184.

CR @alonsogarciapablo @nobuti (both), you were changing this and know better than me the internals.

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
+++ b/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
@@ -126,8 +126,13 @@ module.exports = CoreView.extend({
   },
 
   _parseSQL: function () {
-    var query = this._codemirrorModel.get('content');
-    var altersData = SQLUtils.altersData(query);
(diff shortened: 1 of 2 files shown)


nobuti:
LGTM

alonsogarciapablo:
I'm not very familiar with this part of the code, so probably @nobuti knows more about it.

xavijam:
:+1:, so merging then, frontend tests passed, thanks guys, you are ❤️ .

PR #9239, opened 2016-08-02
Animated aggregation type

### LINKED ISSUE #9010: Delete Animated option and add a new Aggregation method
When style is set as animated some things needs to happen:

- if the style applied does not use any variable it should work like it's working with now
- if we have one style using "by value" 
    - if it's a category attribute: torque style should use `cdb_mode([attribute])` as aggregation and use the turbocarto for the category
    - if it's a ramp attribute: torque style should use `avg([attribute])` as aggregation and use the turbocarto for ramps

- if we have more than one style attribute using "by value" -> we can't do anything with torque here since we don't have multivariable torque <- cc @saleiva. One solution would be to disable animated when this happen but not sure how hard is to explain that limitation to the user.

nobuti:
This PR fixes #9010.

This is still WIP. Things to be done:

- [x] Don't let select "animation" or animated heatmap when there is already a torque layer in the layers collection.
- [x] Style generation.
- [x] Migration from editor to builder (@javitonino)
- [x] Don't remove&add time-series widget when a simple attribute is edited (except for column) (@nobuti).
- [x] Display animated or heatmap tag & style for torque layers (@piensaenpixel).
- ~~Category torque is not working. (Select a 'by value' option) (@xavijam)~~ -> https://github.com/CartoDB/cartodb/issues/9389.
- ~~Torque, with cartodb_id column selected, renders date instead of numbers (@xavijam)~~ -> https://github.com/CartoDB/deep-insights.js/issues/441.
- [x] Change current torque layers to simple if user selects a new animated layer (@xavijam).
- [x] Don't let move torque layers (@xavijam).
- [x] Don't let to move to the first position a non-animated layer if there is an animated one (@xavijam).
- [x] Don't add a new layer on top of the animated one (@xavijam).
- [x] Change to `animation` type instead `animated` (@nobuti).


saleiva:
oooh!

lib/assets/javascripts/cartodb3/editor/style/style-form/style-properties-form/style-labels-properties-form-view.js
```diff
@@ -36,15 +36,15 @@ module.exports = CoreView.extend({
...
-  _initBinds: function () {
-    this._styleModel.bind('change:animated', function () {
-      var isAnimatedEnabled = this._styleModel.get('animated').enabled;
-      if (isAnimatedEnabled) {
-        this._enablerView.setValue(false);
-      }
-    }, this);
-    this.add_related_model(this._styleModel);
-  },
+  // _initBinds: function () {
```
xavijam: Remove the code

lib/assets/javascripts/cartodb3/editor/style/style-form/style-properties-form/style-labels-properties-form-view.js
```diff
@@ -27,7 +27,7 @@ module.exports = CoreView.extend({
       enabler: this._styleModel.get('labels').enabled
     });
 
-    this._initBinds();
+    // this._initBinds();
```
xavijam: Dixit below

lib/assets/test/spec/cartodb3/editor/style/style-form/style-labels-properties-form-view.spec.js
```diff
@@ -39,13 +39,13 @@ describe('editor/style/style-form/style-labels-properties-form-view', function (
...
   });
 
-  describe('bindings', function () {
-    it('should disabled subform when animated is enabled', function () {
-      spyOn(this.view._enablerView, 'setValue');
-      this.styleModel.set('animated', { enabled: true });
-      expect(this.view._enablerView.setValue).toHaveBeenCalledWith(false);
-    });
-  });
+  // describe('bindings', function () {
```
xavijam: We should check that in the heatmap form and don't let choose animated form if there is already a toque layer in the map.

nobuti: I commented out the code when we wanted to delete the `enabled` property. I've put it back.

xavijam:
@javitonino, any doubt we can answer for the backend migration?

javitonino:
Only changes currently in migrator is changing the type of torque/torque_cat to animated. Have there been any additional changes? I heard about removing `animated.enabled` but not sure if that is the only extra change.

xavijam:
we can `fire` enabled property.

nobuti:
@xavijam @CartoDB/design this is in ded06, it should be great if you could take a look, ask me for details if you need it.

josecruz:
Hey guys! I found a few things:

- [ ] Map / Dataset switch appears to be broken, it moves when you change the switch:
bug1

- [ ] When you change the column, a message appear saying that we deleted your widget. This is actually true because we remove it from the UI but I'd rather not show that notification due you're not removing anything.
bug2

- [ ] Torque button stills activated when you change option:
bug3

My two cents 💖 
I would update this comment if I find more things

xavijam:
I don't understand the third one @josecruz, what do you mean with torque button?

xavijam:
Fixed all the things you have commented @josecruz 💋 , let's make a last review tomorrow morning and then deploy.
cc @saleiva

alonsogarciapablo:
Great job guys! 👏

PR #9241, opened 2016-08-02
Fix "nodeTitle is not defined" error when adding "Join columns from 2nd layer" analysis 

### LINKED ISSUE #9240: "nodeTitle is not defined" error when adding "Join columns from 2nd layer" analysis
How to reproduce it:

1. Create a map with one layer.
2. Add a `Join columns from 2nd layer` analysis

The dev console will show the error `nodeTitle is not defined`

viddo:
Fixes #9240 

@javierarce ok?


viddo:
Verified locally and frontend tests passing, so merging to have it fixed now.
@javierarce feel free to open a separate issue to fix things posterior if you think this is insufficient.

PR #9244, opened 2016-08-02
add created_at to visualization properties

### LINKED ISSUE #9191: [Metrics] Add 'age' attribute to deletion-related events
Add new attribute 'age' (calculated as 'deletion time' - 'creation time', in days) for following events:
 - Deleted dataset
 - Deleted map

This shouldn't be difficult as both maps and datasets have a created_at property.

gfiorav:
Fixes #9191 
Fixes https://github.com/CartoDB/cartodb/issues/9271

### DIFF
--- a/lib/carto/tracking/events.rb
+++ b/lib/carto/tracking/events.rb
@@ -4,11 +4,14 @@ module Carto
   module Tracking
     module PropertiesHelper
       def visualization_properties(visualization, origin: nil)
+        created_at = visualization.created_at
+
         properties = {
           vis_id: visualization.id,
           privacy: visualization.privacy,
           type: visualization.type,
-          created_at: visualization.created_at
+          created_at: created_at,
+          lifetime: Time.now.utc - created_at
         }
 
         properties[:origin] = origin if origin
--- a/lib/carto/tracking/events.rb
+++ b/lib/carto/tracking/events.rb
@@ -5,19 +5,23 @@ module Carto
     module PropertiesHelper
       def visualization_properties(visualization, origin: nil)
         created_at = visualization.created_at
-        lifetime_in_days_with_decimal = (Time.now.utc - created_at).to_f / 60 / 60 / 24
+        lifetime_in_days_with_decimals = days_with_decimals(Time.now.utc - created_at)
 
         properties = {
           vis_id: visualization.id,
           privacy: visualization.privacy,
           type: visualization.type,
           object_created_at: created_at,
-          lifetime: lifetime_in_days_with_decimal
+          lifetime: lifetime_in_days_with_decimals
         }
 
         properties[:origin] = origin if origin
 
         properties
+      end
+
+      def days_with_decimals(time_object)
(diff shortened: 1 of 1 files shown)


gfiorav:
Please CR, @javitonino

lib/carto/tracking/events.rb
```diff
@@ -3,11 +3,15 @@
...
+
         properties = {
-          vis_id: table_visualization.id,
-          privacy: table_visualization.privacy,
-          type: table_visualization.type
+          vis_id: visualization.id,
+          privacy: visualization.privacy,
+          type: visualization.type,
+          created_at: created_at,
+          lifetime: Time.now.utc - created_at
```
javitonino: This returns seconds, but days were requested in the ticket. Don't know if it is important (maybe seconds are better, more granularity)

lib/carto/tracking/events.rb
```diff
@@ -3,11 +3,15 @@
...
+        created_at = visualization.created_at
+
         properties = {
-          vis_id: table_visualization.id,
-          privacy: table_visualization.privacy,
-          type: table_visualization.type
+          vis_id: visualization.id,
+          privacy: visualization.privacy,
+          type: visualization.type,
+          created_at: created_at,
```
javitonino: Check that this date format is parseable by Segment. I guess it is, but I don't know how this works exactly.

javitonino:
Take into account the comments you want, and ship it!

lib/carto/tracking/events.rb
gfiorav: yup, we've been using it already! https://github.com/CartoDB/cartodb/blob/master/lib/carto/tracking/events.rb#L56

javitonino: Awesome :sparkles:

lib/carto/tracking/events.rb
gfiorav: True! I asked in the original ticket in case we'd want' to reevaluate granularity. Sublime catch!

gfiorav:
I changed the units to days (with decimal) and added some properties to the event general properties. Review again please @javitonino and let's put this in prod as soon as posible!

lib/carto/tracking/events.rb
```diff
@@ -3,11 +3,16 @@
 module Carto
   module Tracking
     module PropertiesHelper
-      def visualization_properties(table_visualization, origin: nil)
+      def visualization_properties(visualization, origin: nil)
+        created_at = visualization.created_at
+        lifetime_in_days_with_decimal = (Time.now.utc - created_at).to_f / 60 / 60 / 24
```
javitonino: Ruby not having proper timediff objects is a bit disappointing. :cry: 

Anyway, can you extract this to a function? It is repeated in `event_properties`.

javitonino:
Other than the proposed refactor (extract timediff function), nothing else to add. Ship when :green_apple:

javitonino:
:shipit:

Issue #9236, opened 2016-08-02
Migrations for animated aggregation


nobuti:
In order to complete #9010, we need to perform migrations to apply changes from old editor to new one.

PR #9259, opened 2016-08-03
Type for torque and torque cat is animated

javitonino:
Migrations for #9236 

Changes the type for Torque wizards to "animated". Heatmap is untouched.

cc @nobuti @xavijam


javitonino:
Summoning @juanignaciosl for CR duty

juanignaciosl:
Ok, but [original migration spec said the opposite](https://github.com/CartoDB/cartodb/wiki/Wizard-to-style-migration). If this is ok we should delete that wiki page, as it's no longer maintained.

javitonino:
Yes, we changed our minds and we want to mark this as a new type now, to make more explicit where to find this in the UI. I guess we can delete that wiki page, as it is quite outdated anyway.

xavijam:
Merge tomorrow please :)

nobuti:
Lets wait until tomorrow morning, please.

juanignaciosl:
Having a type for something that is an option inside (almost) every type feels weird, but u(i) are the ones who know about that :_) I'll delete the wiki page right now.

javitonino:
Removed `animated.enabled = true` where `type == 'animated'`. Please take a look @juanignaciosl

juanignaciosl:
@javitonino I was waiting for tests to pass, do you want me to CR this anyway?

javitonino:
Yep, there are failing frontend tests, but I think those should be solved in the branch I'm merging to. It doesn't make too much sense for them to fail here.

juanignaciosl:
Ok, on it then. As I saw @nobuti involved I thought that it might be relevant.

app/controllers/carto/api/layer_presenter.rb
```diff
@@ -598,12 +598,12 @@ def generate_dimension_properties(wpp)
...
       }.freeze
 
-      def generate_animated(wpp)
+      def generate_animated(wpp, type)
         animated = {}
 
         apply_direct_mapping(animated, wpp, ANIMATED_DIRECT_MAPPING)
 
-        animated['enabled'] = true unless animated.empty?
+        animated['enabled'] = true if animated.present? && type != 'animated'
```
juanignaciosl: Thank you for reverting the `unless` with the second condition :D

juanignaciosl: PS: shouldn't `animated['enabled']` be `true` if `type == 'animated'`?

javitonino: `animated['enabled']` should not be present if `type` is `animated`. You are right that I added the positive test, but didn't add a negative one. Adding it...

juanignaciosl: Perfect.

javitonino:
Please CR @juanignaciosl (hopefully, last one)

After some back and forths, this is what this is supposed to do:
 - Remove `animated.enabled` property.
 - Remove `animated` block from all non-animated types.
 - For torque and torque_cat, set `animated` type.
 - For heatmap, set `animated` or `heatmap` type depending on whether it was animated or not in the wizard. Keep the `animated` block regardless (a more correct name for `animated` would be `torque` but we are not changing it).

Frontend tests fail (as expected). I will merge this as soon as CR is :+1: to avoid more changes on this one (sorry @juanignaciosl)

app/controllers/carto/api/layer_presenter.rb
```diff
@@ -333,6 +333,8 @@ def migrate(options)
 
         type = if @source_type == 'density'
                  wpp['geometry_type'] == 'Rectangles' ? 'squares' : 'hexabins'
+               elsif @source_type == 'torque_heat'
+                 wpp['heat-animated'] ? 'animated' : 'heatmap'
```
juanignaciosl: Value comparison of `wpp['heat-animated']` is not needed? (`wpp['heat-animated']` == 'true'`, for example).

app/controllers/carto/api/layer_presenter.rb
```diff
@@ -365,10 +367,9 @@ def migrate(options)
         'bubble' => 'simple',
         'choropleth' => 'simple',
         'category' => 'simple',
-        'torque' => 'simple',
-        'torque_cat' => 'simple',
-        'cluster' => 'simple',
-        'torque_heat' => 'heatmap'
+        'torque' => 'animated',
+        'torque_cat' => 'animated',
+        'cluster' => 'simple'
```
juanignaciosl: How come `torque_heat` is removed? What's the expected behaviour if one is migrated?

juanignaciosl:
I've made some comments, probably caused by things that I don't know about the PR. As I've said, this change (having both an `animated` type and `animated` attribute) feels weird with the Builder approach, but it it's good for you it is for me as well... 👍

app/controllers/carto/api/layer_presenter.rb
javitonino: It is handled [above](https://github.com/CartoDB/cartodb/pull/9259/files/3a273684db9467c7b3fd38fca2e43719e30964f7#diff-0cd9478b055c01a5bd92c7b398117bfaR337) as it can migrate to two different types depending if it is animated or not.

app/controllers/carto/api/layer_presenter.rb
javitonino: It comes as a boolean value, so normal boolean logic works here :smile:

PR #9266, opened 2016-08-03
Check if the query is already applied before show the notification.

### LINKED ISSUE #9260: Notification error in SQL editor
Steps to reproduce:

- With the default sql query applied, forced an error. For instance:
`SELECT * + FROM paradas_metro_madrid`
- Save the query, at this point the error shows up.
- Fix the error, removing the `+` and save again.
- A `Applying query` notification is shown but it never ends 💥

nobuti:
This PR fixes #9260.

Because the applying notification listens the vis reload event to change its state (and eventually autoclose it), we check if the query is already applied. In this case, no new map request is made, so no need to show the notification.

cc @xavijam


xavijam:
🇪🇸, but I miss a test.

xavijam:
Merge!

Issue #9214, opened 2016-08-02
Adapt importer connector parameters to changes in odbc_fdw


jgoizueta:
Next release of odbc_fdw has changed the naming of connection parameters (https://github.com/CartoDB/odbc_fdw/pull/22).

The Connector class must prefix connection parameters with  `odbc_` and preserve upper/lowercase characters.

PR #9269, opened 2016-08-03
Adapt to new odbc_fdw API

jgoizueta:
See #9214 

Connection parameters must be now prefixed by `odbc_` in the odbc-fdw options and must
preserve the case (since some parameter names may be case-sensitive)

### DIFF
--- a/services/importer/lib/importer/connector.rb
+++ b/services/importer/lib/importer/connector.rb
@@ -128,17 +128,17 @@ module CartoDB
 
       def server_params
         params = @connection.select { |k, _v| k.downcase.in? SERVER_OPTIONS }
-        conection_options params
+        connection_options params
       end
 
       def user_params
         params = @connection.select { |k, _v| k.downcase.in? USER_OPTIONS }
-        conection_options params
+        connection_options params
       end
 
       def table_params
         params = @connection.reject { |k, _v| k.downcase.in? SERVER_OPTIONS + USER_OPTIONS }
-        conection_options(params).merge(@options)
+        connection_options(params).merge(@options)
       end
 
       # Parse @json_params and extract @params
--- a/services/importer/lib/importer/connector.rb
+++ b/services/importer/lib/importer/connector.rb
@@ -116,7 +116,7 @@ module CartoDB
 
       def fetch_ignoring_case(hash, key)
         if hash
-          k, _v = hash.find { |k, _v| k.to_s.downcase == key.to_s.downcase }
+          k, _v = hash.detect { |k, _v| k.to_s.casecmp(key.to_s) == 0 }
           k
         end
       end
--- a/services/importer/lib/importer/connector.rb
+++ b/services/importer/lib/importer/connector.rb
@@ -238,7 +238,8 @@ module CartoDB
           options = table_params.merge(prefix: foreign_prefix)
           schema = @options['schema'] ||
                    fetch_ignoring_case(@connection, 'schema') ||
-                   fetch_ignoring_case(@connection, 'database')
+                   fetch_ignoring_case(@connection, 'database') ||
+                   'public'
           %{
             IMPORT FOREIGN SCHEMA #{schema}
               FROM SERVER #{server_name}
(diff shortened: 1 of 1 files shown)


services/importer/lib/importer/connector.rb
```diff
@@ -108,27 +108,55 @@ def self.check_availability!(user)
...
-      SERVER_OPTIONS = %w(dsn driver host port database).freeze
+      ACCEPTED_PARAMETERS = %w(connection schema table sql_query sql_count encoding columns).freeze
       REQUIRED_PARAMETERS = %w(connection table).freeze
 
+      SERVER_OPTIONS = %w(dsn driver host server address port database).freeze
+      USER_OPTIONS   = %w(uid pwd user username password).freeze
+
+      def fetch_ignoring_case(hash, key)
+        if hash
+          k, _v = hash.find { |k, _v| k.to_s.downcase == key.to_s.downcase }
```
jgoizueta: Arrg!! my ruboto tells me "Prefer `find` over `detect`! Who decides these things?

jgoizueta:
Please @rafatower  CR this

services/importer/lib/importer/connector.rb
rafatower: https://github.com/CartoDB/cartodb/blob/master/.rubocop.yml ?

services/importer/lib/importer/connector.rb
```diff
@@ -108,27 +108,55 @@ def self.check_availability!(user)
...
 
-      ACCEPTED_PARAMETERS = %w(connection schema table sql_query sql_count encoding).freeze
-      SERVER_OPTIONS = %w(dsn driver host port database).freeze
+      ACCEPTED_PARAMETERS = %w(connection schema table sql_query sql_count encoding columns).freeze
       REQUIRED_PARAMETERS = %w(connection table).freeze
 
+      SERVER_OPTIONS = %w(dsn driver host server address port database).freeze
+      USER_OPTIONS   = %w(uid pwd user username password).freeze
+
+      def fetch_ignoring_case(hash, key)
```
rafatower: I don't understand this function.

If it returns just `k`, shouldn't it be called `fetch_key_ignoring_case`?

If `hash` is `nil` then it will raise an exception ```NameError: undefined local variable or method `k' ```, won't it?

services/importer/lib/importer/connector.rb
```diff
@@ -108,27 +108,55 @@ def self.check_availability!(user)
...
+        if hash
+          k, _v = hash.detect { |k, _v| k.to_s.casecmp(key.to_s) == 0 }
```
rafatower: and now this is supposed to be the right thing to do :smile:

services/importer/lib/importer/connector.rb
jgoizueta: You've spotted a bug; should return v, but the way it's used it was working by chance. thanks!

A nil hash shouldn't be passed to this function, but I'll give it a thought, in case we should be reslient to that.

services/importer/lib/importer/connector.rb
jgoizueta: Arggggg I ended up leaving the wrong one!

services/importer/lib/importer/connector.rb
```diff
@@ -138,23 +166,23 @@ def validate_params!
...
-        elsif params['dsn'].blank? && params['driver'].blank?
+        elsif @dsn.blank? && @driver.blank?
           errors << "Missing required connection parameters"
         end
         errors
       end
 
       def connector_name
-        Carto::DB::Sanitize.sanitize_identifier @params['connection']['dsn'] || @params['connection']['driver']
+        Carto::DB::Sanitize.sanitize_identifier @dsn || @driver
```
rafatower: I did not notice the dsn can act as "connector name"

services/importer/lib/importer/connector.rb
rafatower: just asking, the simplest thing to do is to let it raise an exception and even do not check for "nilness" as it is supposed to be a hash always.

services/importer/lib/importer/connector.rb
jgoizueta: We are just adding some meaningful prefix to the connector name, DSN, if present should be    a good description; otherwise we use the driver name. (Note that if DSN is present, the driver name likely won't be available)

Issue #9224, opened 2016-08-02
Review options vertical spacing


saleiva:
Looks to separated for me

PR #9275, opened 2016-08-03
fix spaces

piensaenpixel:
@nobuti 👓 

image

Close https://github.com/CartoDB/cartodb/issues/9224


nobuti:
👍  not sure if we should add a helper for this in the same manner as `u-tSpace--m` and so on.

piensaenpixel:
we need remove the margin, not change it :( and the component is common for all the builder, isn't it?

piensaenpixel:
@nobuti green light? :D

nobuti:
@piensaenpixel first thing in the morning, I promise 😞

nobuti:
🇪🇸

PR #9287, opened 2016-08-03
9265 add published event

### LINKED ISSUE #9265: [Metrics] Add new 'Published map' event
Generate a new event each time a user generates a new version of the map.

Event properties to track:
 - Usual user properties (id, email, username, plan, ...)
 - Version number (I don't know if we store this)
 - Type of publish (link, embed, carto.js, mobile... I don't know if we can have this)

If the last two properties are costly to get, we can continue without them at this moment.

@gfiorav

gfiorav:
Fixes #9265 

Please CR, @javitonino 

cc/ @jmartinter

### DIFF
--- a/app/controllers/carto/api/mapcaps_controller.rb
+++ b/app/controllers/carto/api/mapcaps_controller.rb
@@ -1,6 +1,8 @@
 # encoding: utf-8
 require_relative '../builder/builder_users_module'
 
+require_dependency 'carto/tracking/events'
+
 module Carto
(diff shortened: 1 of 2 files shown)


javitonino:
If there is no need to run this on the initial visualization creation (which is kind of like a publish), :ship: it.

gfiorav:
Nah, it's cool. Just publish for now 😄 

We have another event for created map already, and that contains info about the privacy.

PR #9290, opened 2016-08-04
Remove tipsy on destroy method.

### LINKED ISSUE #9280: Analysis error tooltip not removed properly
STR:
- Create a new map
- Add an analysis and make it fail to displey the red error cue/dot, (e.g. `Predict trends and volatility` fails for me in local dev env, probably missing some new extension/DB functionality)
- Hover the dot to display the error tooltip, click it when it appears.

Expected result:
- Close the tooltip See the analysis details/form

Actual results:
- Tooltip stays and never goes away, until reloading the page, e.g.:
screen shot 2016-08-03 at 17 41 35

AFAIK the tooltip was implemented in #9084, probably missing some cleanup code somewhere there.

nobuti:
This PR fixes #9280.

cc @viddo

### DIFF
--- a/app/assets/stylesheets/editor-3/block-list.css.scss
+++ b/app/assets/stylesheets/editor-3/block-list.css.scss
@@ -121,21 +121,6 @@
 .HorizontalBlockList-item.is-selected {
   color: $cWhite;
 }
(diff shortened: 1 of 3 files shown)


viddo:
👍

PR #9294, opened 2016-08-04
Replace Sequel method rename_column by SQL code

### LINKED ISSUE #9273: Replace use of Sequel::Postgres::Database#rename_column
The method `rename_column` of a Sequel db connection quotes the whole name passed, so we cannot provided a qualified name (one including an schema) to it:

```
user.in_database.rename_column 'aaa.bbb', 'x', 'y'
# => ALTER TABLE "aaa.bbb" RENAME COLUMN "x" TO "y"

user.in_database.rename_column '"aaa".bbb', 'x', 'y'
# => ALTER TABLE """aaa"".bbb" RENAME COLUMN "x" TO "y"

user.in_database.rename_column '"aaa"."bbb"', 'x', 'y'
# => ALTER TABLE """aaa"".""bbb""" RENAME COLUMN "x" TO "y"
```

In addition, we're passing already quoted schema+table names causing invalid SQL to be generated [here](https://github.com/CartoDB/cartodb/blob/15a105ba2e61631251b9123b795206e27423b859/app/models/synchronization/adapter.rb#L157)

This can be solved simply by replacing that line with:

```
owner.in_database.execute "ALTER TABLE #{qualified_table_name} RENAME COLUMN \"#{THE_GEOM}\" TO \"the_geom_str\""
```

We should replace also other uses of the method to avoid problems.

In the particular line mentioned above we should also consider if we should

Note that this code seems to not have been regularly used since  `ogr2ogr` already handles existing `the_geom` columns, but is now more relevant because connector imports don't go through `ogr2ogr`.

jgoizueta:
Fixes #9273

### DIFF
--- a/app/models/table.rb
+++ b/app/models/table.rb
@@ -926,7 +926,9 @@ class Table
       if Table.column_names_for(user_database, name, self.owner).include?(new_name)
         raise 'Column already exists'
       end
-      user_database.rename_column(name, old_name.to_sym, new_name.to_sym)
+      user_database.execute %{
+          ALTER TABLE "#{name}" RENAME COLUMN "#{old_name}" TO "#{new_name}"
+      }
     end
   end #rename_column
 
@@ -1497,7 +1499,9 @@ class Table
             type = DEFAULT_THE_GEOM_TYPE
           end
         else
-          owner.in_database.rename_column(qualified_table_name, THE_GEOM, :the_geom_str)
+          owner.in_database.execute %{
+              ALTER TABLE #{qualified_table_name} RENAME COLUMN "#{THE_GEOM}" TO "the_geom_str"
+          }
         end
       else # Ensure a the_geom column, of type point by default
         type = DEFAULT_THE_GEOM_TYPE
--- a/app/models/synchronization/adapter.rb
+++ b/app/models/synchronization/adapter.rb
@@ -154,7 +154,9 @@ module CartoDB
         return nil unless the_geom_data
 
         if the_geom_data[:typname] != 'geometry'
-          user.in_database.rename_column(qualified_table_name, THE_GEOM, :the_geom_str)
+          user.in_database.execute %{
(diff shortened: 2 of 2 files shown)


jgoizueta:
Please @ethervoid review this

services/importer/lib/importer/connector.rb
```diff
@@ -286,10 +286,16 @@ def drop_foreign_table_command
       end
 
       def execute_as_superuser(command)
+        puts "\nEXECUTING AS SUPERUSER:"
```
jgoizueta: Ignore changes to this file, I committed them by error, I'll rebase this

ethervoid:
LGTM

app/models/table.rb
```diff
@@ -926,7 +926,9 @@ def rename_column(old_name, new_name='')
       if Table.column_names_for(user_database, name, self.owner).include?(new_name)
         raise 'Column already exists'
       end
-      user_database.rename_column(name, old_name.to_sym, new_name.to_sym)
+      user_database.execute %{
+          ALTER TABLE "#{name}" RENAME COLUMN "#{old_name}" TO "#{new_name}"
```
rafatower: is the name the qualified table name? could this be a problem with org users?

jgoizueta: I'm not sure, but if was working before it should work now: rename_column always quotes the table name. So this would failed for a qualified name (now and before). 
I'll do a quick check to verify Table name is unqualified for org users.

jgoizueta: Verified, Table#name is only the table name, for org users too.

rafatower: then it works as long as it is in the search path. I'd try to rename a column with an org user unless there's some code around following the same.

but IIRC there was another method to return the fully qualified table name (with "schema"."table")

jgoizueta: I'll look at it and open a separate ticket is that should be fixed. (we're not changing current behaviour for this method here.

PR #9296, opened 2016-08-04
Added superadmin method to get database FS usage

rporres:
This method assumes that the following function has to be installed in the database hosts:

``` sql
DROP FUNCTION cartodb.cdb_get_avail_disk_space();
DROP TYPE cartodb.cdb_fs_stats_info;
CREATE TYPE cartodb.cdb_fs_stats_info AS (avail bigint, total bigint);
CREATE FUNCTION cartodb.cdb_get_avail_disk_space()
  RETURNS cartodb.cdb_fs_stats_info
AS $$
  import os

  rv = plpy.execute("select current_setting('data_directory') AS data_directory")
  directory = rv[0]['data_directory']

  if directory:
    vfs = os.statvfs(directory)
    avail = vfs.f_bsize * vfs.f_bavail
    total = vfs.f_bsize * vfs.f_blocks
    return [avail, total]
  else:
    return [-1,-1]
$$ LANGUAGE plpythonu;
```

A bit unsure about how to test this. I guess the best way would be to make sure that the `cartodb.cdb_get_avail_disk_space()` is added to the test database and then create a new spec for superadmin platform controller (I'd rather do this than mocking the call to the db)

@luisbosque: please review

### DIFF
--- a/app/controllers/superadmin/platform_controller.rb
+++ b/app/controllers/superadmin/platform_controller.rb
@@ -72,15 +72,24 @@ class Superadmin::PlatformController < Superadmin::SuperadminController
       return
     end
 
-    connection_params = ::Rails::Sequel.configuration.environment_for(Rails.env)
-                                       .merge('host' => params[:database_host],
-                                              'database' => 'postgres'
-                                             ) { |_, o, n| n.nil? ? o : n }
-    conn = ::Sequel.connect(connection_params)
-    disk_space = conn.fetch('SELECT * FROM cartodb.cdb_get_avail_disk_space();').first
-    close_sequel_connection(conn)
+    begin
+      connection_params = ::Rails::Sequel.configuration.environment_for(Rails.env)
+                                         .merge('host' => params[:database_host],
+                                                'database' => 'postgres'
+                                               ) { |_, o, n| n.nil? ? o : n }
+      conn = ::Sequel.connect(connection_params)
+      message = conn.fetch('SELECT * FROM cartodb.cdb_get_avail_disk_space();').first
+      status = 200
+    rescue => e
+      message = { error: 'Error getting available disk space in DB' }
+      status = 500
+      CartoDB::Logger.error(message: message[:error], exception: e,
+                            database_host: params[:database_host])
+    ensure
+      close_sequel_connection(conn)
+    end
 
-    respond_with(disk_space)
+    respond_with(message, status: status)
   end
 
 end # Superadmin::PlatformController
--- a/app/controllers/superadmin/platform_controller.rb
+++ b/app/controllers/superadmin/platform_controller.rb
@@ -67,7 +67,7 @@ class Superadmin::PlatformController < Superadmin::SuperadminController
 
   def database_host_fs_usage
     unless params[:database_host]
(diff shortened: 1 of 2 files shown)


luisbosque:
LGTM 
/cc @javitonino

app/controllers/superadmin/platform_controller.rb
```diff
@@ -63,4 +65,22 @@ def total_likes
     respond_with({:count => CartoDB::Stats::Platform.new.likes})
   end
 
+  def database_host_fs_usage
+    unless params[:database_host]
```
javitonino: As this is a mandatory parameter, you could pass it in the URL instead and avoid having to do this check. But it's a design choice, so up to you.

app/controllers/superadmin/platform_controller.rb
```diff
@@ -63,4 +65,22 @@ def total_likes
...
+      return
+    end
+
+    connection_params = ::Rails::Sequel.configuration.environment_for(Rails.env)
+                                       .merge('host' => params[:database_host],
+                                              'database' => 'postgres'
+                                             ) { |_, o, n| n.nil? ? o : n }
+    conn = ::Sequel.connect(connection_params)
+    disk_space = conn.fetch('SELECT * FROM cartodb.cdb_get_avail_disk_space();').first
+    close_sequel_connection(conn)
```
javitonino: :+1: Nice to see that closing method.

Can you put it in an `ensure` block so it runs even in case of exception. Something like:
```
begin
   connection_params = ::Rails::Sequel.configuration.environment_for(Rails.env)
                                      .merge('host' => params[:database_host],
                                             'database' => 'postgres'
                                            ) { |_, o, n| n.nil? ? o : n }
   conn = ::Sequel.connect(connection_params)
   disk_space = conn.fetch('SELECT * FROM cartodb.cdb_get_avail_disk_space();').first
   respond_with(disk_space)
rescue => e
  CartoDB::Logger.error(message: 'Error getting available disk space in DB', exception: e, database_host: params[:database_host])
   head 500
ensure
   close_sequel_connection(conn)
end
```

javitonino:
This looks good, but I would like to see the connection closing in an ensure block to be certain that we have no connection leaking (Sequel is a bit tricky with that).

app/controllers/superadmin/platform_controller.rb
rporres: Hmmm, then I would have to escape somehow if database host contains dots. Rails doesn't like routes like `/superadmin/stats/database_host_fs_usage/127.0.0.1`

javitonino: Yeah, didn't think about dots. It will probably still work, but need to add a constraint or something like that. Anyway, not important and may be problematic, so let's just skip it.

rporres:
Fixed as per @javitonino suggestions

javitonino:
:+1: when test pass. Please, do not merge today (we have frozen deploys until tomorrow)

rporres:
retest this please

rporres:
Tagging it as do not merge yet as we need to see how to install the extension mentioned in the summary (will need that to be able to write a test)

PR #9298, opened 2016-08-04
Adding 'opacity' to labels text and halo

### LINKED ISSUE #9289: In labels, text transparency is not showing on the map
Under styles and labels, text transparency updated don't show on the map

captura de pantalla 2016-08-04 a las 10 00 43

xavijam:
AFAIK (please @rochoa correct me if so) there is no opacity attribution for labels (in fact in the old editor we didn't let the user change the opacity in the form), but the good thing is we can make use of the RGBA property, so the result would be:

screen shot 2016-08-04 at 12 21 28
screen shot 2016-08-04 at 12 27 01

CR: @javierarce
Fixes #9289.

Sounds ok?
cc @saleiva @noguerol 

Any possible problem I haven't seen @rochoa?

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/style/style-converter.js
+++ b/lib/assets/javascripts/cartodb3/editor/style/style-converter.js
@@ -1,5 +1,6 @@
 var _ = require('underscore');
 var camshaftReference = require('../../data/camshaft-reference');
+var Utils = require('../../helpers/utils');
 
 var CONFIG = {
   HEATMAP_IMAGE: 'url(http://s3.amazonaws.com/com.cartodb.assets.static/alphamarker.png)',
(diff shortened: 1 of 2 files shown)


lib/assets/javascripts/cartodb3/editor/style/style-converter.js
```diff
@@ -186,12 +187,22 @@ function _labels (props) {
     css['text-face-name'] = "'" + props.font + "'";
     if (props.fill) {
       css['text-size'] = props.fill.size.fixed;
-      css['text-fill'] = props.fill.color.fixed;
+
+      if (props.fill.color.opacity != null && props.fill.color.opacity < 1) {
```
javierarce: `!=` ? With just one equal?

xavijam: Because it checks it is different that `null` and `undefined`.

javierarce: Ok, just to make sure it wasn't a mistake :)

javierarce:
👍

PR #9299, opened 2016-08-04
Add proper fields for the form.

nobuti:
This is a hotfix for time series widget with no proper fields configured to make the form.

cc @xavijam


xavijam:
merge!

Issue #9193, opened 2016-08-01
Switching the basemap very quickly creates multiple labels layers


alonsogarciapablo:
**Steps to reproduce:**

- Enable network throttling in your inspector:
  <img width="416" alt="screen shot 2016-08-01 at 13 15 54" src="https://cloud.githubusercontent.com/assets/390398/17292425/353937da-57ea-11e6-9143-03fc6bec4c01.png">
- Open a map.
- Click on the base layer.
- Switch quickly, and multiple times, between a basemap with labels and a basemap with no labels on top.
- Refresh the page.

**Expected result:**
- There is a single layer with labels.

**Actual result:**
- There are multiple labels layers:
<img width="433" alt="screen shot 2016-08-01 at 13 14 48" src="https://cloud.githubusercontent.com/assets/390398/17292491/bc5b9a1e-57ea-11e6-8b08-459a5143f2b5.png">

I think the problem is the following:

1. A basemap with labels is selected -> A request to create the labels is triggered.
2 A basemap with no labels is selected -> [We try to destroy the labels layer](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js#L207).
3. The request from 1. finishes and the labels layer is there.

In the past, we talked about disabling the "basemap" selector while basemap is begin changed (i.e: requests are still running). I think that's how we solved this in the "old" editor.

Also, this might be why the map in https://github.com/CartoDB/cartodb/issues/9111 got screwed up.

cc: @matallo @xavijam

matallo:
on it

xavijam:
Important in order to prevent those kind of issues, already discussed with master @matallo.

PR #9305, opened 2016-08-04
disable basemaps on basemaps change

matallo:
close https://github.com/CartoDB/cartodb/issues/9193

also:

- add scrollview to infowindows
- add scrollview to basemaps view
- add notifier


lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
```diff
@@ -265,16 +288,20 @@ module.exports = Backbone.Collection.extend({
     return this.size() > 1 && layerTypesAndKinds.isTiledType(this.getLayerOnTop().get('type'));
   },
 
-  _updateLabelsLayer: function (newBaseLayerAttrs) {
+  _updateLabelsLayer: function (newBaseLayerAttrs, opts) {
     var labelsLayer = this.getLayerOnTop();
-    var labelsLayerAttrs = _.extend(newBaseLayerAttrs, {
```
matallo: I had to do this change because otherwise it would set an empty `Id` then it would try to save as a _new_ layer, we don't want this.

xavijam:
Gif or screenshot would be appreciated :)

lib/assets/javascripts/cartodb3/editor/layers/basemap-content-view.js
```diff
@@ -165,14 +171,50 @@ module.exports = CoreView.extend({
...
+      status: 'loading',
+      info: _t('editor.layers.basemap.saving.loading'),
+      closable: false
+    });
+  },
+
+  _onsavingLayersSuccess: function () {
+    this.model.set('disabled', false);
+
+    this.notification.set({
```
xavijam: I feel you can use update method, but @nobuti knows more about that.

lib/assets/javascripts/cartodb3/editor/layers/basemap-content-view.js
```diff
@@ -165,14 +171,50 @@ module.exports = CoreView.extend({
...
+      status: 'error',
+      info: _t('editor.layers.basemap.saving.error'),
+      closable: true
+    });
+  },
+
   _initBinds: function () {
+    this.model.bind('change:disabled', function () {
+      this.$el.find('.js-basemapContent').toggleClass('is-disabled', this.model.get('disabled'));
```
xavijam: `this.$el.find('.js-basemapContent')` === `this.$('.js-basemapContent')`

lib/assets/javascripts/cartodb3/components/mosaic/mosaic-item-view.js
```diff
@@ -41,6 +43,10 @@ module.exports = CoreView.extend({
   },
 
   _onClick: function () {
+    if (this._disabled) {
```
xavijam: aren't we taking into account disabled state when mosaic is rendered?

lib/assets/javascripts/cartodb3/editor/layers/basemap-content-view.js
nobuti: Yes, feel free to use it. It's wrapper for `set`, longer to write but more readable because it has sense of continuity. Also, in the same line, it has individual setters, aka, updateStatus, updateInfo, updateClosable and so on.

lib/assets/javascripts/cartodb3/components/mosaic/mosaic-item-view.js
matallo: actually it's the `basemap-inner-view` which is or isn't disabled, it passes this state to the childs

matallo:
disable

alonsogarciapablo:
Great @matallo! 💪

matallo:
actually there's some drawback I spotted during review, as we're storing state in the `basemap content-view.js` model https://github.com/CartoDB/cartodb/pull/9305/files#diff-b8bdcc7badd276a62ad45b75bda61453R41 each time we go back to layers list and forth to the layer with the baselayer being changed it creates a new model with `disabled: false` so we'd need some higher level model to store this state to the atached layer, for every layer

well, I'm not pretty sure about the best approach for this, it is only my suggestion

opinions @CartoDB/frontend ?

Issue #9308, opened 2016-08-04
Store visualization tokens in the DB


javitonino:
Related to #9174 and #9175

Right now, password protected visualizations have the auth tokens stored in the named map, so it has to be requested to the tiler each time the embed is shown. Better to store it in the DB where it is easier to retrieve and makes more sense (also more consistent with other auth tokens).

PR #9312, opened 2016-08-04
Include only the tokens for the entities in permissions in the named map

### LINKED ISSUE #9175: Too many auth tokens sent to named maps api

javitonino:
Fixes #9175

### DIFF
--- a/app/models/carto/visualization.rb
+++ b/app/models/carto/visualization.rb
@@ -356,7 +356,8 @@ class Carto::Visualization < ActiveRecord::Base
   end
 
   def allowed_auth_tokens
-    permission.entities_with_read_permission.map(&:get_auth_token)
+    entities = [user] + permission.entities_with_read_permission
+    entities.map(&:get_auth_token)
   end
 
   def mapcapped?
--- a/app/models/carto/visualization.rb
+++ b/app/models/carto/visualization.rb
@@ -355,8 +355,8 @@ class Carto::Visualization < ActiveRecord::Base
     CartoDB::Varnish.new.purge(varnish_vizjson_key)
   end
 
-  def all_users_with_read_permission
-    permission.users_with_permissions([CartoDB::Visualization::Member::PERMISSION_READONLY]).push(user)
+  def allowed_auth_tokens
+    permission.entities_with_read_permission.map(&:get_auth_token)
   end
 
   def mapcapped?
--- a/lib/carto/named_maps/template.rb
+++ b/lib/carto/named_maps/template.rb
@@ -227,12 +227,7 @@ module Carto
         method, valid_tokens = if @visualization.password_protected?
                                  [AUTH_TYPE_SIGNED, @visualization.user.get_auth_tokens]
                                elsif @visualization.organization?
-                                 auth_tokens = @visualization.all_users_with_read_permission
-                                                             .map(&:get_auth_tokens)
-                                                             .flatten
-                                                             .uniq
-
-                                 [AUTH_TYPE_SIGNED, auth_tokens]
+                                 [AUTH_TYPE_SIGNED, @visualization.allowed_auth_tokens]
                                else
                                  [AUTH_TYPE_OPEN, nil]
                                end
--- a/spec/models/carto/user_spec.rb
+++ b/spec/models/carto/user_spec.rb
@@ -11,6 +11,10 @@ describe Carto::User do
     def get_user_by_id(user_id)
       Carto::User.where(id: user_id).first
     end
+
+    def create_user
+      FactoryGirl.create(:carto_user)
+    end
   end
 
   describe '#needs_password_confirmation?' do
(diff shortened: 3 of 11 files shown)


javitonino:
Heyo, it's CR time, please @gfiorav

app/models/carto/visualization.rb
```diff
@@ -355,8 +355,9 @@ def invalidate_cache
     CartoDB::Varnish.new.purge(varnish_vizjson_key)
   end
 
-  def all_users_with_read_permission
-    permission.users_with_permissions([CartoDB::Visualization::Member::PERMISSION_READONLY]).push(user)
+  def allowed_auth_tokens
+    entities = [user] + permission.entities_with_read_permission
```
gfiorav: Aren't you getting the owner twice here?

lib/carto/named_maps/template.rb
```diff
@@ -227,12 +227,7 @@ def auth
         method, valid_tokens = if @visualization.password_protected?
                                  [AUTH_TYPE_SIGNED, @visualization.user.get_auth_tokens]
                                elsif @visualization.organization?
-                                 auth_tokens = @visualization.all_users_with_read_permission
-                                                             .map(&:get_auth_tokens)
-                                                             .flatten
-                                                             .uniq
-
-                                 [AUTH_TYPE_SIGNED, auth_tokens]
+                                 [AUTH_TYPE_SIGNED, @visualization.allowed_auth_tokens]
```
gfiorav: This is now redindant right? Password protected and org shared can use `.allowed_auth_tokens`?

What will happen when a user shares a password protected visualization within the org?

spec/models/carto/user_spec.rb
```diff
@@ -11,6 +11,10 @@ def get_twitter_imports_count_by_user_id(user_id)
     def get_user_by_id(user_id)
       Carto::User.where(id: user_id).first
     end
+
+    def create_user
+      FactoryGirl.create(:carto_user)
+    end
```
gfiorav: Why do you need this abstraction?

app/models/carto/visualization.rb
javitonino: No, in fact I had to add this for existing tests to pass (have a :heart: for that). Permissions only includes the permissions granted in the modal, which may not include the owner (and usually doesn't).

spec/models/user_spec.rb
```diff
@@ -25,6 +25,10 @@ def get_twitter_imports_count_by_user_id(user_id)
     def get_user_by_id(user_id)
       ::User.where(id: user_id).first
     end
+
+    def create_user
+      FactoryGirl.create(:valid_user)
+    end
```
gfiorav: Same here

app/models/carto/visualization.rb
gfiorav: Weird! But ok I guess :grin:

lib/carto/named_maps/template.rb
javitonino: Yes, it could be made redundant, but right now it isn't. I will approach this when I do the changes to password protected visualizations #9308. This is only for organizations to keep PR's manageable. Also, this allows me to deploy this without waiting for cartodb.js changes.

gfiorav:
Good tests! 

I have some concerns regarding sharing within org with passwords, so I'll let you review de comments before proceeding.

Nice work!

spec/models/carto/user_spec.rb
javitonino: It is used in the shared examples so the same test applies to both user models: https://github.com/CartoDB/cartodb/pull/9312/files/0e1dc708201d1ec29e35730100b8a1c55cee2c28#diff-4b796520136c6d36d55f0ba58ab80103R447

javitonino:
Password protected visualizations shouldn't have changed at all. If you feel like I may have changed some behaviour regarding them, let me know. I intend to approach them in #9308, but not for now (baby steps).

gfiorav:
Yup, got confused. All right, 📦  it!

PR #9317, opened 2016-08-04
Use category/measure terminology in the Builder

talos:
@saleiva 

In the Observatory, measures are the actual measurements, and they are classified by category.  The vocabulary as originally implemented was confusing -- this brings it into line with the rest of the Observatory.


lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/data-observatory-measure-model.js
```diff
@@ -48,7 +48,7 @@ module.exports = BaseAnalysisFormModel.extend({
   },
 
   _setSegment: function () {
-    var s = this.doMetadata.segment(this.get('area'), this.get('measurement'), this.get('segments'));
+    var s = this.doMetadata.segment(this.get('area'), this.get('categories'), this.get('measure'));
```
xavijam: Hey, we can't change the attribute names as easy as that ^^^ because the analysis will not work due to camshaft will request `measurement` and `segments` attributes, so we can't change that.

saleiva: So is it a problem with the texts on the UI? Or a problem with how it works?

talos: It's a problem with UI.  But it's confusing, and we should fix it before we name more things confusingly.

I will open a PR for Camshaft.

talos: @xavijam I actually think this is safe to deploy without updating Camshaft.  I didn't change the vocabulary around `segment_name`, which is the only part that touches Camshaft (since Camshaft doesn't care about the way the user selects the ID for the measure):

https://github.com/CartoDB/camshaft/blob/master/lib/node/nodes/data-observatory-measure.js#L9

In the future, we should fix the `segment_name` vocab, too, but it's less essential because:

1. It doesn't touch users at all, it's just confusing for devs
2. It requires Camshaft and the UI be deployed at exactly the same time (not sure how these get bundled)

xavijam: Have you tested it in your local env and/or staging?

talos:
@saleiva would it be possible to assign this to someone who could carry it through?  The vocab right now is confusing, and this should be an easy fix.

saleiva:
@javisantana any help here?

xavijam:
Hey mates, UI is gonna change so this ticket will be out-of-date very soon. Closing it ok?

talos:
Fine with me.

xavijam:
:+1:

Issue #9284, opened 2016-08-03
Analysis onboarding notification doesn't remember the forget me checkbox status


javierarce:
Discovered by @viddo in the `trade-area` analysis. 

How to reproduce it:

1. Create and apply a trade-area analysis.
2. Check the checkbox of the "onboarding" window and close the window.
3. Remove and add the same analysis again.

The window will appear, even though the checkbox to skip the screen was checked.

PR #9324, opened 2016-08-05
Fixes "Never show me this message" for certain analysis

javierarce:
This PR fixes an issue which caused the onboarding window for analysis to appear when users checked the "Never show me this message." box. 

The problem was that certain analyses share a common "generic type" and the analysis onboarding launcher wasn't using that generic type, but the analysis type (which is different for every kind of analysis).

Closes: #9284 

CR: @xavijam

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/onboardings/analysis/analysis-launcher.js
+++ b/lib/assets/javascripts/cartodb3/components/onboardings/analysis/analysis-launcher.js
@@ -37,7 +37,7 @@ var launcher = (function () {
     },
 
     launch: function (type, model) {
-      var genericType = ANALYSES_TYPES[type] && ANALYSES_TYPES[type].genericType ? ANALYSES_TYPES[type].genericType : type;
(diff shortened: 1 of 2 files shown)


lib/assets/javascripts/cartodb3/components/onboardings/analysis/analysis-launcher.js
```diff
@@ -37,7 +37,9 @@ var launcher = (function () {
     },
 
     launch: function (type, model) {
-      if (!LocalStorage.get(type) && ANALYSES_TYPES[type]) {
+      var genericType = ANALYSES_TYPES[type] && ANALYSES_TYPES[type].genericType ? ANALYSES_TYPES[type].genericType : type;
```
xavijam: ANALYSES_TYPES[type] && ANALYSES_TYPES[type].genericType || type?

xavijam:
Worth it a test?

javierarce:
I don't think so, that would be throwing out money.

xavijam:
:es:

PR #9336, opened 2016-08-05
text fix on modification of tables

ohasselblad:
cc @xavijam


xavijam:
Thanks @ohasselblad 💋

PR #9341, opened 2016-08-08
9174 remove auth tokens from vizjson

### LINKED ISSUE #9174: Auth tokens in vizjson3 present in privately shared visualizations
A bit of context. Auth tokens are used to authenticate calls to the tiler, and are used in two cases:
 - Password-protected visualizations
 - Private visualizations shared within the organization

In vizjson2, the password protected visualizations include the necessary tokens in the vizjson, and the shared visualizations include it in the embed/public view like so:
```
map_options.auth_token = [];
map_options.auth_token.push('Fv7Wi4BsA6tCNaN7r2qGtg');
map_options.auth_token.push('XVQSox-kPIvk77kJh6ph-g');
map_options.https = true; // when auth_token is used tiles should be fetch using https
```

However, in vizjson3 and the new embed, both cases are handled by including the tokens in the vizjson. This has a serious implication, all the auth tokens allowed for the visualization are included, so you can see the auth tokens for the whole organization in the vizjson.

This was introduced here as a hotfix for password protected embed: #8660 that produces auth tokens in the vizjson for privately shared visualizations. Reverting this (well, without breaking what it was fixing) is not enough, as that leaves no way to pass the auth tokens for privately shared visualizations to cartodb.js.

I don't know exactly how cartodb.js v4 handles this, but there is probably an option to the `createVis` call. We would need to write the tokens to the page and modify the public js to pass it to cartodb.js.

Maybe we can do this entirely in backend, including the tokens from the viewer user in the vizjson, but that makes little conceptual sense to me, and we must be careful with caching in that case. I'd avoid this.

cc @alonsogarciapablo @xavijam @juanignaciosl

### LINKED ISSUE #9276: Text in protected embed
In embed for maps that are password protected, incorrect font:

screenshot_20160803_163635

javitonino:
Moves auth tokens outside the vizjson, to an embed variable.

Uses branch from https://github.com/CartoDB/cartodb.js/pull/1377 (not yet merged)

Closes #9174 
Closes #9276
screenshot_20160808_124815


javitonino:
Please CR @gfiorav

app/controllers/carto/builder/public/embeds_controller.rb
```diff
@@ -34,6 +35,14 @@ def load_visualization
           render_404 unless @visualization
         end
 
+        def load_auth_tokens
+          @auth_tokens = if @visualization.password_protected?
+                           @visualization.get_auth_tokens
+                         elsif @visualization.organization?
+                           current_viewer.get_auth_tokens
```
gfiorav: What if user isn't logged in?

gfiorav:
Sweet stuff! Great tests.

After conflicts and comment it's 📦 ready

app/controllers/carto/builder/public/embeds_controller.rb
javitonino: Great catch. One thing led to another and I ended up with https://github.com/CartoDB/cartodb/pull/9351. I'll recheck this PR after that one is merged. The tests on that one should make this one fail when `current_viewer` is nil.

javitonino:
Please take a second look @gfiorav (although I still have to resolve package.json conflicts). I changed things a bit now that embed authentication is working properly.

app/controllers/carto/builder/public/embeds_controller.rb
```diff
@@ -34,6 +35,14 @@ def load_visualization
           render_404 unless @visualization
         end
 
+        def load_auth_tokens
+          @auth_tokens = if @visualization.password_protected?
+                           @visualization.get_auth_tokens
+                         elsif @visualization.organization?
+                           current_viewer ? current_viewer.get_auth_tokens : []
```
gfiorav: Yep, seems a lot safer 👍

gfiorav:
📦 it for the 🆓 🌍

Issue #9313, opened 2016-08-04
Error in selected basemap


piensaenpixel:
STR
1. Import .carto
2. Go Basemap tab
3. Basemap active is Positron and in the carousel Custom appears as selected
![image](https://cloud.githubusercontent.com/assets/1214201/17409213/801bf3f6-5a6e-11e6-9e05-1e2c44dd0681.png)

matallo:
on it

PR #9344, opened 2016-08-08
basemaps: fix selected basemap

matallo:
close https://github.com/CartoDB/cartodb/issues/9313

- the only drawback for this solution is that custom basempas won't appear as selected in the mosaic (despite the category will) as at the moment of saving the custom basemap in the old editor, it's not stored a selected state
- if we decided to check the selected state with the `urltemplate` we can have the case when using a custom baselayer with the same `urltemplate` than the basemaps list in the config, and that one in the list will appear as selected, not the custombaselayer
- eventually we could run a task to check if the baselayer is custom, then set `selected: true` in the corresponding layer of the custombaselayerscollection /cc @juanignaciosl 

in any case, when adding a new custom baselayer or selecting one from the mosaic in the builder it will appear as selected

CR @viddo /cc @piensaenpixel


viddo:
The implementation looks legit, but I have trouble understanding the explanations you wrote in the PR description to be honest. Not sure who/what they're addressing, and more importantly why?

> the only drawback for this solution is that custom basempas won't appear as selected in the mosaic (despite the category will) as at the moment of saving the custom basemap in the old editor, it's not stored a selected state

Why can't that be solved?

> if we decided to check the selected state with the urltemplate we can have the case when using a custom baselayer with the same urltemplate than the basemaps list in the config, and that one in the list will appear as selected, not the custombaselayer.

Sounds like a problem due to in what order the basemaps are checked/rendered, why not check if layer is a custom basemap before other categories to determine that?

> eventually we could run a task to check if the baselayer is custom, then set selected: true in the corresponding layer of the custombaselayerscollection /cc @juanignaciosl 

What does this _eventually_ depend on to happen? If it's a problem(?) why not solve it as a part of this PR?

lib/assets/javascripts/cartodb3/editor/layers/basemap-content-view.js
```diff
@@ -244,13 +244,15 @@ module.exports = CoreView.extend({
   },
 
   _getBaseLayerCategory: function () {
-    // baseLayer has no category if baseLayer is Plain
+    // baseLayer *has no category* at map creation, if baseLayer is Plain, or if baseLayer is Custom
```
viddo: I'd remove this comment completely, it doesn't really add any value here, it's clear by the code does already.

matallo:
after reading your review you're right I was trying to solve this in a wrong way, using the urlTemplate as the comparision to set the selected

I changed to the `className`, as we do with the config basemaps, it won't collide as their `className` is hardcoded as in "positron_rainbow_labels", also current UI won't let you add a custom basemap you already have, so there can't be a repeated className

finally, I spotted it made no sense having a selected attribute in the `custombaselayer`, take the next case: 

1. user adds a custom basemap, it's saved with selected
1. user creates a new map, that custom basemap appears as selected because it was saved so, despite it has another as a baselayer

updated the  code, thanks for your insight!

viddo:
👍

PR #9354, opened 2016-08-08
9308 viz tokend in db

### LINKED ISSUE #9308: Store visualization tokens in the DB

javitonino:
See #9308

Plan is to deploy this, which makes newly generated tokens to be stored in the database. Then run the rake task to copy tokens from named map to database (~2400 maps) and then, do another PR to make the retrieval of auth_tokens to use the DB instead of the named map.

Please CR @gfiorav

### DIFF
--- a/lib/tasks/viz_maintenance.rake
+++ b/lib/tasks/viz_maintenance.rake
@@ -137,6 +137,20 @@ namespace :cartodb do
       puts "Purge complete. Removed #{purged_count} items"
     end
 
+    desc "Updates visualizations auth tokens from named maps"
+    task update_auth_tokens: :environment do |_|
+      Carto::Visualization.find_each(conditions: "privacy = 'password'") do |visualization|
+        puts "Updating #{visualization.id}"
+        begin
+          tokens = visualization.get_auth_tokens
+          puts "  from #{visualization.auth_token} to #{tokens.first}"
+          visualization.update_column(:auth_token, tokens.first)
+        rescue => e
+          puts "ERROR #{e.inspect}"
+        end
+      end
+    end
+
     private
 
     def inconsistent?(viz)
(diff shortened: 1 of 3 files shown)


db/migrate/20160808114840_add_auth_token_to_visualization.rb
```diff
@@ -0,0 +1,7 @@
+Sequel.migration do
```
gfiorav: Low traffic hour please!

javitonino: YES! Visualization table is always tricky. This migration (adding a column with no default) does not rewrite the table, it is a metadata only operation, but nevertheless I will summon #systems for supervision before running it.

lib/tasks/viz_maintenance.rake
```diff
@@ -137,6 +137,20 @@ namespace :cartodb do
...
     end
 
+    desc "Updates visualizations auth tokens from named maps"
+    task update_auth_tokens: :environment do |_|
+      Carto::Visualization.find_each(conditions: "privacy = 'password'") do |visualization|
+        puts "Updating #{visualization.id}"
+        begin
+          tokens = visualization.get_auth_tokens
+          puts "  from #{visualization.auth_token} to #{tokens.first}"
+          visualization.update_column(:auth_token, tokens.first)
```
gfiorav: are you sure this will commit the change? I always use `.update_attribute(...)`

I guess you tested but better safe than sorry!

gfiorav:
👍 to the plan.

📦 !

lib/tasks/viz_maintenance.rake
javitonino: Yes, this commits the change. The difference is this one goes to the DB directly, skipping before/after hooks.

PR #9373, opened 2016-08-09
Get visualization auth tokens from DB

### LINKED ISSUE #9308: Store visualization tokens in the DB

javitonino:
Closes #9308


javitonino:
Rake already ran, so all password protected visualizations have `auth_token` set. Please CR @gfiorav

gfiorav:
📦 it for the 🆓 🌍 but only once 🍏

PR #9377, opened 2016-08-09
Fixed mobile apps form app_type input if it can't be changed

### LINKED ISSUE #9376: Sending app_type to controller if it can't be changed fails in mobile apps form
That's it.

cc @javitonino @csobier

MariaCheca:
Fixes #9376

Set the input to `disabled` in case it cannot be changed, so it won't be sent to the controller.

CR @javitonino


javitonino:
Does this change the form appearance?

Also, now that we are at it, better to also do it for platform.

MariaCheca:
The form design remains as it was. And it's not necessary change it for platform, as it has already logic to set it as disabled in those cases.

javitonino:
:+1:, ship it then!

MariaCheca:
✨

PR #9378, opened 2016-08-09
Centroide enabler form

nobuti:
This PR applies the enabler form component into the centroid model.

centroid

cc @xavijam


nobuti:
Please @javierarce could you take a look too please?

xavijam:
Let's check that the places where the "Operator" component is being used are not affected by any of these changes.

nobuti:
@xavijam Already checked, the component in the intersection analysis is working as expected. But I'll double check.

xavijam:
Check in style, aggregation options.

nobuti:
Checked 👍

javierarce:
👍  Nice!

PR #9411, opened 2016-08-10
Use simple geom type

viddo:
The raw geometry is not really needed, since we only need the simple geom type.

@xavijam can you review?
cc @gfiorav


xavijam:
Goodbye geometry? fine by me!

xavijam:
:es:

viddo:

PR #9424, opened 2016-08-11
Style option order

### LINKED ISSUE #9423: Max number of buckets on time series widget is too low
It's now 30, it should be 256¿?

nobuti:
This PR:

* Fixes #9423
* Put animation style just before heatmap
* Match fill size for heatmap and component definition, similar issue to first point above.

cc @xavijam


xavijam:
:+1:

PR #9434, opened 2016-08-12
Set avoidNotification on analysis nodes when removing a layer.

### LINKED ISSUE #9124: Deleting a layer with analysis nodes, launches duplicated notifications
When deleting analysis nodes due to the deletion of a layer, we shouldn't raise the "Analysis delated" notification.

screen shot 2016-07-26 at 17 15 02

nobuti:
This PR fixes #9124.

analysis-layer-deletion

cc @xavijam


xavijam:
:+1:

PR #9436, opened 2016-08-12
Add date columns to data view

### LINKED ISSUE #9422: date type columns doesn't appear on the DATA column list
and they should

nobuti:
This PR fixes #9422.

Also fixes a bug when a time series is created and there is another one already. The structure in this scenario was messy at some point.

timeseries-data

cc @xavijam


xavijam:
🇪🇸 , on premise tests running forever, merging due to frontend tests are 👌 .

PR #9443, opened 2016-08-16
fix search and fix colors in close icon

piensaenpixel:
close icon

image

Fix https://github.com/CartoDB/cartodb/issues/9408


javierarce:
👍

PR #9450, opened 2016-08-16
Relax containsNode check for widget def model

viddo:
Fixed the call throwing an error if the source defined for the widget does not exist (e.g. set to `undefined`).

Note that this only fixes a _symptom_ of a map is in an inconsistent state,  the reason for this state is still unknown. 😖 

@javierarce can you review please?


viddo:

Issue #9441, opened 2016-08-16
Select views show a big empty space at the end of the list


javierarce:
I don't know if this was already reported or is already known, but select views are showing a big empty space at the end of the list (spotted in Chrome and Firefox):

![select](https://cloud.githubusercontent.com/assets/4933/17692467/3639bfc2-639a-11e6-8c6e-231a5a3e148d.gif)

javierarce:
cc: @piensaenpixel @xavijam

xavijam:
Could you take a look @piensaenpixel please?

piensaenpixel:
sure

piensaenpixel:
uhm i cant reproduce :( that's production. But i'm checking.
![scroll-custom](https://cloud.githubusercontent.com/assets/1214201/17728933/03f42048-6462-11e6-954f-2ca63a4d67ad.gif)

javierarce:
Could you try to reproduce it with these instructions:  https://github.com/CartoDB/cartodb/issues/9247

piensaenpixel:
Na, dont worry I've just found the problem.

PR #9459, opened 2016-08-17
fix custom list view

piensaenpixel:
@javierarce 
could you take a look?

Fix  https://github.com/CartoDB/cartodb/issues/9441

scroll-custom-fixed


javierarce:
👍

PR #9468, opened 2016-08-17
Adds input placeholders

### LINKED ISSUE #8614: Add placeholders for input texts
The input texts in the Filter analysis should have a default placeholder. 

screen shot 2016-08-17 at 16 51 21

The placeholders depicted in the image above require at two queries, which is expensive. I suggest using a simple text now and decide if we want to request the max and min values from the backend.

javierarce:
This PR adds placeholders for the input fields in the filter analysis: 

screen shot 2016-08-17 at 17 19 39

CR: @piensaenpixel 

Closes #8614


saleiva:
Hey, but shouldn't we add the actual value on those?

javierarce:
Well, I wrote about that in the description of the ticket (#8614). We can do that, but we'll need to request the values. This was a quick solution to show something in the form.

piensaenpixel:
Code 👍

saleiva:
meh ;)
merge then, but we should fill all the inputs with this then :)

PR #9472, opened 2016-08-18
9460 cast geom to simple

### LINKED ISSUE #9460: Improve query-schema-model usage
We are currently (ab)using query-schema-model for _two_ purposes (breaking single-responsibility-principle for the trade-off and benefit of avoiding extra API calls/roundtrips), to get:
 1. The query schema, i.e. the column names and their types
 2. The geometry of the dataset

The problem is, to get both results in one request we do a full `SELECT *` query, which while returns all columns available(1), also includes _all_ data even if we only need the simple type of a geom(2), which if the dataset contains huge `the_geom` or blobs (in magnitude of MBs) doesn't make sense to return to the frontend. Waste of bandwidth and could make the browser (tab) crash. This wasn't seen until last week, so luckily not a urgent problem (yet).

Also, different parts of the UI don't require _both_ these data points, typically it's only `the_geom` that's needed up front, the dataset schema is usually only needed when editing a layer/analysis/widget, so that can be postponed to be fetched on-demand (this is already how it works, so no changes necessary for this).

Considering the common use-cases don't contain huge `the_geom`/blobs I think we could/should tackle this in the following steps:

1. Cast `the_geom` to a the simple type literal consumed by the views (`point`, `polygon`, `line`)
2. Separate query-schema-model in two:
  - `geometry-model`: new model to represent the simple geom ([persisted on/restored per node definition already](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/data/analysis-definition-node-model.js#L16-L18))
  - `query-schema-model`: only representing the schema, so can add a `limit 0` to the query, effectively avoiding to get any unnecessary data
  - These changes should only require changes minimal change to the UI, to use the new geometry-model to get the geometry, there's [only a handful](https://github.com/CartoDB/cartodb/search?utf8=%E2%9C%93&q=simple_geom), and the use-cases are simple, so should not be that severe.

Thoughts, @gfiorav @xavijam @rochoa  ?

viddo:
Fixes #9460 

The solution is _almost_ the same as I proposed in the issue:

> 1. Cast the_geom to a the simple type literal consumed by the views (`point`, `polygon`, `line`)
> 2. Separate query-schema-model in two:
>  - geometry-model: new model to represent the simple geom (persisted on/restored per node definition already)
>  - query-schema-model: only representing the schema, so can add a limit 0 to the query, effectively avoiding to get any unnecessary data
>  - These changes should only require changes minimal change to the UI, to use the new geometry-model to get the geometry, there's only a handful, and the use-cases are simple, so should not be that severe.

… with the difference that doing (1) turned out to require to do (2), due to some parts of the code (dataset/table views) depending/abusing the `rowsSamplesCollection`. On the flip side, solves the issue in one go!

So, summarizing the changes a little bit, to hopefully ease the code review:
- Extracted `simple_geom` and related logic to a `query-geometry-model`, so an analysis node now has two separate models for the two distinct use cases:
  - `nodeDefModel.queryGeometryModel` - lookups the geometry (persisted/restored from the node's `options` data.
  - `nodeDefModel.querySchemaModel` - like before, w/, intended only to lookup the schema (columns and their types etc.)
- Replaced/added `querySchemaModel` w/ `geometryModel` where using wanting the geometry (this is why there are so many changed files)
- `query-schema-model` only requests schema (columns) by default, changed it so to also fetch rows data as part of the query, for the use-cases where `rowsSamplesCollection` was used mentioned above (now renamed to `rowsCollection` btw), e.g.:
```js
var m = new QuerySchemaModel({query: 'SELECT * FROM foo'});
m.set('rows_per_page', 123);
m.fetch() => // => GET ?sql=…&rows_per_page=123
```
- Removed `the_geom_webmercator` usage since it was unused anyway, after talking w/ @rochoa it's my understanding that `the_geom_webmercator` implies `the_geom` existing but not the other way around. And obviously, since we don't use the value anyway it should not matter anyway.
- Also, check the full generated URL, instead of only the sub-query, to (properly) decide on using a GET or POST for the request.

@xavijam 🎁 😘


xavijam:
cc @javierarce and @nobuti, quite important ^^^.

package.json
```diff
@@ -68,6 +68,7 @@
     "jasmine": "^2.4.1",
     "jstify": "0.13.0",
     "load-grunt-tasks": "3.2.0",
+    "multiline": "1.0.2",
```
xavijam: oh, what is it for?

xavijam:
It looks ok, but there are so many changes in so many places, I'd try in staging before merging.

viddo:
>  I'd try in staging before merging.

Agree, deploying on ded03

lib/assets/javascripts/cartodb3/data/query-geometry-model.js
```diff
@@ -0,0 +1,138 @@
+var _ = require('underscore');
+var Backbone = require('backbone');
+var syncAbort = require('./backbone/sync-abort');
+var multiline = require('multiline');
+
+var template = _.template(
+/* eslint-disable */
+multiline(function () {/*!@preserve
```
viddo: @xavijam according to https://github.com/sindresorhus/multiline#minification this kind of comment should preserve the comment, but it was removed anyway by the uglify minification grunt build step… 😢  need to investigate why that is (perhaps I misspelled it or such?). The `eslint-disable` was added because it should have a space in there, perhaps that's enough?

Worst case: 🔥  the multiline and do a normal string concat or similar.

edit: another idea I didn't think of at the time (blackout…), use a standard .tpl file instead, actually probably easier/better since don't need to add yet another dependency.

Cartofante:

Cartofante:

Cartofante:

Cartofante:

xavijam:
@Cartofante, of course!

xavijam:
Found several problems:

- [ ] Geometry parsing in table view.
screen shot 2016-08-19 at 08 55 43

- [ ] Table loading.
screen shot 2016-08-19 at 08 55 47

Cartofante:

Cartofante:

Cartofante:

Cartofante:

Cartofante:

Cartofante:

Cartofante:

Cartofante:

Cartofante:

Cartofante:

xavijam:
Please @nobuti review my last changes. I'm checking everything is ok in staging (), but this is pretty important for future implementations over the Builder.

cc @CartoDB/frontend.

lib/assets/javascripts/cartodb3/data/query-geometry-model.js
```diff
@@ -112,10 +118,10 @@ module.exports = Backbone.Model.extend({
   },
 
   _onChange: function () {
-    if (!this.hasChanged('status') && this.get('status') === 'fetching') {
+    if (!this.hasChanged('status') && this.get('status') === STATUS.fetching) {
       this.fetch(); // If is already fetching just redo the fetch with latest attrs
     } else if (this.hasChanged('query') || this.hasChanged('ready')) {
-      this.set('status', this.get('query') ? 'unfetched' : 'unavailable');
+      this.set('status', STATUS[this.get('query') ? 'unfetched' : 'unavailable']);
```
nobuti: This invalidate the point of having a hash for states. Also, I think this is more readable: `this.get('query') ? STATUS.unfetched : STATUS.unavailable`.

Cartofante:

Cartofante:

lib/assets/javascripts/cartodb3/data/query-geometry-model.js
```diff
@@ -0,0 +1,140 @@
...
+
+  hasValue: function () {
+    return !!this.get('simple_geom');
+  },
+
+  _onChange: function () {
+    if (!this.hasChanged('status') && this.get('status') === STATUS.fetching) {
+      this.fetch(); // If is already fetching just redo the fetch with latest attrs
+    } else if (this.hasChanged('query') || this.hasChanged('ready')) {
+      this.set('status', STATUS[this.get('query') ? 'unfetched' : 'unavailable']);
```
nobuti: `this.get('query') ? STATUS.unfetched' : STATUS.unavailable`

nobuti:
I can't see anything odd, so 👍  to me

Cartofante:

Cartofante:

xavijam:
Tests back to life, thanks for the warning @nobuti :kiss:.

Cartofante:

Cartofante:

Cartofante:

Cartofante:

nobuti:
\o/ grandes @viddo @xavijam

viddo:
🎉 😘 🙇

PR #9477, opened 2016-08-18
Fixed problem related with dataset privacy

### LINKED ISSUE #9465: Change dataset privacy is not working for "standard" users
Checking the problem, we have realized that, for a user that doesn't belong to an organization, the dialog is useless because only the header (for changing the privacy) would appear.

This publish dialog belongs to "standard" user in a map:

screen shot 2016-08-17 at 15 32 50

For a dataset we remove the "central" part, so...

Possibilities:

1. Open a dropdown (not a modal), just for changing the privacy. Bad part is CTA, where to include it?.
2. Open a totally different dialog (with the new design) and only offering the privacy change.

cc @saleiva @urbanphes @piensaenpixel

xavijam:
Dataset view:
screen shot 2016-08-18 at 16 26 55

Share map view:
screen shot 2016-08-18 at 16 28 03

Also, privacy tags have been changed to have the same size (the big ones), looks ok @josecruz?
screen shot 2016-08-18 at 16 28 12
screen shot 2016-08-18 at 16 28 16

cc @josecruz @piensaenpixel 
CR: @javierarce 

Fixes #9465.


josecruz:
Awesome @xavijam. If we're gonna change privacy tag we'll need to replace it on the Dashboard (basically on every dataset and map list) so we maintain consistency between tags, what do you think @saleiva ?

xavijam:
About the dashboard change, yeah, it will happen eventually.

lib/assets/javascripts/cartodb3/components/modals/publish/create-privacy-options.js
```diff
@@ -40,7 +41,7 @@ module.exports = function (visDefinitionModel, userModel) {
       // Set state that depends on vis and user attrs, they should not vary during the lifecycle of this collection
       return _.defaults({
         selected: option.privacy === currentPrivacy,
-        disabled: !(option.alwaysEnable || canSelectPremiumOptions)
+        disabled: !(option.alwaysEnabled || canSelectPremiumOptions)
```
javierarce: If `alwaysEnabled` is going to be always `true` doesn't make sense to remove it?

javierarce: Can `option.alwaysEnabled` be something different than `true`?

javierarce:
👍  (except for a comment)

lib/assets/javascripts/cartodb3/components/modals/publish/create-privacy-options.js
xavijam: Undefined, because it is not defined/expressed in the ALL_OPTIONS object.
The thing is, public privacy can't be disabled, and the only way to do that is check that, in fact it was previously done like that, but there was a bug and it was not working properly.

javierarce: Ah, ok, then 🚢  or :shipit: or 🚀

Issue #9473, opened 2016-08-18
Github oauth


javitonino:
This is the cartodb part (for box login and org signup).

@xavijam @javisantana Do we want to test this without central (OSS/onpremises compatible)? Both login and signup?

xavijam:
Yes, please.

javitonino:
Closed via #9483

PR #9483, opened 2016-08-19
9473 GitHub oauth

javitonino:
See #9473


javitonino:
Please CR @gfiorav 

This includes the changes for user login. Doesn't include org login (disabled on purpose) nor signup (not implemented).

gfiorav:
Look 📦  from a backend point of view, have someone review the front!

PR #9492, opened 2016-08-22
Background-size cover for avatars.

### LINKED ISSUE #9340: Non-square avatars displayed at original proportion in share with users modal
In the modal to share the map with organization users (inside the builder), non-square avatars are not correctly sized.

screenshot_20160808_124401

cc @xavijam @nobuti

nobuti:
This PR fixes #9340.

The best solution would be to use `object-fill` property but IE doesn't implement it, so instead of making a polyfill, the plug and play solution is to use images as background, letting the hard work of managing proportions to css.

cc @piensaenpixel


piensaenpixel:
LGTM

xavijam:
"Dale candelabro"

PR #9503, opened 2016-08-22
Drop table from cdb_importer if failure when renaming to final name

iriberri:
Currently, if there's an exception at `Importer#rename` we leave a stale table when the import finishes with a failure status.

We know at least one scenario in which this method can break: The `ValidTableNameProposer` does not check in the `cdb_importer` schema, so tables names can collide in there. Tables in the `cdb_importer` schema are meant to have a `importer_{hash}` name until they're renamed and moved to the user schema. It is after this rename when collisions with more typical names (common data datasets, or other non-hashed names) occur.

In this PR I'm adding a rescue to the `Importer#rename` method so that the table from `cdb_importer` is dropped after a failure in the `ALTER` statement.

Notice this won't clean the table if the `ALTER` statement succeeded and the method fails anywhere else. After the `importer_hash` name is converted to the real table name we will need to be very careful to drop tables that are actually related to the current import process. (** Note: perhaps we should split this method to have more visibility about table breaking and its name at a specific moment?)

The current `Importer#register` method is as follows (I removed log traces):

```ruby
      def register(result)
        @support_tables_helper.reset

        # Sanitizing table name if it corresponds with a PostgreSQL reseved word
        result.name = Carto::DB::Sanitize.sanitize_identifier(result.name)

        name = rename(result, result.table_name, result.name)
        result.name = name

        move_to_schema(result, name, ORIGIN_SCHEMA, @destination_schema)

        persist_metadata(result, name, data_import_id)
````

So if we have a stale table in `cdb_importer` with a "real" name, we know it passed through the `ALTER` statement in `rename` but it didn't succeed in `move_to_schema`. It could have failed in the instructions in `rename` after the table renaming, because `move_to_schema` has the means to drop the stale table (with its real name).

There's also the possibility that the "real"-named table in the `cdb_importer` comes from a Sync process as we have seen this scenario with the Populated places dataset, although there are also traces in our Rollbar tracker with other user datasets which are not related to common-data (which could be sync tables too). I think this could be the key.

In short: this first PR has the objective of cleaning the stale table in `cdb_importer` after a crash in the situation explained above, but it's clear that we have another scenario which is not covered here (table renamed but still living in cdb_importer).

### CODE REVIEW

**Comment by iriberri** [comment]
Hi @ethervoid could you take a look? I hope the explanation above is clear enough :)

**Comment by iriberri** [comment]
Related: https://github.com/CartoDB/cartodb/issues/8785

File: app/connectors/importer.rb [hunk 1]
```diff
@@ -208,6 +208,9 @@ def rename(result, current_name, new_name)
         end
 
         new_name
+      rescue => exception
+        drop("#{ORIGIN_SCHEMA}.#{current_name}")
```
ethervoid: We should log this exception  in Rollbar to be aware of it

ethervoid:
Why don't we clean stale tables here https://github.com/CartoDB/cartodb/pull/9503/files#diff-048165172f057d68c8c054a742ce0206R91 ?

iriberri:
At that rescue, tables could live in different places:
* cdb_importer.importer_{hash}
* cdb_importer.the_real_name (if it failed after the ALTER statement of `rename`)
* user_schema.the_real_name  (if it failed after `move_to_schema`)

so we wouldn't have the means to know where the table is, and specifically, once it moves to the `user_schema` a drop would be dangerous.

ethervoid:
Ok! lgtm

iriberri:
Merging this as soon as tests passed again. Had to get master for some migrations that were breaking my staging deployments.

PR #9505, opened 2016-08-22
M4564 auth token private visualizations

javitonino:


javitonino:
Please CR @gfiorav

gfiorav:
Do it!

PR #9517, opened 2016-08-23
Fetch org users from dataset view.

nobuti:
This PR fixes https://github.com/CartoDB/cartodb-management/issues/4560

cc @xavijam

### DIFF
--- a/lib/assets/javascripts/cartodb3/dataset.js
+++ b/lib/assets/javascripts/cartodb3/dataset.js
@@ -27,6 +27,7 @@ var VisTableModel = require('./data/visualization-table-model');
 var DatasetUnlockModalView = require('./dataset/dataset-unlock-modal-view');
 var EditorModel = require('./data/editor-model');
 var UserModel = require('./data/user-model');
+var OrganizationUsers = require('./data/organization-users-collection');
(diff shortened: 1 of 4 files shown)


xavijam:
:+1:

xavijam:
Why not fetch them when there are required?
I've just checked that in the builder page, we are requesting those users at the beginning and I feel it is a unnecessary request at that point, only used in the share dialog.

nobuti:
Not exactly, we are showing how many people the visualisation is shared with in the header.

xavijam:
Hum... ok. We could provide the total number or organization users there are in an object, exposed in the html, what do you thing @javitonino @gfiorav ?

javitonino:
Should be easy to add, we can provide a list with all the users and groups in the org in every request. Or we can include only the objects the visualization is shared with, and load the full list in the share dialog (we can do paging then for visualizations with lots of users).

xavijam:
I feel we only need the number of users, with that we will avoid that first request, right @nobuti?

xavijam:
In any case, let's merge this one and open a new ticket about the other thing.

nobuti:
@xavijam yep, only `total_user_entries` excluding the current user by default.

PR #9519, opened 2016-08-23
6002 dataset view read only vis

### LINKED ISSUE #6002: When a table is shared (only READ) with an organization mate, when you try to visit it, it goes to public view and not private view
### STR
1. Having two users (A and B) inside an organization.
2. Being logged with A, shared a dataset (with READ limits) with user B.
3. Log in with user B, check your datasets.
4. Click over the dataset that A shared with B.
5. You are redirected to the private view.

It should go to private view with disabled state.

javitonino:
Closes #6002 

Redirects shared visualization view to new dataset view

### DIFF
--- a/spec/requests/carto/builder/datasets_controller_spec.rb
+++ b/spec/requests/carto/builder/datasets_controller_spec.rb
@@ -60,25 +60,6 @@ describe Carto::Builder::DatasetsController do
       response.location.should end_with public_table_map_path(id: @other_visualization.id)
     end
 
-    describe 'viewer users' do
-      after(:each) do
-        if @user.viewer
-          @user.viewer = false
-          @user.save
-        end
-      end
-
-      it 'redirected to public view for their datasets at the builder' do
-        @user.viewer = true
-        @user.save
-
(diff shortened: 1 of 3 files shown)


javitonino:
CR time @gfiorav

spec/requests/carto/builder/datasets_controller_spec.rb
```diff
@@ -60,25 +60,6 @@
...
 
-    describe 'viewer users' do
-      after(:each) do
-        if @user.viewer
-          @user.viewer = false
-          @user.save
-        end
-      end
-
-      it 'redirected to public view for their datasets at the builder' do
```
gfiorav: Maybe a new test checking the new behaviour would be nice 😄

javitonino: Now there is no different behaviour for viewers, they get exactly the same as the rest (backend-wise). That's why I removed this altogether, didn't know what to write instead.

javitonino:
:green_apple: tests. CR = :+1:?

gfiorav:
make it happen

PR #9521, opened 2016-08-23
Metrics API

gfiorav:
# SYNOPSIS

We need an endpoint where we can send metrics events with a common format (more below) and have the backend connect to any service required (one, another or multiple depending on the event) to trigger the event there.

## Event format

A post should be made to `api/v3/metrics` with the following format:

```json
{
    "name": "<event name>",
    "properties": "<properties>"
}
```

### Event name

This is the name of the event. It's name has to coincide with the backend class that represents it. If you choose a name that corresponds to no class, a friendly error is sent.

 For example: `Potato peeled` -> `Carto::Tracking::Events::PotatoPeeled`.

### Properties

This is a hash that should contain every value that the event needs: ~~no specific per-event requirement compliancy is checked. It's up to you to program it well.~~ Every event is validated accordingly and validation issues are returned. If the event has any record (vis, user) that is not accesible by the logged in user, you get 403.

IMPORTANT: You should never specify any attribute of a model other than it's id. For example, if you want to send a users' username, just specify `user_id: xxx`. There is a helper that will load the model and grab whatever you need from it in the back.

 For example:
```json
{
    "user_id": "abcd-efgh-ijkl-mnopq",
    "visualization_id": "1234-56789-1011-121314",
    "quota_overage": 32.3,
    "origin": "editor"
}
```

## Implementation

Meta, meta, meta, meta. I know, it's not deterministic. I've explored this solution and I recommend to CR this without seeing the diff, just the new stuff.

The paradigm is that we have an internal format for an event, that can be translated into other service's formats. Here is the tree for `lib/carto/tracking`:

```
├── events.rb
├── formats
│   ├── internal.rb
│   └── segment.rb
└── services
    └── segment.rb
```

Please CR @javitonino


app/controllers/carto/api/metrics_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+
+      private
+
+      def load_event_class
+        event_name = params[:name]
+
+        raise Carto::UnprocesableEntityError.new('name not provided') unless event_name
+
+        modulized_name = "Carto::Tracking::Events::#{event_name.parameterize('_').camelize}"
```
javitonino: Question: Why not pass the class name directly to the controller, and avoid the parametrize + camelize? I mean, the original name has to be reconstructed from the class name anyway.

app/controllers/carto/api/metrics_controller.rb
```diff
@@ -0,0 +1,35 @@
+# encoding utf-8
+
+module Carto
+  module Api
+    class MetricsController < ::Api::ApplicationController
```
javitonino: This is complex enough to warrant some tests I feel. Nothing fancy, but make sure the metamagic to get the class name is working at least.

lib/carto/tracking/events.rb
```diff
@@ -1,253 +1,73 @@
...
-        private
-
-        def properties(visualization, views)
-          {
-            map_id: visualization.id,
-            map_name: visualization.fetch.name,
-            mapviews: views
-          }
+        def report
+          methods.select { |method_name| method_name =~ /^report_to/ }.each do |report_method|
```
javitonino: This is the meta thing I'm less sure about. But it is hidden and should not be problematic. I'll try to think if there is some way to break this.

Anyway, I'd use `starts_with?` rather than a regex.

lib/carto/tracking/formats/internal.rb
```diff
@@ -0,0 +1,42 @@
...
+  module Tracking
+    module Formats
+      class Internal
+        def initialize(hash)
+          @hash = hash ? hash.with_indifferent_access : Hash.new
+        end
+
+        # Symbol should be provided as a snake-case'd version the record's model class name.
+        # The id of said record should be provided with key: snake-case'd identifier + '_id'
+        # Only Cart:: records allowed!
```
javitonino: typo

lib/carto/tracking/formats/internal.rb
```diff
@@ -0,0 +1,42 @@
...
+        #  Ex.: :super_duper_mdoel -> Carto::SuperDuperModel; { super_duper_model_id: xxx }
+        def fetch_record(symbol)
```
javitonino: I fear this might be going too far, as we are only using a couple of models. I mean, the code is shorter without this:
```
user = Carto::User.find(@hash[:user_id])
visualization = Carto::Visualization.find(@hash[:visualization_id])
```

app/controllers/carto/api/metrics_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+
+      ssl_required :create
+
+      before_filter :load_event_class, only: :create
+
+      rescue_from Carto::LoadError,
+                  Carto::UnprocesableEntityError, with: :rescue_from_carto_error
+
+      def create
+        @event_class.new(params[:properties]).report
```
javitonino: Crazy idea: Take user from current_user instead of the properties. Don't know if it is always the correct thing, so just think about it.

javitonino:
Left some comments, mainly to open some discussions.

app/controllers/carto/api/metrics_controller.rb
gfiorav: We are on the meta wagon, and it has no brakes.

On a serious note: the reason is to have nice names like `Exported map` and not have to maintain a `map`. The map isn't to bad though.

app/controllers/carto/api/metrics_controller.rb
gfiorav: yup, forgot to commit them 😅

lib/carto/tracking/events.rb
gfiorav: This is the only meta thing I would actually keep! 

Isn't starts_with doing that under the hood?

lib/carto/tracking/formats/internal.rb
gfiorav: Probably. The controlled fetch process can be extracted though, that's why I separated it.

app/controllers/carto/api/metrics_controller.rb
gfiorav: Yeah, I thought about it... I prefer to use the one provided in each case, but I do have the feeling that the whole "user required" stuff is smelly.

app/controllers/carto/api/metrics_controller.rb
javitonino: Not talking about removing the meta (yet...), just the `.parameterize('_').camelize}` and call this with `ExportedMap` instead of `Exported map`.

lib/carto/tracking/events.rb
javitonino: Nope, it doesn't spin up the regex engine, instead it just compares the first few bytes of the string. More efficient and secure.

gfiorav:
Added some stuff:

+ Per-event property validation
+ Records concerned by the properties are tested to be accesible by the user logged in.

gfiorav:
Tests are 🍏 

I've added validation and security concerns.

Please have a quick look @javitonino

lib/tasks/notify_trending_maps.rake
```diff
@@ -23,7 +23,9 @@ namespace :cartodb do
                                                                                           300)
         unless simulation
           trending_maps_lib.notify_trending_map(visualization_id, views, preview_image)
-          Carto::Tracking::Events::ScoredTrendingMap.new(visualization.fetch.user, visualization, views).report
+          Carto::Tracking::Events::ScoredTrendingMap.new(user_id: visualization.fetch.user.id,
+                                                         visualization_id: visualization.id,
+                                                         mapviews: views).report
```
javitonino: Super minor: maybe `map_views` for consistency?

lib/resque/tracking_jobs.rb
```diff
@@ -0,0 +1,22 @@
+# encoding utf-8
+
+module Resque
+  module TrackingJobs
+    module SendSegmentEvent
+      @queue = :tracker
+
+      def self.perform(user_id, name, properties)
+        return unless key = Cartodb.config[:segment]['api_key']
```
javitonino: This crashes if `Cartodb.config[:segment]` is nil. Not a big deal since there is a check in the enqueuing, but easily solved with `Cartodb.get_config(:segment, 'api_key')`.

lib/carto/tracking/validators/user.rb
```diff
@@ -0,0 +1,15 @@
+# encoding utf-8
+
+module Carto
```
javitonino:
```
#
 #
  #
   #
  #
 #
#
```

gfiorav: you mean
```
#
 #
  #
   #
   #
  #
 #
#
```

lib/carto/tracking/formats/internal.rb
javitonino: Typo not gone :stuck_out_tongue:

lib/carto/tracking/formats/internal.rb
```diff
@@ -0,0 +1,56 @@
...
+          record_id_key = "#{symbol_string}_id".freeze
+          record_id = @hash[record_id_key]
+
+          record_class_name.constantize.find(record_id)
+        rescue
+          record_id ? (raise Carto::LoadError.new("#{record_class_name} not found")) : nil
+        end
+
+        # Find anything like xxx_id and tries to fetch a record for it.
+        def concerned_records
```
javitonino: I cannot find this in use anywhere.

lib/carto/tracking/events.rb
```diff
@@ -1,253 +1,180 @@
...
-            Carto::Tracking::Events::FailedConnection.new(user, parameters)
-          end
-        end
+      class LikedMap < Event
+        include Carto::Tracking::Services::Segment
+
+        include Carto::Tracking::Validators::Visualization::Readable
+        include Carto::Tracking::Validators::User
+
+        required_properties :user_id, :visualization_id, :action
```
javitonino: Question, Are `:user_id, :visualization_id` required to be here? Maybe they can be deleted. Not saying they should, depends a bit on validation errors (it can be valuable to have it break sooner, if fields are missing, before trying to fetch it). 

Also curious, does the validator crash if we remove this and `visualization_id` is missing? Reading `fetch_record!` it seems like it might.

lib/carto/tracking/validators/visualization.rb
```diff
@@ -0,0 +1,35 @@
+# encoding utf-8
+
+module Carto
+  module Tracking
+    module Validators
+      module Visualization
+        module Helpers
+          def visualization
+            @format.fetch_record!(:visualization)
```
javitonino: Seeing that something very similar to this is also used in `lib/carto/tracking/formats/internal.rb to_segment` it may make more sense to move this method there. Not sure if they can easily be made compatible.

app/controllers/api/json/visualizations_controller.rb
```diff
@@ -277,13 +303,17 @@ def remove_like
...
         @stats_aggregator.timing('destroy') do
-          vis.remove_like_from(current_viewer.id)
+          vis.remove_like_from(current_viewer_id)
              .fetch
              .invalidate_cache
         end
 
-        Carto::Tracking::Events::DislikedMap.new(current_viewer, vis).report
+        Carto::Tracking::Events::LikedMap.new(current_viewer_id,
+                                              user_id: current_viewer_id,
```
javitonino: Having this duplication of `current_viewer_id` seems a bit overkill. Maybe it could be somehow refactored (say set `user_id` to `reporter_id` if no `user_id` is specified).

javitonino: On second thought, I like the explicit way more than adding more magic to this. Disregard.

lib/carto/tracking/events.rb
```diff
@@ -1,253 +1,180 @@
...
         def report
-          Carto::Tracking::SegmentWrapper.new.send_event(@user, @name, @properties.merge(event_properties))
+          report!
         rescue => exception
-          CartoDB::Logger.warning(message: 'Carto::Tracking: Event couldn\'t be reported',
-                                  exception: exception,
-                                  properties: @properties,
-                                  user: @user)
+          CartoDB::Logger.error(message: 'Carto::Tracking: Couldn\'t report event',
+                                exception: exception)
```
javitonino: You could add more information about the event here. Yes, it can be guessed from the backtrace, but having the event name, and the internal `@format` would be nice, I feel.

lib/carto/tracking/events.rb
```diff
@@ -1,253 +1,180 @@
...
-      end
+      class FailedConnection < Event
+        include Carto::Tracking::Services::Segment
 
-      class CreatedDataset < TrackingEvent
-        def initialize(user, table_visualization, origin: 'blank')
-          super(user, 'Created dataset', visualization_properties(table_visualization, origin: origin))
-        end
-      end
+        include Carto::Tracking::Validators::Visualization::Writable
```
javitonino: Do failed imports have a visualization assigned to them? I read the importer code and didn't seem like it.

lib/carto/tracking/formats/segment.rb
```diff
@@ -0,0 +1,113 @@
...
+        end
+
+        def to_hash
+          properties = event_properties
+
+          properties.merge!(user_properties) if @user
+          properties.merge!(visualization_properties) if @visualization
+          properties.merge!(connection_properties) if @connection
+          properties.merge!(map_liking_properties) if @action
+          properties.merge!(trending_map_properties) if @mapviews
```
javitonino: If a property is repeated in this methods, it can be overwritten in some cases. It doesn't seem to be a problem as the ones with the same name, always have the same value :+1:

javitonino:
Wrote a bunch of stuff. I think nothing really serious, mostly suggestions (feel free to ignore).

PR #9526, opened 2016-08-23
Prepend table name with owner's username.

### LINKED ISSUE #9516: Prefix sql query with user when the dataset is shared inside org
If an owner is working with a dataset, the default query (or custom) doesn't include the user prefix for tables' names. If this user shares the dataset with another user, the data visualisation fails because `the relation doesn't exist`. It's necessary to prefix the table's name with the owner's name in order to work properly.

~~Also, there is an inconsistency between the view when the dataset is shared with read permission and write and read permission. In the first scenario, the user is redirected to the public view.~~

nobuti:
This PR fixes #9516.

It prepends the owner's username to table name if the dataset it's shared.

cc @xavijam

### DIFF
--- a/lib/assets/javascripts/cartodb3/dataset.js
+++ b/lib/assets/javascripts/cartodb3/dataset.js
@@ -31,7 +31,6 @@ var OrganizationUsers = require('./data/organization-users-collection');
 var HeaderView = require('./dataset/dataset-header/header-view');
 var Notifier = require('./components/notifier/notifier');
 var UserGroupFetcher = require('./data/users-group-fetcher');
-var SQLUtils = require('./helpers/sql-utils');
 
 var userData = window.userData;
 var visData = window.visualizationData;
@@ -88,13 +87,6 @@ UserGroupFetcher.track({
 
 var tableName = tableModel.getUnqualifiedName();
 var query = layerModel.get('sql') || 'SELECT * FROM ' + Utils.safeTableNameQuoting(tableName);
-var isOwner = tableModel.isOwner(userModel);
-var userName;
-
-if (!isOwner) {
-  userName = tableModel.get('permission').owner.username;
-  query = SQLUtils.prependTableName(query, tableName, userName);
-}
 
 var querySchemaModel = new QuerySchemaModel({
   query: query,
--- a/lib/assets/javascripts/cartodb3/dataset.js
+++ b/lib/assets/javascripts/cartodb3/dataset.js
@@ -74,7 +74,15 @@ UserGroupFetcher.track({
   acl: visModel.getPermissionModel().acl
 });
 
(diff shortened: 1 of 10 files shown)


lib/assets/javascripts/cartodb3/helpers/sql-utils.js
```diff
@@ -60,6 +60,18 @@ module.exports = {
 
   _removeLineBreaks: function (sql) {
     return sql.replace(/\r?\n|\r/g, ' ');
-  }
+  },
+
+  prependTableName: function (sql, tableModel, userModel) {
+    var isOwner = tableModel.isOwner(userModel);
```
xavijam: Not sure about having this logic in this class. For me we should prepend the user name before the table is needed, but we should check if it is necessary before using it.

nobuti: Do you mean in `tableModel::getUnquotedName()`? But in any case, we have to make a replace in the query that comes from the backend right? How does it solve this case?

xavijam: Yeah, replace is needed, and I like the idea of having a method to do that here. But I would not check if the action is needed here, I'd do it before.

lib/assets/javascripts/cartodb3/dataset.js
```diff
@@ -73,8 +74,10 @@ UserGroupFetcher.track({
   acl: visModel.getPermissionModel().acl
 });
 
+var query = SQLUtils.prependTableName(layerModel.get('sql') || 'SELECT * FROM ' + Utils.safeTableNameQuoting(tableModel.getUnquotedName()), tableModel, userModel);
```
xavijam: Basically here, for example, check if sql has to be qualified.

lib/assets/javascripts/cartodb3/dataset/dataset-options/dataset-options-view.js
```diff
@@ -275,7 +275,8 @@ module.exports = CoreView.extend({
   },
 
   _defaultSQL: function () {
-    return 'SELECT * FROM ' + this._tableModel.getUnquotedName();
+    var defaultQuery = 'SELECT * FROM ' + this._tableModel.getUnquotedName();
```
xavijam: And here.

lib/assets/javascripts/cartodb3/dataset.js
nobuti: 👍  It's more or less how I had it initially, but in order to not repeat the `isOwner` condition calculation every time the method is called , I moved it inside the helper.

xavijam: It is only in two places, not that bad. I wanted to have sql-utils kind of easy and simple and not knowing about rest of the things outside.

xavijam:
I feel we need it in the builder view too.

xavijam:
:kiss:

xavijam:
Come on!

PR #9553, opened 2016-08-25
Fixing problem related with categories and changes over schema

### LINKED ISSUE #9540: Category and torque category wizards listen table metadata changes although they are hidden
## STR
- Create a map using the editor with this [file](https://dl.dropboxusercontent.com/u/931536/untitled_table_4.geojson).
- Go to the map view.
- Open the wizards, and select the category one, let's select `column_3`.
- New legend will appear with `column_3` values.
- Go to the filters section and add a new one, choosing `column_2`.
- Check that legend has changed and it shows now `column_1` values.
- Category wizard has changed as well, but only category items, column is `column_3`, but CartoCSS has changed.

If we try the same steps with torque-category, happens the same.

The problem basically is that we don't "remove" those views when visit between panel tabs (like filters, css, sql, etc), and a binding `this.cartoProperties.bind('change:metadata', this._getMetadata, this);` keeps listening table metadata changes anywhere, like it happens when a filter has been added.

cc @javisantana

xavijam:
Fixes #9540.

CR: @javisantana


xavijam:
Closing!

Issue #9475, opened 2016-08-18
inconsistent data tab in editor panel when sql error, error when change from analysis tab


matallo:
first of all sorry for the long title, I didn't know how to best write it

1. this is a series of issues that while related, they can be taken apart separately, first of all I find it weird when applying an error SQL query I can go to the data in the map and see the error

however, when I disable the sql the data tab still shows data from before appying the query, so if we show an error in one side I think we should be consistent at least not showing the data

---

2. besides this, if instead of directly disable sql you go to "analysis" and then back to "data" instead of showing the sql panel it shows the data panel indeed now with no data but an infinite load

![sqlerror](https://cloud.githubusercontent.com/assets/36676/17772681/763286b6-654a-11e6-8f72-fe347d38b6db.gif)

/cc @nobuti for the sql panel
/cc @saleiva for inconsistency issue
/cc @xavijam  for general awareness

not sure who to assign tho, maybe this can be tackled at the same time idk

nobuti:
Closed by #9563.

PR #9563, opened 2016-08-25
SQL query error message

### LINKED ISSUE #9497: Data panel doesn't change its state when sql query fails
I've checked that problem before, but I don't remember how to make it fail so, I've decided to reproduce it like:

## STR
1. Create a map with any layer, and go to the builder.
2. Once there, go to your dev environment and stop the SQL API (fast and easy instead of looking for a new way of making it fail).
3. Go back to the builder UI.
4. Click over one layer, and check that panel doesn't change its state. It should show an error message and the loading should be hidden.

screen shot 2016-08-22 at 11 41 30

nobuti:
This PR fixes #9497 and #9475.

I added a direct link to the message, so the user can easily access to the sql editor to fix the query. Please @urbanphes @saleiva confirm this is ok.

Also, in order to reduce the noise about notifications, we make the error notification fixed (not closable), so when the user changes the active tab, the notification remains with the error to add some context.

sql-error

cc @xavijam

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
+++ b/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
@@ -76,11 +76,6 @@ module.exports = CoreView.extend({
 
     this._configPanes();
     this._initBinds();
-    this._initData();
-
-    // If there are errors in the SQL, open the sql editor by default
-    // we defer it to not interfere with the high order tabpane item selection
-    setTimeout(this._redirectIfErrors.bind(this), 0);
   },
 
   render: function () {
@@ -97,14 +92,6 @@ module.exports = CoreView.extend({
       this._editorModel.set({
         edition: true
       });
-    }
-  },
-
-  _initData: function () {
-    if (!this._hasFetchedQuerySchema()) {
-      this._querySchemaModel.fetch();
-    } else {
-      this._onQuerySchemaStatusChange();
     }
   },
 
@@ -408,6 +395,10 @@ module.exports = CoreView.extend({
 
     this.$el.append(panelWithOptionsView.render().el);
     this.addView(panelWithOptionsView);
+
+    // If there are errors in the SQL, open the sql editor by default
+    // we defer it to not interfere with the high order tabpane item selection
+    setTimeout(this._redirectIfErrors.bind(this), 0);
   },
 
   clean: function () {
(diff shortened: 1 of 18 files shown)


xavijam:
If we do so (perfect for me), let's disable CartoCSS buttons in style panel.

nobuti:
sql-error-disable

nobuti:
btw in ded06

lib/assets/javascripts/cartodb3/editor/style/style-content-view.js
```diff
@@ -1,12 +1,23 @@
...
 var StyleFormView = require('./style-form/style-form-view');
 var StylesFactory = require('./styles-factory');
 var CarouselFormView = require('../../components/carousel-form-view');
 var CarouselCollection = require('../../components/custom-carousel/custom-carousel-collection');
 var styleFormNotReadyTemplate = require('./style-form-not-ready.tpl');
+var styleSQLErrorTemplate = require('./style-content-sql-error.tpl');
 var OverlayView = require('../components/overlay/overlay-view');
 var Notifier = require('../../components/notifier/notifier');
+var QuerySanity = require('../layers/query-sanity-check');
+var ACTION_ERROR_TEMPLATE = _.template("<button class='CDB-Text u-actionTextColor js-fix-sql'><%- label %></button>");
```
xavijam: Why not using a .tpl?

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
```diff
@@ -393,6 +404,10 @@ module.exports = CoreView.extend({
 
     this.$el.append(panelWithOptionsView.render().el);
     this.addView(panelWithOptionsView);
+
+    // If there are errors in the SQL, open the sql editor by default
+    // we defer it to not interfere with the high order tabpane item selection
+    setTimeout(this._redirectIfErrors.bind(this), 0);
```
xavijam: 😱  but 👌

xavijam:
Not much to say, perfect! 💋

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
nobuti: Agree, I need to investigate a better way of doing this kind of things because it's really ugly. Bu at the time, I don't know how to solve it. Even with that, I think the final result is not totally good.

PR #9566, opened 2016-08-25
Show formula widget for cartodb_id.

### LINKED ISSUE #8130: Cartodb_id in DATA tab should be a formula widget that says count
The normal flow would be to add a widget counting a number of features. Widget name should be "Number of" and get COUNT by default as agg method

nobuti:
This PR fixes #8130.

number-of

I'm deleting the % of null because for cartodb_id I guess it doesn't make sense. Correct me if I'm wrong.

cc @saleiva @xavijam


xavijam:
:+1:

saleiva:
About the name. @javierarce what do you think?
Number of Polygons/points/lines??

javierarce:
Yes. Or `polygon/point/line count`?

saleiva:
+1 to that

xavijam:
:+1:

PR #9569, opened 2016-08-26
Move copy privileges after cartodbfy, as it may rewrite the table

### LINKED ISSUE #9527: Shared permissions seem to be lost every time dataset changes
I've created a dataset by importing a csv file from dropbox. The dataset is configured to be sync'ed every hour. I'm read-only sharing the dataset with other account (user2) in same org so it can be accessed by something similar to:
```
https://user2.carto.com/api/v2/sql?api_key=xxxxxxxx&q=SELECT%20*%20FROM%20%22user1%22.bug_test
```
Problem is every time the dataset changes (new data is added), then the query returns this error:
```
{"error":["permission denied for relation bug_test"]}
```
Reactivating share permissions removes the error... until next update : (

Let me know when working on it so I can provide you real usernames and api key to reproduce the issue.

javitonino:
Fixes #9527 

Moved copy privileges after cartodbfy, as under some circunstances, that recreates the table (and loses all the privileges).

Please CR @iriberri

### DIFF
--- a/app/models/synchronization/adapter.rb
+++ b/app/models/synchronization/adapter.rb
@@ -27,12 +27,12 @@ module CartoDB
             data_for_exception << "1st result:#{runner.results.first.inspect}"
             raise data_for_exception
           end
-          copy_privileges(user.database_schema, table_name, result.schema, result.table_name)
           index_statements = generate_index_statements(user.database_schema, table_name)
           move_to_schema(result)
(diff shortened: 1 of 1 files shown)


iriberri:
LGTM! I just finished testing it in staging, first against the current master branch (where I reproduce the bug) and then with this change, and it's working and it didn't break the sync neither a map I had created from it.

👍

javitonino:
Great! Thanks for your help with this one, much appreciated

PR #9570, opened 2016-08-26
Fix dataset sync view to correctly check state

javitonino:
- Considers the queued state as part of the syncing process so the modal doesn't immediately disappear.
 - Correctly hides the syncing modal only when the status changes to success/failure, but not when changing from queued to syncing


javitonino:
Please CR @xavijam

xavijam:
Looks good to me! thanks mate! 💋

PR #9573, opened 2016-08-26
Fix locale attributes to translate.

### LINKED ISSUE #9531: Convert column from number to text placeholders from en.json seem to not be filled with actual names
https://github.com/CartoDB/cartodb/blob/0ab43dd05804a18a9e58d948eae04be1e50fc5a9/lib/assets/locale/en.json#L887

I edited a column from number to text and in the bottom bar that appeared in the Builder during the process, I was able to see the `%{columname}` instead of the name of my column.

It seems that it only happens in the from number to text scenario.

**STR**
* Open a dataset
* Create a numeric column
* Change column type to text

**Screenshot**
image

cc @xavijam

nobuti:
This PR fixes #9531.

screen shot 2016-08-26 at 13 24 37

cc @xavijam


xavijam:
Much appreciated

PR #9603, opened 2016-08-29
Fetch the row if the geometry is not loaded.

### LINKED ISSUE #9592: error editing cells in dataset with polygon geometries
When a cell of a dataset is edited by doing double click (or choosing the "Edit this cell" option), CARTO returns a http 400 bad request error and a message indicating that it can't edit the cell because the geometry of the row is invalid.

screen shot 2016-08-29 at 12 41 21

I've tried to edit the cells of the world borders dataset and barris de barcelona dataset (both from the data library). I can create maps form both datasets and the geometries are correctly rendered.

screen shot 2016-08-29 at 12 42 30

This doesn't happen with datasets that have point geometries.

nobuti:
This PR fixes #9592.

It fetches the row information when the geom is not loaded, that happens for polygons and lines, but not points. It uses the same approach as we had in the editor.

cc @gfiorav @xavijam

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/table/body/table-body-view.js
+++ b/lib/assets/javascripts/cartodb3/components/table/body/table-body-view.js
@@ -341,13 +341,7 @@ module.exports = CoreView.extend({
     }
   },
 
-  _editCell: function (rowModel, attribute) {
-    // If geom is not fetched, then bring the complete row information and wait
-    if (!rowModel.isGeomLoaded()) {
-      this._fetchRow(rowModel, attribute);
-      return;
-    }
-
+  _doCellEdition: function (rowModel, attribute) {
     var $tableRow = this.$('[data-model="' + rowModel.cid + '"]');
     var $tableCell = $tableRow.find('[data-attribute="' + attribute + '"]');
     var $options = $tableCell.find('.js-options');
@@ -402,6 +396,11 @@ module.exports = CoreView.extend({
       },
       position
     );
+  },
+
+  _editCell: function (rowModel, attribute) {
+    var callback = this._doCellEdition.bind(this, rowModel, attribute);
+    rowModel.fetchRowIfGeomIsNotLoaded(callback);
   },
 
   _removeRow: function (rowModel) {
--- a/lib/assets/javascripts/cartodb3/components/table/body/table-body-view.js
+++ b/lib/assets/javascripts/cartodb3/components/table/body/table-body-view.js
@@ -342,6 +342,12 @@ module.exports = CoreView.extend({
   },
 
   _editCell: function (rowModel, attribute) {
+    // If geom is not fetched, then bring the complete row information and wait
+    if (!rowModel.isGeomLoaded()) {
+      this._fetchRow(rowModel, attribute);
+      return;
+    }
+
     var $tableRow = this.$('[data-model="' + rowModel.cid + '"]');
     var $tableCell = $tableRow.find('[data-attribute="' + attribute + '"]');
     var $options = $tableCell.find('.js-options');
@@ -434,6 +440,15 @@ module.exports = CoreView.extend({
     this.$('.js-tbody').animate({
       scrollTop: tbodyHeight
     }, 'slow');
+  },
+
+  _fetchRow: function (rowModel, attribute) {
+    // This method is call from the editCell call, so in order to complete the flow
+    // we call it back again when the row is fetched
+    var cb = this._editCell.bind(this, rowModel, attribute);
+    rowModel.fetch({
+      success: cb
+    });
   },
 
   clean: function () {
--- a/lib/assets/javascripts/cartodb3/components/table/body/table-body-view.js
+++ b/lib/assets/javascripts/cartodb3/components/table/body/table-body-view.js
@@ -441,15 +441,6 @@ module.exports = CoreView.extend({
     }, 'slow');
   },
 
-  _fetchRow: function (rowModel, attribute) {
-    // This method is call from the editCell call, so in order to complete the flow
-    // we call it back again when the row is fetched
-    var cb = this._editCell.bind(this, rowModel, attribute);
-    rowModel.fetch({
-      success: cb
-    });
-  },
-
   clean: function () {
     this._destroyScrollBinding();
     CoreView.prototype.clean.apply(this);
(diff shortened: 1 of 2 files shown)


lib/assets/javascripts/cartodb3/data/query-row-model.js
```diff
@@ -39,6 +48,11 @@ module.exports = Backbone.Model.extend({
     return !this.has('cartodb_id');
   },
 
+  isGeomLoaded: function () {
+    var geojson = this.get('the_geom');
+    return (geojson !== 'GeoJSON' && geojson !== -1 && !_.contains(WKT_TYPES, geojson));
```
gfiorav: Interesting! What are the two first checks for?

nobuti:
This is in ded06.
Please @alonsogarciapablo CR 🙏

lib/assets/javascripts/cartodb3/components/table/body/table-body-view.js
```diff
@@ -342,6 +342,12 @@ module.exports = CoreView.extend({
   },
 
   _editCell: function (rowModel, attribute) {
+    // If geom is not fetched, then bring the complete row information and wait
+    if (!rowModel.isGeomLoaded()) {
```
alonsogarciapablo: Question: why is `the_geom` not loaded in the first place (when the row is loaded)?

nobuti: We are not talking about the type, but the geometry data. For points, this data is short, only two points but for lines and polygons, this data is usually much bigger. In order to not overload the payload (we are loading 40 rows for every page of a dataset), for these kind of geometry, this data is deferred until the user needs it.

alonsogarciapablo: I see. Thanks for the explanation!

lib/assets/javascripts/cartodb3/data/query-row-model.js
```diff
@@ -39,6 +48,11 @@ module.exports = Backbone.Model.extend({
     return !this.has('cartodb_id');
   },
 
+  isGeomLoaded: function () {
+    var geojson = this.get('the_geom');
+    return !_.contains(WKT_TYPES, geojson);
```
alonsogarciapablo: If `the_geom` contains the json that describes the geometry (as mentioned in https://github.com/CartoDB/cartodb/pull/9603/files/0d69305eb1e9050e1e3613c68cbe7ae40647a2e7#r76755972)... Wouldn't we need to be doing something like `return !_.contains(WKT_TYPES, geojson['type'])` here?

nobuti: By default, if `the_geom` doesn't include the data, is set to the literal constant, ie `MULTIPOLYGON`. If data exists, it's a json object. So, if `geojson` it's included in the array, it's need to be loaded.

alonsogarciapablo: Out of curiosity... where's that literal being set?

lib/assets/javascripts/cartodb3/components/table/body/table-body-view.js
```diff
@@ -436,6 +442,15 @@ module.exports = CoreView.extend({
     }, 'slow');
   },
 
+  _fetchRow: function (rowModel, attribute) {
```
alonsogarciapablo: Suggestion... we could move this function to `/lib/assets/javascripts/cartodb3/data/query-row-model.js`

```
// lib/assets/javascripts/cartodb3/components/table/body/table-body-view.js

...

  _editCell: function (rowModel, attribute) {
    rowModel.fetchRowIfGeomIsNotLoaded(function() {
      // do whatever _editCell is supposed to do
    })
  },

...

// lib/assets/javascripts/cartodb3/data/query-row-model.js

...

  fetchRowIfGeomIsNotLoaded: function (callback) {
    if (this.isGeomLoaded()) {
      callback();
      return;
    }

    this.fetch({
      success: callback
    });
  },

...
```

What do you think?

lib/assets/javascripts/cartodb3/data/query-row-model.js
nobuti: Not sure if this helps but this the call in order to render the table:

screen shot 2016-08-30 at 12 48 31

lib/assets/javascripts/cartodb3/components/table/body/table-body-view.js
nobuti: 💯  on it

lib/assets/javascripts/cartodb3/data/query-row-model.js
alonsogarciapablo: It does help. I guess we're not loading `the_geom` initially for performance reasons.

> if the_geom doesn't include the data, is set to the literal constant, ie MULTIPOLYGON. If data exists, it's a json object...

I guess we could also do something like:

```
 isGeomLoaded: function () {
  return _.isObject(this.get('the_geom'));
 }
```

nobuti: Oops, it seems I was wrong and it's a string :(

alonsogarciapablo: Yes. Then we could do something like:

```
isGeomLoaded: function () {
  try {
    JSON.parse(this.get('the_geom'));
    return true;
  } catch {
    return false;
  }
}
```

nobuti: I love it, but as we actually aren't catching any error, I'll switch that catch for finally.

alonsogarciapablo: `JSON.parse` will throw an error if `this.get('the_geom')` is a string that's why I was suggesting to use `catch`... If you can do it with a `finally`, perfect for me. The "big" win here is not having to check the geometry type explicitly 😄 .

lib/assets/javascripts/cartodb3/components/table/body/table-body-view.js
alonsogarciapablo: We can get rid of this method now, right? :)

alonsogarciapablo:
👍 after you remove `_fetchRow`. Thanks for your patience @nobuti!

lib/assets/javascripts/cartodb3/components/table/body/table-body-view.js
nobuti: Oops, right, sorry!

nobuti:
🚀  thanks to you @alonsogarciapablo!

gfiorav:
Thanks guys!

PR #9611, opened 2016-08-30
Fixes subsample of rows onboarding window to show a rounded percentage

### LINKED ISSUE #9610: Subsample of rows onboarding window shows an unrounded percentage
screen shot 2016-08-29 at 17 32 42

javierarce:
Closes #9610

CR: @nobuti

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/onboardings/analysis/analyses/sampling.tpl
+++ b/lib/assets/javascripts/cartodb3/components/onboardings/analysis/analyses/sampling.tpl
@@ -3,5 +3,5 @@
 </p>
 
 <p class="CDB-Text Onboarding-description">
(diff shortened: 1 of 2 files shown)


lib/assets/javascripts/cartodb3/components/onboardings/analysis/analyses/sampling.tpl
```diff
@@ -3,5 +3,5 @@
 </p>
 
 <p class="CDB-Text Onboarding-description">
-  <%- _t('analyses-onboarding.sampling.description', { percentage: sampling * 100 }) %>
+  <%- _t('analyses-onboarding.sampling.description', { percentage: Math.round(sampling * 100) }) %>
```
javierarce: I don't like to include logic in the templates, but this is the most practical way to do it in this particular case.

nobuti:
👍

PR #9612, opened 2016-08-30
Disable the option of adding widgets if no geometry data

### LINKED ISSUE #9560: Don't let widgets to be added when not having geometries on the data
rel with: https://github.com/CartoDB/cartodb/issues/9546

nobuti:
This PR fixes #9560.

widget-geometry

If the layer (aka the dataset) doesn't include geometry data:
- the data panel view of that layer doesn't show any column to add the widgets, showing the message you see above.
- the widgets view still invites the user to add widgets, but only for those layers with geometry data in its dataset.
- when there is no geometry data at all, this view shows the message too.

@javierarce could you take a look? Also, not convinced about the message, maybe (I'm sure actually) you can improve it :)

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/modals/add-widgets/create-tuples-items.js
+++ b/lib/assets/javascripts/cartodb3/components/modals/add-widgets/create-tuples-items.js
@@ -26,6 +26,7 @@ module.exports = function (analysisDefinitionNodesCollection, layerDefinitionsCo
 
       var querySchemaModel = nodeDefModel.querySchemaModel;
       if (!querySchemaModel) return tuplesItems;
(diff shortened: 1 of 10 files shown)


javierarce:
Code is 👍 

Regarding the error message, I'd prefer to have something more direct and actionable, and also similar to the other message we have there (`You have not added any widgets yet. Add new widgets to discover new things.`). 

I also think we should have two messages, one for the widget pane and another one for the data pane:

* **Data pane**: `Your dataset doesn't contain any geometry data`.
* **Widget pane**: `Your dataset lacks of geometry data and it's not possible to add a widget`.

They aren't great, but I think they'll do for now.

Issue #9615, opened 2016-08-31
Logging in Syncs is generating escaped & single-line huge logs in some specific scenarios


iriberri:
**STR**:

* Create a sync
* Let it sync one without data updates (probably not needed)
* Let it sync now with data updates successfully
* Break the file (add some nasty HTML+JS instead of your CSV)-> UnsupportedFileError
* Fix the file and let it sync
* Break the file -> UnsupportedError

Due to the fact that the exceptions raised in loader.rb are using the whole log as exception message, the log keeps growing and growing. Also, some incorrect line feed characters are provoking the backslashes to be escaped and the log to be a single line, which ends up blocking the sync as the log ends up being impossible to edit due to its size.

iriberri:
To start with, we found a probably unrelated bug here: https://github.com/CartoDB/cartodb/blob/62bfd6b1149ed861848d65c065d6754ef93e75fb/services/importer/lib/importer/job.rb#L37

Append method parameters are disordered. Check: https://github.com/CartoDB/cartodb/commit/d42e9079298fed18caeca6861d06572676b2c2ce#diff-048d440a8838cbe46765a9504863c889

jgoizueta:
We have several places in which we're using `'\n'` instead of `"\n"`. The first is a slash followed by an "n", not an end-of-line.

* https://github.com/CartoDB/cartodb/blob/f8c33235c974a49bc70b8cc0a393d79edcf25379/app/models/log.rb
* https://github.com/CartoDB/cartodb/blob/f8c33235c974a49bc70b8cc0a393d79edcf25379/app/models/synchronization/member.rb#L204
* https://github.com/CartoDB/cartodb/blob/f8c33235c974a49bc70b8cc0a393d79edcf25379/services/geocoder/lib/hires_geocoder.rb#L107
* https://github.com/CartoDB/cartodb/blob/f8c33235c974a49bc70b8cc0a393d79edcf25379/services/geocoder/lib/hires_geocoder.rb#L134

jgoizueta:
This is also unrelated to the error, but this is causing out of order lines in long logs:

https://github.com/CartoDB/cartodb/blob/master/app/models/log.rb#L163

We should replace that line by:

```
(@fixed_entries_half + @circular_entries_half[@circular_index..-1] + @circular_entries_half[0...@circular_index]).join('')
```

To keep the correct order of the lines in the circular buffer.

jgoizueta:
[...]

PR #9622, opened 2016-08-31
Fix large sync logs problems

jgoizueta:
Fix some issues in long synchronization logs, see #9615

### DIFF
--- a/app/models/log.rb
+++ b/app/models/log.rb
@@ -16,7 +16,7 @@ module CartoDB
     ENTRY_FORMAT = "%s: %s#{ENTRY_POSTFIX}"
     ENTRY_REHYDRATED_FORMAT = "%s#{ENTRY_POSTFIX}"
 
-    HALF_OF_LOG_MARK = '===LOG HALF===\n'
+    HALF_OF_LOG_MARK = "===LOG HALF===\n"
     END_OF_LOG_MARK = '===LOG END==='
 
     TYPE_DATA_IMPORT     = 'import'.freeze
--- a/app/models/log.rb
+++ b/app/models/log.rb
@@ -16,8 +16,8 @@ module CartoDB
     ENTRY_FORMAT = "%s: %s#{ENTRY_POSTFIX}"
     ENTRY_REHYDRATED_FORMAT = "%s#{ENTRY_POSTFIX}"
 
-    HALF_OF_LOG_MARK = "===LOG HALF===\n"
-    END_OF_LOG_MARK = '===LOG END==='
+    HALF_OF_LOG_MARK = "===LOG HALF===\n".freeze
+    END_OF_LOG_MARK  = '===LOG END==='.freeze
 
     TYPE_DATA_IMPORT     = 'import'.freeze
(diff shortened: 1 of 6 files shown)


jgoizueta:
I'm excluding fixes for two other problems detected in #9615 to expedite solving the problem which is hurting some users.

jgoizueta:
@iriberri @ethervoid please review

app/models/log.rb
```diff
@@ -16,7 +16,7 @@ class Log < Sequel::Model
     ENTRY_FORMAT = "%s: %s#{ENTRY_POSTFIX}"
     ENTRY_REHYDRATED_FORMAT = "%s#{ENTRY_POSTFIX}"
 
-    HALF_OF_LOG_MARK = '===LOG HALF===\n'
+    HALF_OF_LOG_MARK = "===LOG HALF===\n"
```
iriberri: Agree :D

services/importer/lib/importer/exceptions.rb
```diff
@@ -16,7 +16,11 @@ def initialize(message, error_code=nil)
     end
 
     # Generic/unmapped errors
-    class GenericImportError < StandardError; end
+    class GenericImportError < StandardError;
```
iriberri: Yep!

iriberri: Actually with the change here, the KmlWithoutStyleIdError exception will now return an error which is not very accurate.

services/importer/lib/importer/loader.rb
```diff
@@ -306,24 +306,24 @@ def check_for_import_errors
...
         end
 
         if ogr2ogr.unsupported_format?
-          raise UnsupportedFormatError.new(job.logger)
+          raise UnsupportedFormatError.new
         end
 
         if ogr2ogr.kml_style_missing?
-          raise KmlWithoutStyleIdError.new(job.logger)
+          raise KmlWithoutStyleIdError.new
```
iriberri: I'd add a message here too. Otherwise this is returning the default message. This error is for KMLs which don't have the StyleID tag which is mandatory.

services/importer/lib/importer/loader.rb
```diff
@@ -293,9 +293,9 @@ def try_fallback(append_mode)
       end
 
       def check_for_import_errors
-        raise DuplicatedColumnError.new(job.logger) if ogr2ogr.duplicate_column?
-        raise InvalidGeoJSONError.new(job.logger) if ogr2ogr.invalid_geojson?
-        raise TooManyColumnsError.new(job.logger) if ogr2ogr.too_many_columns?
+        raise DuplicatedColumnError.new if ogr2ogr.duplicate_column?
```
iriberri: Same as for the KML thing, not sure if it would be worthy to add an explicit message here. In the end this is just for logging and the exception class will be printed so... the message wouldn't really matter actually, as far as the exception is clear.

iriberri:
I added some minor comments, but overall it looks good! :)

Issue #9624, opened 2016-08-31
Import/Sync log lines can be out of order


jgoizueta:
Each time `entries` is generated (when an updated `Log` is stored), the circular buffer used when the number of log lines is large (over 500 lines) is dumped in the stored order, but entries starting at the circular array pointer `@circular_index` are actually older than the preceding entries in the array (since they're most recently overwritten).

The problem is [here](https://github.com/CartoDB/cartodb/blob/1b6dac2daf2810d43904653d6558a1d3ef52de34/app/models/log.rb#L163)

See https://github.com/CartoDB/cartodb/issues/9615#issuecomment-243737746

Also, I think the 'minus one' [here](https://github.com/CartoDB/cartodb/blob/1b6dac2daf2810d43904653d6558a1d3ef52de34/app/models/log.rb#L163) should be removed (as the index points to the place where the next appended line will be stored). And I'd also add an  `else` to that conditional, setting `@circular_index` to zero (we start adding lines at the start of an empty buffer).

Issue #9623, opened 2016-08-31
Lines appended via Job to Import/Sync logs have wrong timestamp, truncation


jgoizueta:
As detected in https://github.com/CartoDB/cartodb/issues/9615#issuecomment-243726968

The order of the `Log#append` parameters [here](https://github.com/CartoDB/cartodb/blob/62bfd6b1149ed861848d65c065d6754ef93e75fb/services/importer/lib/importer/job.rb#L37) is wrong. (this [partial revert](https://github.com/CartoDB/cartodb/commit/d42e9079298fed18caeca6861d06572676b2c2ce#diff-048d440a8838cbe46765a9504863c889) caused the inconsistency).

As a result, log lines generated by `Job#log` have true/false (the value of `truncate`) for the timestamp at the start of the line, and they're never truncated.

PR #9631, opened 2016-09-01
Log fixes

jgoizueta:
See #9624 #9623

### DIFF
--- a/spec/models/log_spec.rb
+++ b/spec/models/log_spec.rb
@@ -42,8 +42,9 @@ describe CartoDB::Log do
       text5 = "5"
       text6 = "6"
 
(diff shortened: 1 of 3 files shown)


jgoizueta:
@iriberri please review

iriberri:
LGTM, let's see if @ethervoid can add something else! :)

spec/models/log_spec.rb
```diff
@@ -15,8 +15,8 @@
       timestamp = Time.now.utc
 
       log = Log.new({ type: type })
-      log.append(text1, timestamp)
```
ethervoid: What about to remove the timestamp because is the same that the default value in the method definition

ethervoid:
Looks good to me. Good catch!

PR #9665, opened 2016-09-05
Check alternative_names is not undefined before accessing properties.

nobuti:
This PR fixes this annoying bug:

bug-empty-panel

Not sure if it hides any deeper issue, but it fixes the empty panel.
CR @matallo 🙏


matallo:
ooop, LGTM

PR #9674, opened 2016-09-05
Connectors API Revamp

jgoizueta:
The (database) connectors API is changed and the code is refactor to make it easy to have specific provider-dependent variations (e.g. different parameters) and to add new providers.

Previously all the logic of database connectors was handled in the `Connector` class, and the API was transparent to (ODBC) connection parameters: the driver and all the driver-dependent parameters were directly passed to the ODBC backend.

Now we have a `provider` parameter which determines which `Provider`-derived class will be used by `Connector` to handle the connection. These classes can validate accepted parameters, map the exposed parameter names to the internal parameters of the ODBC driver and fix some internal parameters (e.g. the ODBC driver).

The new code organisation makes it easy to add non-ODBC providers, based on other FDWs.


services/importer/spec/unit/connector_spec.rb
```diff
@@ -0,0 +1,1167 @@
...
+          }]
+        }, {
+          # DROP FOREIGN TABLE
+          mode: :superuser,
+          sql: [{
+            command: :drop_foreign_table_if_exists,
+            table_name: foreign_table_name,
+            cascade: /CASCADE/i
+          }]
+        }, {
```
jgoizueta: I resist to remove those brackets in these cases, for legibility. This is not the typical case where the last argument is a hash of options, which is clearer removing the brackets.

jgoizueta:
Please review this: @ethervoid @iriberri

services/importer/lib/importer/connector.rb
```diff
@@ -125,85 +120,18 @@ def last_modified
...
-
-      def connection_params_errors
-        errors = []
-        if @connection.blank?
-          errors << "Connection parameters are missing"
-        elsif @dsn.blank? && @driver.blank?
-          errors << "Missing required connection parameters"
-        end
-        errors
+        @provider_name = Support.fetch_ignoring_case(@params, 'provider')
```
ethervoid: There is no way to get a key from a hash ignoring case?

services/importer/lib/importer/connector.rb
```diff
@@ -216,7 +144,7 @@ def foreign_prefix
       end
 
       def result_table_name
-        Carto::DB::Sanitize.sanitize_identifier @options['table']
+        Carto::DB::Sanitize.sanitize_identifier @provider.table_name
```
ethervoid: You could add the sanitize inside the table_name method of provider.

services/importer/lib/importer/connector.rb
```diff
@@ -125,85 +120,18 @@ def last_modified
...
-          errors << "Missing required connection parameters"
-        end
-        errors
+        @provider_name = Support.fetch_ignoring_case(@params, 'provider')
       end
 
       def connector_name
         max_len = MAX_PG_IDENTIFIER_LEN - @unique_suffix.size - MIN_TAB_ID_LEN - 1
-        name = Carto::DB::Sanitize.sanitize_identifier @dsn || @driver
+        name = Carto::DB::Sanitize.sanitize_identifier @provider_name
```
ethervoid: Maybe you can use the @provider.name method instead, for the sake of consistency

services/importer/lib/importer/connector.rb
jgoizueta: No that I know of.

services/importer/lib/importer/connector.rb
jgoizueta: Yep. I'll do

services/importer/lib/importer/connector.rb
jgoizueta: Or maybe add result_table name to provider doing the sanitize, while keeping the original table_name from the table parameter unsanitized (since it can refer to a remote table)...

services/importer/lib/importer/connector.rb
```diff
@@ -55,12 +49,13 @@ def initialize(connector_source, options = {})
...
         @tracker = tracker
         @job.log "Connector #{@json_params}"
-        # TODO: deal with schemas, org users, etc.,
         # TODO: logging with CartoDB::Logger
         table_name = @job.table_name
         qualified_table_name = %{"#{@job.schema}"."#{table_name}"}
         @job.log "Creating Server"
         execute_as_superuser create_server_command
+        @job.log "Creating Usermap"
+        execute_as_superuser create_usermap_command
```
iriberri: Not really important, but I see you create here the user mapping for the FDW and then you drop it in cascade together with the server. Perhaps the user mapping could be deleted more explicitly, although this wouldn't make much difference besides the readability of dropping the same things you create in the code.

services/importer/lib/importer/connector/providers/base.rb
```diff
@@ -0,0 +1,93 @@
...
+          @params = params
+        end
+
+        private
+
+        include CartoDB::Importer2::Connector::Support
+        include FdwSupport
+
+        def must_be_defined_in_derived_class
+          raise "Method #{caller_locations(1, 1)[0].label} must be defined in derived class"
```
ethervoid: For these cases I tend to use http://ruby-doc.org/core-2.2.0/NotImplementedError.html

services/importer/lib/importer/connector/providers/base.rb
```diff
@@ -0,0 +1,93 @@
...
+          required_parameters + optional_parameters
+        end
+
+        def initialize(params = {})
+          @params = params
+        end
+
+        private
+
+        include CartoDB::Importer2::Connector::Support
```
ethervoid: For me is better to put the includes below the class name

services/importer/lib/importer/connector/support.rb
```diff
@@ -0,0 +1,20 @@
+module CartoDB
+  module Importer2
+    class Connector
+      module Support
```
ethervoid: I think is better to have a ParamsSupport or ParamsHelper class that encapsulates the parameters management logic

services/importer/lib/importer/connector/providers/generic_odbc.rb
```diff
@@ -0,0 +1,45 @@
...
+      # Generic ODBC provider passing through any ODBC connection attributes in the 'connection' parameter.
+      # Either a 'driver' or a 'dns' parameter must be present in 'connection'.
+      #
+      # This is not meant for public use.
+      #
+      class GenericOdbcProvider < OdbcProvider
+
+        def initialize(params)
+          super
+          @connection = Support.fetch_ignoring_case(@params, 'connection')
```
ethervoid: One example of use for the Params class is to raise an exception if the connection parameter is not setted

services/importer/lib/importer/connector/providers/generic_odbc.rb
```diff
@@ -0,0 +1,45 @@
...
+          if @connection
+            @dsn        = Support.fetch_ignoring_case(@connection, 'dsn')
+            @driver     = Support.fetch_ignoring_case(@connection, 'driver')
+          end
+        end
+
+        private
+
+        def connection_errors
```
ethervoid: This logic should be encapsulated in a Parameters class

services/importer/lib/importer/connector/providers/base.rb
jgoizueta: I'd do that for the public interface, but since this is implementation detail I prefer to bury it behind the `private` :grin:

services/importer/lib/importer/connector/providers/base.rb
jgoizueta: Nice, thx!

services/importer/lib/importer/connector/providers/generic_odbc.rb
jgoizueta: I think the Parameters class is a good idea, I'll give it a try

services/importer/lib/importer/fdw_support.rb
```diff
@@ -0,0 +1,101 @@
...
+          #{options_clause(options)};
+       }
+    end
+
+    def fdw_drop_server(server_name)
+      "DROP SERVER IF EXISTS #{server_name} CASCADE;"
+    end
+
+    def fdw_drop_foreign_table(schema_name, table_name)
+      %{DROP FOREIGN TABLE IF EXISTS #{qualified_table_name(schema_name, table_name)} CASCADE;}
```
ethervoid: Why do we want to execute cascade here?

ethervoid:
There is one thing we could add here. We could provide the JSON/whatever structure with the required and optional parameter structure to have it ready for the backend implementation.

What do you think?

ethervoid:
I've been reading everything and it looks fine but I have to say the there's a lot of logic so in one CR is hard to tell everything is ok.

services/importer/lib/importer/fdw_support.rb
iriberri: Is it because of the metadata table? As I said above I'd probably get rid of all cascades and make the user mapping and this one explicit.

jgoizueta: I'll get rid of the cascades and drop user mappings

jgoizueta:
I had in mind we'd need to add things like the parameters information to implement the Backend. Now I see we actually now what we need from our sketch of the Backed API, so I think I'll add it right away.

services/importer/lib/importer/connector.rb
jgoizueta: On second thought I wont :grin:

The Provider classes don't have names; we assign *provider names* to Provider classes in `connector/provider.rb`.

services/importer/lib/importer/connector/providers/pg_fdw.rb
```diff
@@ -0,0 +1,138 @@
...
+    class Connector
+
+      # PostgreSQL provider using Postgres-FDW: https://www.postgresql.org/docs/current/static/postgres-fdw.html
+      #
+      # options:
+      # * https://www.postgresql.org/docs/current/static/postgres-fdw.html#AEN174711
+      # * https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-PARAMKEYWORDS
+      #
+      # Requirements:
+      #   * postgres_fdw extension must be installed in the user database
```
rafatower: So this is finally going through that instead of odbc, just surprised about it. Hope it never causes trouble with queries such as bringing big tables locally instead of executing the whole thing in the remote end.

ethervoid: Check line 28

rafatower: :+1: thanks for the info :smile:

Issue #9588, opened 2016-08-29
Import timeouts are happening from the import_cleanup process at Table model


iriberri:
This part of the code is using connections to the database "as superuser". Perhaps we should switch it by the new methods (direct connections or transactions with timeout) we've been using lately.

https://github.com/CartoDB/cartodb/blob/6f060448fb68288d7f9695906df4b556c213b6fc/app/models/table.rb#L302

iriberri:
cc @oriolbx

iriberri:
Started to check this. The import_cleanup process is actually flagged as potential code to be removed because the CartoDBfication already manages similar issues at `_CDB_Has_Usable_Primary_ID`. I'm gonna study it to see if it can be completely dropped, or otherwise treat better the timeout.

iriberri:
There's a difference between this code and CartoDBfy: the ruby part gets rid of other known columns, generated usually by Ogr2ogr which are used as PK, but CartoDBfication doesn't. Removing the code without changing the PK name in ogr2ogr would provoke that exporting and importing a table (with cartodb_id) would add an extra gid column on each import, because this extra column wouldn't be dropped.

rafatower:
Already solved in the mentioned PR, thus closing this.

PR #9685, opened 2016-09-06
Move sql statements into transaction_with_timeout (Table#import_cleanup)

iriberri:
WIP probably, not proud of this diff, so very open to suggestions. Also a note: this code should be deleted as soon as we apply some pending tweaks to ogr: setting `cartodb_id` as ogr primary key.

====

In #9588 I detected timeouts for the statements inside `import_cleanup`. Then I noticed that these were not using our "timeout improvements" that we applied months ago to other methods which ran queries to the database and can be expensive if the dataset is big (I reproduced it with ~1M rows dataset).

We want to use `transaction_with_timeout`, which is the only way we have right now to go through pgbouncer and be able to set a statement timeout without it getting overriden. Previously, the whole method was running queries via `in_database as superuser` but this doesn't take care of timeouts.

The problem now is: when using `transaction_with_timeout`, we need to split the code into several transactions because `ALTER TABLE` statements (not all, but the ones here) are using an exclusive access lock over the table, and this lock is not closed until the transaction commits.

I have used  `transaction_with_timeout` splitting it as soon as an `ALTER` statement occurs -- you can see I aggregated a set of update/selects/alter because locks don't cause a trouble there.

### DIFF
--- a/app/models/table.rb
+++ b/app/models/table.rb
@@ -300,7 +300,6 @@ class Table
   # TODO: basically most if not all of what the import_cleanup does is done by cartodbfy.
   # Consider deletion.
   def import_cleanup
-
       # When tables are created using ogr2ogr they are added a ogc_fid or gid primary key
       # In that case:
       #  - If cartodb_id already exists, remove ogc_fid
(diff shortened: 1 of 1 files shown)


iriberri:
Oooops, the new blocks are defining variables which then are not in the scope when used. This should be reviewed.

iriberri:
Hey @ethervoid and @jgoizueta, I'd appreciate your thoughts (and review) here! :-)

jgoizueta:
:+1:

xavijam:
Ready to be merged?

iriberri:
Nope @xavijam :*

@rafatower could you give us your thoughts? :)

rafatower:
:+1: consider what hound says (mostly cosmetic) and proper testing in staging (minicloud if possible) I take for granted that you tested it locally.

thanks! :D

iriberri:
Thanks!

I already tested it in staging too (not the mincloud, though) but given that this has been opened for some time already I'd better check again a bit more exhaustively. I'll take this out next Monday.

iriberri:
Still testing this in the minicloud with the file with which I was able to reproduce the problem.

I just got something I never saw before:

````
2016-09-19 09:35:43 UTC: Exception: PG::Error: ERROR:  CDB(_CDB_Add_Indexes:53200:out of memory): ALTER TABLE block_buildings_wgs84_metadataremoved ADD PRIMARY KEY (cartodb_id)
````

Issue #9696, opened 2016-09-07
CartoDB login loads 10 sec later


icetorm:
Hi everyone, 
I installed CARTO in ubuntu 12.04 and it works smoothly with some tweaks that i made. However when i load  http://localhost:3000/login the background load instantly but the login takes 10 seconds to load everytime.

Thanks. Your help will be really appreciated.

iriberri:
This is happening because of the Google Plus login option. In order to avoid it, you might want to comment or remove the  `google_plus` line inside `oauth` in your `app_config.yml` file. Explicitly, these ones [here](https://github.com/CartoDB/cartodb/blob/47e653514133c241a5e630a85484fe21a697e773/config/app_config.yml.sample#L299-L302):

```yml
    #google_plus:
      #client_id: ''
      #cookie_policy: 'single_host_origin'
      #cookie_policy: 'https://cartodb-staging.com'
```

icetorm:
Thanks @iriberri , my problem is solved.

iriberri:
That's great @icetorm! By the way, I'm going to comment those lines in the sample app config probably, as it won't work unless configured anyway.

PR #9697, opened 2016-09-07
Comment google_plus oauth config

iriberri:
It will not work unless configured and we've got several questions/issues about this, such as https://github.com/CartoDB/cartodb/issues/9696.


iriberri:
Retest this please

PR #9698, opened 2016-09-07
Check styles from layer when app is initialized

### LINKED ISSUE #9684: Set simple style if current type is NONE and there is no analysis running
cc @saleiva @rochoa @javisantana

xavijam:
So basically the problem found is:

1. Create a map with a dataset of points (so many of them).
2. Apply a complex analysis, it will take several hours.
3. Close the builder, because you need that analysis finished in order to keep going.
4. Open the builder after X hours, and check that the analysis has finished but the styles are pink, because the style type is the NONE one (@saleiva's favourite).

We used to change the style when an analysis has finished, but if the tab is closed... 🙈 .

About the approach, I don't like the idea of change the style when the application has started and without a user change, but looks like the "way" of doing it.

CR: @alonsogarciapablo 
cc @rochoa @saleiva @javisantana 

Fixes #9684.

- [x] Pending tests.


lib/assets/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -68,6 +69,32 @@ var F = function (opts) {
   VisNotifications.track(this._vis());
 };
 
+F.prototype._checkStyle = function (layerDefModel) {
+  var nodeDefModel = layerDefModel.getAnalysisDefinitionNodeModel();
+  var analysisCollection = this._analysis();
+  var nodeModel = analysisCollection && analysisCollection.findNodeById(layerDefModel.get('source'));
+  var isAnalysisNode = nodeModel && nodeModel.get('type') !== 'source';
+  var isRunning = nodeModel && nodeModel.get('status') !== 'ready';
```
alonsogarciapablo: Would [`nodeModel.isDone()`](https://github.com/CartoDB/cartodb.js/blob/v4/src/analysis/analysis-model.js#L107) work here?

lib/assets/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -68,6 +69,32 @@ var F = function (opts) {
   VisNotifications.track(this._vis());
 };
 
+F.prototype._checkStyle = function (layerDefModel) {
```
alonsogarciapablo: In general, I find too many details in this method (notice that we have a lot of `something.somethingelse.somethingelsemore.something(...)`). Instead of having all these details in this one method, we could push some of them up in the hierarchy to simplify this method and make it a bit more understandable. eg:

```
F.prototype._checkStyle = function (layerDefModel) {
  var nodeDefModel = layerDefModel.getAnalysisDefinitionNodeModel();
  var nodeModel = this._analysis().findNodeById(layerDefModel.get('source'));
  if (layerDefModel.hasDefaultStyles() && !nodeModel.isSource() && nodeModel.isDone()) {
    layerDefModel.resetStylesToSimple();
  }
},
```

:point_up: basically moves some of the details to `layerDefModel.hasDefaultStyles()` and `layerDefModel.resetStylesToSimple()`.

Also, `_checkStyles` is a bit generic IMO. It's doing more than just a "check": it's applying some simple default styles to the layer if it has some "none" default styles and has an analysis that is done... Perhaps we could change the name to something like `_applySimpleStylesIfStylesCouldNotBeApplied` or `_applySimpleStylesIfHasDefaultStyles`... you can probably think of a better name...

lib/assets/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -68,6 +69,32 @@ var F = function (opts) {
...
+    var simpleGeom = nodeDefModel.querySchemaModel.get('simple_geom');
+
+    var applyDefaultStyles = function () {
+      simpleGeom = nodeDefModel.querySchemaModel.get('simple_geom');
+      layerDefModel.styleModel.setDefaultPropertiesByType('simple', simpleGeom);
+      nodeDefModel.querySchemaModel.unbind(null, null, self);
+    };
+
+    if (!simpleGeom) {
+      nodeDefModel.querySchemaModel.bind('change:simple_geom', applyDefaultStyles, this);
```
alonsogarciapablo: I'm guessing `simple_geom` might not have been fetched initially, and that's why we have to way. Not sure if we are fetching that geom right away or it could be fetched due to something unrelated... eg: the analysis that took X hours is removed and geom changes, and `applyDefaultStyles` is still bound to the `querySchemaModel`... could this happen?

Also, could we use `.once` here so that we don't need to `.bind` and `.unbind`?

alonsogarciapablo:
Left some comments/questions. About the approach, It's scary, but I couldn't think about anything better to solve it this. I was worried that this approach wouldn't work if the user went directly to see the embed after X hours, but I guess the map needs to be re-published anyway... so 👍 

I guess you'll be adding some minimal testing @xavijam. Thanks!

lib/assets/javascripts/cartodb3/deep-insights-integrations.js
xavijam: Yeah working \o/.

lib/assets/javascripts/cartodb3/deep-insights-integrations.js
xavijam: I would not add anything about styles into layerDefModel, in my opinion. Good to have the isSource method in order to know the type.

xavijam: And of course about the change of the function.

lib/assets/javascripts/cartodb3/deep-insights-integrations.js
xavijam: I forgot to add the proper `fetch` after the bind.

lib/assets/javascripts/cartodb3/deep-insights-integrations.js
xavijam: `isSource` should be available in CartoDB.js, because that nodeDefModel comes from there.

xavijam:
Added tests and applied feedback

lib/assets/test/spec/cartodb3/deep-insights-integrations.spec.js
```diff
@@ -136,6 +137,92 @@ describe('deep-insights-integrations', function () {
...
+      this.integrations._resetStylesIfNoneApplied(this.layerDefModel);
+      expect(nodeDef.querySchemaModel.fetch).toHaveBeenCalled();
+      expect(this.layerDefModel.styleModel.setDefaultPropertiesByType).not.toHaveBeenCalled();
+    });
+
+    it('should reset styles if layer has none styles', function () {
+      var nodeMod = this.analysis._analysisCollection.at(1);
+      nodeMod.isDone.and.returnValue(true);
+      this.integrations._resetStylesIfNoneApplied(this.layerDefModel);
+      expect(this.layerDefModel.styleModel.setDefaultPropertiesByType).toHaveBeenCalled();
```
alonsogarciapablo: I'd personally be a bit more explicit a bit this... eg:

```
expect(this.layerDefModel.hasNoneStyles()).toBeTruthy();
this.integrations._resetStylesIfNoneApplied(this.layerDefModel);
expect(this.layerDefModel.hasNoneStyles()).toBeFalsy();
```

alonsogarciapablo:
A bit better now (although I still think that `_resetStylesIfNoneApplied ` has too much detail). 

👍

lib/assets/test/spec/cartodb3/deep-insights-integrations.spec.js
xavijam: Agree!

xavijam:
:joy: come on! it is really simplified, and If it is possible to create more helper methods, I'd do it.

PR #9712, opened 2016-09-08
Add rake task to adapt syncs to API changes

jgoizueta:
The rake taks will adapt existing connector synchronizations
to work with the new API defined by #9674


jgoizueta:
Please @iriberri take a look

iriberri:
LGTM

PR #9714, opened 2016-09-08
Set properly the style model definition when a new layer is added (from import or selection)

xavijam:
Basically we used to set the CartoCSS properly [here](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/data/user-actions.js#L31) but we were forgetting the style model definition for the style forms. So there was a problem like the image below:

screen shot 2016-09-08 at 12 39 11

Requesting the geometry when a new dataset was imported did the trick.

Also, fixed problem with negative progress when a new file was imported + correctly sorted import steps thanks to @iriberri.

CR: @nobuti 
cc @javisantana


nobuti:
LGTM

xavijam:
Frontend tests passing, merging! 💋

lib/assets/test/spec/cartodb3/data/editor-background-polling-model.spec.js
```diff
@@ -1,12 +1,22 @@
...
 var ImportsCollection = require('../../../../javascripts/cartodb3/data/background-importer/background-importer-imports-collection');
 var UserActions = require('../../../../javascripts/cartodb3/data/user-actions');
 
 describe('cartodb3/data/editor-background-polling-model', function () {
   beforeEach(function () {
-    var configModel = 'c';
+    var configModel = new Backbone.Model();
+    configModel.urlVersion = function () {};
+
+    spyOn($, 'ajax').and.callFake(function (req) {
```
nobuti: sweet!

PR #9729, opened 2016-09-09
Provide a default style form definition 

xavijam:
So we have managed the problem about CartoCSS when a new dataset is imported thanks to @guido in #9726, but unfortunately we have to do the same with the style form. We have to make use of the `style_properties` in the layer generation and fill it with the `default-form-styles.json` + `default-cartography.json`.

### Pending
- [x] Add the necessary logic in the backend in order to give the form style definition for the provided geometry (@gfiorav 🙏  👼 ).


xavijam:
cc @juanignaciosl

xavijam:
:kiss: ❤️

gfiorav:
Please @javitonino have a look!

lib/carto/form.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    def forms_definition
+      @forms_definition ||= Carto::Definition.instance
+                                             .load_from_file(DEFAULT_FORMS_DEFINITION_LOCATION)
+    end
+
+    def style_definition
+      return @style_definition if @style_definition
+
+      @style_definition = style.default_definition
```
javitonino: Maybe remove the `return` and just `@style_definition ||= style.default_definition`

javitonino:
You could expand the tests from the previous PR to check that `style_options` is set for builder users but not set for editor users.

Apart from that and a minor comment (feel free to ignore), looks ready to me.

gfiorav:
Added 'em. Have a final look please!

javitonino:
:ship: it!

PR #9733, opened 2016-09-12
C1452 remove dev mobile app type

javitonino:
Removing dev app_type, we will give open apps and most features to all users, so it doesn't make much sense to have this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cartodb-ui",
-  "version": "4.1.43",
+  "version": "4.1.44",
   "description": "CartoDB UI frontend",
   "repository": {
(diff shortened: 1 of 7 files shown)


javitonino:
Please CR @gfiorav

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cartodb-ui",
-  "version": "4.1.43",
+  "version": "4.1.44",
```
gfiorav: I'd bump this after merge, but if you're quick enough it's ok!

gfiorav:
Looks correct. You could leave a comment explaining very briefly why we're removing dev though hehe

📦  it!

PR #9747, opened 2016-09-13
Removed polyglot requirement from style-form-components-dictionary

### LINKED ISSUE #9474: Remove polyglot setup from non-entry file
https://github.com/CartoDB/cartodb/blob/327610c5436a3168f293acddf9f2d7bf0b2f3490/lib/assets/javascripts/cartodb3/editor/style/style-form/style-form-components-dictionary.js#L2-L10

This should only be setup in an entry file, based on current implicit usage convention. This could/would leave to subtle bugs and problems down the line if/when translation files changes to not use the same paths etc.

The test environments should already be prepared ([cartodb3](https://github.com/CartoDB/cartodb/blob/14c5d7448e9b6aff306799fe86025674dec6e3fd/lib/assets/test/spec/SpecHelper3.js#L36-L38), [CLI](https://github.com/CartoDB/cartodb/blob/14c5d7448e9b6aff306799fe86025674dec6e3fd/lib/assets/test/jasmine/helpers/SpecHelper.js#L3-L10)), so tests should not require this either. If the problem are assertions I'd recommend to relax them, e.g.:
```js
expect(this.view.el.innerHTML).toContain('check-the.token.instead')
``` 
or even
```js
expect(this.view.$('.some-element').html().trim()).not.toEqual('')
```

xavijam:
Nothing else, fixes #9474.

CR: @nobuti

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/style/style-form/style-form-components-dictionary.js
+++ b/lib/assets/javascripts/cartodb3/editor/style/style-form/style-form-components-dictionary.js
@@ -1,13 +1,4 @@
 var _ = require('underscore');
-var ACTIVE_LOCALE = window.ACTIVE_LOCALE;
-var Polyglot = require('node-polyglot');
-var Locale = require('../../../../../locale/index');
-var polyglot = new Polyglot({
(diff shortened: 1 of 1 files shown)


nobuti:
🎏

PR #9748, opened 2016-09-13
Refactor Connector classes

jgoizueta:
The original `CartoDB::Importer2::Connector` class was taking care of two different responsibilities:

* Execute the table import process using connectors (acting as a substitute for `CartoDB::Importer2::Runner`.
* Implementing the connector logic (based on Providers)

Now this is separated into two different classes:

* `CartoDB::Importer2::ConnectorRunner` managing the import process
* `Carto::Connector` handling connector operations via Providers

With this separation we can now implement connector-related operations (like list tables) which are not related to the import process in `Carto::Connector`.

Note that all the Connector stuff that used to be in `services/importer/lib/importer/connector` is now in `lib/carto/connector`. This is all related to the `Carto::Connector` and will not only used from the Import API.

- [x] Refactor  `Connector` -> `ConnectorRunner`, `Connector`
- [x] Implement limits + app_config limits/availability (per user/provider will come later)
- [x] Add `list_tables` functionality to `Connector`


jgoizueta:
@ethervoid this is the refactor previous to adding the list tables, and limits functionality that will be used from the new connector endpoints for the UI.

/cc @rafatower

jgoizueta:
Having moved all those files, the diff here is of not much use to review the changes 😢

jgoizueta:
Note: some minor changes to the Hive provider class have slipped here. This is not good practice, but well... I've left them as they were motivated by running the new tests and are not too relevant.

lib/carto/connector/providers/pg_fdw.rb
```diff
@@ -0,0 +1,136 @@
...
+        {
+          "list_tables": true,
+          "list_databases": false,
+          "sql_queries": false
+        }
+      end
+
+      private
+
+      ATTRIBUTE_NAMES = {
```
jgoizueta: Hey houndci what's wrong with you? it is frozen!

lib/carto/connector/providers.rb
```diff
@@ -0,0 +1,22 @@
...
+require_relative 'providers/generic_odbc'
+require_relative 'providers/mysql'
+require_relative 'providers/postgresql'
+require_relative 'providers/sqlserver'
+require_relative 'providers/hive'
+require_relative 'providers/pg_fdw'
+
+module Carto
+  class Connector
+    PROVIDERS = {
```
jgoizueta: I know, I know... as soon as I figure out how to add another provider for testing I'll fix this

lib/carto/connector/parameters.rb
```diff
@@ -0,0 +1,159 @@
...
+        Hash[hash.map { |name, value| [normalized_key(name), value] }]
+      end
+
+      def normalized_array(array)
+        array.map { |name| normalized_key(name) }
+      end
+
+      # returns [actual_name, value]
+      def fetch(name)
+        @params.find { |internal_name, _value| internal_name.to_s.casecmp(name.to_s).zero? }
```
jgoizueta: I prefer find. :expressionless:

lib/carto/connector/providers.rb
ethervoid: Nor sinner, you haven't a .freeze at the end of the provider Hash

lib/carto/connector.rb
```diff
@@ -0,0 +1,162 @@
...
+      @params = Parameters.new(parameters)
+
+      @name = @params[:provider]
+      @name ||= DEFAULT_PROVIDER
+
+      raise InvalidParametersError.new("Provider not defined") if @name.blank?
+      @provider = PROVIDERS[@name].try :new, @params
+      raise InvalidParametersError.new("Invalid provider: #{@name}") if @provider.blank?
+
+      # @schema = @user.database_schema
```
ethervoid: Remove unused comments

lib/carto/connector.rb
```diff
@@ -0,0 +1,162 @@
...
+    def remote_data_updated?
+      # TODO: can we detect if query results have changed?
+      true
+    end
+
+    def remote_table_name
+      @provider.table_name
+    end
+
+    def self.check_availability!(user)
```
ethervoid: We have to think how to make this more fine-grained

lib/carto/connector/providers.rb
```diff
@@ -0,0 +1,22 @@
...
+
+require_relative 'providers/generic_odbc'
+require_relative 'providers/mysql'
+require_relative 'providers/postgresql'
+require_relative 'providers/sqlserver'
+require_relative 'providers/hive'
+require_relative 'providers/pg_fdw'
+
+module Carto
+  class Connector
```
ethervoid: I think is better to have a Factory instead to make a pseudo-factory through a Hash and make the instance in the initialize of the Connector. Moreover in any case we're coupling the connector with the provider which should be a dependency.

jgoizueta: Agreed, I'll do it in the upcoming changes to support limits, because I just needed to add more info to PROVIDERS (actually I had added a few methods to interface with it, but I'll rethink it as a provider factory.

lib/carto/connector.rb
```diff
@@ -0,0 +1,293 @@
...
+    def check_availability!
+      Connector.check_availability!(@user)
+      if !available?
+        raise ConnectorsDisabledError.new(user: @user, provider: @name)
+      end
+    end
+
+    # Limits for the user/provider
+    def limits
+      Connector.limits provider: @name, user: @user
```
ethervoid: Why don't use the old fashion way to call methods?

lib/carto/connector.rb
```diff
@@ -0,0 +1,293 @@
...
+    # By default all parameters are validated
+    def validate!(only: nil)
+      @provider.validate!(only: only)
+    end
+
+    def log(message, truncate = true)
+      @logger.append message, truncate if @logger
+    end
+
+    MAX_PG_IDENTIFIER_LEN = 60
```
ethervoid: Should be 63 isn't it?

lib/carto/connector/errors.rb
```diff
@@ -0,0 +1,30 @@
+module Carto
+  class Connector
+
+    class ConnectorError < StandardError
+      attr_reader :user_name
+
+      def initialize(message: 'General error', user: nil, provider: nil)
+        @user_name = user && user.username
```
ethervoid: Cool way to do it :)

lib/carto/connector/fdw_support.rb
```diff
@@ -0,0 +1,114 @@
...
+
+      def fdw_grant_select(schema_name, table_name, user_name)
+        %{
+          GRANT SELECT ON #{qualified_table_name(schema_name, table_name)} TO "#{user_name}";
+         }
+      end
+
+      def fdw_create_foreign_table(server_name, schema_name, table_name, columns, options)
+        %{
+          CREATE FOREIGN TABLE #{qualified_table_name(schema_name, table_name)} (#{columns * ','})
```
ethervoid: Interesting way to make the join but for the sake of readability, I'd have used join

lib/carto/connector/fdw_support.rb
```diff
@@ -0,0 +1,114 @@
...
+      def quote_option_name(option)
+        option = option.to_s
+        if option && option.to_s.downcase != option.to_s
+          %{"#{option}"}
+        else
+          option
+        end
+      end
+
+      def options_clause(options)
```
ethervoid: I'm wondering if we're not going to have SQL injection problems. That kind of problems are mitigated using the Sequel/ActiveRecord prepared statement but here we're making some quote escaping tasks but could be insufficient

lib/carto/connector/providers.rb
ethervoid: I keep preferring a real Factory. Too much magic adding this as methods for the Connector class. I dunno...

services/importer/lib/importer/connector_runner.rb
```diff
@@ -0,0 +1,139 @@
...
+        @id = @job.id
+        @unique_suffix = @id.delete('-')
+        @json_params = connector_source
+        extract_params
+        @connector = Carto::Connector.new(@params, user: @user, logger: @log)
+        @results = []
+        @tracker = nil
+        @stats = {}
+        @warnings = {}
+        @connector.check_availability!
```
ethervoid: Could be nil?

services/importer/lib/importer/connector_runner.rb
```diff
@@ -0,0 +1,139 @@
...
+        # This method is needed to make the interface of ConnectorRunner compatible with Runner
+        []
+      end
+
+      # General availability of connectors for a user
+      def self.check_availability!(user)
+        Carto::Connector.check_availability! user
+      end
+
+      def etag
```
ethervoid: Maybe a nonimplementerror to avoid problems

config/app_config.yml.sample
```diff
@@ -497,6 +497,22 @@ defaults: &defaults
     min_rows: 2000000
     statement_timeout: 1800000
     tolerance_px: 1.0
+  connectors:
+    odbc:
+      available: false
+      max_rows: nil
+    mysql:
+      available: true
+      max_rows: 500000
```
jgoizueta: Since this is just a "sample" (used for CI tests, though), I'd leave it as is.

lib/carto/connector.rb
```diff
@@ -0,0 +1,293 @@
...
+
+      raise InvalidParametersError.new(message: "Provider not defined") if @name.blank?
+      @provider = Connector.provider_class(@name).try :new, @params
+      raise InvalidParametersError.new(message: "Invalid provider", provider: @name) if @provider.blank?
+    end
+
+    def copy_table(schema_name:, table_name:)
+      log "Connector Copy table  #{schema_name}.#{table_name}"
+      validate!
+      # TODO: logging with CartoDB::Logger
```
jgoizueta: And it's been there for ages!

lib/carto/connector/fdw_support.rb
jgoizueta: In this case the options are mapped by the Provider classes, so thet cannot be injected through the input parameters.

services/importer/lib/importer/connector_runner.rb
jgoizueta: Nope (`Connector.new` never returns nil)

services/importer/lib/importer/connector_runner.rb
jgoizueta: Currently the import process is calling this method even if the rsult not really used...

jgoizueta:
To expedite merging and deploying this I'm leaving some minor glitches to be handled later:

MAX_PG_IDENTIFIER_LEN should be 63 not 60
Connector#provider_name instead of Connector#name

lib/carto/connector/providers.rb
```diff
@@ -0,0 +1,80 @@
...
+        name: 'PostgreSQL',
+        class: PostgreSQLProvider,
+        public: true
+      },
+      'mysql' => {
+        name: 'MySQL',
+        class: MySqlProvider,
+        public: true
+      },
+      'sqlserver' => {
```
rafatower: why not symbol notation for these keys instead of `to_s`?

lib/carto/connector/providers.rb
rafatower: As far as I can see, this class is used basically for 2 different things:
- (a) to define a namespace used in other files:
```ruby
module Carto
  class Connector
    class Whatever
```

- and (b) as a bag of static data accessible to other classes.

Correct me if I'm wrong

Please bear in mind this pattern is not very "rubyst" (it just happens to use some ruby feature - the eigenclass) nor very "patternish" so anyone not familiar with this particular code will need to go through the code and spend some minutes repeating my investigation. I'd change it to something else.

I'd just define an "interface" (abstract class or whatever) that the providers can respond to. Another possibility but maybe a bit overgoing: https://en.wikipedia.org/wiki/Abstract_factory_pattern

rafatower:
:+1: ok to the merge but please kindly read the comments and consider creating tickets for things that may take more than 5 minutes if you deem them worth your while. Use your best judgement :smile:

lib/carto/connector/providers/generic_odbc.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      def initialize(params)
+        super
+        if @connection
+          @dsn        = @connection[:dsn]
+          @driver     = @connection[:driver]
+        end
+      end
+
+      def errors(only: nil)
```
jgoizueta: With the default value of nil for `only`, all errors, corresponding to all the required parameters are returned. With only this can be narrowed to get only the errors corresponding to some of the parameters (passed in an array, e.g. `errors(only: [:connection])`.

The reason for this is that we want to validate only the parameters that define the connection for some operations (list_tables) which don't need the parameters that define the table/query to be copied.

lib/carto/connector/providers.rb
jgoizueta: I guess just because we get the parameter that defines the provider as an String and i was too lazy to apply `to_sym` to it. 😛

jgoizueta: Regarding the design concerns about the `Connector` class, I was seeing it as a the provider of 'connector' services (i.e. access to remote databases) and decided to namespace all the implementation artifacts inside it (Provider classes, etc.). I kinda like that "connector parameters" are handled by class `Connector::Parameters` :grin: I wasn't aware of this obfuscating the design, but this probably deserves more thought, and it also is related to @ethervoid concerns about using a proper Factory pattern for the Providers, so I'll take note for tha backlog.

PR #9755, opened 2016-09-14
Fixed problem related with operator component + tests

### LINKED ISSUE #9599: Aggregation selection component doesn't let apply changes until changing the selected column
STR

- Aggregate by square for example
- Click on SUM(cartodb_id)
- Click on AVG(cartodb_id) - You cannot as cartodb_id is already selected.

xavijam:
Fixed problems identified with custom operator component.

operator

Fixes #9599.

CR: @nobuti

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/form-components/editors/operators/operators-view.js
+++ b/lib/assets/javascripts/cartodb3/components/form-components/editors/operators/operators-view.js
@@ -1,5 +1,6 @@
 var Backbone = require('backbone');
 var _ = require('underscore');
+var $ = require('jquery');
 var OperatorListView = require('./operators-list-view');
 var OperatorListCollection = require('./operators-list-collection');
 var template = require('./operators.tpl');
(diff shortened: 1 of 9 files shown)


lib/assets/javascripts/cartodb3/components/form-components/editors/operators/operators-view.js
```diff
@@ -22,50 +23,49 @@ Backbone.Form.editors.Operators = Backbone.Form.editors.Base.extend({
...
   _initViews: function () {
     var value = this.model.get(this.options.keyAttr);
-    this.$el.html(
-      template({
-        name: this._getOperatorsTextValue(),
-        disabled: this.options.disabled,
-        keyAttr: this.options.keyAttr
-      })
+
+    this.$el.append(
```
nobuti: Should not be this in the template somehow?

xavijam: No, because the template changes the "input"/button. So, if there is any change over the values, we only need to run `_setOperatorTemplate` method and 💫 .

lib/assets/javascripts/cartodb3/components/form-components/editors/operators/operators-view.js
```diff
@@ -22,50 +23,49 @@ Backbone.Form.editors.Operators = Backbone.Form.editors.Base.extend({
...
-      }
-    );
-
-    var optionsCollection = opts.schema && opts.schema.options || opts.options || [];
-    this.collection = new OperatorListCollection(optionsCollection);
-    this.setValue(this.model.get(this.options.keyAttr));
     Backbone.Form.editors.Base.prototype.initialize.call(this, opts);
+    this._setOptions(opts);
+
+    this.collection = new OperatorListCollection(this.options.options);
```
nobuti: `this.options.options` it's weird, could be `opts.options` better?

lib/assets/javascripts/cartodb3/components/form-components/editors/operators/operators-view.js
xavijam: Basically, if we include that in the template, we would need to overwritten template each time the vulues change, and that would destroy the list view.

lib/assets/javascripts/cartodb3/components/form-components/editors/operators/operators-view.js
xavijam: Uhm... not really, if not an alone field will not work (using it in a form yes). We use that for the select view and it is inherited from the base.js view.

xavijam: And yes, it is weird, I should fix it for any component that needs a list :(.

nobuti:
👏  🇪🇸

PR #9761, opened 2016-09-14
Placeholder text when style form can't be rendered

### LINKED ISSUE #9700: Style form text not rendering if geometry type is undefined
If we have a layer with no geometry, the [translation text doesn't exist](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/editor/style/style-form/style-properties-form/style-properties-form.tpl#L6).

As a solution we should render a "your layer has no geometries" message and not render the style form.

screen shot 2016-09-07 at 18 12 48

MariaCheca:
Fixes #9700 

Added template that will render instead of the style form in case there's not known geometry in the layer.

screen shot 2016-09-14 at 17 47 39

CR @xavijam


lib/assets/javascripts/cartodb3/editor/style/style-content-view.js
```diff
@@ -87,11 +88,15 @@ module.exports = CoreView.extend({
...
-      if (this._getCurrentSimpleGeometryType() === 'point') {
-        this._renderCarousel();
+      if (this._queryGeometryModel.hasValue()) {
+        if (this._queryGeometryModel.get('simple_geom') === 'point') {
+          this._renderCarousel();
+        }
+        this._renderForm();
+        this._renderOverlay();
+      } else {
+        this.$el.append(styleFormPlaceholderTemplate());
```
xavijam: What would you use, `append` or `html` and why?

xavijam: I'd add a test checking that state please :kiss:.

MariaCheca: Maybe it's better to use `.html()` instead of `.append()` in order to make sure the `$el` is always empty when adding the template.

MariaCheca:
@xavijam tests are passing now, could you take a look again please?

lib/assets/test/spec/cartodb3/editor/style/style-content-view.spec.js
```diff
@@ -144,6 +148,19 @@ describe('editor/style/style-content-view', function () {
...
+  describe('._renderForm', function () {
+    it('should not render when data has no geometry', function () {
+      StyleContentView.prototype.render.calls.reset();
+      this.queryGeometryModel.set({
+        status: 'fetched',
+        simple_geom: ''
+      });
+
+      this.view.render();
+      expect(_.size(this.view._subviews)).toBe(0);
```
xavijam: I'd check that the view you want to show is being displayed.

xavijam: Not only that there is no subviews added in the current view.

MariaCheca: I improved that test, now it's more as you suggested :)

PR #9763, opened 2016-09-14
Persist source node geometry when a query is applied

xavijam:
Basically, after updating everything with the PR about `query-geometry-model`, I missed the case when a query was applied and it could change the geometry type (for example, not requesting the `the_geom_webmercator` column).

So the idea is, wait until query-schema-model and query-geometry-model fetch has completed (successfully or badly in this last case), and then run `_saveSQL`, where analysis is saved ([here](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js#L233)). That means geometry_type will be included in the node definition and it will be stored in the analysis definition.

About nodes those are not source type, we check geometry when the Builder app has started, due to checks from CartoDB.js, so I'd not store those for the moment, it doesn't make sense.

Sounds cool @nobuti?

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
+++ b/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
@@ -188,10 +188,9 @@ module.exports = CoreView.extend({
       });
 
       this._queryGeometryModel.set({
-        status: 'unfetched',
-        query: currentQuery,
-        simple_geom: ''
-      }, { silent: true });
+        simple_geom: '',
+        query: currentQuery
+      });
 
       SQLNotifications.showNotification({
         status: 'loading',
@@ -199,11 +198,9 @@ module.exports = CoreView.extend({
         closable: false
       });
 
-      this._querySchemaModel.fetch({
-        success: this._saveSQL.bind(this)
-      });
-
-      this._queryGeometryModel.fetch();
+      var saveSQL = _.after(2, this._saveSQL.bind(this));
+      this._querySchemaModel.fetch({ success: saveSQL });
+      this._queryGeometryModel.fetch({ complete: saveSQL });
     }
   },
 
--- a/lib/assets/javascripts/cartodb3/data/analysis-definition-node-model.js
+++ b/lib/assets/javascripts/cartodb3/data/analysis-definition-node-model.js
@@ -21,8 +21,6 @@ module.exports = Backbone.Model.extend({
       simple_geom: simpleGeom,
       status: simpleGeom ? 'fetched' : 'unfetched'
     }, modelOpts);
-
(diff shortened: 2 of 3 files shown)


lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
```diff
@@ -188,22 +188,19 @@ module.exports = CoreView.extend({
...
         info: _t('notifications.sql.applying'),
         closable: false
       });
 
-      this._querySchemaModel.fetch({
-        success: this._saveSQL.bind(this)
-      });
-
-      this._queryGeometryModel.fetch();
+      var saveSQL = _.after(2, this._saveSQL.bind(this));
```
nobuti: wow! I didn't know that, in what manner is that better than promises? Just curious.

nobuti: What happens if any of those can't fetch?

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
```diff
@@ -188,22 +188,19 @@ module.exports = CoreView.extend({
...
+      this._querySchemaModel.fetch({ success: saveSQL });
+      this._queryGeometryModel.fetch({ complete: saveSQL });
```
nobuti: complete as option? I thought only success and error callbacks were supported, not sure if you meant something like `this._queryGeometryModel.fetch().complete(saveSQL)`

xavijam:
Complete callback exists as well, not only as a promise. About your question about if any of those don't fetch, what do you mean? Both have to fetch if a new query is applied. If querySchemaModel one fails, execution goes for other thread, so ok. If queryGeometry fetch fails, it doesn't matter, complete callback will be launched in any case.

nobuti:
Oops I can't see it in the backbone's source code.

My point is, fetch for those models test if fetch is able (canFetch) before to perform any requests. We are checking here if the number of requests is 2. But I think (maybe I'm wrong) if an error raises or the request can't make it, the success callback is not called. So in this case, that check will fail.

xavijam:
About complete callback, take into account request are done using $.ajax so...
If the query-schema-model request fails, we have this bind `this._querySchemaModel.on('change:query_errors', this._showErrors, this);`, so the execution will take that path, nothing new because it was like that.

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
xavijam: Both should be fetched, because if we check when they can't be fetched, the function is:

```js
  canFetch: function () {
    return this.get('query') && !this.isFetched();
  },
```

And we set the query and the status some lines above so...

nobuti:
I see so :+1:

PR #9768, opened 2016-09-15
Usage metrics api

javitonino:


javitonino:
CR @gfiorav

config/routes.rb
```diff
@@ -610,6 +610,17 @@
     namespace :superadmin do
       resources :user_migration_exports, only: [:show, :create]
       resources :user_migration_imports, only: [:show, :create]
+      resources :users, only: [] do
+        member do
+          get '/usage' => 'users#usage'
+        end
+      end
```
gfiorav: Why not

```ruby
namespace :users
  get '/usage' => 'users#usage'
end
```
or
```ruby
namespace :users do
  match 'usage', to: 'users#usage', via: :get
end
```

and save an indentation level?

config/routes.rb
```diff
@@ -610,6 +610,17 @@
...
+
+      resources :organizations, only: [] do
+        member do
+          get '/usage' => 'organizations#usage'
+        end
```
gfiorav: same here

lib/carto/metrics/mapviews_usage_metrics.rb
```diff
@@ -0,0 +1,28 @@
...
+    VALID_SERVICES = [
+      :mapviews
+    ].freeze
+
+    def initialize(user, orgname)
+      @user = user
+      @organization = Carto::Organization.where(name: orgname).first
+    end
+
+    def get(_service, _metric, date)
```
gfiorav: Why have an unused `_service` param?

lib/carto/metrics/twitter_imports_retriever.rb
```diff
@@ -0,0 +1,24 @@
...
+  class TwitterImportsRetriever
+    def metrics
+      [:retrieved_items]
+    end
+
+    def services
+      [:twitter_imports]
+    end
+
+    def get_range(user, _org, _service, _metric, date_from, date_to)
```
gfiorav: Same comment for all `_xxx` params

lib/carto/metrics/twitter_imports_retriever.rb
```diff
@@ -0,0 +1,24 @@
...
+      tweets_query = if user
+                       user.search_tweets
+                     else
+                       _org.users.joins(:search_tweets).reorder(nil)
```
gfiorav: But then you use it! 🤔

lib/carto/metrics/usage_metrics_retriever.rb
```diff
@@ -0,0 +1,25 @@
+module Carto::Metrics
+  class UsageMetricsRetriever
+    def initialize(cls)
+      @cls = cls
```
gfiorav: Here come the men in meta! 😛

app/controllers/carto/superadmin/organizations_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      def usage
+        usage = get_usage(nil, @organization)
+
+        respond_with(usage)
+      end
+
+      private
+
+      def last_billing_cycle
```
javitonino: From MetricsControllerHelper: https://github.com/CartoDB/cartodb/pull/9768/files#diff-a6865e4cd867617e3b69022977da3738R28

config/routes.rb
javitonino: I thought I'd prepare this for eventual migration of more controllers to the Carto namespace, and this way we have the resource ready to go. Can change it no proble if you'd like it better.

lib/carto/metrics/mapviews_usage_metrics.rb
javitonino: For compatibility with the other usage metrics classes. In this one is not needed (as there is only a possible service/metric).

lib/carto/metrics/twitter_imports_retriever.rb
javitonino: AAAAARGGG. Don't look at it! Shame!

Changing!

lib/carto/metrics/twitter_imports_retriever.rb
javitonino: Same reason as above, for compatibility with UsageMetricsRetriever.

lib/carto/metrics/mapviews_usage_metrics.rb
```diff
@@ -0,0 +1,27 @@
...
+      if @organization
+        @organization.users.map { |user|
+          CartoDB::Stats::APICalls.new.get_api_calls_from_redis_source(
+            user.username, 'mapviews', from: date, to: date).values[0]
```
gfiorav: Agree!

app/controllers/carto/superadmin/organizations_controller.rb
gfiorav: Uh! There's something weird with having a method in a controller to be used from a helper. Do you think you could move this method to the helper?

config/routes.rb
gfiorav: Ok! What I don't get is why use member?

lib/carto/metrics/mapviews_usage_metrics.rb
javitonino: This is outdated ¬¬. I already fixed it: https://github.com/CartoDB/cartodb/pull/9768/commits/8eb49d23df437e2bcf9ae46cf5c811a3481adec5

lib/carto/metrics/twitter_imports_retriever.rb
gfiorav: Kinda weird!

If it's to much work, leave it. If not, just namespace the hell out of this and break compatibility. Having new methods with unused args just seems wrong! For me the limit is something like 
```ruby
whatever_hash.each do |_key, value|
  (...)
end
```

because you can't avoid having it, but here it seems avoidable

app/controllers/carto/superadmin/organizations_controller.rb
javitonino: Nope, can't do. It's different for user and organizations (i.e: the organization calls this on the owner).

The other option I considered would be to move the part that used it (`params[:from]` parsing) to the controller (outside the helper), but it seemed strange to do this here and the rest of the parameter parsing in the helper.

Maybe another option is to add it as a parameter to `get_usage`. Do you think that would be better? It makes sense now that it only has two parameters (before it already had like 8 and was crazy to add another one).

config/routes.rb
javitonino: I have to add member so the route is under `users/:id`
What about?
```
resources :users, only: [] do
        get '/usage' => 'users#usage', on: :member
      end
```
That saves an indentation levels. Best of both worlds?

lib/carto/metrics/twitter_imports_retriever.rb
javitonino: It's actually compatibility with my own class that I created here, so I can change it. But `TwitterImportsRetriever` and `UsageMetricsRetriever` should have the same interface. The second one needs `service` and `metric` while the first one doesn't.

app/controllers/carto/superadmin/organizations_controller.rb
gfiorav: Yup, I'd totally make it a param for `get_usage`. Then `last_billing_cycle` is directly used by this controller and it makes more sense. 

Thanks <3

config/routes.rb
gfiorav: Ok sorry!

Scope it?

```ruby
scope 'user/:user_id' do
  match 'usage', to: 'users#usage', via: :get # (or however you want to match)
end
```

lib/carto/metrics/twitter_imports_retriever.rb
gfiorav: But you should know beforehand what type you're going to use right? If not it's too meta IMHO. I mean:

```
Fruit#peel(a,b,c,d) -> Apple#peel(a,b)
                    -> Orange#peel(b,c,d)
```

Then 
```ruby
class Fruit
  def peel(a, b, c, d)
    (...)
  end
end

class Apple < Fruit
  def peel(a, b)
    super(a, b, ?, ?)
  end
end

class Orange < Fruit
  def peel(b, c, d)
    super(?, b, c, d)
  end
end
```

where `?` is a value you give to that required parameter that you don't need (nil, optional param, a given value, etc). You think you could avoid these unused params by doing this?

config/routes.rb
javitonino: I prefer the resources approach as it lets Rails auto-generate the URL's (btw, it's `users/:id`, not `user/:user_id`). Also, I don't have to rewrite all the tests with the new url generator method.

lib/carto/metrics/twitter_imports_retriever.rb
javitonino: Exactly, it's a bit of meta. It loops over an array of `*Retriever` classes, calling all of them with the same signature.

PR #9770, opened 2016-09-15
default form should be point

gfiorav:
Please CR @javitonino

### DIFF
--- a/lib/carto/form.rb
+++ b/lib/carto/form.rb
@@ -9,13 +9,6 @@ module Carto
 
     def initialize(geometry_type)
       @geometry_type = geometry_type
-
-      unless style_class
-        message = "Carto::Forms: No form for geometry type"
-        CartoDB::Logger.error(message: message, geometry_type: geometry_type)
-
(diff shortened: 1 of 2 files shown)


spec/lib/carto/form_spec.rb
```diff
@@ -4,8 +4,170 @@
...
+          "overlap" => true,
+          "placement" => "point"
+        }
+      }
+    end
+
+    it 'return point form if inexisting type is specified' do
+      form = Carto::Form.new('geometry')
+
+      form.to_hash.should eq Carto::Form.new('point').to_hash
```
javitonino: Why not `point_form`?

lib/carto/form.rb
```diff
@@ -43,7 +36,11 @@ def style_definition
     end
 
     def style_class
-      @style_class ||= Carto::Styles::Style.style_for_geometry_type(@geometry_type)
+      return @style_class if @style_class
```
javitonino: Undoing the last change (`||=`) I see. :laughing:

PR #9782, opened 2016-09-16
add check for query before rewrite

gfiorav:
CR @javitonino


javitonino:
:+1: Ship it like it's :fire:

PR #9795, opened 2016-09-16
Adding color-picker for Codemirror panel as an addon

xavijam:
screen shot 2016-12-09 at 17 23 02

### Pending things
- [x] Tests.
- [x] Dark theme (cc @piensaenpixel).
- [x] Colorize color variables in the editor content.
- [x] Close color-picker with ESC.
- [x] Open color-picker above if it is close to the end of the document.


saleiva:
@piensaenpixel I think this was supposed to be dark themed, right? You guys are the best!

nobuti:
@xavijam I made some tests, they seem to pass in the browser (isolated) but not in phantomjs, I would need a hand 🙏

xavijam:
Ok, let's take a look on Friday 💋

lib/assets/javascripts/cartodb3/components/code-mirror/colorpicker.code-mirror.js
```diff
@@ -53,6 +53,14 @@ module.exports = CoreView.extend({
     this._editor.off(null, null, this);
   },
 
+  _enableUpdateBind: function () {
+    this._editor.on('update', _.debounce(this._updateColors.bind(this), 100));
+  },
+
+  _disableUpdateBind: function () {
+    this._editor.off('update', _.debounce(this._updateColors.bind(this), 100));
```
nobuti: Be careful here, debounce returns a new function, so I think this is going to take no effect. Storing a reference to the unique debounced function should be better, something like this:
```
...
this._updateColors = _.debounce(this._updateColors.bind(this), 100);
...
_enableUpdateBind: function () {
  this._editor.on('update', this._updateColors);
},
_disableUpdateBind: function () {
   this._editor.off('update', this._updateColors);
}
```

xavijam: Yeah, just realized that, thanks @nobuti 💋 , #putoamo.

nobuti:
colorpicker

Ready for a round!

xavijam:
What do you think @CartoDB/design ?

arianaescobar:
love it <3

xavijam:
I can't approve the PR because I started it :joy:. But, I approve it!
Let's go to acceptance.

xavijam:
Any @CartoDB/design er available for testing this small but cool feature?

arianaescobar:
@xavijam I can

xavijam:
Thanks @arianaescobar, being deployed at ded04 servers 💋 .

arianaescobar:
Hi guys, a few things:

We messed up by not giving you a mockup for this when you asked for it and although @nobuti did a splendid work at applying our styles to the component, we'd like to tweak it just a bit if it's possible, like this:   

artboard

Notice that in production there's a line below the color viewer that shouldn't be there and a slightly noticeable border surrounding the color chip that also shouldn't be there.

artboard2

Then I wanted to ask about why aren't we adding the opacity slider to this component, if it's because of a technical limitation and we **can't** add it, we'd have to go back to the designs and try to make the color slider look nicer without all that margin below it.

Continuing with the opacity parameter, if I have set my opacity to let's say 0,66 in cartoCSS, it's a bit confusing that when I open the color properties, the alpha input always shows a 1. Couldn't we link both things? If not, we should also take a look at it in design to remove that input from there. 

Now, in terms of behavior, there are a few issues of this component not-closing. Specifically, if I have it open and I switch back to VALUE on the toggle at the bottom. Examples: 

screen shot 2017-03-15 at 12 12 29

screen shot 2017-03-15 at 13 35 43

screen shot 2017-03-15 at 13 37 51

xavijam:
Ok, let's face those problems on a Leapfrog day then :D. Removing it from Kanban due to we have several other things to do before this 💋 .

nobuti:
@arianaescobar I'm afraid the alpha's behaviour is different in this, at code panel level. For instance, in the style panel, if you are editing the stroke, the color slider acts on `marker-line-color` while the alpha slider acts on `marker-line-opacity`. In the code view, we have knowledge only of the color the user has clicked on. We don't know if the color belongs to `line-color`, `polygon-fill` or other color properties. So, in this case, the alpha should act on the color itself, resulting in the end in a `rgba` color, instead a `hex` one. Not sure about the implications of this.

In the meantime, my inner designer has updated the popup, tweaking the component to get rid of those weird spacing and subcomponents we are not using:

screen shot 2017-03-17 at 17 06 51

Also, all the issues you found should be fixed.

noguerol:
Love it

urbanphes:
Hey sorry for the delay,

I think the height of the color bar is too much. I would leave it with normal height. We have the same case on the white side and the size is the same. Here's the sample:

slice

The rest is perfect. I like how the inputs for HEX and RGB take advantage of the full width.

arianaescobar:
+1 to the `Love it` and also to juanma's comment 🙃

Also, just a small thing, but please remember to fix that line below the color viewer (in green on the last screenshot)

nobuti:
Ready for another round.

screen shot 2017-04-04 at 18 01 13

arianaescobar:
Looks great @nobuti

ivanmalagon:
Just found a weird behaviour that happens sometimes. Not only in codemirror but in regular ones too.

It applies to the color spectrum panel the selected color.

colorpicker bug

Getting it back to development to take a look on this.

ivanmalagon:
Everything OK! PR accepted. :1st_place_medal: 

I'm going to fix conflicts (package.json and shrinkwrap) and it'll be ready for production.

ivanmalagon:
Deployed. Already live.

PR #9796, opened 2016-09-16
Added rake tasks to modify the services providers for users and organizations

ethervoid:
Please @iriberri check this

### DIFF
--- a/lib/tasks/services.rake
+++ b/lib/tasks/services.rake
@@ -1,0 +1,60 @@
+namespace :cartodb do
+  namespace :services do
+
+    # usage example:
+    #   bundle exec rake cartodb:service:set_user_provider['username','geocoder','mapzen']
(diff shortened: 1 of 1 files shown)


lib/tasks/services.rake
```diff
@@ -0,0 +1,60 @@
...
+      PROVIDERS=['here', 'google', 'mapzen']
+
+      username = args[:username]
+      service = args[:service]
+      provider = args[:provider]
+
+      raise 'Please specify the username of the user to be modified' if args[:username].blank?
+      raise 'Please specify the service to set the provider' if args[:service].blank?
+      raise "Unkown service. Please use one of the accepted services: #{SERVICES.join(',')}" if not SERVICES.include? args[:service]
+      raise 'Please specify the provider to be setted' if args[:provider].blank?
```
iriberri: To be set :)

PR #9808, opened 2016-09-19
Goodbye cdb.god

xavijam:
We don't need it, we don't use it, long live the `cdb.god`.

cc @CartoDB/frontend


matallo:
good by old friend, you served well 👋

Issue #9812, opened 2016-09-19
Importer results are not being propagated to DataImport in case of uncontrolled exception


iriberri:
https://github.com/CartoDB/cartodb/blob/e63431259cad23152007185a3eae4f87c5bfff81/app/models/data_import.rb#L713-L725

If an exception is raised here, `store_results` never occurs and the DataImport finishes with results = `[]`.

This is affecting the MailNotifier: https://github.com/CartoDB/cartodb/issues/9805

As the `importer` (and therefore its results) do only make sense inside the `execute_importer` method, we should rescue exceptions there to store the importer results and then raise the exception to continue with the normal behaviour.

iriberri:
I'm not going through the `store_results` scenario because it might store false positives if the state of the importer wasn't set to false. I'll call `set_error` which will setup the default 99999 (missing until now) and build a results array out of this. Unfortunately, this won't provide the table name though, but will return a:

````
[<CartoDB::Importer2::Result @name=, @schema=, @extension=, @tables=, @success=false, @error_code=99999, @support_tables=, @original_name=>]
````

MariaCheca:
Related to https://github.com/CartoDB/cartodb/issues/9805

Issue #9805, opened 2016-09-19
Incorrect email template when import fails without error_code


MariaCheca:
When an import fails without error_code, the template rendered is the one for successful or partially successful (in cases when the user as uploaded more than one files), when it should be the warning one.

![screen shot 2016-09-19 at 11 44 24](https://cloud.githubusercontent.com/assets/2141690/18628038/8bbc14a2-7e5e-11e6-9d3a-28a79d04c5b6.png)

Thank you for reporting it @iriberri :*

iriberri:
> (in cases when the user as uploaded more than one files)

I didn't tell, but I'm uploading a Shapefile with a single layer :)

MariaCheca:
I supposed, that's why it shouldn't show that "x out of x were blahblah" :/

iriberri:
Deploying the empty results fix!

iriberri:
Deployed. Unless there is any other scenario which I haven't noticed, you should not receive empty results anymore. Anyway, I think it'd be good to check it just in case.

PR #9814, opened 2016-09-20
Rescue untreated exceptions to set error 99999

iriberri:
Closes https://github.com/CartoDB/cartodb/issues/9812

=====

If an exception is raised here, `store_results` never occurs and the DataImport finishes with results = `[]`.

This is affecting the MailNotifier: https://github.com/CartoDB/cartodb/issues/9805

Calling `set_error` will setup the default 99999 (missing until now) and build a results array out of this. Unfortunately, this won't provide the table name though or any other metadata, but will return a non-empty results with a non-null error code:

````
[<CartoDB::Importer2::Result @name=, @schema=, @extension=, @tables=, @success=false, @error_code=99999, @support_tables=, @original_name=>]
````

This should fix also another problem we noticed which is that some imports end up with null error_code because it's never treated. Wondering if it'd be worthy to send the whole `importer.results` (whenever importer is defined) and manipulate those by setting the error code if null, and the success to false (it's generally true in Data Import because Importer raised the exception without treating the results, or because an error happened above the Importer layer).

### DIFF
--- a/app/models/data_import.rb
+++ b/app/models/data_import.rb
@@ -720,8 +720,14 @@ class DataImport < Sequel::Model
       importer.run(tracker)
       log.append 'After importer run'
     end
+
     store_results(importer, runner, datasource_provider, manual_fields)
     importer.nil? ? false : importer.success?
+  rescue => e
+    # Note: If this exception is not treated, results will not be defined
+    # and the import will finish with a null error_code
+    set_error(manual_fields.fetch(:error_code, 99999))
+    raise e
   end
 
   # Note: Assumes that if importer is nil an error happened


iriberri:
@ethervoid what do you think?

This is currently a problem for the Mailer because for some untreated exceptions it receives an empty result (I was for example able to reproduce this with some weird Out Of Memory error from Postgres).

Do you think it's worth it ti struggle a bit more to get more details from the import process itself? One idea I had was to add another parameter to `set_error` with the importer.results hash and try to recover info from there, but perhaps it could get messy with multifile uploads.

iriberri:
Actually I guess we couldn't know from which of the importer results is the failure coming. And even if it's an exception in register or persist_metadata, the Ghost Tables process could make the table accessible later on -- as it just happened in my local testing.

ethervoid:
It'd be great to know where we're missing to treat the exception and give a proper error message or at least try to recover from it but until them this solution looks like a good approach to at least provide some information to the user.

iriberri:
Thanks! I'll go on. :)

Just noticed that for example, in my usecase, the code was raising an exception in the Table model after the Data Import saves the table: https://github.com/CartoDB/cartodb/blob/671d27fce116612fdfdaa1478aeaa7a19000a50b/app/models/data_import.rb#L530

So the most possible scenario here is that Importer ended successfully, but something went wrong on the callback to create the Table in the model: https://github.com/CartoDB/cartodb/blob/be5e75b133ebbba4d51c91c8ac3eac915feae7f6/app/models/table.rb#L377-L416

PR #9819, opened 2016-09-20
Added "Your workflow" title with count of analyses applied

### LINKED ISSUE #9508: Analysis first step title is not completely right
The original designs says "Your Workflow (X)" while X is the number of steps already applied.

screen shot 2016-08-22 at 17 43 26

MariaCheca:
Fixes #9508

- Changed step title to "Workflow" to "Your workflow"
- Added analyses count.

untitled

CR @xavijam

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-view.js
+++ b/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-view.js
@@ -34,7 +34,8 @@ module.exports = CoreView.extend({
       deleteLabel: !this._nodeDefModel()
         ? _t('editor.layers.layer.cancel-delete-analysis')
         : _t('editor.layers.layer.delete'),
-      selectedNodeId: this.viewModel.get('selectedNodeId')
+      selectedNodeId: this.viewModel.get('selectedNodeId'),
+      layerAnalysisCount: this.analysisFormsCollection.length
     });
   },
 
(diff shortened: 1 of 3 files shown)


lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-view.js
```diff
@@ -34,7 +34,8 @@ module.exports = CoreView.extend({
       deleteLabel: !this._nodeDefModel()
         ? _t('editor.layers.layer.cancel-delete-analysis')
         : _t('editor.layers.layer.delete'),
-      selectedNodeId: this.viewModel.get('selectedNodeId')
+      selectedNodeId: this.viewModel.get('selectedNodeId'),
+      layerAnalysisCount: this.analysisFormsCollection.length
```
xavijam: I would use the proper method within a Backbone.Collection.

xavijam: One question: Having an analysis already applied. If another analysis is added, but not applied, workflow should say 2 or 1.

MariaCheca: I would say it should display "2", because wether the second analysis is applied or not, it is represented in the workflow the same way as the first one (already applied). Furthermore, you can switch between analyses no matter one is not applied yet.

xavijam: 👍

MariaCheca:
@xavijam after talking with @piensaenpixel, the workflow count should show only applied analyses, so I replaced the `.length` solution I used with a method owned by the layer definition model that returns the number of *currently applied* analyses.

PR #9837, opened 2016-09-21
replace instagram description in twitter connector fallback

matallo:
- the string seems to be wrong (see attached)

screen shot 2016-09-21 at 14 20 19

fixed with the proper localized string for twitter

CR @xavijam


xavijam:
Hey, but the description is wrong too, could you reviewed too please @matallo :kiss:

xavijam:
Ah the image was the error one, ooook! 💋

PR #9844, opened 2016-09-21
Fixed negative workflow count

### LINKED ISSUE #9841: Avoid negative workflow count
If layer nodes are 0 (that happens when a layer source is another layer's analysis) avoid showing this -1.

screen shot 2016-09-21 at 16 16 40

MariaCheca:
Fixes #9841 

In case the value returned is 0, return it without the -1 operation to exclude from the result the layer source.

CR @xavijam


MariaCheca:
retest this please

MariaCheca:
please retest this

PR #9847, opened 2016-09-22
Migration for Connector configuration

jgoizueta:
This is the migration for the upcoming Connector Configuration changes

### DIFF
--- a/db/migrate/20160916131013_create_connector_providers_configurations.rb
+++ b/db/migrate/20160916131013_create_connector_providers_configurations.rb
@@ -20,7 +20,7 @@ Sequel.migration do
 
       foreign_key :user_id, :users, type: :uuid, null: true, on_delete: :cascade
       foreign_key :organization_id, :organizations, type: :uuid, null: true, on_delete: :cascade
-      foreign_key :connector_provider_id, :connector_providers, type: :uuid, null: false, on_delete: :cascade
+      foreign_key :connector_provider_id, :connector_providers, type: :uuid, null: false, on_delete: :restrict
     end
 
     alter_table :connector_configurations do
--- a/db/migrate/20160916131013_create_connector_providers_configurations.rb
+++ b/db/migrate/20160916131013_create_connector_providers_configurations.rb
@@ -1,0 +1,37 @@
+Sequel.migration do
+  up do
+    create_table :connector_providers do
+      Uuid     :id, primary_key: true, default: 'uuid_generate_v4()'.lit
+      DateTime :created_at, default: Sequel::CURRENT_TIMESTAMP
+      DateTime :updated_at, default: Sequel::CURRENT_TIMESTAMP
+      String   :name, null: false
+    end
+
+    alter_table :connector_providers do
+      add_index :name
+    end
+
+    create_table :connector_configurations do
+      Uuid     :id, primary_key: true, default: 'uuid_generate_v4()'.lit
+      DateTime :created_at, default: Sequel::CURRENT_TIMESTAMP
+      DateTime :updated_at, default: Sequel::CURRENT_TIMESTAMP
+      Boolean  :enabled, null: false
+      Integer  :max_rows, null: true
+
+      foreign_key :user_id, :users, type: :uuid, null: true, on_delete: :cascade
+      foreign_key :organization_id, :organizations, type: :uuid, null: true, on_delete: :cascade
+      foreign_key :connector_provider_id, :connector_providers, type: :uuid, null: false, on_delete: :cascade
+    end
+
+    alter_table :connector_configurations do
+      add_index :user_id
+      add_index :organization_id
+      add_index :connector_provider_id
+    end
+  end
+
+  down do
+    drop_table :connector_configurations
+    drop_table :connector_providers
+  end
+end


jgoizueta:
@ethervoid @rafatower please review

/cc @javitonino @gfiorav can you keep an eye on this? the code that uses these tables is in #9849

db/migrate/20160916131013_create_connector_providers_configurations.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    create_table :connector_configurations do
+      Uuid     :id, primary_key: true, default: 'uuid_generate_v4()'.lit
+      DateTime :created_at, default: Sequel::CURRENT_TIMESTAMP
+      DateTime :updated_at, default: Sequel::CURRENT_TIMESTAMP
+      Boolean  :enabled, null: false
+      Integer  :max_rows, null: true
+
+      foreign_key :user_id, :users, type: :uuid, null: true, on_delete: :cascade
```
gfiorav: Is a `users` FK safe @luisbosque ?

cc/ @juanignaciosl

db/migrate/20160916131013_create_connector_providers_configurations.rb
```diff
@@ -0,0 +1,37 @@
...
+      foreign_key :organization_id, :organizations, type: :uuid, null: true, on_delete: :cascade
+      foreign_key :connector_provider_id, :connector_providers, type: :uuid, null: false, on_delete: :cascade
+    end
+
+    alter_table :connector_configurations do
+      add_index :user_id
```
gfiorav: Also, add indexes concurrently please!

Syntax would be 
```ruby
Rails::Sequel.connection.run %{
  CREATE INDEX CONCURRENTLY "table_foreign_id_index"
    ON "table" ("foreign_id"); }
```

in your case:

```ruby
Rails::Sequel.connection.run %{
  CREATE INDEX CONCURRENTLY "connector_configurations_user_id_index"
    ON "connector_configurations" ("user_id"); }
```

NOTE: Watch out with the famous 63 char limit on `connector_configurations_user_id_index` and others!

db/migrate/20160916131013_create_connector_providers_configurations.rb
```diff
@@ -0,0 +1,37 @@
...
+  up do
+    create_table :connector_providers do
+      Uuid     :id, primary_key: true, default: 'uuid_generate_v4()'.lit
+      DateTime :created_at, default: Sequel::CURRENT_TIMESTAMP
+      DateTime :updated_at, default: Sequel::CURRENT_TIMESTAMP
+      String   :name, null: false
+    end
+
+    alter_table :connector_providers do
+      add_index :name
```
gfiorav: same concurrent indexes please!

db/migrate/20160916131013_create_connector_providers_configurations.rb
jgoizueta: I thought that wouldn't be necessary for a new, empty table.

db/migrate/20160916131013_create_connector_providers_configurations.rb
jgoizueta: OK, I'll wait for confirmation. No problem removing it.

rafatower: 200K rows for the FK shall be no problem at all, confirmed with Luisico

db/migrate/20160916131013_create_connector_providers_configurations.rb
rafatower: no need for concurrent creation on a new empty table, basically it will cost 0

rafatower:
IMO the migration code is good enough taking into account it creates a brand new table.

db/migrate/20160916131013_create_connector_providers_configurations.rb
ethervoid: I agree, this is for the user's table, should be no problem to execute this migration

javitonino: We did some before with little issue. But please monitor Rollbar for user creation error. Sometimes they fail (can be fixed from Central).

db/migrate/20160916131013_create_connector_providers_configurations.rb
```diff
@@ -0,0 +1,37 @@
...
+    create_table :connector_configurations do
+      Uuid     :id, primary_key: true, default: 'uuid_generate_v4()'.lit
+      DateTime :created_at, default: Sequel::CURRENT_TIMESTAMP
+      DateTime :updated_at, default: Sequel::CURRENT_TIMESTAMP
+      Boolean  :enabled, null: false
+      Integer  :max_rows, null: true
+
+      foreign_key :user_id, :users, type: :uuid, null: true, on_delete: :cascade
+      foreign_key :organization_id, :organizations, type: :uuid, null: true, on_delete: :cascade
+      foreign_key :connector_provider_id, :connector_providers, type: :uuid, null: false, on_delete: :cascade
```
javitonino: I feel cascade here might be a bit aggressive. I would prefer the deletion of a provider to fail if it's in use, rather than silently delete all configurations. We can still do it manually if we delete/deprecate a provider. Maybe being overcautious here, so up to you.

javitonino:
Left a couple of comments, but nothing serious. :+1:

db/migrate/20160916131013_create_connector_providers_configurations.rb
jgoizueta: Removing an existing provider shouldn't be a "normal" think. I agree it'd be more cautious to fail, so I'll change it to `:restrict`

PR #9849, opened 2016-09-22
Connector configuration

jgoizueta:
# Connectors configuration

This uses two tables to keep Connector configuration. The migration to create the tables is in #9847

Configuration for a connector determines if it is available (`enabled` configuration parameter) and will also hold its limits (currently only maximum number of rows, `max_rows`).

There are three levels of configuration:

* A default configuration for each connector (technically for each connector provider, such as 'mysql')
* A configuration for an organization and provider; overrides the default configuration of the provider, and
  applies to all the org users which don't have specific configuration.
* A user-specific configuration for a user and provider. Overrides any other configuration.

The configurations are stored in a metadata table `connector_configurations`.
The provider default configuration is stored in `app_config`. If we need more dynamic provider-defaults, we could use the connector_configurations table to override this provider defaults too,
but to avoid complexity/possible confusions we'll use app_config for the moment.

## Rake tasks

There are some rake tasks to show and change the organization and user configuration. (to specify providers, users or organization either the name or id can be used).

show configuration of a user:

```
rake cartodb:connectors:show_user_config[provider,user]
```
show configuration for an organization:

```
rake cartodb:connectors:show_org_config[provider,organization]
```

show default configuration for a connector:

```
rake cartodb:connectors:show_org_config[provider]
```

change configuration for a user: (enabled values: true or false; max_rows: leave empty (or 0) for unlimited)

```
 rake cartodb:connectors:set_user_config[provider,user,enabled,max_rows]
````

remove user configuration (reset to organization or provider defaults):

```
rake cartodb:connectors:set_user_config[provider,user,default]
```

change configuration for an organization

```
rake cartodb:connectors:set_org_config[provider,organization,enabled,max_rows]
```

remove org configuration (reset to provider defaults):

```
rake cartodb:connectors:set_org_config[provider,organization,default]
````

### Example

If connector `sqlserver` is disabled by default, and we need to make it
available for members of organization 'team', except for 'unreliable_staff', and available too for user 'big_customer':

```
rake cartodb:connector:set_org_config[mysql,team,true,1000000]
rake cartodb:connector:set_user_config[mysql,unreliable_staff,false,1000000]
rake cartodb:connector:set_user_config[mysql,big_customer,true,10000000]
````

### DIFF
--- a/app/models/carto/connector_configuration.rb
+++ b/app/models/carto/connector_configuration.rb
@@ -43,11 +43,21 @@ class Carto::ConnectorConfiguration < ActiveRecord::Base
     config
   end
 
+  def self.for_organization(organization, provider)
+    if provider
+      config = where(organization_id: organization.id, connector_provider_id: provider.id).first
+      if config.blank?
+        config = default(provider)
+      end
+      config
+    end
+  end
+
   def self.for_user(user, provider)
     if provider
       config = where(user_id: user.id, connector_provider_id: provider.id).first
       if config.blank? && user.organization_id.present?
-        config = where(organization_id: user.organization_id, connector_provider_id: provider.id).first
+        config = for_organization(user.organization, provider)
       end
       if config.blank?
         config = default(provider)
--- a/app/models/carto/connector_configuration.rb
+++ b/app/models/carto/connector_configuration.rb
@@ -1,0 +1,58 @@
+# encoding: UTF-8
+
+require 'carto/connector'
+
+class Carto::ConnectorConfiguration < ActiveRecord::Base
+  belongs_to :connector_provider, class_name: Carto::ConnectorProvider
+  belongs_to :user, class_name: Carto::User
+  belongs_to :organization, class_name: Carto::Organization
+
+  # A ConnectorConfiguration can belong to either user or an organization, but not both;
+  # it may also be a default configuration (no user or organization).
+  # Only one configuration may exist for any valid combination of user, organization and provider.
+  # So we have 3 kinds of configuration records:
+  # * default configuration for provider: user.nil? && organization.nil? FIXME: remove this? (only app_config defaults)
+  # * organization configuration: user.nil? && organization.present?
+  # * user configuration: organization.nil? && user.present?
+  validates :connector_provider_id, presence: true
+  validates :user_id, uniqueness: { scope: [:connector_provider_id, :organization_id] }, if: 'organization_id.nil?'
+  validates :organization_id, uniqueness: { scope: [:connector_provider_id, :user_id] }, if: 'user_id.nil?'
+  validate :not_user_and_organization_simultaneously
+
+  def not_user_and_organization_simultaneously
+    if user_id.present? && organization_id.present?
+      errors.add(:user_id, "can't assign to an organization simultaneously with a user")
+    end
+  end
+
+  # columns:
+  # enabled boolean
+  # max_rows integer
+
+  def self.default(provider)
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

**Comment by jgoizueta** [comment]
@ethervoid @rafatower please review

File: app/models/carto/connector_configuration.rb [hunk 1]
```diff
@@ -0,0 +1,58 @@
...
+    end
+  end
+
+  # columns:
+  # enabled boolean
+  # max_rows integer
+
+  def self.default(provider)
+    # Look for default provider configration
+    config = where(user_id: nil, organization_id: nil, connector_provider_id: provider.id).first
```
jgoizueta: I was in doubt  about removing this line (and @ethervoid voted to remove it):
If we keep it we can change the defaults for a provider by writing to the `connector_configurations` table instead of having to change app_cofig.yml.

If we remove it we should disallow `connector_configurations` with both  `user_id` and `organization_id` nil (with a validation).

What do you think?

ethervoid: For me is confusing to have entries in the database with no user nor organization and use it for default values. For me is not an inconvenience to change the app_config to set default values.

app/models/carto/connector_configuration.rb
```diff
@@ -0,0 +1,58 @@
+# encoding: UTF-8
+
+require 'carto/connector'
+
+class Carto::ConnectorConfiguration < ActiveRecord::Base
```
ethervoid: We have to duplicate the models: one in the models/carto just for reading-only purposes and one in the app/models using sequel for write purposes

app/models/carto/connector_configuration.rb
```diff
@@ -0,0 +1,58 @@
...
+  validates :organization_id, uniqueness: { scope: [:connector_provider_id, :user_id] }, if: 'user_id.nil?'
+  validate :not_user_and_organization_simultaneously
+
+  def not_user_and_organization_simultaneously
+    if user_id.present? && organization_id.present?
+      errors.add(:user_id, "can't assign to an organization simultaneously with a user")
+    end
+  end
+
+  # columns:
```
ethervoid: I don't think we need this comment, basically because is going to be outdated with the next property we add here

app/models/carto/connector_configuration.rb
```diff
@@ -0,0 +1,58 @@
...
+      config = new(
+        connector_provider: provider,
+        enabled:  Cartodb.get_config(:connectors, provider.name, 'enabled') || false,
+        max_rows: Cartodb.get_config(:connectors, provider.name, 'max_rows')
+      )
+    end
+    config
+  end
+
+  def self.for_user(user, provider)
```
ethervoid: For previous experience is better to have org and user separated. I'll make other method only for organization

app/models/carto/helpers/has_connector_configuration.rb
```diff
@@ -0,0 +1,8 @@
+module Carto
```
ethervoid: Missing #encoding: UTF-8

app/models/carto/connector_provider.rb
```diff
@@ -0,0 +1,10 @@
+# encoding: UTF-8
```
ethervoid: The same than before, if we're going to persist using the models you have to add it to app/models as well

app/models/carto/connector_configuration.rb
ethervoid: nevermind

PR #9853, opened 2016-09-22
Hide the Boundaries option available through the UI in the Enrichment analysis

### LINKED ISSUE #9604: Boundary enrichment option available through the UI
# STR

`.carto` file to reproduce the issue: 
[Boundaries (on 2016-08-29 at 14.51.35).carto.zip](https://github.com/CartoDB/cartodb/files/442654/Boundaries.on.2016-08-29.at.14.51.35.carto.zip)

1 - Add a layer
2 - Add an analysis and select `Enrich from Data Observatory`
3 - Analysis parameters allow to select `Boundaries` as the measure, although that feature is still not working from UI
screen shot 2016-08-29 at 16 52 50
4 - Analysis fails, as expected. 

# What's expected

Boundary enrichment should not be available through the UI. 

cc/ @saleiva

javierarce:
Closes #9604 

CR: @xavijam 
CC: @javisantana @ernesmb


xavijam:
No much to say :es:

PR #9862, opened 2016-09-23
Refactor and add warnings for connectors max number of rows

ethervoid:


ethervoid:
Tested and working.

Screenshot of the warning message to be reviewed by @brittanymicek or @csobier :)

captura de pantalla 2016-09-29 13 53 34

csobier:
Suggested edit:
"Maximum limit reached
You have reached the maximum limit of 90000 rows. Some of your data may not be imported from the database connector."

brittanymicek:
+1 to @csobier with just one additional edit:
Maximum limit reached
You have reached the maximum limit of 90,000 rows. Some of your data may not be imported from the database connector.

ethervoid:
I want to make a point about the intentionality of the message. When we reach this point we're pretty sure that the user has reached the limit of rows but we don't know for sure, hence the `possibly` part of the message

csobier:
Hmm, okay, how about:
"You may have reached the maximum limit
For a database connector import, the number of rows allowed is 90,000. Some of your data may not be imported."

ethervoid:
Looks good for me, changing!

PR #9874, opened 2016-09-26
8892 widget sorting on export import

### LINKED ISSUE #8892: The widgets lost the order and there is no way to sort them out

javitonino:
Fixes #8892 by exporting/importing widgets order


javitonino:
Please CR @gfiorav

PR #9878, opened 2016-09-26
9788 viewer in eumapi

gfiorav:
Please CR @javitonino

### DIFF
--- a/app/controllers/carto/api/organization_users_controller.rb
+++ b/app/controllers/carto/api/organization_users_controller.rb
@@ -37,6 +37,9 @@ module Carto
         account_creator.with_password(create_params[:password]) if create_params[:password].present?
         account_creator.with_quota_in_bytes(create_params[:quota_in_bytes]) if create_params[:quota_in_bytes].present?
 
+        param_viewer = create_params[:viewer]
+        account_creator.with_viewer(param_viewer) if param_viewer
+
         if create_params[:soft_geocoding_limit].present?
           account_creator.with_soft_geocoding_limit(create_params[:soft_geocoding_limit])
         end
--- a/app/controllers/carto/api/organization_users_controller.rb
+++ b/app/controllers/carto/api/organization_users_controller.rb
@@ -143,12 +143,13 @@ module Carto
 
       # TODO: Use native strong params when in Rails 4+
       def create_params
-        permit(COMMON_MUTABLE_ATTRIBUTES.merge(:username, :viewer))
+        @create_params ||=
+          permit(COMMON_MUTABLE_ATTRIBUTES + [:username, :viewer])
       end
 
       # TODO: Use native strong params when in Rails 4+
       def update_params
-        permit(COMMON_MUTABLE_ATTRIBUTES)
+        @update_params ||= permit(COMMON_MUTABLE_ATTRIBUTES)
       end
     end
   end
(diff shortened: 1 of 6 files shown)


app/controllers/carto/api/organization_users_controller.rb
```diff
@@ -130,16 +133,26 @@ def destroy
...
-               :soft_obs_snapshot_limit, :soft_obs_general_limit, :soft_twitter_datasource_limit)
+        @create_params ||=
+          permit(COMMON_MUTABLE_ATTRIBUTES + [:username, :viewer])
       end
 
       # TODO: Use native strong params when in Rails 4+
       def update_params
-        permit(:email, :password, :quota_in_bytes, :soft_geocoding_limit, :soft_here_isolines_limit,
-               :soft_obs_snapshot_limit, :soft_obs_general_limit, :soft_twitter_datasource_limit)
+        @update_params ||= permit(COMMON_MUTABLE_ATTRIBUTES)
```
javitonino: Why can't you update the viewer field?

app/helpers/organization_users_helper.rb
```diff
@@ -39,7 +39,7 @@ def current_viewer_is_owner?
   def permit(*permitted)
     hardened_params = params.dup
 
-    hardened_params.keep_if { |k, _v| permitted.include?(k.to_sym) }
+    hardened_params.keep_if { |k, _v| permitted.flatten.include?(k.to_sym) }
```
javitonino: If this is not used elsewhere, you could remove `flatten` here and remove the `*` from the method signature. Not important though.

app/models/carto/user_creation.rb
```diff
@@ -210,6 +211,7 @@ def initialize_user
     @cartodb_user.soft_obs_snapshot_limit = soft_obs_snapshot_limit unless soft_obs_snapshot_limit.nil?
     @cartodb_user.soft_obs_general_limit = soft_obs_general_limit unless soft_obs_general_limit.nil?
     @cartodb_user.soft_twitter_datasource_limit = soft_twitter_datasource_limit unless soft_twitter_datasource_limit.nil?
+    @cartodb_user.viewer = viewer if viewer
```
javitonino: `if viewer` is unnecessary as the `viewer` field is not-nullable, right?

app/controllers/carto/api/organization_users_controller.rb
gfiorav: It's a mess apparently. Not as simple as updating the value :/

app/helpers/organization_users_helper.rb
gfiorav: Since it preserves the methods functionality and the abstractions make sense, I'd leave it!

app/models/carto/user_creation.rb
gfiorav: I'm not sure that if you expressly set something to null, the default thingies work as expected. Like: In a method with optional params, if you set one to nil, default values don't apply.

javitonino: Not what I meant. Let me try again. `viewer` here is reading the field from the database (user_creations table) and that field cannot be null, so I don't think `viewer` can return `nil` here under any circumstances.

I think the `if  viewer` will always be true and thus is not needed. But anyway, not important, so feel free to ignore it.

PR #9888, opened 2016-09-26
Updated font list for Builder

makella:
@xavijam here is the updated list for Builder fonts. I tested each one as typed out in staging so all should be good.

thanks!
mamata


xavijam:
Yay! :)

xavijam:
BTW, don't wait for hound, it is a bit crazy today.

PR #9891, opened 2016-09-27
9758 delete dataset dependencies

### LINKED ISSUE #9758: Delete a dataset warns about map that no longer points to it
# STR

1. Upload twice the same dataset
2. Create a map with one
3. Change sql on the layer to point to other dataset
4. Delete original dataset

# Expected result

Dataset is deleted

# Actual result

A warning is issued about the map being deleted and it's effectively deleted upon confirmation.

NOTE: It also warns when deleting the right dataset.

javitonino:
Fixes #9758


javitonino:
Please CR @juanignaciosl

PR #9899, opened 2016-09-27
Don't remove time-series when layer has changed

xavijam:
After checking a use-case with @saleiva, we have found this problem:

- Create a map with a layer of points (having a date column).
- Add a time series widget and filter it, just select a small block.
- Add a widget over that layer, for example, centroid one.
- A JS error appears, but the problem is related with the behaviour of removing the widget and then trying to add it again.

I don't really remember why we need to do this with the "static" time series widget, I could understand that for the torque one, but I'm not sure. Could you help me @nobuti refreshing me why we took that decision? 💋


nobuti:
To be honest, I don't remember either, as far I can see, it comes from the animated type change we introduced some weeks ago. Probably it was wrong and I missed it.

xavijam:
Basically the idea would be to check if previous style change comes from the animation type in order to decide if remove or not the widget.

xavijam:
Merging due to travis tests are passing.

PR #9902, opened 2016-09-27
9779 dataset sharing queries

### LINKED ISSUE #9779: Dataset sharing errors
Description:

Christophe shared a dataset with me.

I got the email from CARTO with the notification.
screenshot 2016-09-15 18 49 57

When I open the link the dataset doesn't load:

screenshot 2016-09-15 18 49 22

When I try to download it doesn't work (I have read only access):

screenshot 2016-09-15 18 49 31

It is surprising that you can see the query to the database but nice for those who understand SQL.

javitonino:
Closes #9779 

 - Qualifies table name with user schema on import (for organizations)
 - Fixes [SQL] tag display on dataset view and builder (to take into account the schema)

This does not rewrite queries for existing datasets, although it will mark them as "SQL applied" and clearing the SQL will qualify them. I think that's enough, but we can consider rewriting existing queries in the backend either via rake, or when sharing a dataset.

The change only affects organizations, lone users are unchanged.


javitonino:
Please CR @xavijam @gfiorav

app/model_factories/layer_factory.rb
```diff
@@ -33,6 +33,9 @@ def self.get_default_data_layer(table_name, user, geometry_type)
       data_layer.tooltip ||= {}
       data_layer.tooltip['fields'] = []
 
+      schema_prefix = user.organization.present? ? "#{user.sql_safe_database_schema}." : ""
```
gfiorav: supremely minor: use `''` for non interpolated strings!

javitonino:
Re @gfiorav review (any way to reply to CR comments?)

That's a good point. Yes, initializing the analysis takes the query from the canonical visualization layer. However, when a new layer to a map (not creating a map from a layer), the query is reset, and right now it's created without qualifying it if you own the layer.

I don't think that's a problem for maps where qualifying the query is not always necessary (as you can't share them with write permissions). Anyway, I feel changing it makes everything more consistent. Also, I tested this and I need to make some adjustments to the UI to handle this case correctly.

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
```diff
@@ -284,11 +284,15 @@ module.exports = CoreView.extend({
...
     var query = 'SELECT * FROM ' + tableName;
 
-    if (isShared) {
-      query = SQLUtils.prependTableName(query, tableName, userName);
+    if (isInOrganization) {
+      if (userName) {
+        query = SQLUtils.prependTableName(query, tableName, userName);
+      } else {
+        query = 'SELECT * FROM ' + this._layerDefinitionModel.getTableName();
+      }
```
javitonino: This is awful but works. It is actually solving a different issue which is present in master. If you add a layer from another user to an existing map, and go to the SQL editor, you can clear the query and it will unqualify the table name.

Basically due to `table_name` and `user_name` option handling. When we import, we set both, but when adding a layer in the builder, `table_name` is set with user.table.

javitonino:
Fixed the case when adding a layer after the map is created and some little fun thing along the way. The code is ugly, can be improved, but I want to talk about the `user_name` and `table_name` options (which of those we are going to use and how) to see if we can simplify.

javitonino:
Arg, as expected, this keeps on giving. Adding a default qualified query confuses the old dataset editor (it goes to read-only mode).

lib/assets/javascripts/cartodb3/dataset/dataset-content/dataset-content-options-view.js
```diff
@@ -90,16 +90,14 @@ module.exports = CoreView.extend({
 
   _isCustomQueryApplied: function () {
     var tableName = this._tableModel.getUnqualifiedName();
-    var originalQuery = 'SELECT * FROM ' + Utils.safeTableNameQuoting(tableName);
-    var isOwner = this._tableModel.isOwner(this._userModel);
-    var userName;
+    var defaultQuery = 'SELECT * FROM ' + Utils.safeTableNameQuoting(tableName);
```
xavijam: :+1: to these changes over this class.

lib/assets/javascripts/cartodb3/dataset/dataset-header/header-view.js
```diff
@@ -368,8 +368,15 @@ module.exports = CoreView.extend({
   },
 
   _isCustomQueryApplied: function () {
-    var originalQuery = 'SELECT * FROM ' + this._tableModel.getUnquotedName();
-    return !SQLUtils.isSameQuery(originalQuery, this._querySchemaModel.get('query'));
+    var tableName = this._tableModel.getUnqualifiedName();
```
xavijam: Same here :+1:.

lib/assets/javascripts/cartodb3/dataset/dataset-options/dataset-options-view.js
```diff
@@ -297,11 +297,8 @@ module.exports = CoreView.extend({
     var tableName = this._tableModel.getUnqualifiedName();
     var defaultQuery = 'SELECT * FROM ' + tableName;
 
-    var isOwner = this._tableModel.isOwner(this._userModel);
-    var userName;
-
-    if (!isOwner) {
-      userName = this._tableModel.get('permission').owner.username;
+    if (this._userModel.isInsideOrg()) {
```
xavijam: :+1: too.

lib/assets/javascripts/cartodb3/dataset/dataset-options/preview-map-view.js
```diff
@@ -39,8 +39,14 @@ module.exports = CoreView.extend({
   },
 
   render: function () {
-    var originalQuery = 'SELECT * FROM ' + this._tableModel.getUnquotedName();
-    var isCustomQueryApplied = !SQLUtils.isSameQuery(originalQuery, this._querySchemaModel.get('query'));
+    var tableName = this._tableModel.getUnqualifiedName();
```
xavijam: Venga niño!

lib/assets/javascripts/cartodb3/editor/layers/analysis-views/source-layer-analysis-view.js
```diff
@@ -12,13 +12,25 @@ module.exports = CoreView.extend({
...
+  },
+
   options: {
     showId: true
   },
 
   render: function () {
-    var originalQuery = 'SELECT * FROM ' + this.model.get('table_name');
-    var isCustomQueryApplied = !SQLUtils.isSameQuery(originalQuery, this.model.get('query'));
+    var tableName = this.model.get('table_name');
```
xavijam: This will qualify all datasets, no matter if user belongs to an organization or not. We could provide user-model at this point if needed.

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
xavijam: We could change that in the "layer" addition instead of doing this check, how does it sound?

lib/assets/javascripts/cartodb3/editor/layers/analysis-views/source-layer-analysis-view.js
javitonino: This is not used for qualifying, just for checking whether to display the [SQL] tag or not in the source analyses. This will indeed hide the tag for both qualified or unqualified datasets, which I think it might even be good.

But yes, passing the user model here might be even better.

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
javitonino: If we can always add `user_name` to the layer it might be good. Didn't know if there was any reason for removal.

lib/assets/javascripts/cartodb3/editor/layers/analysis-views/source-layer-analysis-view.js
xavijam: Uhm, ok!

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js
xavijam: We have agreed a better option.

xavijam: Changes coming.

javitonino:
Deployed in staging cloud and everything seems to work fine, except for issues #9933 and #9927 which I think are best left for another PR.

PR #9909, opened 2016-09-27
Connector endpoints

ethervoid:
- [X] List providers endpoint
- [X] Show provider information endpoint
- [x] Test provider connection endpoint (need tests)
- [X] List tables for provider


ethervoid:
@jgoizueta please take a look

app/controllers/carto/api/connectors_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+        end
+      end
+
+      private
+
+      def build_connection_parameters(request_params)
+        parameters = {}
+        parameters[:provider] = request_params[:provider_id]
+        parameters[:connection] = {}
+        parameters[:connection][:server] = request_params[:server]
```
jgoizueta: We should not hardcode the names of the connection parameters; they are defined by each provider. I think you could pass all parameters present into `:connection` and let Connector validate them.

I would have also expected parameters here as passed to the Import API, i.e. all connection parameters  inside `:connection` just as they are being relayed here to the Connector. That will probably simplify this code (although calls to the API may be a little less convenient).

lib/carto/connector.rb
```diff
@@ -99,7 +104,7 @@ def self.limits(provider_name:, user:)
     #     }
     #   }, ...
     # }
-    def self.information(provider_name, user = nil)
+    def self.information(provider_name)
```
jgoizueta: Oops, what was that `user` was doing there? (my bad)

ethervoid:
@jgoizueta take a look now

app/controllers/carto/api/connectors_controller.rb
```diff
@@ -26,40 +26,45 @@ def show
...
-        parameters[:connection][:server] = request_params[:server]
-        parameters[:connection][:port] = request_params[:port]
-        parameters[:connection][:database] = request_params[:database]
-        parameters[:connection][:username] = request_params[:username]
-        if request_params[:password].blank?
-          parameters[:connection][:password] = ''
-        else
-          parameters[:connection][:password] = request_params[:password]
+        provider_information = Carto::Connector.information(provider_id)
+        provider_information[:parameters]["connection"].each do |key, _value|
```
jgoizueta: you can iterate the keys with
```
provider_information[:parameters]["connection"].keys.each do |key|
```

And you could replace the loop by:

```
keys = provider_information[:parameters]["connection"].keys
parameters[:connection] = Hash[keys.map{|k| [k.to_sym, request_params[key.to_sym]]}]
```
...but that probably makes it less legible 😛

Issue #9223, opened 2016-08-02
Add publishing option for hiding lateral(blue) bar


saleiva:

saleiva:
@xavijam can we prioritize this?

saleiva:
moved, just in case.

xavijam:
@nobuti will make his magic adding it in the current PR about legends config too, so...

saleiva:
OMG

nobuti:
This is going to be shipped with #9911

saleiva:
@noguerol is going to be soooooo happy.

nobuti:
Already in production.

PR #9911, opened 2016-09-28
Legends map options

### LINKED ISSUE #9751: Add legends options to map options
After #9584, we need to add the option in order to customise the public view.

nobuti:
This PR fixes #9751 #9223.

legends-map-options

It changes a bit the way we save the layer selector option. Because it's coupled with the legend block, it doesn't make sense to handle it different. Remember that overlays create views with order and positions. Until now, the layer selector was an overlay in cartodb.js, and but not legends. This PR treats them both as they would be outside overlays. In the builder this means we are saving both options in the map definition model. So the `vizjson` need to reflect this change and it should add the proper `layer_selector`.

~~If we decide this is the way to go, the changes in the vizjson need to be done in order to persist the user's option.~~

Also, as part of this PR, I fixed how we initialise the custom legend model, in the same manner of the rest of models. This avoid an empty categories in the form when the custom option is rendered.

This PR depends on https://github.com/CartoDB/cartodb/pull/9926 and viceversa. Also depends on https://github.com/CartoDB/cartodb.js/pull/1425

cc @xavijam


xavijam:
I'd remove the "layer selector" option because as you have pointed, it does not make any sense to treat them in two different ways if they are in the same component.

nobuti:
This is ready for another pass.

lib/assets/javascripts/cartodb3/editor.js
```diff
@@ -111,7 +111,7 @@ UserGroupFetcher.track({
 // from the given vizJSON, but enabling legends behind a feature flag
 vizJSON = _.extend(
   vizJSON, {
-    legends: userModel.featureEnabled('builder-legends')
+    legends: userModel.featureEnabled('builder-legends') && vizJSON.legends
```
xavijam: Let's remove feature flag please :)

lib/assets/javascripts/cartodb3/public_editor.js
```diff
@@ -1,16 +1,11 @@
...
-  {
-    legends: false
-  }
-);
+var vizJSON = window.vizJSON;
 var dashboardOpts = {
   no_cdn: false,
   cartodb_logo: false,
-  renderMenu: true,
+  renderMenu: vizJSON.embed_options.dashboard_menu,
```
xavijam: cc @saleiva

xavijam: And enable legends if they are not.

nobuti:
Getting the do not merge label back again. This PR depends on two other PR, so we have to wait for them.

nobuti:
Current behaviour after fixing it:

map-legends-options

nobuti:
Frontend tests passing 🚀

PR #9912, opened 2016-09-28
Added title value to column name element in dataset view

### LINKED ISSUE #9880: It's difficult to identify columns with long column names in the dataset view
In the Editor, as we had horizontal scroll in the Dataset View, column names were rendered totally even if long:

image

In the Builder, long column names get trimmed (with no alt message):

image

and it looks like if you click over them, they "expand", but the name is still not visible:

image

I noticed about this while working with @talos in the Data Observatory use case, where columns usually have long names so the navigation could be a bit difficult.

MariaCheca:
Fixes #9880 

In cases where the column name is too long and the text is truncated with `...` it wil be still accessible with the `title` tag.

screen shot 2016-09-28 at 09 32 33

CR @nobuti please :)


iriberri:
Yayyyy!

talos:
👍

PR #9914, opened 2016-09-28
Add check_connection method to Connector

jgoizueta:
This adds a method to check a connection intended to validate the parameters without needed to fetch the list of available tables (which could not be implemented for some providers).


jgoizueta:
One detail left to decide: currently it returns `true` for a valid connection and may return `false` or raise an exception for invalid ones. Should avoid raising exceptions? return error messages?

jgoizueta:
@ethervoid this is tested in staging and ready to go, can you review?

PR #9923, opened 2016-09-28
Fixed in package dependencies.

nobuti:
This PR fixes the dev dependencies of a dependency 🙉  🔫 

We had to fork [backbone-model-file-upload](https://github.com/CartoDB/backbone-model-file-upload/tree/1.0.0)  in order to fix the version, then to update the package pointing to that branch.

cc @xavijam @rochoa

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "dependencies": {
     "backbone": "1.2.3",
     "backbone-forms": "0.14.0",
-    "backbone-model-file-upload": "git://github.com/CartoDB/backbone-model-file-upload.git#1.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
   "dependencies": {
     "backbone": "1.2.3",
     "backbone-forms": "0.14.0",
-    "backbone-model-file-upload": "1.0.0",
+    "backbone-model-file-upload": "git://github.com/CartoDB/backbone-model-file-upload.git#1.0.0",
```
xavijam: could we make the same as we do for our own repos?
```json
…
"cartodb/backbone-model-file-upload#1.0.0"
…
```

PR #9926, opened 2016-09-28
Add embed options to map

### LINKED ISSUE #9916: Add `embed_options` to Map
Add a new validated JSON called `embed_options` to store different embed options.

gfiorav:
Fixes #9916


gfiorav:
Please CR @javitonino !

xavijam:
Ping!

gfiorav:
@nobuti and me are testing integration, I don't want to call on CR until I'm sure it won't change @xavijam !

nobuti:
@gfiorav We need the info in the `vizjson` in order to initialize the frontend, but I can't see any `embed_options` there.

gfiorav:
Please CR @javitonino

db/migrate/20160928095559_add_embed_options_to_map.rb
```diff
@@ -0,0 +1,15 @@
...
+                 type: 'json'
+    end
+  end
+
+  down do
+    alter_table :maps do
+      drop_column :embed_options
+    end
+  end
+end
```
javitonino: Should be safe (no table scan nor rewrites) :sparkles:

config/routes.rb
```diff
@@ -645,6 +606,75 @@
     get '/superadmin/stats/total_likes' => 'platform#total_likes'
   end
 
+  UUID_REGEXP = /([0-9a-f]{8})-([0-9a-f]{4})-([0-9a-f]{4})-([0-9a-f]{2})([0-9a-f]{2})-([0-9a-f]{12})/
+
+  scope module: Carto::Api, path: '(/user/:user_domain)(/u/:user_domain)/api/', format: :json do
```
javitonino: Does `Carto::Api` do the same thing as `'carto/api'`?

config/routes.rb
```diff
@@ -645,6 +606,75 @@
...
+    scope 'v3/' do
```
javitonino: It seems like you have not deleted the old `v3` scope, and many of this routes are duplicated!

lib/assets/javascripts/cartodb3/data/map-definition-model.js
```diff
@@ -21,7 +21,7 @@ module.exports = Backbone.Model.extend({
 
   urlRoot: function () {
     var baseUrl = this._configModel.get('base_url');
-    return baseUrl + '/api/v1/maps/';
+    return baseUrl + '/api/v2/maps/';
```
javitonino: Remember to package bump on deploy. Also, can this break if tries to access the `v1` route? That is, a client running with old javascript doing requests against the v1 controller in the new code.

app/models/carto/map.rb
```diff
@@ -28,6 +28,15 @@ class Carto::Map < ActiveRecord::Base
     center:          [30, 0]
   }.freeze
 
+  serialize :embed_options, ::Carto::CartoJsonSerializer
+  validates :embed_options, carto_json_symbolizer: true
+
+  validate :validate_embed_options
+
+  after_initialize :ensure_embed_options
+  after_find :ensure_embed_options
```
javitonino: This seems redundant according to docs: http://guides.rubyonrails.org/active_record_querying.html#retrieving-objects-from-the-database. `after_initialize` should also get called in this case. I tried it out and seemed to work with `where`, `find` and `first` at least.

app/models/carto/map.rb
```diff
@@ -123,8 +132,42 @@ def update_dataset_dependencies
...
+  end
+
+  def layer_selector=(value)
+    embed_options[:layer_selector] = value
+  end
+
+  def layer_selector
+    embed_options[:layer_selector]
+  end
+
```
javitonino: No getters/setters for legends/scrollwheel?

app/controllers/carto/api/vizjson3_presenter.rb
```diff
@@ -78,23 +78,22 @@ def calculate_vizjson(https_request: false, forced_privacy_version: nil)
...
-          scrollwheel:    map.scrollwheel,
-          legends:        map.legends,
-          map_provider:   map.provider,
           bounds:         bounds_from(map),
           center:         map.center,
-          zoom:           map.zoom,
-          updated_at:     map.viz_updated_at,
+          datasource:     datasource_vizjson(options, forced_privacy_version),
+          description:    html_safe(@visualization.description),
+          embed_options:  map.embed_options,
```
javitonino: Hard to see what was added with all the reordering :dizzy_face: Not a fan of the reorder, it was easier for me to read the previous version. Of course, I wrote it and the order makes sense to me, so I am not impartial at all here.

app/controllers/carto/api/maps_controller.rb
```diff
@@ -3,37 +3,56 @@
...
-      def is_uuid?(text)
-        !(Regexp.new(%r{\A#{UUIDTools::UUID_REGEXP}\Z}) =~ text).nil?
+      def owners_only
+        unless @map.writable_by_user?(current_viewer)
+          raise Carto::LoadError.new('Map not found')
+        end
       end
 
+      def update_params
+        @update_params ||= params.slice(:bounding_box_ne,
```
javitonino: I would remove the memoification here. This value is not currently reused within the same call, and it's not a costly operation anyway.

config/routes.rb
```diff
@@ -645,6 +606,75 @@
     get '/superadmin/stats/total_likes' => 'platform#total_likes'
   end
 
+  UUID_REGEXP = /([0-9a-f]{8})-([0-9a-f]{4})-([0-9a-f]{4})-([0-9a-f]{2})([0-9a-f]{2})-([0-9a-f]{12})/
```
javitonino: :heart:

app/controllers/carto/api/maps_controller.rb
```diff
@@ -3,37 +3,56 @@
...
+        @map = Carto::Map.find(params[:id])
+      rescue ActiveRecord::RecordNotFound
+        raise Carto::LoadError.new('Map not found')
       end
 
-      # TODO: remove this method and use  app/helpers/carto/uuidhelper.rb. Not used yet because this changed was pushed before
-      def is_uuid?(text)
-        !(Regexp.new(%r{\A#{UUIDTools::UUID_REGEXP}\Z}) =~ text).nil?
+      def owners_only
+        unless @map.writable_by_user?(current_viewer)
```
javitonino: Old version used `current_user`. I don't think there is any difference with `current_viewer` for this purpose. Have you checked?

app/controllers/carto/api/maps_controller.rb
```diff
@@ -3,37 +3,56 @@
...
+                                        :bounding_box_sw,
+                                        :center,
+                                        :embed_options,
+                                        :provider,
+                                        :view_bounds_ne,
+                                        :view_bounds_sw,
+                                        :zoom)
+      end
+
+      def map_presentation(map: @map)
```
javitonino: This parameter is never used with non-default values. I'd remove it.

config/routes.rb
gfiorav: _"yes"_ ™️ ©️*


*: might not.

Checking to be safer but I'm kinda sure

config/routes.rb
gfiorav: :/ changing!

lib/assets/javascripts/cartodb3/data/map-definition-model.js
gfiorav: Old path will lead to old controller, which is what was used before. No issues here!

app/models/carto/map.rb
gfiorav: Oh, read that wrong. Thanks for checking!

app/controllers/carto/api/maps_controller.rb
javitonino: It's not likely to be a huge memory drain. But it's adding a non-zero overhead (not only in memory, but also in CPU to call the memory allocator) for no gain (as it's never called twice in the same request).

It's a minor impact, but seeing as it has no benefit, I am for removing it.

gfiorav:
Have a second look please!

spec/requests/carto/api/maps_controller_spec.rb
```diff
@@ -5,35 +5,100 @@
...
+      end
+    end
+
+    it 'validates on update' do
+      payload[:embed_options] = { spammy: 'hell', dashboard_menu: true }
+      put_json create_show_map_url, payload do |response|
+        response.status.should eq 422
+      end
+    end
+
```
javitonino: Can you add a test to check that vizjson/named map invalidation is called? Doesn't have to be complicated, just to make sure that the invalidation code is being called.

app/controllers/carto/api/maps_controller.rb
gfiorav: I think it's safer to use current_viewer no?

javitonino: Kinda. The only difference is if you are logged in but looking at stuff from another user (different subdomain). As this is for ownership, it shouldn't matter.

config/routes.rb
```diff
@@ -645,6 +571,75 @@
...
+
+  scope module: 'carto/api', path: '(/user/:user_domain)(/u/:user_domain)/api/', format: :json do
+    scope 'v3/' do
+      scope 'maps/:map_id/layers/:map_layer_id', constraints: { map_id: /[^\/]+/, map_layer_id: /[^\/]+/ } do
```
javitonino: These IDs should also probably be UUID_REGEXP, but not sure if this would break anything. Not a thing I'd do in this PR anyway.

gfiorav: Yup, some tests use 'wadus' for a fake id instead of `random_uuid`.

gfiorav:
Hey @javitonino, thanks for the review. I've added the tests you suggested!

xavijam:
ping!

gfiorav:
@xavijam this can't be deployed without the front end changes, so waiting on integration to be ready on @nobuti side!

xavijam:
Ooook! waiting for them then 💋

alonsogarciapablo:
Question... why did we chose `embed_options` as the name for this key? In the context of the vizjson, I don't think using the term "embed" has too much sense... In general, vizjson files are used to display "maps"/dashboards (using cdb.js), no matter what context these maps are used in... IMO, the term "embed" is something internal that we use to refer to the public map, and it doesn't make to much sense for the external world. IMO, these are options are more like `settings`... thoughts? Thx!

gfiorav:
@alonsogarciapablo admit this is revenge for `definition`!! 

Haha! Jokes aside:

The real question I think is: why are we using vizjson to store options that are only used for embeds?

alonsogarciapablo:
😛 

> The real question I think is: why are we using vizjson to store options that are only used for embeds?

What's an embed for you @gfiorav?

That's exactly my point... these options are not only used for "embeds", but for any map...

gfiorav:
hey @alonsogarciapablo I have an idea: let's call it `options`. It even follows the standard I think? no? 😛

alonsogarciapablo:
As long as it doesn't have `embed` on it, I'm cool with whatever we decide: `options`, `settings`... are there any other ideas?

xavijam:
I agree about changing the name, options or params should fit great.

gfiorav:
Changed! Thanks for the comments. Let's see this go 🍏 , have @javitonino have a last look, have migrations on #10009 out, and then deploy this.

alonsogarciapablo:
Thanks for getting rid of `embed_options` @gfiorav! 👏

app/models/carto/map.rb
```diff
@@ -123,8 +131,42 @@ def update_dataset_dependencies
...
+    self.options ||= {
+      dashboard_menu: true,
+      layer_selector: true,
+      legends: legends,
+      scrollwheel: scrollwheel
+    }
+  end
+
+  def validate_options
+    location = "#{Rails.root}/lib/formats/map/options.json"
```
javitonino: This file is still named embed_options.json

gfiorav: yup!

gfiorav:
This is being merged through https://github.com/CartoDB/cartodb/pull/9911

PR #9936, opened 2016-09-28
fix width

piensaenpixel:
@MariaCheca could you take a look?

image


MariaCheca:
Looks great, thank you @piensaenpixel :)

Issue #9927, opened 2016-09-28
Cannot delete sync from dataset view


javitonino:
Using the new dataset view from the builder, if you try to delete a synchronization (set interval to "never") it sets the interval to 0 instead of deleting it.

 1. Create a sync table (e.g: import from the data library)
 2. Open in the builder's dataset view
 3. In the sync bar, select "View options", select "Never" and save

Expected behaviour: DELETE to synchronizations endpoint (same as editor).
Actual behaviour: PUT to synchronizations endpoint setting interval to 0. The dataset still shows as synced in the dashboard, and actually syncs from time to time.

xavijam:
👍

PR #9942, opened 2016-09-29
Destroy synchronization when user selects interval 0

xavijam:
Fixes https://github.com/CartoDB/cartodb/issues/9927.

CR: @matallo 
cc @javitonino


xavijam:
I've spotted one thing in the builder side, let me fix it before CR @nobuti.

xavijam:
Moving CR to @matallo, please man 💋 .

matallo:
roger!

xavijam:
ping-pong!

lib/assets/test/spec/cartodb3/data/query-rows-collection.spec.js
```diff
@@ -1,7 +1,7 @@
-// var _ = require('underscore');
+// var _ = require('underscore')
```
matallo: shall we keep commented lines?

xavijam: no, let me remove it.

PR #9944, opened 2016-09-29
Escape regex parameters in `ensure_unique_name`

javitonino:
Closes https://github.com/CartoDB/cartodb-central/issues/1482


javitonino:
Please CR @gfiorav

Issue #9048, opened 2016-07-21
UI is not rendered when initial request to Maps API fails


alonsogarciapablo:
When the initial request to the Maps API fails (from cartodb.js), the UI doesn't render the sidebar, etc, so users don't know what happened (and there's nothing they can do but going back using the brower's back button):

![screen shot 2016-07-21 at 12 48 44](https://cloud.githubusercontent.com/assets/390398/17020364/7bb49378-4f41-11e6-91af-e2e9c03228bd.png)

We should render the UI and display some errors to inform users that something went wrong.

cc: @xavijam

PR #9959, opened 2016-09-29
Changes over Builder initialization

xavijam:
Changes done, mentioning people after each point:

- Removed unnecessary background-geocoding files: they are useless because geocoding now run with analyses. Found that because we were using `vis.map` from CartoDB.js in those views, so removed dependency and views (@nobuti).
- Populated node **status** to Builder node-definition-model: because we were passing CartoDB.js analysis collection to several views in order to know the status of the analysis. Better to "set" that status in our Builder node model. Node status is not stored/persisted in any moment (@alonsogarciapablo).
- Map-definition-model needs vis map, so instead of wait for CartoDB.js initialization in order to create the model, we set that "map" object to the model when it is ready (@nobuti).
- Same logic applied for overlays-collection (@nobuti).
- Removed CartoDB.js `vis.analysis` dependency from several views. I feel they were not changed when deep-insights-integrations was created, or we didn't want to use our Builder instances due to the lack of node `status` (now populated as I mentioned above) (@alonsogarciapablo).
- Not restoring CartoCSS (autostyle thing) always when there is a change over the layer, only when auto-style is enabled within the widget (@alonsogarciapablo @rochoa).
- Avoiding initial fetchs from mapscaps and overlays collections (thanks @gfiorav).
- And of course changes over Builder entry file moving several Classes out of the CartoDB.js dashboard initialization.

Fixes https://github.com/CartoDB/cartodb/issues/9048.
cc @rochoa


xavijam:
I have to review failing tests and add new ones, but I feel this could be a good start 💃 .

nobuti:
Currazo 💯

lib/assets/javascripts/cartodb3/editor.js
```diff
@@ -125,69 +294,18 @@ deepInsights.createDashboard('#dashboard', vizJSON, {
...
-  var layerDefinitionsCollection = new LayerDefinitionsCollection(null, {
-    configModel: configModel,
-    analysisDefinitionNodesCollection: analysisDefinitionNodesCollection,
-    mapId: mapId
-  });
-
   var vis = dashboard.getMap();
   var dashboardView = dashboard.getView();
 
+  mapDefModel.setVis(vis);
```
alonsogarciapablo: It looks like we initialise most of the builder models/collections with data coming from the backend (eg: `layerDefinitionsCollection.resetByLayersData(layersData);`). Why are we using [vizjson](https://github.com/CartoDB/cartodb/pull/9959/files#diff-d27b00749653f54e193fd10402d14d8dR199) and cartodb.js models in this case?

xavijam: I agree with you, but this model works a bit different that the rest, because it needs info from the vis map when it is going to be saved.

lib/assets/javascripts/cartodb3/editor.js
```diff
@@ -125,69 +294,18 @@ deepInsights.createDashboard('#dashboard', vizJSON, {
...
+  overlaysCollection.setMirrorCollection(vis.overlaysCollection);
```
xavijam: Yeah, we can do that easily with this one.

lib/assets/javascripts/cartodb3/editor.js
xavijam: those come from vis.map, yes

nobuti:
It keeps looking good, I like the fact we are putting together all the cartodb.js integration under the same file, but not sure if this file will become unmanageable some day.  While this day comes 👍

xavijam:
Yeah, that day will come sooner or later, but in the mean time, we are safe. Take a look at my last changes please, I feel I'm covering what @alonsogarciapablo mentioned in the feedback. I need to add some tests but basically, the approach is there.

lib/assets/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -521,6 +534,34 @@ F.prototype._setLayerError = function (layerDefinitionModel, cdbError) {
...
+  this._vis().overlaysCollection.add(m.toJSON());
+};
+
+F.prototype._onOverlayDefinitionRemoved = function (mdl) {
+  var collection = this._vis().overlaysCollection;
+  var overlay = collection.findWhere({ type: m.get('type') });
+  overlay && collection.remove(overlay);
+};
+
+F.prototype._onNewBasemapSet = function () {
```
alonsogarciapablo: We have the map definition model in this file too. Since we're already updating the map when  a base layer is saved [here](https://github.com/CartoDB/cartodb/pull/9959/files#diff-7317d638b3d159e8ceda11ed714f6acfR39), we could also do something like:

```
opts.mapDefinitionModel.on('change:minZoom change:maxZoom', this._onMapZoomChanged, this);
```

Just an idea...

alonsogarciapablo:
👍 . I like how this is turning out @xavijam! I left an idea...

lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
```diff
@@ -167,18 +167,22 @@ module.exports = Backbone.Collection.extend({
...
-          success: function () {
-            deferred.resolve();
-          },
-          error: function () {
-            deferred.reject();
-          }
-        });
-      }
+      } else if (querySchemaModel.canFetch()) {
+        if (status !== 'fetching') {
```
alonsogarciapablo: What this is telling me is that the querySchemaModel can be fetched unless `status === 'fetching'`... Couldn't that condition be part of `canFetch`?

```
if (querySchemaModel.isFetched()) {
  // resolve the promise
} else if (querySchemaModel.canFetch()) {
  // resolve or reject the promise
}
```

Which also raises some questions/ideas:

1. what if the querySchemaModel is not fetched and for some reason can not be fetched. Would this be possible. Wouldn't the promise be left unresolved/unrejected?
2. the fact that we're querying multiple times to querySchemaModel (`isFetched`, `canFetch`, `get('status')`), before actually telling the object what to do (`fetch`), makes me wonder if we could move this logic to the querySchemaModel, so would just have to do something like the following here...

  ```
  // perhaps we can find a better name, but you see my idea...
  querySchemaModel.fetchOrForward({
    success: function () { deferred.resolve(); },
    error: function () { deferred.reject(); }
  });
  ```

lib/assets/javascripts/cartodb3/deep-insights-integrations.js
xavijam: That function should only be triggered when a new basemap is set, the thing is that the event is not super descriptive and probably the comment starts from there.

xavijam: Agree with you finally, better to have map changes centralized in this place instead of listening layer-definitions-collection event. 💋

lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
xavijam: If it is fetching it can be fetch again, because we abort the previous petition and done. But in this case we can't cancel it because it could break our promises. I would not go into details with this thing. It is an edge case I've discovered doing some testing.

xavijam: About questions:
- If request fails, promise is rejected. But all of them will be tried.
- That's a not bad idea, but I'd not add more things to this PR and this is something a bit out of the scope.

If we consider we should leave it for other PR I could remove those changes.

xavijam:
@gfiorav, check backend changes please, although we have changed together, just in case.
I'm deploying changes to ded02 in order to test them, I encourage everybody in the @CartoDB/frontend to take a look at the changes because we will initialize the Builder in a different way, and I hope we don't add new bugs.

lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js
alonsogarciapablo: Thanks for the reply! As I said, feel free to tackle this is a future PR :).

About the promise not being resolved/rejected, what I actually meant, is that there's one branch of these (:point_down:) ifs/elses that we're mssing/not considering (I added it at the bottom):
 
```
if (querySchemaModel.isFetched()) {
  deferred.resolve();
} else if (querySchemaModel.canFetch()) {
  if (status !== 'fetching') {
    querySchemaModel.fetch({
      success: function () {
        deferred.resolve();
      },
      error: function () {
        deferred.reject();
      }
    });
  } else {
    deferred.resolve();
  }
} else {
  // WARNING: we're not resolving/rejecting the promise in this case...
}
```

xavijam: :+1: thanks man! 💋

xavijam:
Ready to go then guys?

nobuti:
Checked in staging and I couldn't see anything 💯

javitonino:
:+1: to backend

xavijam:
It will be merged when TRAVIS gives a green hello.

alonsogarciapablo:
👏 👏 👏  Great job @xavijam!

xavijam:
Merrrrging!!! 😱

PR #9965, opened 2016-09-30
Fix connectors rake

jgoizueta:
Fix a couple of issues with the rake tasks for connectors:

* bug with organizations task, not being recognized
* failure to create provider records for non public providers (that we need also to configure for internal use)


jgoizueta:
@ethervoid, please review/approve

PR #9969, opened 2016-10-01
Leapfrog jasmine

nobuti:
This PR adds [mock-ajax](http://jasmine.github.io/2.5/ajax.html) to our test stack. TBH this is completely selfish because I have troubles testing ajax callbacks and promises, so I'm tired of it. This helper eases those kind of testing. As an example, I've updated one of out tests to cover fully the case of saving in the backend (mocking the response of course).

Feel free to discard this PR, but in this case I demand a proper explanation about how we should test this kind of thins 😁 

cc @xavijam @CartoDB/frontend


package.json
```diff
@@ -65,7 +65,8 @@
     "grunt-mustache": "0.1.7",
     "grunt-postcss": "0.5.5",
     "grunt-timer": "0.3.3",
-    "jasmine": "2.5.0",
+    "jasmine": "2.5.2",
+    "jasmine-ajax": "git://github.com/nobuti/jasmine-ajax.git#master",
```
nobuti: It has a bug related to modern browsers when one header is not present, see https://github.com/jasmine/jasmine-ajax/issues/153. So I forked and fixed. The I realised there was a PR open to fix it yet.

xavijam: 😂

xavijam:
I really like the idea, so 👍

alonsogarciapablo:
This seems pretty useful! 👍

matallo:
I feel your pain, and actually I asked @javisantana assistance with something similar before, he pointed me to that same plugin

I finally solved it via:

```
spyOn($, 'ajax');
```

and

```
$.ajax.calls.argsFor(0)[0].error();
$.ajax.calls.argsFor(0)[0].success();
```

I always get stuck at this point so 👍

PR #9972, opened 2016-10-03
9939 remove analysis cache on deletion

### LINKED ISSUE #9939: Remove analysis tables for org user before deleting it
This is supposed to avoid issues like:

```
User was not deleted. PG::Error: ERROR:  cannot drop schema user_xxx because other objects depend on it
DETAIL:  table user_xxx.analysis_foobarsz depends on schema user_xxx
table user_xxx.analysis_foobarsz1 depends on schema user_xxx
table user_xxx.analysis_foobarsz2 depends on schema user_xxx
table user_xxx.analysis_foobarsz3 depends on schema user_xxx
table user_xxx.analysis_foobarsz4 depends on schema user_xxx
table user_xxx.analysis_foobarsz5 depends on schema user_xxx
table user_xxx.analysis_foobarsz6 depends on schema user_xxx
table user_xxx.analysis_foobarsz7 depends on schema user_xxx
table user_xxx.analysis_foobarsz8 depends on schema user_xxx
table user_xxx.analysis_foobarsz9 depends on schema user_xxx
table user_xxx.analysis_foobarsz10 depends on schema user_xxx
HINT:  Use DROP ... CASCADE to drop the dependent objects too.
```

We can use the analysis catalog for this.

cc/ @javitonino @rochoa

javitonino:
Fixes #9939

### DIFF
--- a/app/models/user/db_service.rb
+++ b/app/models/user/db_service.rb
@@ -390,6 +390,7 @@ module CartoDB
             # If user is in an organization should never have public schema, but to be safe (& tests which stub stuff)
             unless @user.database_schema == SCHEMA_PUBLIC
               database.run(%{ DROP FUNCTION IF EXISTS "#{@user.database_schema}"._CDB_UserQuotaInBytes()})
+              drop_analysis_cache
               drop_all_functions_from_schema(@user.database_schema)
               database.run(%{ DROP SCHEMA IF EXISTS "#{@user.database_schema}" })
             end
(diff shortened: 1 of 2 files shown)


javitonino:
CR @gfiorav

app/models/user/db_service.rb
```diff
@@ -815,6 +816,17 @@ def revoke_all_memberships_on_database_to_role(conn, role)
         end
       end
 
+      def drop_analysis_cache
+        list_sql = "SELECT DISTINCT unnest(cache_tables) FROM cdb_analysis_catalog WHERE username = '#{@user.username}'"
+        delete_sql = "DELETE FROM cdb_analysis_catalog WHERE username = '#{@user.username}'"
```
gfiorav: Why not cleanup after each viz? This is fine, it might be a better solution. I'm just curious!

spec/models/user_spec.rb
```diff
@@ -1518,6 +1518,17 @@ def overquota(delta = 0)
         end.to raise_error(/role "#{role}" does not exist/)
         db.disconnect
       end
+
+      it 'deletes temporary analysis tables' do
```
gfiorav: Boss! 😎

app/models/user/db_service.rb
javitonino: Good idea! I did not think of that (I was just trying to solve the user deletion problem).

However, I don't think I can do that. There is nothing matching to `visualization_id` (or similar) from the `cdb_analysis_catalog`. I am not sure exactly how it works in camshaft, but I think it matches using the analysis definition with some unique id's.

PR #9985, opened 2016-10-03
Enhance margins dataset editor title

### LINKED ISSUE #9806: Enhance margins in the Dataset editor options
We'd like to improve a bit the visual in the Dataset editor, specifically the margin between the dataset title and the options icon. 

Right now it's like this: 

screen shot 2016-09-19 at 12 03 36

And we'd like to make this changes, so there's always 24px of separation in between elements. 

image

Let me know if you have any doubts :)
And thanks!

MariaCheca:
Fixes #9806 

This branch includes changes concerning a bug when the user changes the dataset name in the edition view. In cases when the dataset had a query applied, after changing the name and refreshing the views the query had to be _removed_, but it wasn't persisted.

screen shot 2016-10-03 at 16 32 52

CR @nobuti


MariaCheca:
@nobuti yep, I'm currently fixing the tests :) Thanks!

PR #10006, opened 2016-10-04
Migration to add version to visualizations

javitonino:
Also created technical debt issue: https://github.com/CartoDB/cartodb/issues/10005 to do after we remove common data from the visualizations table.

CR @gfiorav

### DIFF
--- a/db/migrate/20161004130412_add_version_to_visualizations.rb
+++ b/db/migrate/20161004130412_add_version_to_visualizations.rb
@@ -1,0 +1,8 @@
+Sequel.migration do
+  change do
+    alter_table :visualizations do
+      # This should be non-nullable and default=2 but that forces a db downtime
+      add_column :version, :integer, null: true
+    end
+  end
+end


db/migrate/20161004130412_add_version_to_visualizations.rb
```diff
@@ -0,0 +1,8 @@
+Sequel.migration do
+  change do
+    alter_table :visualizations do
+      # This should be non-nullable and default=2 but that forces a db downtime
+      add_column :version, :integer, null: true
```
gfiorav: are you sure you don't want to go with a `x.y.z` versioning scheme?

javitonino: Seems overkill to me and it's harder to parse (string vs integer). I think this should be used for breaking changes on visualizations, and I store what would be the major version number (x). Easy enough to implement a mroe complicated version scheme in this column if we eventually feel the need (030101).

PR #10009, opened 2016-10-04
Migration for #9926

gfiorav:
Migration for #9926

CR @javitonino


javitonino:
:shipit: but let's monitor the time it takes to migrate (stats).

PR #10026, opened 2016-10-05
10025 drag layer nodes sql indicator

### LINKED ISSUE #10025: SQL appears when dragging out the original layer node from an analysis
### STR

1. Select a dataset and click on "CREATE MAP".
2. Add an analysis to that layer. For instance, "Create Areas of Influence".
3. Drag out and drop the original layer from the analysis. 
4. An unexpected behaviour happens, the layer node shows that there is an `SQL` query applied. When checking, just the schema is added. You can click on "CLEAR" to remove this addition.

javitonino:
Fixes #10025


javitonino:
Please CR @xavijam

xavijam:
:es:

PR #10027, opened 2016-10-05
organizations: placeholder avatar

matallo:
- adds a placeholder avatar for organizations, when they have not uploaded theirs yet 

screen shot 2016-10-05 at 11 45 13

CR @javierarce 

/cc @santisaez @carlostallon


app/views/admin/organizations/settings.html.erb
```diff
@@ -33,7 +33,22 @@
         </div>
         <div class="FormAccount-rowData FormAccount-avatar">
           <div class="FormAccount-avatarPreview">
-            <img src="<%= @organization.avatar_url %>" title="" alt="" class="FormAccount-avatarPreviewImage" />
+            <% if @organization.avatar_url.present? %>
+              <img src="<%= @organization.avatar_url %>" title="" alt="" class="FormAccount-avatarPreviewImage" />
```
javierarce: Super minor thing. Does it make sense to have the `title` and `alt` empty? Couldn't we remove them or fill them with some data?

matallo: empty tile makes no sense, I'll remove it from the element I replicated, too

empty alt is recommended

doing these ^^^

PR #10036, opened 2016-10-05
Mark visualizations as vizjson2

javitonino:


javitonino:
Please CR @gfiorav 

cc @luisbosque

app/controllers/carto/api/visualizations_controller.rb
```diff
@@ -204,6 +205,13 @@ def to_hash(visualization)
         VisualizationPresenter.new(visualization, current_viewer, self).to_poro
       end
 
+      def mark_as_vizjson2
+        referer_host = URI.parse(request.referer).host
+        unless referer_host.ends_with?('carto.com') || referer_host.ends_with?('cartodb.com')
+          @visualization.mark_as_vizjson2
+        end
+      rescue
```
gfiorav: I'd capture the specific exception. Better not shadow any other errors!

app/models/carto/visualization.rb
```diff
@@ -431,6 +433,14 @@ def state
     super ? super : build_state
   end
 
+  def mark_as_vizjson2
+    $tables_metadata.SADD(V2_VISUALIZATIONS_REDIS_KEY, id)
```
gfiorav: Id use `SETNX`.

app/models/carto/visualization.rb
```diff
@@ -431,6 +433,14 @@ def state
...
+  end
+
+  def uses_vizjson2?
+    $tables_metadata.SISMEMBER(V2_VISUALIZATIONS_REDIS_KEY, id) > 0
```
gfiorav: Why not `GET`? Also, something tells we should avoid using this method massively. We don't have the use case now, but we might in the future. Maybe adding a comment?

spec/requests/carto/api/visualizations_controller_spec.rb
```diff
@@ -32,6 +32,17 @@ def vizjson_vx_version
       def attributions_from_vizjson(visualization)
         visualization['layers'][1]['options']['attribution'].split(',').map(&:strip)
       end
+
+      it 'marks visualizations as using vizjson2' do
```
gfiorav: good test!

app/models/carto/visualization.rb
javitonino: I was considering using that (id in key) but finally used sets. Some discussion in https://github.com/CartoDB/cartodb-platform/issues/2506

gfiorav: Ok! Then disregard the first part of the next comment! 😄

app/controllers/carto/api/visualizations_controller.rb
javitonino: In fact, I just realised the rescue should mark the visualization :(

Adding tests for that as first step

javitonino:
Made some changes and added some tests. Can you take a second look?

PR #10045, opened 2016-10-05
Fixes problem with undo and redo buttons in Firefox

### LINKED ISSUE #9947: Query history buttons do not work in Firefox
In the builder, the buttons to go back/forward in the query history, do not work in Firefox. No console errors reported.

cc @xavijam

xavijam:
Fixes #9947.

cc @javitonino


javitonino:
Tested and it works, thank you!

PR #10057, opened 2016-10-05
Check if edition before to show the notification.

### LINKED ISSUE #10052: cartocss error should not appear in the notifier if the CartoCSS panel is open.
its too much to see the same error twice at the same time.

nobuti:
This PR fixes #10052.

CR @xavijam


xavijam:
:+1:

PR #10063, opened 2016-10-06
Do not lookup vizjson visualization by user table uuid

javitonino:
It is not being used anymore.


javitonino:
Please CR @juanignaciosl

PR #10080, opened 2016-10-07
Use correct layout for embed_map for orgs

javitonino:
This fixes an error in development that for some reason doesn't happen in production.

STR:
 1. Create a map with editor
 2. Open the embed
 3. It shows the dashboard header but no embed (and this incorrect version is cached)

Alternatively:
 1. Reset the embed cached (e.g: change privacy)
 2. Open the embed in private browsing mode
 3. The embed loads (and cached the correct version)

cc @donflopez @javisantana


javisantana:
EPA :es:

PR #10082, opened 2016-10-07
Filter analyses issues

xavijam:
Several fixes found checking filter analysis.
cc @javierarce


javierarce:
Sorry, I just saw this.

PR #10087, opened 2016-10-07
9981 builder redirections

### LINKED ISSUE #9981: Builder and dataset redirections
So:

- If a user has viz.json v2 visualizations, they should be opened with the Editor. Public maps should be opened with the public editor map, and builder map can't be reached.
- If a user has "migrated" its map to Builder, public or embed Editor map should redirect to Builder public map.

What about the table/dataset view?
- If a user has Builder enabled, dataset view.
- If a user has Editor enabled, table view (GME users for example).
- For public viewers, table view

Take into account this [premise](https://github.com/CartoDB/cartodb/issues/9978).

javitonino:
- Embed redirections (based on visualization version flag)
 - Auto migration (version update) for safe visualizations
 - Open vizjson2 visualizations in editor

Closes #9981

### DIFF
--- a/app/controllers/admin/visualizations_controller.rb
+++ b/app/controllers/admin/visualizations_controller.rb
@@ -82,7 +82,7 @@ class Admin::VisualizationsController < Admin::AdminController
     elsif !@visualization.has_write_permission?(current_user)
       return redirect_to CartoDB.url(self, 'public_visualizations_public_map',
                                      id: request.params[:id], redirected: true)
-    elsif current_user.force_builder?
+    elsif current_user.force_builder? && !@visualization.open_in_editor?
       return redirect_to CartoDB.url(self, 'builder_visualization', id: request.params[:id])
     end
 
--- a/app/controllers/admin/visualizations_controller.rb
+++ b/app/controllers/admin/visualizations_controller.rb
@@ -39,8 +39,7 @@ class Admin::VisualizationsController < Admin::AdminController
   before_filter :redirect_to_builder_embed_if_v3, only: [:embed_map, :show_organization_public_map,
                                                          :show_organization_embed_map, :show_protected_public_map,
                                                          :show_protected_embed_map,
-                                                         :public_map, :show_protected_public_map
-                                                        ]
+                                                         :public_map, :show_protected_public_map]
 
   after_filter :update_user_last_activity, only: [:index, :show]
   after_filter :track_dashboard_visit, only: :index
--- a/app/controllers/admin/visualizations_controller.rb
+++ b/app/controllers/admin/visualizations_controller.rb
@@ -36,6 +36,9 @@ class Admin::VisualizationsController < Admin::AdminController
                                                          :show_protected_public_map, :show_protected_embed_map]
 
   before_filter :resolve_visualization_and_table_if_not_cached, only: [:embed_map]
+  before_filter :redirect_to_builder_embed_if_v3, only: [:embed_map, :show_organization_public_map,
+                                                         :show_organization_embed_map, :show_protected_public_map,
+                                                         :show_protected_embed_map]
 
   after_filter :update_user_last_activity, only: [:index, :show]
   after_filter :track_dashboard_visit, only: :index
@@ -692,5 +695,13 @@ class Admin::VisualizationsController < Admin::AdminController
     Carto::Tracking::Events::VisitedPrivatePage.new(current_user_id,
                                                     user_id: current_user_id,
                                                     page: 'dashboard').report
+  end
+
+  def redirect_to_builder_embed_if_v3
+    # @visualization is not loaded if the embed is cached
+    # Changing version invalidates the embed cache
+    if @visualization && @visualization.version == 3
+      redirect_to CartoDB.url(self, 'builder_visualization_public_embed', visualization_id: @visualization.id)
+    end
   end
 end
--- a/app/controllers/admin/visualizations_controller.rb
+++ b/app/controllers/admin/visualizations_controller.rb
@@ -38,7 +38,9 @@ class Admin::VisualizationsController < Admin::AdminController
   before_filter :resolve_visualization_and_table_if_not_cached, only: [:embed_map]
   before_filter :redirect_to_builder_embed_if_v3, only: [:embed_map, :show_organization_public_map,
                                                          :show_organization_embed_map, :show_protected_public_map,
-                                                         :show_protected_embed_map]
+                                                         :show_protected_embed_map,
+                                                         :public_map, :show_protected_public_map
+                                                        ]
 
   after_filter :update_user_last_activity, only: [:index, :show]
   after_filter :track_dashboard_visit, only: :index
--- a/app/controllers/carto/admin/visualization_public_map_adapter.rb
+++ b/app/controllers/carto/admin/visualization_public_map_adapter.rb
@@ -20,7 +20,8 @@ module Carto
         :password_protected?, :varnish_key, :related_tables, :password_valid?, :get_auth_tokens, :table, :name,
         :overlays, :created_at, :updated_at, :description, :mapviews, :geometry_types, :privacy, :tags,
         :surrogate_key, :has_password?, :total_mapviews, :is_viewable_by_user?, :is_accesible_by_user?,
-        :can_be_cached?, :is_privacy_private?, :source, :kind_raster?, :has_read_permission?, :has_write_permission?
+        :can_be_cached?, :is_privacy_private?, :source, :kind_raster?, :has_read_permission?, :has_write_permission?,
+        :open_in_editor?
       ] => :visualization
 
       attr_reader :visualization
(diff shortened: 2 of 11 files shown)


javitonino:
Please CR @gfiorav

app/controllers/admin/visualizations_controller.rb
```diff
@@ -36,6 +36,9 @@ class Admin::VisualizationsController < Admin::AdminController
                                                          :show_protected_public_map, :show_protected_embed_map]
 
   before_filter :resolve_visualization_and_table_if_not_cached, only: [:embed_map]
+  before_filter :redirect_to_builder_embed_if_v3, only: [:embed_map, :show_organization_public_map,
```
gfiorav: Test: 

0. Password protected map on a non migrated viz
1. Go to password prompt (for v2)
2. Put wrong password

Are you properly redirected?

app/controllers/admin/visualizations_controller.rb
```diff
@@ -693,4 +696,12 @@ def track_dashboard_visit
                                                     user_id: current_user_id,
                                                     page: 'dashboard').report
   end
+
+  def redirect_to_builder_embed_if_v3
+    # @visualization is not loaded if the embed is cached
```
gfiorav: So if cached there will be no auto migration?

app/controllers/carto/admin/visualization_public_map_adapter.rb
```diff
@@ -20,7 +20,8 @@ class VisualizationPublicMapAdapter
         :password_protected?, :varnish_key, :related_tables, :password_valid?, :get_auth_tokens, :table, :name,
         :overlays, :created_at, :updated_at, :description, :mapviews, :geometry_types, :privacy, :tags,
         :surrogate_key, :has_password?, :total_mapviews, :is_viewable_by_user?, :is_accesible_by_user?,
-        :can_be_cached?, :is_privacy_private?, :source, :kind_raster?, :has_read_permission?, :has_write_permission?
+        :can_be_cached?, :is_privacy_private?, :source, :kind_raster?, :has_read_permission?, :has_write_permission?,
+        :open_in_editor?, :version
```
gfiorav: Minor, off topic: we should really stop worsening our viz<->map identity problem with all these delegates 😂

app/controllers/carto/builder/public/embeds_controller.rb
```diff
@@ -8,7 +8,8 @@ class EmbedsController < BuilderController
 
         ssl_required :show, :show_protected
 
-        before_filter :load_visualization, only: [:show, :show_protected]
+        before_filter :load_visualization,
+                      :redirect_to_old_embed_if_v2, only: [:show, :show_protected]
```
gfiorav: Same testing proposed (wrong password on embed protected when not yet migrated).

app/controllers/carto/builder/visualizations_controller.rb
```diff
@@ -12,8 +12,9 @@ class VisualizationsController < BuilderController
 
       ssl_required :show
 
-      before_filter :redirect_to_editor_if_forced,
-                    :load_derived_visualization, only: :show
+      before_filter :load_derived_visualization,
```
gfiorav: why the change of order? curious

app/models/carto/overlay.rb
```diff
@@ -23,6 +23,9 @@ class Overlay < ActiveRecord::Base
       'header', 'search', 'layer_selector', 'share', 'zoom', 'logo', 'loader', 'fullscreen'
     ].freeze
 
+    BUILDER_COMPATIBLE_TYPES = ['search', 'layer_selector', 'share', 'fullscreen', 'loader', 'logo', 'zoom'].freeze
+    scope :builder_incompatible, where('type NOT IN (?)', BUILDER_COMPATIBLE_TYPES)
```
gfiorav: nice trick!

app/models/visualization/member.rb
```diff
@@ -364,6 +364,11 @@ def tags=(tags)
         super(tags)
       end
 
+      def version=(version)
+        self.dirty = true
```
gfiorav: explain!

spec/models/table_spec.rb
```diff
@@ -736,9 +736,9 @@ def @data_import.data_source=(filepath)
 
     id = table.table_visualization.id
     CartoDB::Varnish.any_instance.expects(:purge)
-                                 .times(4)
-                                 .with(".*#{id}:vizjson")
-                                 .returns(true)
+                    .at_least_once
```
gfiorav: This test is now vulnerable to overlooking missing invalidations!

spec/models/table_spec.rb
```diff
@@ -2298,7 +2298,7 @@ def @data_import.data_source=(filepath)
       CartoDB::Visualization::Member.stubs(:new).with(has_entry(id: derived.id)).returns(derived)
       CartoDB::Visualization::Member.stubs(:new).with(has_entry(type: 'table')).returns(table.table_visualization)
 
-      derived.expects(:invalidate_cache).once
+      derived.expects(:invalidate_cache).at_least_once
```
gfiorav: Overlooking posible redundant invalidations. Although super minor.

spec/requests/admin/visualizations_spec.rb
```diff
@@ -161,6 +161,36 @@ def polymorphic_path(*args)
...
           last_response.status.should eq 200
         end
+
+        it 'never for vizjson2 visualizations' do
+          @user.stubs(:builder_enabled).returns(true)
+          @user.stubs(:builder_enabled?).returns(true)
+          Carto::Visualization::any_instance.stubs(:uses_vizjson2?).returns(true)
+
+          login_as(@user, scope: @user.username)
+          get "/viz/#{@id}", {}, @headers
```
gfiorav: No path names? Helpful in case we change routes in the future!

spec/requests/admin/visualizations_spec.rb
```diff
@@ -161,6 +161,36 @@ def polymorphic_path(*args)
...
+        it 'embed redirects to builder for v3 when needed' do
+          # These two tests are in the same testcase to test proper embed cache invalidation
+          @user.stubs(:builder_enabled).returns(false)
+          @user.stubs(:builder_enabled?).returns(false)
+          visualization = CartoDB::Visualization::Member.new(id: @id).fetch
+          visualization.version = 2
+          visualization.store
+
+          login_as(@user, scope: @user.username)
+          get "/viz/#{@id}/embed_map", {}, @headers
```
gfiorav: Same comment about routes.

app/controllers/carto/builder/visualizations_controller.rb
```diff
@@ -49,7 +50,9 @@ def builder_notifications
       end
 
       def redirect_to_editor_if_forced
-        redirect_to CartoDB.url(self, 'public_visualizations_show_map', id: params[:id]) if current_user.force_editor?
+        if current_user.force_editor? || @visualization.open_in_editor?
```
gfiorav: I've checked the code for `.force_editor?`. Why isn't it the same as `!.force_builder?`

app/controllers/carto/admin/visualization_public_map_adapter.rb
javitonino: This is kind of another thing. This is a wrapper around Carto::Visualization to make it look like a Memberino, so it can be used in controllers/presenters without the need to rewrite everything. ¯\_(ツ)_/¯

app/controllers/admin/visualizations_controller.rb
javitonino: Auto migration only happens from the builder, never from the embed. Summarizing:

- Embed redirections only check the version field. 2 -> old embed, 3 -> new embed
- Builder redirections check:
  - User is builder (if not -> editor)
  - Visualization uses vizjson 2 (if yes -> editor)
  - Can be automigrated (if yes -> do it!)
  - Else: show builder with version = 2 -> Opens manual migration modal

app/controllers/carto/builder/public/embeds_controller.rb
javitonino: Working!

app/controllers/carto/builder/visualizations_controller.rb
javitonino: Before, the redirection only depended on the user. Now, it also depends on the visualization itself (it uses vizjson2 or not). So I need to load the visualization before deciding to redirect. I could have kept the order and added a new filter for the viz-based redirection, but I prefer having all the redirection logic in the same place.

app/controllers/carto/builder/visualizations_controller.rb
javitonino: Because there is the option where nothing is force (`User.builder_enabled` is nil) and you can go to either editor and builder depending on the URL (ooooold behaviour). This will be removed soon-ish (next week most likely).

spec/models/table_spec.rb
javitonino: Uhm, didn't think that through. You are right! Changing back to 5.

spec/models/table_spec.rb
javitonino: I think this one is alright, as it is checking that it's invalidated. It shouldn't be a problem if it gets invalidated twice.

spec/requests/admin/visualizations_spec.rb
javitonino: I'm lazy and copied it from the other specs. Changing.

spec/requests/admin/visualizations_spec.rb
javitonino: Same

app/models/visualization/member.rb
gfiorav: Hacky! Can we make this into a method with a friendly name? 😁

gfiorav:
Apart from that `self.dirty` stuff,

📦 it for the 🆓 🌍

Issue #10091, opened 2016-10-07
Legends still appearing even tho I'm turning them off in Map Options


arianaescobar:
There's no error on the Console (nothing happens, really) and the legends are still there even tho I'm turning them off. 

![screen shot 2016-10-07 at 17 27 26](https://cloud.githubusercontent.com/assets/16247628/19195862/8f967046-8cb3-11e6-80ee-4f2b84280bab.png)

I'm not sure if this is the correct repo, sorry if it isn't!

xavijam:
Already being tackle by @gfiorav.

arianaescobar:
That was fast!

PR #10092, opened 2016-10-07
Add map options to exporter

gfiorav:
Fixes https://github.com/CartoDB/cartodb/issues/10091

Please CR @javitonino


javitonino:
:fire: it!

PR #10098, opened 2016-10-07
Legends migrator

### LINKED ISSUE #9983: Legends migration
Migrate old legends to the new ones. Important for the rollout.

cc @CartoDB/builder-backend

gfiorav:
Fixes #9983


xavijam:
Go legends migrator go!

gfiorav:
Please CR @javitonino !

lib/carto/legend_migrator.rb
```diff
@@ -0,0 +1,232 @@
...
+    attr_reader :layer_id, :legend
+
+    def initialize(layer_id, legend)
+      @layer_id = layer_id
+      @legend = legend
+    end
+
+    def build
+      new_type, new_definition = type_and_definition
+      title = title if title.present? && legend['show_title']
```
javitonino: Please, make the local variable and method names different.

lib/assets/javascripts/cartodb3/data/legends/legend-base-definition-model.js
```diff
@@ -21,6 +21,9 @@ module.exports = Backbone.Model.extend({
       _.omit(r, 'created_at', 'updated_at', 'definition', 'pre_html', 'post_html')
     );
 
+    if (r.definition && r.definition.html) {
+      attrs.rawHTML = r.definition.html;
+    }
```
javitonino: What's this about? Something I need to know?

app/controllers/carto/api/vizjson3_presenter.rb
```diff
@@ -331,6 +337,11 @@ def as_data
         }
       end
 
+      def migrate_legends(old_legend)
+        Carto::LegendMigrator.new(@layer.id, old_legend).build.save
+        @layer.reload
```
javitonino: If you want to avoid this `reload`, you could do something like `@layer.legends.create(...)`. Alternatively, something easier is `@layer.legends.reload`, which will only fetch the legends (no need to refetch the layer itself).

lib/carto/legend_migrator.rb
```diff
@@ -0,0 +1,232 @@
...
+      custom_definition = Hash.new
+
+      categories = items.each_with_index.map do |item, index|
+        title = item['name'].to_s || "Category #{index + 1}"
+        color = item['value']
+
+        category_definition = { title: title }
+
+        if color && color =~ COLOR_REGEXP
+          category_definition[:color] = color
```
javitonino: What happens if the original did not have any color applied? Default on cartodb.js side?

lib/carto/legend_migrator.rb
```diff
@@ -0,0 +1,232 @@
...
+        end
+
+        category_definition
+      end
+
+      custom_definition[:categories] = categories
+      custom_definition
```
javitonino: I think you can delete the first line of this method, as well as this, and just return `{ categories: categories }`.

lib/carto/legend_migrator.rb
```diff
@@ -0,0 +1,232 @@
...
+
+      if stripped.length == 3
+        characters = stripped.map { |character| [character, character] }
+        stripped = characters.flatten.join
+      end
+
+      [stripped[0..1].hex, stripped[2..3].hex, stripped[4..5].hex]
+    end
+
+    def formmated_string_number(number)
```
javitonino: Typo: formatted

lib/carto/legend_migrator.rb
```diff
@@ -0,0 +1,232 @@
...
+      html += %(  </div>\n)
+      html += %(</div>\n)
+
+      { html: html }
+    end
+
+    def labels_for_items
+      return @labels_for_items if @labels_for_items
+
+      first_item = items.first
```
javitonino: What about doing something like this instead:

`items.select { |i| i['type'] == 'text' }.map { |i| i['value'] }` (and ensure there are at least 2 elements in the result)

I am thinking of the case where the two labels are not the two first ones (which I do not know if it is possible).

lib/carto/legend_migrator.rb
```diff
@@ -0,0 +1,232 @@
...
+      html += %(  <p class="CDB-Text CDB-Size-small">#{right_label}</p>\n)
+      html += %(</div>\n)
+      html += %(<div class="Legend-choropleth" style="#{style}"></div>\n)
+
+      { html: html }
+    end
+
+    def style_for_gradient
+      first_color_index = labels_for_items.compact.count
+      item_colors = items[first_color_index..-1].map do |item|
```
javitonino: What about `items.select { |i| i['type'] == 'color' }`. Seems more robust than trying to count the number of `text` items.

lib/carto/legend_migrator.rb
```diff
@@ -0,0 +1,232 @@
...
+
+      range = right - left
+      step = range / 6
+
+      first = formmated_string_number(left)
+      second = formmated_string_number(left + (2 * step))
+      third = formmated_string_number(left + (3 * step))
+      fourth = formmated_string_number(left + (4 * step))
+      fifth = formmated_string_number(left + (5 * step))
+      sixth = formmated_string_number(right + step)
```
javitonino: This `right + step` seems wrong, but is probably right. Can you explain why this is not `left + 6*step` or simply `right`.

lib/carto/legend_migrator.rb
```diff
@@ -0,0 +1,232 @@
...
+
+      sixth_mark = 100
+      fifth_mark = 80
+      fourth_mark = 60
+      third_mark = 40
+      second_mark = 20
+      first_mark = 0
+
```
javitonino: This method would look much nicer using arrays.
```
values = 5.downto(0) do |i|
  value = formmated_string_number(left + i*step),
  mark = 100 + 20*i
  [value, mark]
end


array.each |i|
html += %(    <li class="Bubble-numbersItem CDB-Text CDB-Size-small" style="bottom: #{array[1]}%">#{array[0]}</li></li>\n)
end
```

But not sure this is exactly the same (it appears to be pretty much the same for each element, but I might be overlooking something). Ignore if this doesn't work.

lib/carto/legend_migrator.rb
```diff
@@ -0,0 +1,232 @@
...
+      end
+
+      if hex_green.length == 1
+        hex_green = [hex_green, hex_green].join
+      end
+
+      if hex_blue.length == 1
+        hex_blue = [hex_blue, hex_blue].join
+      end
+
```
javitonino: Maybe you can use an `[r,g,b]` array instead of three variables, and just wrap everything in `rgb_array.map do ...`

spec/lib/carto/legend_migrator_spec.rb
```diff
@@ -0,0 +1,408 @@
...
+      end
+
+      it 'migrates old custom to new custom' do
+        @old_legend = old_custom
+      end
+
+      after(:each) do
+        new_legend = Carto::LegendMigrator.new(@layer.id, @old_legend).build
+
+        new_legend.type.should eq 'custom'
```
javitonino: Have you tried if asserts work in `after` code? I know exceptions are silently eaten.

spec/lib/carto/legend_migrator_spec.rb
```diff
@@ -0,0 +1,408 @@
...
+    describe('#bad legends') do
+      let(:bad_legend) do
+        {
+          patata_pochada: 'buena',
+          pero: 'mejor que pochada',
+          frita: 'a que si!'
+        }
+      end
+
+      let(:bad_category) do
```
javitonino: Not used?

lib/assets/javascripts/cartodb3/data/legends/legend-base-definition-model.js
gfiorav: Checking for definition in addition to html to support the case where there are no legends.

lib/carto/legend_migrator.rb
gfiorav: `nil` won't match the regex. It get's compacted out.

lib/carto/legend_migrator.rb
gfiorav: Yup, forgot to refactor.

lib/carto/legend_migrator.rb
gfiorav: Since I'm filling out a template, I value template readability over compactness. This is the same philosophy for `.erb`s (no line limits).

lib/carto/legend_migrator.rb
gfiorav: It's shouldn't be possible. Here we run into the dilema: should I get paranoid and validate everything?

My goal here is to be a able to handle some errors i.e, not break a vis on a bad legend. In this code, it won't break. It could be more compact, I agree.

lib/carto/legend_migrator.rb
gfiorav: Since it's a recurrent theme in this review: sure, it can be more compact. I do value readability since this is way off to normal backend stuff.

lib/carto/legend_migrator.rb
gfiorav: Thanks!

spec/lib/carto/legend_migrator_spec.rb
gfiorav: > I know exceptions are silently eaten.

Really? 

> Have you tried if asserts work in after code?

Yes, I wen't from red to green. :)

lib/carto/legend_migrator.rb
gfiorav: Can't really tell you, reviewing 🤔

lib/carto/legend_migrator.rb
javitonino: I see. If you don't write the color to the new legend, I guess the builder has a default for that, right?

lib/carto/legend_migrator.rb
javitonino: Yeah, I think this might be worth it, to avoid depending on items ordering. But I don't think it will matter to much (as all legends are created by the same JS code, that surely puts everything in the same order always).

lib/carto/legend_migrator.rb
javitonino: Yeah, as I said in the global comment, do not worry too much about the refactor requests. :sparkles:

lib/carto/legend_migrator.rb
gfiorav: No, the validation fails and the legend is not added.

gfiorav:
Second round, fight!

lib/carto/legend_migrator.rb
javitonino: Oh, got it! :+1:

lib/carto/legend_migrator.rb
javitonino: Still pending or is GH trolling me?

Issue #9482, opened 2016-08-18
give warning for .carto downloads if datasets are private


andy-esch:
Export of .carto files works even if the datasets are set to private, which causes confusion about the contents of the file once it is downloaded. There should be a warning and an on boarding screen about the contents of the file before a user downloads it.

ztephm:
Also if a .carto file is made from private datasets/private map but it is imported to a Free account, those datasets and map become public.

cc @csobier b/c of first Importing Rules bullet [here](https://carto.com/docs/carto-builder/managing-your-map/#exportingimporting-carto-files)

saleiva:
yes @ztephm. The owner of the data can always export it, but we can't/shouldn't control how it gets imported since it is an open format (you can always unzip the file and will get all stuff there)

ohasselblad:
I should have said above... My main reason for filing this ticket is that if a user with only public permissions tries to download a .carto file from another user, but the datasets are set to private, the user _will_ be able to download the file but it will not have the data for the private tables. This is the confusing part to me and deserving of more explanation from the interface.

csobier:
I clarified some of the exporting/importing rules in the docs for now. (Also, see the Slack #Builder channel discussion from today. There are plans to improve this functionality in the future, but not anytime soon.  Let me know if there is anything I can do.)

ztephm:
Hiya @ohasselblad  @csobier - Maybe behavior changed recently but I was able to see and use data from the private-dataset 'With link' Builder map .carto files when I imported it to my Free (Editor) account:

1. I exported the .carto file from within Builder by clicking the vertical ellipses next to map name & then Export map. 
2. I imported that to Free account by dragging/dropping .carto file over Maps page.

I was also able to see/use private dataset when I exported a .carto file from [public map link](https://team.carto.com/u/stephaniemongon/viz/1995f784-6884-11e6-9130-0e233c30368f/public), then drag/dropped that into my Free account

Here's my test:
Builder 'With Link' map using private dataset: https://team.carto.com/u/stephaniemongon/builder/1995f784-6884-11e6-9130-0e233c30368f/embed

.carto file:
[cartozip.zip](https://github.com/CartoDB/cartodb/files/435055/cartozip.zip)

Free account Editor map using previously private datasets:
https://ztephm.carto.com/viz/3a261818-ce77-459c-857b-4558214cf055/public_map

csobier:
[...]

PR #10101, opened 2016-10-09
9482 private datasets warning

hacheka:
This pull request has been created as a result of a challenge sent by @javisantana and @xavijam 
It resolves CartoDB/cartodb#9482 

There are changes in two places:
- In private builder, when exporting a map with private datasets, a warning shows up.
- In a map public page, when exporting the map it shows different warning depending on the private datasets ownership: it they are yours, it warns you that the private datasets will be exported. If you don't own them, the dialog warns you that not all the data will be exported.

You can find the GIFs of the different scenarios below.

Some notes about this pull request:
- There are tests for the new view logic. Anyway, no tests have been created related to code where no previous tests exist. This code only makes minor changes in existing logic, so no new specs have been created because it'd been implied to write tests for logic not related to the issue.
- An iteration of this PR could be to apply these changes to organizations, where users can have access to datasets not owned by them.
## Changes explained

There have been changes in private builder page and map public page.
### Private builder

New render option passed to modal-export-map-view indicating if the map contains private datasets.

**Map with public datasets**

public dataset

**Map with private datasets**

private dataset
### Map public page

Two data attributes rendered at the export map link: `data-private-datasets` and `data-owning-private-datasets` These attributes are filled at controller level and retrieved in the link event handling function. Passed to the modal view constructor.

**Map with public datasets**

public export

**Map with private datasets owned by the current user**

owning private

**Map with private datasets owned by another user**

not owning private

I guess that @xavijam is the one to review this PR. Be nice! ;)

Regards,
Ivan

### DIFF
--- a/app/controllers/admin/visualizations_controller.rb
+++ b/app/controllers/admin/visualizations_controller.rb
@@ -201,6 +201,7 @@ class Admin::VisualizationsController < Admin::AdminController
   end
 
   def public_map
+
     if current_user.nil? && !request.params[:redirected].present?
       redirect_url = get_corrected_url_if_proceeds(for_table=false)
       unless redirect_url.nil?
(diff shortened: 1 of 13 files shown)


Cartofante:

Cartofante:

Cartofante:

Cartofante:

xavijam:
Closing since we would manage in a short term properly.

PR #10102, opened 2016-10-10
Proper redirect to organization URL

javitonino:
Avoids double redirection going through user subdomain

Fixes point 1 of https://github.com/CartoDB/cartodb/issues/10090


javitonino:
Please CR @gfiorav

javitonino:
In this case, we must user `current_user` as it is redirecting to the owner's domain. `current_viewer` would go to the logged user domain, which of course, does not have the map when accessing shared maps and such stuff.

gfiorav:
thanks for the explanation!

📦 it for the 🆓 🌍

PR #10114, opened 2016-10-10
Delete affected widgets by a source change and then recreate them when new layer and source are created

### LINKED ISSUE #10099: Problem changing layer_id for a widget
## STR
- Create a map of points.
- Apply a category widget.
- Apply an analysis (subsample is ok).
- Drag out the source node to a new layer.
- Go to widgets list and check that the widget is pointing to the incorrect layer (check layer color and so on).

cc @saleiva @javisantana

xavijam:
Basically, when a node that has widgets associated is moved to a new layer, those widgets have to be remove and then recreated with the new layer_id and source.
Changed tests and added a notification that mentions widgets are being restored...

cc @javisantana @nobuti @javierarce @matallo 

Fixes #10099


xavijam:
If you see something weird, let me know...

xavijam:
@nobuti, one important thing. I've realized while testing this functionality that we were fetching query-schema-model for the widgets-view, and we needed to fetch the query-geometry-model. NFI why I didn't change that when I was doing last changes over those premises, but anyway, fixed. Take a look in case you can find something wrong :kiss:.

nobuti:
@xavijam yeah, because if query-schema model has a sql error, we don't allow the user to interact with that view, adding or deleting widgets.

xavijam:
Arg, so... ha... we will need to fetch both then?

nobuti:
I guess so...

xavijam:
😱 , ok!

nobuti:
Confirmed, nothing related with query errors. all we need is geometry https://github.com/CartoDB/cartodb/issues/9653

xavijam:
:+1:, so after merging master, do you see any problem with those changes?

nobuti:
No, it's looking good 👍

xavijam:
Fixing the last (hopefully) bug I found with this issue. That one fixed the problem creating widgets belonging to a layer where the head node was moved. 💋

xavijam:
Passing specs 🚀

PR #10123, opened 2016-10-11
10115 index database hosts

### LINKED ISSUE #10115: Add index on users.database_host
We need the following index on `users` table:

```sql
CREATE INDEX CONCURRENTLY ON users(database_host);
```

Before:

```sql
cartodb_production=# EXPLAIN SELECT COUNT(*) AS count FROM users U WHERE U.enabled = true AND U.database_host = 'a.b.c.d' ;
                             QUERY PLAN
---------------------------------------------------------------------
 Aggregate  (cost=16508.63..16508.64 rows=1 width=0)
   ->  Seq Scan on users u  (cost=0.00..16508.60 rows=75 width=0)
         Filter: (enabled AND (database_host = 'a.b.c.d'::text))
(3 rows)
```

The same query with the index created:

```sql
cartodb_production=# EXPLAIN SELECT COUNT(*) AS count FROM users U WHERE U.enabled = true AND U.database_host = 'a.b.c.d' ;
                                           QUERY PLAN
------------------------------------------------------------------------------------------------
 Aggregate  (cost=280.71..280.71 rows=1 width=0)
   ->  Index Scan using users_database_host_idx on users u  (cost=0.08..280.67 rows=74 width=0)
         Index Cond: (database_host = 'a.b.c.d'::text)
         Filter: enabled
(4 rows)
```

cc/ @javitonino

javitonino:
Fixes #10115 

Please CR @santisaez

### DIFF
--- a/db/migrate/20161011074820_add_index_to_database_hosts.rb
+++ b/db/migrate/20161011074820_add_index_to_database_hosts.rb
@@ -1,0 +1,15 @@
+Sequel.migration do
+  no_transaction # indexes need to be created outside of the transaction
+
+  up do
+    Rails::Sequel.connection.run %{
+      CREATE INDEX CONCURRENTLY "users_database_host_index" ON "users" ("database_host");
+    }
+  end
+
+  down do
+    Rails::Sequel.connection.run %{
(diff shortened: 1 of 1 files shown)


santisaez:
Thanks @javitonino, LGTM.

Issue #10124, opened 2016-10-11
Geometry Management


alonsogarciapablo:
Designs are here: https://projects.invisionapp.com/share/HK8NV15DB#/screens/195548487
- [ ] Adding a feature (drawing) - 5
  - Icon and modal to select the type of geometry that will be drawn.
  - While in "drawing" mode... display an "Edit Geometry" form (displays attributes fields)
  - When Geometry is "complete", fill in the geoJSON and enable de Accept button on the form.
  - If "Cancel" is clicked, discard the geometry
  - if "Accept" is clicked, persist GeoJSON and attributtes using SQL API.
  - Reload map.
- [ ] Editing a feature (Form -> UI) - 7
  - When a feature is clicked, determine if it's editable (shouldn't be editable if layer has some   analysis)
  - Display "Edit" button icon next to the point where user has clicked and editable feature
  - When "Edit" button is clicked:
    - Fetch info about the feature using the SQL API
    - Display "Edit Geometry" form with info about the geometry
    - Add editable geometry to the map using CARTO.js
    - Upon clicking "Save":
      - Persist changes using SQL API
    - Remove geometry via CARTO.js
    - Re-load map
- [ ] Editing a feature (UI -> Form) - 2
  - While in "Edit feature mode".
  - When the user modifies the geometry (drag and drop):
    - Update the form
    - Update the GeoJSON of the geometry (debounced)

cc: @xavijam @rochoa @javisantana @matallo

xavijam:
cc @matallo

saleiva:
Closing this since its being tracked in many others

PR #10137, opened 2016-10-11
edit geometry

matallo:
first commit for the edit geometry panel https://github.com/CartoDB/cartodb/issues/10124

working on the forms, already added the panel to the stack layout, and the header (see attached)

edit geometry panel

waiting for the new overlays https://github.com/CartoDB/cartodb.js/pull/1426, just added a new provisional button to trigger event and show panel

rough CR to check everything is ok and move on with this panel @alonsogarciapablo


matallo:
updates on the panel, I'm already sending the geometries (they are hardcoded in the buttons) and checking what type they are to show a form or other, in the end the model geometry it's always updated

geommodel


I'll go on with the properties form, and I will start thinking about getting the geometries from map features at some point

/cc @alonsogarciapablo

lib/assets/javascripts/cartodb3/data/feature-definition-model.js
```diff
@@ -43,6 +45,15 @@ var FeatureDefinitionModel = Backbone.Model.extend({
     });
   },
 
+  save: function () {
+    var columnsCollection = this._layerDefinitionModel.getAnalysisDefinitionNodeModel().querySchemaModel.columnsCollection;
+    var columns = columnsCollection.pluck('name');
+    var attrs = _.pick(this.toJSON(), _.difference(columns, BLACKLISTED_COLUMNS));
+
+    this._queryRowModel.set(attrs);
+    this._queryRowModel.save();
```
alonsogarciapablo: [Backbone's `save` accepts attributes](http://backbonejs.org/#Model-save) so we could just do: `this._queryRowModel.save(attrs)`.

lib/assets/javascripts/cartodb3/editor/layers/edit-feature-content-views/edit-feature-controls-view.js
```diff
@@ -65,13 +61,8 @@ module.exports = CoreView.extend({
...
 
   _saveFeature: function () {
-    var columns = _.map(this._columnsCollection.models, function(mdl) {
-      return mdl.get('name');
-    });
-    var attrs = _.pick(this._featureModel.toJSON(), _.without(columns, 'cartodb_id', 'created_at', 'the_geom_webmercator', 'updated_at'));
-
-    this._row.set(attrs);
-    this._row.save();
+    this._featureModel.fetch();
```
alonsogarciapablo: Why are fetching here?

matallo:
fixed the featureType, if it can't be induced by `the_geom` it goes to the featuretype set at the moment of creation

not sure this is the best solution but works both with creation and edition, as well as when there's a legit `type` attribute (happened to me with a dataset)

/cc @alonsogarciapablo

xavijam:
- [ ] If user edits a point (move it in the map), save it and then go back, the maps is not refreshed.
- [ ] If a query is applied, we should not allow to edit or add a new feature.
- [ ] If you type something in the form, it is rendered again and focus is lost.
- [ ] When a new feature is added (correctly drawn) and you click over save, cartodb_id is not set in the form data.
- [ ] Cancel button is not present.
- [ ] Edit feature button is not correct.
- [ ] Text inputs are longer than expected (Number ones are ok).
- [ ] Loading and error states are not present when form is being saved.
- [ ] Date edition should appear for date type.

matallo:
prevent feature edition when analysis is applied, as well as when is query applied; also bind the changes in the query to show the add geometries buttons

changes in https://github.com/CartoDB/cartodb/pull/10137/commits/c07e231c8b6566667eb78cc6f2b8ea8a748567e0

(realized I need to do the same when analysis change)

alonsogarciapablo:
About 
- [ ] If user edits a point (move it in the map), save it and then go back, the map is not refreshed.

edit-point

@xavijam could you please try to reproduce it again? Thanks!

lib/assets/javascripts/cartodb3/editor/layers/layer-content-view.js
```diff
@@ -260,7 +260,7 @@ module.exports = CoreView.extend({
     this._tableView = TableManager.create({
       queryGeometryModel: this._queryGeometryModel,
       querySchemaModel: this._querySchemaModel,
-      tableName: this._getTableName(),
+      tableName: this._nodeDefModel.get('table_name'),
```
matallo: this can be empty when layer is an analysis

lib/assets/javascripts/cartodb3/editor/layers/layer-content-view.js
```diff
@@ -274,12 +274,8 @@ module.exports = CoreView.extend({
     var sanitizeName = function (str) {
       return str.replace(/"/g, '').toLowerCase();
     };
-    var query = this._nodeDefModel.get('query');
-    return sanitizeName(query) !== sanitizeName('select * from ' + this._getTableName());
-  },
-
-  _getTableName: function () {
-    return this._nodeDefModel.get('table_name');
+    var query = this._querySchemaModel.get('query');
```
matallo: `this._nodeDefModel` doesn't have `query`attr, I created an issue for this purpose https://github.com/CartoDB/cartodb/issues/10334

lib/assets/javascripts/cartodb3/editor/layers/layer-content-view.js
```diff
@@ -274,12 +274,8 @@ module.exports = CoreView.extend({
...
+    return sanitizeName(query) !== sanitizeName('select * from ' + this._layerDefinitionModel.get('table_name'));
```
matallo: had to take the table name somewhere, we'd better use Utils for this

matallo:
fixed number fields width without slider, despite it works ™️ we should properly take a look as text fields and number fields are soo different (see attached)

screen shot 2016-10-25 at 19 28 22

lib/assets/javascripts/cartodb3/editor/layers/layer-content-view.js
```diff
@@ -265,28 +268,46 @@ module.exports = CoreView.extend({
...
+
     // TODO: check the behaviour, place and template of the buttons
     $('.CDB-Dashboard-canvas').append(
       changeViewButtons({
         isThereOtherWidgets: this._widgetDefinitionsCollection.isThereOtherWidgets(),
-        isThereTimeSeries: this._widgetDefinitionsCollection.isThereTimeSeries()
+        isThereTimeSeries: this._widgetDefinitionsCollection.isThereTimeSeries(),
+        queryGeometryModel: queryGeometryModel.get('simple_geom'),
+        isSourceType: isSourceType,
+        isCustomQueryApplied: this._isCustomQueryApplied()
```
alonsogarciapablo: we should use `[this._layerDefinitionModel.hasCustomQueryApplied](https://github.com/CartoDB/cartodb/pull/10137/files#diff-96ed1f8a6f25d43fa3d5758b58158c56R224)` here  ☝️

lib/assets/javascripts/cartodb3/data/layer-definition-model.js
```diff
@@ -217,6 +221,23 @@ module.exports = Backbone.Model.extend({
     return count;
   },
 
+  hasCustomQueryApplied: function () {
```
xavijam: Hey, one thing, for me this is not correct, because query is not dependant of the layer anymore, this logic should be placed in the analysis source node model, in my opinion.

alonsogarciapablo: But a layer still points to one analysis definition node model right? I'm okay with adding a `hasCustomQueryApplied` to the source node model too, and moving this logic there, but still, geometries are currently linked to layers... so that's why we do [this](https://github.com/CartoDB/cartodb/pull/10137/files#diff-6a9e0aec74c52158c7ccf6e2cfc9beafR53). As I said, perhaps FeatureDefinitionModels should be linked to Analysis nodes directly...

lib/assets/javascripts/cartodb3/editor/layers/layer-content-view.js
alonsogarciapablo: This is currently failing right after a new analysis is added... To reproduce:

1. Create any analysis.
2. This view gets rendered again.
3. Schema is not fetched yet, but `this._isCustomQueryApplied` tries check if a custom query is applied.
4. QuerySchemaModel is `undefined` and 💥

lib/assets/javascripts/cartodb3/data/layer-definition-model.js
```diff
@@ -217,6 +221,23 @@ module.exports = Backbone.Model.extend({
...
+    var sanitizeName = function (str) {
+      return str.replace(/"/g, '').toLowerCase();
+    };
+    var query = this._getQuerySchemaModel().get('query');
```
matallo: roger

lib/assets/javascripts/cartodb3/data/feature-definition-model.js
```diff
@@ -101,6 +101,17 @@ var FeatureDefinitionModel = Backbone.Model.extend({
     });
   },
 
+  destroy: function (options) {
+    this._getQueryRowModel().destroy({
```
matallo: shall we do anything more than this, removing the row?

lib/assets/javascripts/cartodb3/editor/layers/edit-feature-content-view.js
```diff
@@ -85,7 +86,10 @@ module.exports = CoreView.extend({
     this._headerView = new EditFeatureHeaderView({
       url: this._url,
       tableName: this._tableName,
-      model: this._featureModel
+      model: this._featureModel,
+      modals: this._modals,
+      isNew: this._featureModel.isNew(),
+      backAction: this.clean.bind(this)
```
matallo: passing the backAction as a parameter, as it'll go back after removing the feature, do you have a better naming for this?

lib/assets/javascripts/cartodb3/editor/layers/edit-feature-content-view.js
```diff
@@ -130,6 +139,59 @@ module.exports = CoreView.extend({
...
+      id: NOTIFICATION_ID,
+      status: 'loading',
+      info: '_onSaveFeature',
+      closable: false
+    });
+  },
+
+  _onFeatureSaved: function () {
+    this.notification.set({
+      id: NOTIFICATION_ID,
```
nobuti: No need to update the id, but it's ok.

alonsogarciapablo:
@matallo, I fixed most of the tests. I left some for you! :P

matallo:
💚

nobuti:
Vamos! 👏 👏

alonsogarciapablo:
👏

xavijam:
💋 @matallo @alonsogarciapablo

PR #10141, opened 2016-10-11
Adds isolines/track option to buffer analysis

### LINKED ISSUE #10001: 'Tracts' setting is applied in distance AOI analysis
# STR

- Create a points map with the builder
- Add AOI analysis
- Open `Time` type parameters and set `2` tracts

screen shot 2016-10-04 at 12 01 37

- Go back to `Units` type settings, change anything (or not) and Apply the analysis. The result is showing **2 tracts** even if the setting is not present for that type of AOI analysis.

screen shot 2016-10-04 at 12 03 32

## What's expected

Either the `tracts` setting should only apply on the `Time` AOIs.

Or the `tracts` setting should also be available for `Units` AOIs

## .carto file to reproduce the issue

[Populated Places (on 2016-10-04 at 10.07.48).carto.zip](https://github.com/CartoDB/cartodb/files/508148/Populated.Places.on.2016-10-04.at.10.07.48.carto.zip)

javierarce:
buffer

Closes #10001

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/area-of-influence-buffer.js
+++ b/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/area-of-influence-buffer.js
@@ -15,8 +15,10 @@ module.exports = {
       formAttrs.radius = DistanceConverter.toDistance(nodeAttrs.radius, nodeAttrs.distance);
       formAttrs.distance = nodeAttrs.distance || DEFAULT_DISTANCE;
       formAttrs.dissolved = nodeAttrs.dissolved === undefined ? false : nodeAttrs.dissolved;
+      formAttrs.isolines = nodeAttrs.isolines || 1;
     } catch (err) {
       formAttrs.radius = 100;
+      formAttrs.isolines = 1;
       formAttrs.distance = DEFAULT_DISTANCE;
       formAttrs.dissolved = false;
     }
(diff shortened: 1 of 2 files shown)


javierarce:
CR: @xavijam 
CC: @ernesmb @dgaubert

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/area-of-influence-buffer.js
```diff
@@ -15,8 +15,10 @@ module.exports = {
       formAttrs.radius = DistanceConverter.toDistance(nodeAttrs.radius, nodeAttrs.distance);
       formAttrs.distance = nodeAttrs.distance || DEFAULT_DISTANCE;
       formAttrs.dissolved = nodeAttrs.dissolved === undefined ? false : nodeAttrs.dissolved;
+      formAttrs.isolines = nodeAttrs.isolines || 1;
```
xavijam: Be careful with this, if isolines is 0, it will apply 1 by default.
```js
formAttrs.isolines = nodeAttrs.isolines != null ? nodeAttrs.isolines : 1;
```

javierarce: Good point, but if I'm not mistaken I think the buffer analysis [considers only two cases](https://github.com/CartoDB/camshaft/blob/90bda2c063295c6c1ba00945a4d19f4773efa4ee/lib/node/nodes/buffer.js#L18): > 1 and the rest (so 0 or 1). 

cc: @rochoa @dgaubert

xavijam: aha!, that's a good one. So, in this case, what I've said doesn't make sense 😂 . 💋

Issue #10110, opened 2016-10-10
Mapzen routing quota: Migration for the new fields


rafatower:
In order to implement mapzen routing quota, the following fields are to be added to `users` and `organizations` tables:
- `mapzen_routing_quota`
- `mapzen_routing_block_price`
- `soft_mapzen_routing_limit`

PR #10144, opened 2016-10-11
Migration to enable mapzen routing quota #10110

rafatower:
@ethervoid can you please review?

Any special precautions to deploy this?

### DIFF
--- a/db/migrate/20161011162345_mapzen_routing_quota.rb
+++ b/db/migrate/20161011162345_mapzen_routing_quota.rb
@@ -9,11 +9,11 @@ Sequel.migration do
   end
 
   down do
-    drop_column :users, :mapzen_routing_quota, :integer, null: false, default: 0
-    drop_column :users, :mapzen_routing_block_price, :integer
-    drop_column :users, :soft_mapzen_routing_limit, :boolean
-    drop_column :organizations, :mapzen_routing_quota, :integer, null: false, default: 0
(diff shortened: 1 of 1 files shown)


rafatower:
retest this, please

rafatower:
cc/ @jgoizueta

db/migrate/20161011162345_mapzen_routing_quota.rb
```diff
@@ -0,0 +1,19 @@
+Sequel.migration do
+  up do
+    add_column :users, :mapzen_routing_quota, :integer, null: false, default: 0
```
jgoizueta: Could we do without non-null and default values? That would make the migration much lighter.
(see  https://github.com/CartoDB/cartodb-management/wiki/Deploy-information#warn-devops-about-migrations)
If we remove that we'd need to make the code resilient to this, defaulting to 0 in each use of the quota; any idea how much work that'd be? would it be worthwhile?

rafatower: I don't know at DB level how that would make the migration simpler, probably I'm missing something. You mean because of the `not null` constraint?

jgoizueta: Both the not null and default value constraints will be costly on the large users table. both are on the list of things to avoid in large production tables in the link above.

PR #10149, opened 2016-10-11
Persist node geometry when it changes, not when the analysis is saved

### LINKED ISSUE #10143: Geometry type is not being fetched when there is an analysis applied
Discovered by @saleiva

xavijam:
Basically we used to make a request asking for the geometry for those layers which has node analysis when the Builder started, but we are not doing that anymore so we have to "store" the geometry in order to "draw" the geometry type in the layers-view.

Fixes #10143.
cc @saleiva
CR: @nobuti


nobuti:
LGTM 👍

xavijam:
Merging 🚀

PR #10162, opened 2016-10-13
Add error trace to legend migrator

gfiorav:
CR @javitonino


lib/carto/legend_migrator.rb
```diff
@@ -17,6 +17,13 @@ def build
                  title: legend_title,
                  type: new_type,
                  definition: new_definition)
+    rescue => exception
+      CartoDB::Logger.debug(message: 'Carto::LegendMigrator: couldn\'t migrate',
```
javitonino: I feel like this makes more sense in the caller. I mean, if there is a validation error saving the model, this won't catch it, will it?

Same with the default return below this, that will fail validation, right?

javitonino:
I am not sure if this will catch what you are trying to debug. Is it an error when initializing the legend model, or when saving it?

lib/carto/legend_migrator.rb
javitonino: :+1: I thought this was a FactoryGirl `.build` somehow and I got confused.

PR #10171, opened 2016-10-13
Update deep-insights.js & cartodb.js

### LINKED ISSUE #10159: Carto parser fails with conditionals in autostyling
When set autostyling, the maps api fails in some cases like this example:

```
#layer {
 [mapnik-geometry-type=point] {
    marker-fill: #FF6600;
    marker-opacity: 1;
    marker-width: 12;
    marker-line-color: white;
    marker-line-width: 3;
    marker-line-opacity: 0.9;
    marker-placement: point;
    marker-type: ellipse;marker-allow-overlap: true;
  }
 }
```

With the autostyling it is changed to this:

```
#layer { polygon-fill: ramp([type], (#7F3C8D, #11A579), ('FRANCE', 'ESPAÑA')); 

 [mapnik-geometry-type=point] {
    marker-opacity: 1;
    marker-width: 12;
    marker-line-color: white;
    marker-line-width: 3;
    marker-line-opacity: 0.9;
    marker-placement: point;
    marker-type: ellipse;marker-allow-overlap: true;
  }
 }
```

If you remove the conditional [] in the initial styles it seems to work and if you copy and rewrite the variable with the devtools it works too with exactly the same string.

Could it be an encoding error?

donflopez:
This PR fix #10159 tested in staging

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "cartodb-ui",
-  "version": "4.3.20",
+  "version": "4.3.22",
   "dependencies": {
     "backbone": {
       "version": "1.2.3",
@@ -9,7 +9,8 @@
     },
     "backbone-forms": {
       "version": "0.14.0",
-      "from": "backbone-forms@0.14.0"
+      "from": "backbone-forms@0.14.0",
+      "resolved": "https://registry.npmjs.org/backbone-forms/-/backbone-forms-0.14.0.tgz"
     },
     "backbone-model-file-upload": {
       "version": "1.0.0",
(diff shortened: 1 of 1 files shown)


npm-shrinkwrap.json
```diff
@@ -9,7 +9,8 @@
     },
     "backbone-forms": {
       "version": "0.14.0",
-      "from": "backbone-forms@0.14.0"
+      "from": "backbone-forms@0.14.0",
+      "resolved": "https://registry.npmjs.org/backbone-forms/-/backbone-forms-0.14.0.tgz"
```
xavijam: Could you check why we are changing those resolved attributes?

donflopez: I saw it with @rochoa and it only add the resolved attribute, he told me that looks good because it uses the same version of the package, maybe the resolved attr is added by another version of npm (I have `2.14.0`)?

xavijam: Fair enough!

PR #10172, opened 2016-10-13
Use default cartocss for geometry type

### LINKED ISSUE #9769: Add proper CartoCSS and form styles for tables with 'geometry' as geometry type
For CartoCSS, we should add all styles, just in case. Something like:

```css
#layer['mapnik::geometry_type'=1] {
  marker-width: 7;
  marker-fill: #FFB927;
  marker-fill-opacity: 0.9;
  marker-line-color: #FFF;
  marker-line-width:  1;
  marker-line-opacity: 1;
  marker-placement: point;
  marker-type: ellipse;
  marker-allow-overlap: true;
}
#layer['mapnik::geometry_type'=2] {
  line-color: #3EBCAE;
  line-width: 1.5;
  line-opacity: 1;
}
#layer['mapnik::geometry_type'=3] {
  polygon-fill: #374C70;
  polygon-opacity: 0.9;
  polygon-gamma: 0.5;
  line-color: #FFF;
  line-width: 1;
  line-opacity: 0.5;
  line-comp-op: soft-light;
}
```

For the style form, I'd go for the point one.

gfiorav:
Fixes #9769

Please CR @javitonino

### DIFF
--- a/spec/lib/carto/styles/geometry_spec.rb
+++ b/spec/lib/carto/styles/geometry_spec.rb
@@ -11,24 +11,24 @@ module Carto
           "  marker-width: 7;\n"\
           "  marker-fill: #FFB927;\n"\
           "  marker-fill-opacity: 0.9;\n"\
-          "  marker-line-width: 1;\n"\
           "  marker-line-color: #FFF;\n"\
+          "  marker-line-width: 1;\n"\
           "  marker-line-opacity: 1;\n"\
           "  marker-placement: point;\n"\
           "  marker-type: ellipse;\n"\
           "  marker-allow-overlap: true;\n"\
           "}\n"\
           "#layer['mapnik::geometry_type'=2] {\n"\
-          "  line-width: 1.5;\n"\
           "  line-color: #3EBCAE;\n"\
+          "  line-width: 1.5;\n"\
           "  line-opacity: 1;\n"\
           "}\n"\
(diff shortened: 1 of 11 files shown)


spec/lib/carto/styles/geometry_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+          "  marker-line-width: 1;\n"\
+          "  marker-line-color: #FFF;\n"\
+          "  marker-line-opacity: 1;\n"\
+          "  marker-placement: point;\n"\
+          "  marker-type: ellipse;\n"\
+          "  marker-allow-overlap: true;\n"\
+          "}\n"\
+          "#layer['mapnik::geometry_type'=2] {\n"\
+          "  line-width: 1.5;\n"\
+          "  line-color: #3EBCAE;\n"\
```
javitonino: I think this is going to fail with the property reordering.

javitonino:
I would add a comment to this class explaining the magic in a sentence. Something like:
`For multi-geometry tables. Generates a default CartoCSS for all possible geometry types`

Otherwise, great job! :clap: It's awesome to see how easy and natural this change feels, due to the previous design. I :heart: 

I left a comment about tests, I think they will fail due to that last-minute commit. But merge when :green_apple:

gfiorav:
Thanks! you're right!! I updated the tests

Thanks 😄

PR #10175, opened 2016-10-13
Adds "false" label instead of Null for boolean columns

### LINKED ISSUE #10154: 'false' boolean values are being treated as nulls when creating a category map
Steps to reproduce:

1. Import this dataset: https://team.carto.com/u/eschbacher/tables/users_copy_1/public?redirected=true
2. Create a map from it
3. Make sure the `is_outlier` column is boolean
4. Style by value off of the `is_outlier` field
5. Notice that the suggested values are `"true"` or `Null` instead of true or false (there are no null values in that column):

screen shot 2016-10-12 at 10 07 33

this same bug seems to break a 'torque category' map:

1. import that data
2. create an animated map
3. style by the `is_outlier` field
4. observe that the map fails:

screen shot 2016-10-12 at 10 48 11

replacing the true/false values with 1s or 0s fixes the problems and the map works as expected, although the legend does not load with the values.

cc @xavijam

javierarce:
Closes #10154 

screen shot 2016-10-13 at 17 06 59


xavijam:
Green light!?
:)

javierarce:
Green light now. 

[This change](https://github.com/CartoDB/cartodb/pull/10175/commits/8fabb87644ef66234e8d860b585ee7a897fe6181) is the important bit. `_.uniq` was considering both `null` and `false` to be equal, so they were being removed from the array. I've also improved the `quote` method so it returns the values `true`, `false` & `null`, and not `"true"`, `null` &`null` 

This will be the result now: 

screen shot 2016-10-19 at 16 34 11

xavijam:
Ole! ok, 💃

PR #10190, opened 2016-10-14
Accept builder_enabled from Central

javitonino:
Migration plan here: #10187


javitonino:
Please CR @juanignaciosl

PR #10192, opened 2016-10-14
Distinguish between Builder and Editor for event_origin

### LINKED ISSUE #10169: Identify if events are sent from builder or from editor
All events sent from carto app have an attribute 'event_origin' to identify the app which is launching them. At present, this attributed is always hardcoded as 'Editor'. 

We need to distinguish between events sent from new Buider and events sent from old Editor and fulfil this 'event_origin' attribute accordingly. This will allow us to get specific stats about new builder usage.

cc'ed @gfiorav @xavijam

gfiorav:
Fixes #10169

### DIFF
--- a/lib/carto/tracking/formats/segment.rb
+++ b/lib/carto/tracking/formats/segment.rb
@@ -58,6 +58,7 @@ module Carto
           {
             username: @user.username,
             email: @user.email,
+            event_origin: @user.force_builder? ? 'Builder' : 'Editor',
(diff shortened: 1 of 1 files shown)


gfiorav:
Please CR @javitonino 

cc/ @jmartinter

javitonino:
This works if we want dashboard events to count as whatever tool the user is configured to be using :+1:

gfiorav:
Yup! That's the idea 👍

PR #10197, opened 2016-10-14
Fixing problems with null values from the fill component and torque categories

### LINKED ISSUE #10113: sql_wrap in Builder does not include case for NULL
STR:
 1. Import a dataset with a column with some NULL values
 2. Create map with that dataset
 3. Style the layer: Animated, fill by value based on the column with NULLs

sql_wrap looks like:
`"select *, (CASE WHEN "conversion" = 0 THEN 1 WHEN "conversion" = 1 THEN 2  END) as value FROM (<%= sql %>) __wrapped"`
It should be
`"select *, (CASE WHEN "conversion" = 0 THEN 1 WHEN "conversion" = 1 THEN 2 WHEN "conversion" IS NULL THEN 3 END) as value FROM (<%= sql %>) __wrapped"`

xavijam:
Fixes #10113.
CR: @javierarce

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/form-components/editors/fill/input-color/input-categories/input-color-categories.js
+++ b/lib/assets/javascripts/cartodb3/components/form-components/editors/fill/input-color/input-categories/input-color-categories.js
@@ -133,7 +133,7 @@ module.exports = CoreView.extend({
     }).slice(0, MAX_VALUES).sort();
 
     if (this.model.get('attribute_type') !== 'number') {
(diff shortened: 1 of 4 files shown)


javierarce:
:+1: Once it's merged I'll merge into https://github.com/CartoDB/cartodb/pull/10175 to fix the boolean representation.

xavijam:
Ok @javierarce, fixed failing tests 💋 .

xavijam:
Frontend tests passed, merging! 🚀

PR #10198, opened 2016-10-14
Fine tune layer selector

### LINKED ISSUE #10093: Legends should hide when their layer is hidden
Right now, if I hide a layer, the legend keeps on showing. 

screen shot 2016-10-07 at 17 45 39

The expected behaviour is that whenever I hide a layer, the legends of that layer should hide as well.

### LINKED ISSUE #10109: Legend does not fade out when disabling view from layer list
### STR

* Add a legend to a layer via changing its style.
* Disable layer view. The layer selector checkbox is disabled, but the legend does not fade out (as you can check in this [embed](https://solutionscdb.carto.com/u/cdbsol-admin/builder/9c060f60-8ee0-11e6-b675-0ee66e2c9693/embed)). 
* In addition, when turning on the checkbox on the layer selector, the layer continues to be on "non-view" status. So it looks like the layer selector and the layer views are not quite well connected.

nobuti:
This PR fixes #10093 and fixes #10109 on builder side.

layer-selector

It depends on https://github.com/CartoDB/cartodb.js/pull/1447

CR @xavijam

### DIFF
--- a/lib/assets/javascripts/cartodb3/deep-insights-integrations.js
+++ b/lib/assets/javascripts/cartodb3/deep-insights-integrations.js
@@ -94,7 +94,11 @@ var F = function (opts) {
   // In order to sync layer selector and layer visbility
   this._getLayers().on('change:visible', function (layer, visible) {
     var layerDefModel = opts.layerDefinitionsCollection.findWhere({id: layer.id});
-    layerDefModel && layerDefModel.set({visible: visible});
+    if (layerDefModel) {
+      if (layerDefModel.get('visible') !== visible) {
+        layerDefModel.save({visible: visible});
+      }
+    }
   }, this);
 
   VisNotifications.track(this._vis());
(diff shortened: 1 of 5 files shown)


xavijam:
LGTM! 🇪🇸

lib/assets/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -87,6 +91,16 @@ var F = function (opts) {
 
   this.visMap().on('change', _.debounce(this._saveStateDefinition.bind(this), 500), this);
 
+  // In order to sync layer selector and layer visbility
+  this._getLayers().on('change:visible', function (layer, visible) {
```
xavijam: Don't we need to add this bind each time a layer is added?
Unbind when remove too, of course.

nobuti: We are binding on the collection.

xavijam: 🇪🇸

PR #10199, opened 2016-10-14
Removed problem with several requests to layer endpoint

### LINKED ISSUE #10195: Duplicated maps/v1/xxx/layers/yyy calls first time changing style color
screen shot 2016-10-14 at 14 31 28

STR

1. Open dev console and filter by XHR and "maps"
2. Open style dialogue in layer
3. Change color

Expected result:

There is a request to the layers endpoint

Actual result:

There are 2 requests, the first one appears "cancelled" **but it's actually reaching the controller**. It's creating problems when adding hooks at the controller.

cc/ @xavijam

xavijam:
1. We have this listener [here](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/editor/style/style-content-view.js#L129), and it was being thrown twice the first time a user set a new style, because we had those set in the `onStyleChange` method.
2. We were saving the analysis definition each time this method was called, unnecessarily, because `saveLayer` within `userActions` Class was saving the layer and the analyses related to the layer.

Fixes #10195.
CR: @nobuti 
cc @gfiorav


xavijam:
Passing frontend tests 🚀 , ❤️ , 🌞 .

PR #10207, opened 2016-10-17
Correctly generate empty mapcaps collection

### LINKED ISSUE #10206: Map appears always as "Published a few seconds ago"
- Create a map of points, for example.
- Check that publish info says that it was created several seconds ago, :+1:.
- Wait at least one minute and reload.
- It keeps saying "Published a few seconds ago" 😱 .

*Checked in production.

cc @juanignaciosl 
And it is likely related with https://github.com/CartoDB/cartodb/issues/10201.

javitonino:
Fixes #10206

With no mapcap, it incorrectly generated `[{}]`. Now it generates `[]`.


javitonino:
Please CR @juanignaciosl

PR #10213, opened 2016-10-17
Update cartodb.js dependency.

### LINKED ISSUE #10182: Legends not working with negative values
Legends in the Builder do not work with negative values. I loaded a dataset of values that has negative and positive values in the range. The choropleth band renders fine, however, the average line falls to the left of the choropleth bar outside of the bar itself since the average is a negative value.

nobuti:
This PR fixes #10182.

cc @xavijam

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "cartodb-ui",
-  "version": "4.3.22",
+  "version": "4.3.26",
   "dependencies": {
     "backbone": {
       "version": "1.2.3",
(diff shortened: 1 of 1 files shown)


xavijam:
DALE NENO!

xavijam:
Hey, wait, it brings several changes in torque and other packages.

nobuti:
I don't know where the others come from, it supposed it only has to update cartodb.js. Maybe @donflopez can throw some light for the torque thing. About the rest of the changes, I think we are safe, it only changes the resolve attribute, pretty sure it's the npm version.

donflopez:
Torque has this new [PR](https://github.com/CartoDB/torque/pull/266). I need to try it in staging, but the changes are for a feature that we are not using.

nobuti:
@donflopez so do I need to freeze this PR? are we safe merging it?

donflopez:
It seems to work without any problem @nobuti!

Thanks 🙂

xavijam:
:+1:, if you are sure @donflopez, let's go! 🚢 🇮🇹 .

nobuti:
This will be shipped as part of https://github.com/CartoDB/cartodb/pull/10198

Issue #10111, opened 2016-10-10
Mapzen routing quota: add changes to the models, presenters and "synchronizable"


rafatower:
In order to implement the mapzen routing, add the changes in the models, presenters and "synchronizable" concerns. As a reference example I'll use the changes applied to implement the `here_isolines_quota`.

NOTE: https://github.com/CartoDB/cartodb/issues/10110 must be deployed first.

PR #10214, opened 2016-10-17
Implement mapzen routing quota

jgoizueta:
see #10111
requires migration #10110


jgoizueta:
Note that we are using the prefix/infix `mapzen_routing` for all method and column names related to the Mapzen routing service (in consistency, e.g. with `here_isolines`), but the key `routing_mapzen` was already in use for the metrics kept in redis, and we're maintaining that usage.

PR #10227, opened 2016-10-18
Mitigating disqus problem

xavijam:
CR: @javitonino


javitonino:
Bye bye disqus, we won't miss you. :shipit:

Issue #10202, opened 2016-10-14
Don't reset styles after analysis has finished


xavijam:
### STR

- Comment unnecessary query-schema/geometry fetchs in table-view.
- Create a map of points.
- Apply a cluster of points over the layer (a1).
- Then apply group points into polygons (a2).
- Drag out node a2 below.
- Style points layer (a1), making points bigger and red.
- Change cluster node, apply 6 clusters, for example.
- Wait until analyses finish correctly.
- Then go to style tab.
- Styles have been reseted 😱 .

cc @javisantana

PR #10231, opened 2016-10-18
Changes over how to proceed when an analysis is running and when it has finished

xavijam:
Things that have changed:

- Don't keep using` QuerySchemaModel` for independent request in the Builder app (found 7) (CR: @javierarce).
- Unbind query-geometry-model changes before resetting styles if the query-geometry-model status is reseted.
- Don't make schema or geometry requests if the Node is not ready (changes over query-schema-model and query-geometry-model) (CR: @nobuti).
- Render loading within table-body-view if Node is not ready.

Let me add several reviewers, because it touches so many things and it is a bit critical.

Fixes https://github.com/CartoDB/cartodb/issues/10202.

cc @javisantana @rochoa


javierarce:
It's this ready to be reviewed? 👼

xavijam:
Yep!

xavijam:
Deploying last version in ded02.

javierarce:
👍  Reviewed the removal of the QuerySchemaModel in the Builder and tested in staging: changes LGTM.

lib/assets/javascripts/cartodb3/components/modals/add-widgets/tablestats.js
```diff
@@ -36,68 +36,72 @@ TableStats.prototype = {
...
+      sql.execute(
+        'with a as (select reltuples from pg_class where relname = \'' + table + '\'), b as (select * from pg_stats where tablename = \'' + table + '\' and schemaname = \'' + schema + '\') select * from a,b',
+        null,
+        {
+          rows_per_page: 40,
+          page: 0
+        }
+      ).done(function (data) {
+        data = data || {};
+        var rows = data.rows.map(function (r) {
```
xavijam: It will not "draw" anything, I guess. Take into account this only replaces the same what we already had.

nobuti: I bet for `map doesn't exist for undefined`

lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -125,27 +125,37 @@ module.exports = function (params) {
...
       var CHANGE_STATUS = 'change:status';
       var saveDefaultStylesIfStillRelevant;
       saveDefaultStylesIfStillRelevant = function () {
-        if (queryGeometryModel.get('status') !== 'fetched') return; // wait until fetched
+        var unbindNodeChangeStatus = function () {
+          queryGeometryModel.unbind(CHANGE_STATUS, saveDefaultStylesIfStillRelevant);
+        };
 
-        queryGeometryModel.off(CHANGE_STATUS, saveDefaultStylesIfStillRelevant); // only do this once
+        if (queryGeometryModel.isFetching()) { // Wait until fetched
```
xavijam: Probably not the best code ever, but take into account there are 3 cases, and with that one we could not cover all of them.

nobuti: Oh, I've just seen that `return` 👍

xavijam:
Travis passing, merging! 🚀

PR #10233, opened 2016-10-18
Adds bins to style by size

### LINKED ISSUE #10148: Styling by size is missing the Classification method
It should work like in by color.

javierarce:
Closes #10148

Adds a bins pane so we can style by size using the classification method:

screen shot 2016-10-18 at 12 33 41
screen shot 2016-10-18 at 12 33 53

cc: @saleiva @makella @rochoa


javierarce:
CR: @xavijam

xavijam:
The CartoCSS generation is working properly=

javierarce:
> Is the CartoCSS generation working properly?

Yes, the `makeWidthRamp` method of the style-converter.js file [was already supporting the number of bins](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/editor/style/style-converter.js#L64).

xavijam:
The future was written... scared. Ok, :es:.

rochoa:
Could we use:

```
marker-width: ramp([ccc], range(1, 20), headtails(7));
```

javierarce:
> Could we use:
> marker-width: ramp([ccc], range(1, 20), headtails(7));

Sure. Much better syntax. I'll make the change.

javierarce:
Ok, syntax updated. Have a look [here](https://github.com/CartoDB/cartodb/pull/10233/commits/599fb710e6d23072a297909bfd6cf3af32a949a1), please.

lib/assets/javascripts/cartodb3/editor/style/style-converter.js
```diff
@@ -58,16 +58,24 @@ function makeColorRamp (props, isTorqueCategory) {
...
   }
+
   if (props.quantification) {
-    c.push(props.quantification.toLowerCase());
+    var quantification = props.quantification.toLowerCase();
+
+    if (props.bins) {
+      c.push(quantification + '(' + props.bins + ')');
+    } else {
+      c.push(quantification);
```
rochoa: You can use `quantification()` and it will use the default number of bins as well.

javierarce: That's good to know… but it would produce something like this, right?

```marker-width: ramp([ccc], range(1, 20), headtails());```

If that's the case, I'd rather use the syntax without the params and just have `quantification` as it looks simpler.

makella:
i just want to say this is the most awesome thing. this is something that we haven't done properly in the past for proportional symbol maps. i can't say thank you enough for getting this in!! it is amazing.

PR #10234, opened 2016-10-18
Move binding to the editor in order to provide a good solution for geometry storage

### LINKED ISSUE #10219: PUT to analysis when opening builder
I am seeing a PUT to the analysis endpoint when opening the builder. It seems related to geometry type refresh.

STR:
 1. Create a new map
 2. Open in builder (so source analysis are created)
 3. Refresh the builder

Expected: No requests to backend
Actual: PUT to analysis endpoint

### LINKED ISSUE #10232: SQL query automatically resets when no analysis applied
STR
 1. Create a map
 2. Change the SQL query of the source analysis
 3. Refresh the page

Expected: The query is kept
Actual: The query is reset to the default one

This only happens when there is no analysis (aside from source) applied to the layer.

I tracked the issue to https://github.com/CartoDB/cartodb/blob/8382-styles-per-node/lib/assets/javascripts/cartodb3/data/layer-definitions-collection.js#L87 `o.source === sourceId`, which is always true for single-layer, no analysis maps (the single node will always be `a0`).

Removing that comparison should fix this, but I'm not sure about secondary effects. I don't think it's problematic since source analysis are now initialized by the backend. What do you think @xavijam?

xavijam:
We have found several problems after the change I introduced here https://github.com/CartoDB/cartodb/pull/10149. So, I've though that it doesn't make any sense to add the `layerDefinitionsCollection` to another collection (analysisDefinitionsCollection), if the desired behaviour should not be placed there, it should be in a parent Class. Weird thing is that Class is the parent one but I feel it should be the right place, what do you think @nobuti?

Fixes #10232.
Fixes #10219.
CR: @nobuti 
cc @javitonino


xavijam:
Ah, that's true, I like it 💋 .

xavijam:
Moved to a tracker man!

nobuti:
Sweet 👏 👏

xavijam:
Frontend tests passing, merging! 💋

PR #10251, opened 2016-10-19
Fix add button state when coming from the analysis detail view

### LINKED ISSUE #10239: Coming back from the add analysis breadcrumb shows a disabled 'Add button'
#### How to reproduce the problem:

1. Open the add analysis modal.
2. Click on the more info link of an analysis.
3. Use the breadcrumb navigation to go back.

#### What should happen?

Since there's a selected analysis (the one you got information for), the add button should be selected?

#### What happens instead?

The button is disabled.

javierarce:
Coming back from the add analysis breadcrumb shows a disabled 'Add button' instead of an enabled one. See issue #10239 to see a full description of the problem.

This PR also: 

- Removes some unnecessary variables.
- Adds specs for the AnalysisViewPane. 

Closes #10239


javierarce:
CR: @xavijam

lib/assets/test/spec/cartodb3/components/modals/add-analysis/analysis-view-pane.spec.js
```diff
@@ -0,0 +1,112 @@
+var Backbone = require('backbone');
```
xavijam: Specs, how many lives you save.

PR #10254, opened 2016-10-19
Sync GH user id with central

javitonino:


javitonino:
Please CR @juanignaciosl

PR #10264, opened 2016-10-19
Legends pixel aggregation

### LINKED ISSUE #10224: Legends for pixel aggregated maps don't work.
No data available.

It should be the same than the choropleth legend.

#SMOKE_MG31

cc/ @xavijam

### LINKED ISSUE #10247: creating a category with a numeric column creates the wrong legend
steps:

- load a table with a numeric column with 3 values (0, 1 and 2 for example)
- style by value using that column, switch to a category map

screen shot 2016-10-19 at 11 41 17 am
#SMOKE_MG29

nobuti:
This PR fixes #10224 and fixes #10247

category-map

We look at quantification in order to generate color legends.

custom-choropleth

It implements a custom choropleth legend in builder side. The form associated is the same as a regular choropleth, keeping the prefix and suffix fields. Could you confirm @saleiva?

Also, I've observed some race conditions when changing the map aggregation carousel. In order to minimize them, a buffer strategy is implemented between the saving in the backend and operating with the legend model.

CR @xavijam

### DIFF
--- a/app/models/carto/legend.rb
+++ b/app/models/carto/legend.rb
@@ -10,7 +10,7 @@ module Carto
 
     belongs_to :layer, class_name: Carto::Layer
 
-    VALID_LEGEND_TYPES = %(html category bubble choropleth custom).freeze
+    VALID_LEGEND_TYPES = %(html category bubble choropleth custom custom_choropleth).freeze
 
     serialize :definition, ::Carto::CartoJsonSerializer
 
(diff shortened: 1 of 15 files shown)


nobuti:
@jatorre please take a look at the backend part 🙏

javitonino:
I've got to ask for some tests in `spec/requests/carto/api/legends_controller_spec.rb` (guid0 made me promise I'll take care of his tests :laughing:). It should be easy enough, just follow the example of any legend payload (e.g: `custom_legend_payload`).

It's basically just adding an example JSON, and then adding it to a couple of places, but the machinery to test the result is already there.

After that, :green_apple:

PR #10265, opened 2016-10-19
fix modal

piensaenpixel:


javierarce:
On Safari the main analysis view works fine, but the info view shows a blank page :(

piensaenpixel:
@javierarce  i think is fixed, could you take a look? also we should review all the modals :(

javierarce:
👍  🎏

javierarce:
Ok, last changes fixes the issue and doesn't have side effects in other modals (tested in safari, firefox, chrome and ie11). Good to go!

piensaenpixel:
I cant't believe :)

PR #10273, opened 2016-10-20
10259 update named maps on permission change

### LINKED ISSUE #10259: sharing a map with a team mate raise instanciation error
- create a map
- keep it private and share with a team mate
- open the map as the team mate

you have access to it but it raises and instanciation error

#SMOKE_MM07

javitonino:
Fixes #10259


javitonino:
Please CR @juanignaciosl

PR #10283, opened 2016-10-20
Fixed problem managing extra params for queries

xavijam:
We were adding wrongly extra parameters using CartoDB.js SQL.
And it includes https://github.com/CartoDB/cartodb.js/pull/1460.
Fixes https://github.com/CartoDB/cartodb-management/issues/4670#event-830501189.
CR: @nobuti


xavijam:
Fixed a problem with categories in the fill component, we were sorting those and the list of the results and the legend results were different, cc @javierarce.

screen shot 2016-10-20 at 19 36 02

xavijam:
And we need this issue in order to get the same results from the Maps API -> https://github.com/CartoDB/Windshaft-cartodb/issues/581

PR #10293, opened 2016-10-21
Use CDB_QueryTablesText as it doesn't limit return length

### LINKED ISSUE #10274: Export .carto is not downloading involved datasets in analysis where it is required a second node
Like in JOINs operations.

javitonino:
Fixes #10274


javitonino:
Please CR @juanignaciosl

PR #10306, opened 2016-10-24
Fix mapzen_routing_quota default being replaced by 0

jgoizueta:
Two kind of fixes here:
* Applying to_i when a numerical quota value is needed (so the default nil is taken as 0)
* Not requiring a non null quota for users/orgs and avoiding setting a default of 0 before saving

### DIFF
--- a/app/models/carto/user.rb
+++ b/app/models/carto/user.rb
@@ -289,7 +289,7 @@ class Carto::User < ActiveRecord::Base
     if organization.present?
       remaining = organization.remaining_mapzen_routing_quota(options)
     else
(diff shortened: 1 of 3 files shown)


jgoizueta:
@rafatower please take a look at this fix

jgoizueta:
Note that before this patch organizations and users which were saved had the default mapzen routing quota (NULL) replaced by 0. So, after applying it we should revert those 0s to NULLs (there are not many of them: less than 2000 in the biggest cloud)

rafatower:
:+1:

PR #10311, opened 2016-10-24
Fixes analysis error tooltip

### LINKED ISSUE #10222: The red dot that signals an error on an analysis is missing the tooltip message
#SMOKE_MG39

Updated by @javierarce: 
Image of the original tooltip that is missing now: 

image

javierarce:
Closes #10222 

screen shot 2016-10-24 at 14 07 38

This PR fixes an issue with the error tooltip of the analyses. The problem with the original code was that each time the object `AnalysisTooltip` [was used](https://github.com/CartoDB/cartodb/pull/10311/files#diff-e9bc12d36aa74c9775586d3b37b1a9aaL27), the status message was being overwritten (we weren't creating a new instance, but using the same object every time). 

The solution is transform the object to a regular Backbone view and create a new instance when needed. This way, every tooltip is independent and linked to each analysis.

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/layers/analysis-views/analyses-tooltip-error.js
+++ b/lib/assets/javascripts/cartodb3/editor/layers/analysis-views/analyses-tooltip-error.js
@@ -1,25 +1,24 @@
 var $ = require('jquery');
+var cdb = require('cartodb.js');
+var CoreView = require('backbone/core-view');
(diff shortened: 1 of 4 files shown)


javierarce:
CR: @xavijam 
CC: @saleiva @javisantana, since the original issue was linked to #SMOKE_MG39.

PR #10319, opened 2016-10-24
Fixes an issue that prevented running an analysis after an error

### LINKED ISSUE #10280: Unable to apply/run analysis after error
## STR

1. Create a map with a points dataset.

1. Add analysis: Calculate clusters of points.

1. Apply.
screen shot 2016-10-20 at 15 55 19

1. The analysis fails due to a statement timeout limit.

## Current result

1. There is no error message associated with the failed status.

1. It is no longer possible to apply the analysis with other parameters.

1. After reloading the builder, the message failed status message appears.

1. However, it's still not possible to apply an analysis with different parameters.

## Expected result

1. I would like to see the failed status error message without reloading.
1. I would like to apply the analysis with different parameters, with and without reloading the builder.

cc @javisantana @xavijam

javierarce:
Closes #10280

This PR fixes one of the reported errors: 
> I would like to apply the analysis with different parameters, with and without reloading the builder.
> I would like to see the failed status error message without reloading.

Please, read this thread to understand the reason of the second issue and the solution to fix it.

### DIFF
--- a/lib/assets/javascripts/cartodb3/analysis-notifications.js
+++ b/lib/assets/javascripts/cartodb3/analysis-notifications.js
@@ -87,16 +87,20 @@ var AnalysisNotifications = {
     var message = _t('notifications.analysis.failed', {
       nodeId: analysisNode.get('id').toUpperCase()
     });
+
     var error = analysisNode.get('error');
-    if (error && error.message) {
-      message += ': ' + error.message;
+
+    if ((error && error.message) || analysisNode.get('error_message')) {
+      message += ': ' + ((error && error.message) || analysisNode.get('error_message'));
     }
+
     var notificationAttrs = {
       status: 'error',
       info: message,
       closable: true,
       delay: DEFAULT_DELAY
     };
+
     this._addOrUpdateNotification(analysisNode.cid, notificationAttrs);
   },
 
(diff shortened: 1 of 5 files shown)


javierarce:
@xavijam, @javisantana: any idea how to obtain a timeout error applying an analysis?

javisantana:
do `SELECT cartodb.CDB_Conf_SetConf('analysis_quota_factor', '0');` as postgres user in the database (don't do that in production)

then  `SELECT cartodb.CDB_Conf_SetConf('analysis_quota_factor', '2');` to go back to normal

javierarce:
@javisantana: cool. Thanks!

javierarce:
Ok, just learned that the backend returns the error in the key `error_message` 

screen_shot_2016-10-25_at_09_38_13

… and we are printing the error that comes in the [error property of the analysis node](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/analysis-notifications.js#L90).

xavijam:
:(, when did we change that? cc @dgaubert and @alonsogarciapablo.

dgaubert:
I think, it was: https://github.com/CartoDB/Windshaft-cartodb/commit/d0980a2872ddaa86a7ceecddd3faf35b633b12e4

javierarce:
Ok, where should we fix that?

In the frontend we could account for that and simply do:
```
if ((error && error.message) || analysisNode.get('error_message')) {
   message += ': ' + ((error && error.message) || analysisNode.get('error_message'));
}
```

javierarce:
Ok, I'll fix it with the solution I just described. Let me know if you prefer other solution.

xavijam:
Ok, fine by me @javierarce 🤗

xavijam:
Nice! :es:

PR #10335, opened 2016-10-25
fix overflow

piensaenpixel:

modals

Fix https://github.com/CartoDB/cartodb/issues/10279


javierarce:
👍

PR #10336, opened 2016-10-25
Avoid relaunching the analysis onboarding by interacting with a widget

### LINKED ISSUE #10316: onboarding dialog is shown everytime a widget to the source changes
STR:
1) add a points table
2) add a sampling analysis (10%)
3) add a wiget to the source table

everytime you filter the builder shows the onboarding dialog

#SMOKE_AN05

javierarce:
Closes #10316 

This PR fixes an issue where an analysis onboarding appeared after a widget attached to the layer with the analysis was changed.

The difficulty of solving this issue is that both widgets and user interactions can trigger the analysis [changed binding](https://github.com/CartoDB/cartodb/compare/10316-onboarding-reopen?expand=1#diff-3d74f8298f087839dd188fe931ea5322L195), and we can't really guess which of them did it by just looking to the analysis node definition. 

The solution I propose is to set a flag (`USER_SAVED`) to indicate that the user updated the analysis. This way, anytime we need to show the onboarding, we can check that this flag was set (I'm using a flag instead of setting a property to the model to avoid storing it permanently in the model).

Are you OK with this approach @xavijam & @javisantana?

### DIFF
--- a/lib/assets/javascripts/cartodb3/data/analysis-definition-model.js
+++ b/lib/assets/javascripts/cartodb3/data/analysis-definition-model.js
@@ -16,6 +16,8 @@ module.exports = Backbone.Model.extend({
   initialize: function (attrs, opts) {
     if (!opts.analysisDefinitionNodesCollection) throw new Error('analysisDefinitionNodesCollection is required');
 
+    this.USER_SAVED = false;
+
     this._analysisDefinitionNodesCollection = opts.analysisDefinitionNodesCollection;
   },
 
--- a/lib/assets/javascripts/cartodb3/data/analysis-definition-model.js
+++ b/lib/assets/javascripts/cartodb3/data/analysis-definition-model.js
@@ -16,7 +16,7 @@ module.exports = Backbone.Model.extend({
   initialize: function (attrs, opts) {
     if (!opts.analysisDefinitionNodesCollection) throw new Error('analysisDefinitionNodesCollection is required');
 
-    this.USER_SAVED = false;
+    this.USER_SAVED = false; // flag to indicate if the user saved the node and avoid triggering the onboarding model if they didn't
 
     this._analysisDefinitionNodesCollection = opts.analysisDefinitionNodesCollection;
   },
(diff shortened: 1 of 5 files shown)


javisantana:
a question, when i reload the builder i will get the messages again, right?

javierarce:
No, you shouldn't get it, since the flag won't be set then (it only gets set whenever you save the analysis again and it's not stored permanently).

javierarce:
CR: @xavijam

xavijam:
If we agree this decision (which is fine by me) I'd add a test .

javierarce:
Pushed some specs. 🙏  CR: @xavijam

lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -178,6 +178,7 @@ module.exports = function (params) {
       var nodeDefModel = analysisDefinitionNodesCollection.get(analysisFormModel.id);
 
       if (nodeDefModel) {
+        nodeDefModel.USER_SAVED = true;
```
xavijam: Not sure if we should move this "local" variable to directly analysis-definition-model and set it in the initialize, what do you think @javierarce?

javierarce: Yes. You mean put it on the `initialize` method and set it `false` by default, right?

javierarce: Here's the change: https://github.com/CartoDB/cartodb/pull/10336/commits/6ed57e6bdca13ed4cdac1a0ab606a13de90af739

lib/assets/javascripts/cartodb3/data/analysis-definition-model.js
```diff
@@ -16,6 +16,8 @@ module.exports = Backbone.Model.extend({
   initialize: function (attrs, opts) {
     if (!opts.analysisDefinitionNodesCollection) throw new Error('analysisDefinitionNodesCollection is required');
 
+    this.USER_SAVED = false;
```
xavijam: :+1:, could we add a description over the variable in order to know why we took that decision? thanks man and sorry to be picky.

javierarce:
Ready to go?

xavijam:
Fine by me

PR #10342, opened 2016-10-25
Ui notifier

piensaenpixel:
Basically small changes in the visual and group by status (loading or finish)
cc/ @nobuti 

notifier


app/assets/stylesheets/editor-3/notifier.css.scss
```diff
@@ -59,10 +97,11 @@
   margin-top: 1px;
 }
 
-// Loader styles
+// Loader styles/
+/*
```
javierarce: If this is not going to be used… could we remove it?

PR #10344, opened 2016-10-25
10321 batch queries statement timeout

rafatower:
This change uses redis as storage directly, avoiding duplication and potential de-syncing with the metadata DB. Therefore this PR does not require a migration.

Check how it is used from the batch api:
https://github.com/CartoDB/CartoDB-SQL-API/blob/861b9bb037c1c17a9c0cdd1c2c98839bc893978e/batch/job_runner.js#L65-L79

I can write a few tests for this but prefer to get a first pass through your :eyes: before going any further

@jgoizueta and/or @ethervoid can you please review?

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -20,6 +20,7 @@ require_dependency 'cartodb/redis_vizjson_cache'
 require_dependency 'carto/bolt'
 require_dependency 'carto/helpers/auth_token_generator'
 require_dependency 'carto/helpers/has_connector_configuration'
+require_dependency 'carto/helpers/batch_queries_statement_timeout'
 
 class User < Sequel::Model
   include CartoDB::MiniSequel
@@ -29,6 +30,7 @@ class User < Sequel::Model
   include DataServicesMetricsHelper
   include Carto::AuthTokenGenerator
   include Carto::HasConnectorConfiguration
+  include Carto::BatchQueriesStatementTimeout
 
   self.strict_param_setting = false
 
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -1649,6 +1649,33 @@ def new_visualizations_version
     builder_enabled? ? 3 : 2
   end
 
+  def batch_queries_statement_timeout
+    if !defined?(@batch_queries_statement_timeout)
+      timeout_str = $users_metadata.HMGET(batch_limits_key, 'timeout').first
```
ethervoid: Could this return nil and make the .first raise an exception?

app/models/user.rb
```diff
@@ -1649,6 +1649,33 @@ def new_visualizations_version
...
+      @batch_queries_statement_timeout = timeout_str.nil? ? nil : timeout_str.to_i
+    end
+    @batch_queries_statement_timeout
+  end
+
+  def batch_queries_statement_timeout=(timeout_ms)
+    unless !timeout_ms.present? || timeout_ms.to_i > 0
```
ethervoid: In case the timeout is nil, where is the default value? in sql-api?

ethervoid: Ok, that's what I saw yesterday.

app/models/user.rb
rafatower:
```
irb(main):017:0> $users_metadata.HMGET('rails:users:cdb', 'geocoder_provider').first
=> "mapzen"

```

ethervoid: Ok  cool!

rafatower:
Changes since you've last seen this:
- Created tests. Most of them for real cases I had to test manually and tweak the code.
- Moved the code to a common module than `::User` and `Carto::User` can use.
- Please Hound

@jgoizueta or @ethervoid can you please take a final look? I'll wait for tests then deploy this. Thanks!

PR #10381, opened 2016-10-26
Show affected visualizations and users in delete dataset modal

### LINKED ISSUE #10147: Deleting a dataset doesnt show the affected maps anymore on the warning modal
# SMOKE_MG04

MariaCheca:
Closes #10147 
- There will be shown a maximum of 3 affected maps and 20 affected users.
- Maps affected are visualizations owned by the user.
- Affected users are (in case of organization) those the owner of the dataset has given access to, to write or just read.

screen shot 2016-10-26 at 13 32 45
screen shot 2016-10-26 at 13 33 20
screen shot 2016-10-26 at 13 34 07
screen shot 2016-10-26 at 13 34 59

Front CR @xavijam Back CR @javitonino 
@carlostallon please review the design


app/controllers/carto/api/user_table_presenter.rb
```diff
@@ -39,6 +41,12 @@ def to_poro
           size: row_count_and_size[:size],
           row_count: row_count_and_size[:row_count]
         }
+
+        if owned_dependent_derived_visualizations && context
+          poro[:owned_dependent_derived_visualizations] = @user_table.owned_dependent_derived_visualizations.map { |v| Carto::Api::VisualizationPresenter.new(v, @current_viewer, context).to_poro }
```
javitonino: Hound is right :trollface: 

We are trying to obey the damn dog in new code, so please try to shorten this line, maybe extracting to a helper method.

app/models/carto/user_table.rb
```diff
@@ -64,6 +64,10 @@ def dependent_visualizations
       affected_visualizations.select(&:dependent?)
     end
 
+    def owned_dependent_derived_visualizations
+      affected_visualizations.map { |v| (v.has_read_permission?(self.user) && v.type == 'derived') ? v : nil }.compact
```
javitonino: To filter from a list, it's better to use `select` like this (I also changed it to use the derived? helper method):

`affected_visualizations.select { |v| v.derived? ? v.has_read_permission?(user) }.compact`

Also, as this is filtering only maps, I would rename this to `accessible_dependent_derived_maps` or something along this lines, to better reflect what it does.

And a question. This does not return public maps created by other users, maybe we should include those as well?

lib/assets/javascripts/cartodb3/helpers/mapcard-preview.js
```diff
@@ -0,0 +1,26 @@
+var _ = require('underscore-cdb-v3');
+/**
+* Mapcard preview url generator
+*/
+
+module.exports = {
+  url_for_static_map: function (maps_api_template, visualization, width, height) {
```
xavijam: Camel case please :).

lib/assets/javascripts/cartodb3/helpers/mapcard-preview.js
```diff
@@ -0,0 +1,26 @@
+var _ = require('underscore-cdb-v3');
```
xavijam: As we have talked offline, these packages are for editor, not for Builder. Please review those we are doing the same in the Builder.

app/models/carto/user_table.rb
```diff
@@ -64,6 +64,10 @@ def dependent_visualizations
       affected_visualizations.select(&:dependent?)
     end
 
+    def accessible_dependent_derived_maps
+      affected_visualizations.select { |v| (v.has_read_permission?(self.user) && v.derived?) ? v : nil }.compact
```
javitonino: Do it like so:

`affected_visualizations.select { |v| v.has_read_permission?(user) && v.derived }`

If you use `select`, you don't need `compact` (used to remove nils). The only way select can return a `nil` if it there was one in the original list (but that won't happen here).

The block passed to `select` should return true/false if you want to keep/reject the result. Of course, your ternary operator works because a Visualizations is *truthy* and nil is *falsy*, but it's better to just use the conditional here.

You can also remove `self.user` here, as `user` can not refer to anything else so it is unnecessary.

lib/assets/javascripts/cartodb3/components/modals/remove-dataset/remove-dataset-view.js
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ *  Remove dataset modal dialog
+ */
+module.exports = ConfirmationView.extend({
+  className: 'Dialog-content',
+
+  events: function () {
+    return _.extend({}, ConfirmationView.prototype.events, {
+      'click [data-event=exportMapAction]': '_exportMap'
```
xavijam: Where is the exportMap?

MariaCheca: Just something that shouldn't be there 🙃

lib/assets/javascripts/cartodb3/components/modals/remove-dataset/remove-dataset-view.js
```diff
@@ -0,0 +1,125 @@
...
+      visModel: self._visModel
+    });
+  },
+
+  _onSuccessDestroyDataset: function () {
+    window.location = this._configModel.get('base_url') + '/dashboard';
+  },
+
+  _onErrorDestroyDataset: function () {
+    this._modalModel.destroy();
```
xavijam: Why do we destroy the modal when the deletion fails?

xavijam:
@hacheka, this is your time man!, take the PR!

app/assets/stylesheets/common/maps-list.css.scss
```diff
@@ -33,6 +33,10 @@
   width: 300px;
   margin: 20px 10px 0 10px;
 }
+.MapsList-item--wRightMargins {
+  margin: 0 20px 0 0;
```
hacheka: There is no other way to express a right only margin. There is no shorthand for that using less than 4 values.

saleiva:
kudos guys!

PR #10415, opened 2016-10-27
Listen widgets changes in DI.

### LINKED ISSUE #10383: Update state when widgets change
Some time ago there was a bind so that the state would be updated every time the dashboard's widgets changed (create, update, destroy). https://github.com/CartoDB/cartodb/pull/9374/files#diff-f4a97f58cc2854a50cbd5efe6294e959R44

We have to add it again so state is updated with widgets stuff.

nobuti:
This PR fixes #10383. Also it updates DI to fix the histogram state after refresh.

cc @xavijam


lib/assets/javascripts/cartodb3/deep-insights-integration/state-tracker.js
```diff
@@ -0,0 +1,27 @@
+var _ = require('underscore');
```
xavijam: Could we not create new tracker for this one please? I rather to have all logic in the same file although it is becoming a huge thing.

xavijam: Forget it, leave it as it is.

xavijam: It is fine :es:

xavijam:
Merge it!

PR #10421, opened 2016-10-27
10396 end color bug

### LINKED ISSUE #10396: legends with custom html fail to be migrated
1) create a visualization using the editor, add a layer with points and create a choropleth (attached)
2) add a legend and then edit it (add some text)
3) enable builder and open that visualization with the builder

THe builder shows a choropleth legend but not the custom html

[migrate (on 2016-10-27 at 06.53.31).carto.zip](https://github.com/CartoDB/cartodb/files/555126/migrate.on.2016-10-27.at.06.53.31.carto.zip)

#SMOKE_MG33

gfiorav:
Fixes #10396

CR please @javitonino 

Note: A test has been removed as it no longer makes sense (non parsable colours due to malformed legends will be considered icon url paths rather than be avoided).

### DIFF
--- a/spec/lib/carto/legend_migrator_spec.rb
+++ b/spec/lib/carto/legend_migrator_spec.rb
@@ -364,13 +364,6 @@ module Carto
         }
       end
 
-      it 'omits badly formatted colors' do
-        truncated = old_intensity.dup
-        truncated['items'].last['value'] = '#fatal#fatal#fatal'
-
-        @old_legend = truncated
-      end
-
       it 'migrates old custom with template to new html' do
         @old_legend = old_custom
       end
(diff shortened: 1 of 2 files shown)


spec/lib/carto/legend_migrator_spec.rb
```diff
@@ -413,10 +406,24 @@ module Carto
...
-        @old_legend = nil
 
         new_legend.type.should eq 'html'
         new_legend.valid?.should be_true
+
+        unless @old_legend['type'] == 'bubble' || @old_legend['type'] == 'custom'
+          html = new_legend.definition[:html]
+
+          # Result should have labels, colors, and icons somewhere in generated
+          # html
```
javitonino: Those 80 columns :P

PR #10424, opened 2016-10-27
save old legend and delete it to avoid remigration

### LINKED ISSUE #10410: Legends are migrated if no legends set always

gfiorav:
Fixes #10410 

Please CR @javitonino

### DIFF
--- a/app/controllers/carto/api/vizjson3_presenter.rb
+++ b/app/controllers/carto/api/vizjson3_presenter.rb
@@ -338,12 +338,12 @@ module Carto
       def migrate_old_legend
         Carto::LegendMigrator.new(@layer.id, @layer.legend).build.save
         mark_old_legend_migrated
+        @layer.legends.reload
       end
 
       def mark_old_legend_migrated
         @layer.options[:legend][:migrated] = true
         @layer.save
-        @layer.legends.reload
       end
 
       def as_torque
--- a/app/controllers/carto/api/vizjson3_presenter.rb
+++ b/app/controllers/carto/api/vizjson3_presenter.rb
@@ -335,7 +335,11 @@ module Carto
       end
 
       def migrate_legends(old_legend)
+        @layer.options[:backup_legend] = old_legend
         Carto::LegendMigrator.new(@layer.id, old_legend).build.save
+        @layer.options[:legend] = nil
+
+        @layer.save
         @layer.legends.reload
       end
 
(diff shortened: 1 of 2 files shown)


app/controllers/carto/api/vizjson3_presenter.rb
```diff
@@ -335,7 +335,11 @@ def as_data
       end
 
       def migrate_legends(old_legend)
+        @layer.options[:backup_legend] = old_legend
         Carto::LegendMigrator.new(@layer.id, old_legend).build.save
+        @layer.options[:legend] = nil
```
javitonino: `@layer.options.remove(:legend)` ?

gfiorav:
The thing is that there might already be a guy in the builder who's legends need to be migrated because we didn't back in the day or we need to do it again. I hadn't thought on version to be honest. What do you think?

javitonino:
Uhm, you are right that this offers some advantages. :+1:. Then, maybe instead of renaming `legend` -> `backup_legend`, do something like `options[:legend][:migrated] = true`, so it's easier to go back to the editor (just in case).

gfiorav:
Have a quick second look please, @javitonino and I'll merge!

javitonino:
LGTM

javitonino:
Quick thought: Maybe it is worth it to mark legends as migrated as soon as we see something in the new legend model. My scenario is that current users that have migrated a map (previous to this patch) and then delete all legends, will get a legend migration (just once) which may be annoying. But it is only one time, so...

gfiorav:
Yup, totally. Changing that, thanks!

gfiorav:
How about now?

gfiorav:
@javitonino Ok I hope it's good now!

app/controllers/carto/api/vizjson3_presenter.rb
```diff
@@ -334,8 +335,14 @@ def as_data
...
-        Carto::LegendMigrator.new(@layer.id, old_legend).build.save
+      def migrate_old_legend
+        Carto::LegendMigrator.new(@layer.id, @layer.legend).build.save
+        mark_old_legend_migrated
+      end
+
+      def mark_old_legend_migrated
+        @layer.options[:legend][:migrated] = true
+        @layer.save
         @layer.legends.reload
```
javitonino: I would move `@layer.legends.reload` to `migrate_old_legend`. Not a big deal, but seems to make more sense.

gfiorav:
Thanks, one last look.

javitonino:
:+1:

PR #10430, opened 2016-10-28
Avoid autoclosing error notifications

### LINKED ISSUE #10398: Let's start taking into account node analysis states for the different panels
- If a node is running, we should show the loading state in any panel (style, data, legends,...) except in the analysis panel, where we should show the stuff we show nowadays.
- If an analysis has failed, we should show the error state in each panel (similar to the loading one but with a red infobox at the bottom), except for the analysis panel, where we should show the error with the red dot.

cc @javisantana

javierarce:
Closes #10398 

This PR implements a new option to avoid autoclosing a notification message and enables it for the failed analyses messages, so they remain open and appear in the Builder tabs (data, style, etc.).


javierarce:
CR: @xavijam 

It turns out that the notifier spec wasn't being picked by jasmine (at least locally on my machine)

xavijam:
:+1:

PR #10442, opened 2016-10-28
10440 template aware legend migrator

gfiorav:
Ok so we have to choose. In the editor, any legend could be customized. If the user customised the legend, we have no option but to "copy and paste" the whole thing into a new place in the builder if we have to respect anything the user made. It's a shame, because I made the migrator to port non-custom old stuff to the new builder look and feel. 

So here is an example.

If we "respect" custom stuff:

screen shot 2016-10-28 at 16 38 44

If we use the "intelligent" migrator I made and drop custom labels etc:

screen shot 2016-10-28 at 16 39 48

I think that maybe we could respect custom stuff when the legend type is custom, but for bubbles, chropleths and so on we might want to "correct" them a bit. If not, it totally breaks the builder design.

What do you think?

cc/ @xavijam @javisantana @saleiva

### DIFF
--- a/spec/lib/carto/legend_migrator_spec.rb
+++ b/spec/lib/carto/legend_migrator_spec.rb
@@ -13,6 +13,49 @@ module Carto
       @layer.destroy
     end
 
+    describe('#with templates') do
+      let(:old_legend_with_template) do
+        {
+          "show_title" => false,
+          "title" => "",
+          "template" => "<h1>Manolo Escobar</h1>",
+          "visible" => true,
+          "items" => []
+        }
+      end
+
+      it 'should migrate to html for type custom' do
+        @old_legend = old_legend_with_template.merge(type: 'custom')
+      end
+
+      it 'should migrate to html for type category' do
+        @old_legend = old_legend_with_template.merge(type: 'category')
+      end
+
+      it 'should migrate to html for type bubble' do
+        @old_legend = old_legend_with_template.merge(type: 'bubble')
+      end
+
+      it 'should migrate to html for type choropleth' do
+        @old_legend = old_legend_with_template.merge(type: 'choropleth')
+      end
+
+      it 'should migrate to html for type intensity' do
+        @old_legend = old_legend_with_template.merge(type: 'intensity')
+      end
+
+      it 'should migrate to html for type density' do
+        @old_legend = old_legend_with_template.merge(type: 'density')
+      end
+
+      after(:each) do
+        new_legend = Carto::LegendMigrator.new(@layer.id, @old_legend).build
+
+        new_legend.definition[:html].should include('<h1>Manolo Escobar</h1>')
+        new_legend.type.should eq 'html'
+      end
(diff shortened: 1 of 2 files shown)


javisantana:
Good report,

@piensaenpixel could we do something with the styling as we did with infowindows?

To be clear, being smarter is going to be really a PITA and it's hard to do that, let's see if we can tweak stuff with styling

piensaenpixel:
@gfiorav could you send me the .carto?

gfiorav:
@piensaenpixel slack'd

piensaenpixel:
this is all i can do @gfiorav  @javisantana  :(

image

gfiorav:
Pretty good, @piensaenpixel ! 

No one edits custom bubbles in the way I did on that map, to be honest. It was a very hard one. The rest looks fine. I'd go on with the plan then. Good work @piensaenpixel !

piensaenpixel:
uhmm let me think about bubbles :(

piensaenpixel:
@gfiorav then can i release the styles?

gfiorav:
Even without the bubble fix it looks production ready to me so I'd say yes

gfiorav:
Since it looks like this is going forward, please CR @javitonino

spec/lib/carto/legend_migrator_spec.rb
```diff
@@ -381,10 +386,6 @@ module Carto
         @old_legend = old_bubble
       end
 
-      it 'migrates old bubble with custom labels to new html' do
```
javitonino: Why delete this test? I can create a bubble legend with custom labels and the template does not change.

gfiorav: It's duplicated with the new template tests.

javitonino: Legends with custom labels does not automatically mean they have a custom `template`. This case covers the case where the labels are strings (not numbers) which used to fail while calculating intermediate values (only in bubbles, we could extend it).

gfiorav:
As per @javisantana and @xavijam request, ping @saleiva. 

In the Editor any legend's HTML could be customised. This poses a problem when migrating edited legends to the Builder. Our current effort yields:

Before:

screen shot 2016-11-02 at 12 22 14

After:

screen shot 2016-11-02 at 12 22 44

gfiorav:
As per @javisantana and @xavijam request, ping @saleiva. 

In the Editor any legend's HTML could be customised. This poses a problem when migrating edited legends to the Builder. Our current effort yields:

Before:

screen shot 2016-11-02 at 12 22 14

After:

screen shot 2016-11-02 at 12 44 50

gfiorav:
@javitonino I brought the tests back, could you take a look?

xavijam:
So @saleiva and @javisantana, merging this in order to "improve" the current migrated legends, and then still checking stats about people with custom legends, sounds ok?

PR #10447, opened 2016-10-28
Adding an error notification when layer has errors

### LINKED ISSUE #10432: Front end error is missing when messing with two tables after applying query
### STR

* Create a map with `table1` dataset.
* Style `table1` layer based upon `field1`.
* Change query from `select * from table1` to `select * from table2`. Where `table2` does not have `field1`.
* Apply. The visualization breaks (styles, legends...). The console points out the error:

```javascript
{
error: [
"column "field1" does not exist"
],
hint: "Perhaps you meant to reference the column "table2.field2"."
}
```

But the UI does not tell you what is happening.

### Expected results

UI should show error and hint.

xavijam:
So, in short, we weren't getting the error state from the map instantiation (https://github.com/CartoDB/cartodb.js/pull/1478) and the error was not displayed in the proper layer. That's fixed, and the tooltip is appearing, but one important thing I've noticed is we should show the error in the notification, why? Because the error could belong to the Style panel but it could be "triggered" from the SQL panel, as @ramiroaznar has pointed in https://github.com/CartoDB/cartodb/issues/10432. Of course, once the layer is ok and the map is rendering again, the notification would disappear.

So, the things would look like:

screen shot 2016-10-28 at 18 29 25
screen shot 2016-10-28 at 18 33 29
screen shot 2016-10-28 at 18 33 38

Closes #10432
CR: @nobuti 

cc @saleiva and @javisantana (any smoke test related to that?)

### DIFF
--- a/app/assets/stylesheets/editor-3/layers-list.css.scss
+++ b/app/assets/stylesheets/editor-3/layers-list.css.scss
@@ -61,7 +61,6 @@
   border: 2px solid $cWhite;
   border-radius: 50px;
   background-color: $cError;
-  content: ' ';
 }
 
 .Editor-ListLayer-item--placeholder {
--- a/app/assets/stylesheets/editor-3/layers-list.css.scss
+++ b/app/assets/stylesheets/editor-3/layers-list.css.scss
@@ -51,19 +51,17 @@
   margin-top: 1px;
 }
 
-.Editor-ListLayer-item.has-error {
-  &::after {
-    display: block;
-    position: absolute;
-    top: -5px;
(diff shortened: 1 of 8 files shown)


xavijam:
@javierarce, if you can "improve" the error parsing in your next PR would be great 💋 .

javierarce:
@xavijam: I'm not sure if in this case it would be make sense to improve the error in the front. The changes would be minimal… I think the message that appears in the image works ok and it's informative for the users, so I'd leave it as it's now.

xavijam:
:+1:

PR #10457, opened 2016-10-31
Fixing broken torque-categories for string columns with more than 10 results

### LINKED ISSUE #10433: Torque-category breaks with >10 categories
### .carto file:

[torque_category (on 2016-10-28 at 10.26.23).carto.zip](https://github.com/CartoDB/cartodb/files/558153/torque_category.on.2016-10-28.at.10.26.23.carto.zip)
### STR
- Create a map from a dataset with categorical data. 
- Select 'animated' aggregation
- Style by value and select a categorical column with 10+ distinct values
- The map breaks: 

screen shot 2016-10-28 at 12 36 13
- Instantiation call (`map?config=...`)returns:  

```
"TorqueRenderer: column "value" does not exist"
```
- There is another call (`map?lzma=...`) that is returning: 

```
"TorqueRenderer: column "ja" does not exist"
```

**NOTE** that `ja` is one of the values in the categorical column, precisely the _tenth_ category

Using a column with less distinct categories works just fine

xavijam:
Looks like with the last changes over the styles rendering and how we manage the occurrences with ramps and so on, we broke the case with torque-categories for string columns, always having more than 10 results (others case).
Also fixed the problem with others for any other type of column, the `ELSE` of the query wrapper should always point to the `Others` value number, in this case `11`.

CR: @javierarce 
Fixes #10433

### DIFF
--- a/lib/assets/test/spec/cartodb3/editor/style/style-converter-fixtures.spec.js
+++ b/lib/assets/test/spec/cartodb3/editor/style/style-converter-fixtures.spec.js
@@ -702,7 +702,7 @@ module.exports = [
             attribute: 'columnA',
             attribute_type: 'number',
             range: ['#555', '#5B3F95', '#1D6996', '#129C63', '#73AF48', '#EDAD08', '#E17C05', '#C94034', '#BA0040', '#8E1966', '#6F3072'],
-            domain: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11],
+            domain: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
             opacity: 1
           },
           size: {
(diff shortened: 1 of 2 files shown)


lib/assets/test/spec/cartodb3/editor/style/style-converter-fixtures.spec.js
```diff
@@ -702,7 +702,7 @@ module.exports = [
             attribute: 'columnA',
             attribute_type: 'number',
             range: ['#555', '#5B3F95', '#1D6996', '#129C63', '#73AF48', '#EDAD08', '#E17C05', '#C94034', '#BA0040', '#8E1966', '#6F3072'],
-            domain: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11],
+            domain: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
```
xavijam: When Others is included, we don't send 11 elements in the domain anymore, so fixed that too.

xavijam:
Ping :)

PR #10458, opened 2016-10-31
More meaningful errors

### LINKED ISSUE #10444: map api errors to meaningful messages in the UI
1) set user batch timeout to 100ms
2) run an analysis that takes more than 100ms

you get this mesage:

```
Analysis B1 failed: RuntimeError: Execution of function interrupted by signal. 
```

It would be nice to translate to something meaningful for the user

Other messages:

```
Analysis B1 failed: canceling statement due to statement timeout. 
```

javierarce:
Closes #10444 

This PR makes a couple of analysis error messages more informative for the users and improves the [AnalysisNotificationErrorHandler](https://github.com/CartoDB/cartodb/pull/10458/files#diff-e6a8444de39e547a9cd30632dab795bfR1) so it can not only extracts the meaningful error message, but replace the contents of certain errors with a more readable text.

My proposal is to "translate" these two messages:

- `RuntimeError: Execution of function interrupted by signal.`
- `canceling statement due to statement timeout.`

With: 

- `The execution of the analysis timed out and was stopped.`


javierarce:
CR: @xavijam 
CC: @javisantana

lib/assets/javascripts/cartodb3/analysis-notification-error-message-handler.js
```diff
@@ -1,3 +1,5 @@
+var Messages = require('./analysis-notification-error-messages.json');
```
xavijam: We should not add those files at cartodb3 root level. We should move it to a more significance place.

javierarce: I put it there because `analysis-notifications.js` in the same path. I'll move the notification files to a better place.

xavijam:
Move both please.

javierarce:
Files moved (to `editor/layers/analysis-views/`) and improve the coverage of the spec with an extra case.

xavijam:
Perfect, thanks man 🇪🇪

PR #10461, opened 2016-10-31
retry geometry type when the_geom is not found

javisantana:


javisantana:
CR @xavijam

xavijam:
Tested in staging, looks great man, thanks! 💋 
It doesn't solve the main problem but it is part of it, merging due to tests are passing and there was only a package version change 💋

PR #10462, opened 2016-10-31
Resetting styles when a new query is applied

### LINKED ISSUE #10413: Style UI should be updated after applying a query with geom transformation
### STR
- Apply a query which transforms the layer geometry (from polygons to points) such as this one:

``` sql
select
cartodb_id,
st_centroid(the_geom) as the_geom,
st_centroid(the_geom_webmercator) as the_geom_webmercator
from
table_name
```
- Go to style.
- Style UI displays polygon styling instead of points styling.
### Expected results

Style UI should display the right geometry styling, so after applying a query such as that one, style UI should be updated.
#SMOKE_WD14
#SMOKE_WD15

xavijam:
Fixes #10413 
CR: @nobuti 

cc @ramiroaznar @saleiva @javisantana


javisantana:
the problem I see here is that style reset should not be driven by the query but by the geometry tpye and the schema.

If I have "select * from table" and then apply "select * from table limit 10" the styles are going to be reset but I don't want that.

javisantana:
In the editor we reset styles when:
- geometry type changes
- columns needed are not in the query

xavijam:
Ok about adding that logic, if so, it will take us a bit more time.

xavijam:
So, I feel we can test it tomorrow morning in staging @javitonino.
cc @saleiva @javisantana 
Also, I have to check something about legends with you @nobuti 💋 .

xavijam:
@nobuti, take a look because things have changed a little bit 😬 .

lib/assets/javascripts/cartodb3/data/user-actions.js
```diff
@@ -189,6 +142,8 @@ module.exports = function (params) {
             type: analysisFormModel.attributes['type']
           }
         });
+
+        this._resetStylePerNode(nodeDefModel, layerDefModel, false, true);
```
nobuti: what does it mean the last `true`? The method definition only defines 3 params, right?

xavijam: no, it defines 3, there is an explanation in the proper method/class. `resetQueryReady`

xavijam: You are right, I was not passing the 4th parameter to the reset-style-per-node

xavijam: ❤️

lib/assets/javascripts/cartodb3/helpers/reset-style-per-node.js
```diff
@@ -0,0 +1,110 @@
...
+
+    if (!queryGeometryModel.isFetched()) {
+      queryGeometryModel.bind(CHANGE_STATUS, saveDefaultStylesIfStillRelevant);
+    }
+
+    if (!querySchemaModel.isFetched()) {
+      querySchemaModel.bind(CHANGE_STATUS, saveDefaultStylesIfStillRelevant);
+    }
+
+    if (!queryGeometryModel.isFetching() && queryGeometryModel.canFetch()) {
```
nobuti: Any reason for not including the `isFetching` condition as part of the `canFetch` method? Just curiosity, I always see them both...

nobuti: 👍

lib/assets/javascripts/cartodb3/data/query-schema-model.js
```diff
@@ -138,6 +142,22 @@ module.exports = Backbone.Model.extend({
     this.set('status', this.get('query') ? STATUS.unfetched : STATUS.unavailable);
   },
 
+  // Basically checks if an schema is different than the current one
+  hasDifferentSchemaThan: function (schemaArray) {
+    var foundAnyColumnNotIncluded = false;
+
+    _.each(schemaArray, function (columnObj) {
```
xavijam: Applied, thanks man 💋

nobuti:
@xavijam @javierarce @saleiva I don't know if it makes sense some kind of notification to the user informing that the styles have been reset because the geometry changes / analysis applied, probably for another issue.

xavijam:
Travis tests passing, so merging guys.
cc @javitonino

PR #10470, opened 2016-10-31
Launchs onboarding on creation

### LINKED ISSUE #10463: Analysis onboarding launcher is not being displayed when a new analysis is added
## STR
1. Create a map of points.
2. Apply a new analysis of subsample, for example.
3. Analysis onboarding doesn't appear.
4. Change the percentage and select, for example 50% and apply the analysis.
5. Analysis onboarding appears.

javierarce:
Fixes #10463

The flag was being set in the wrong place (it was only set when the node already existed)

CR: @xavijam


javierarce:
CR: @xavijam

xavijam:
🇪🇪

PR #10488, opened 2016-11-02
Update install.rst

chriswhong:
Per [this recommendation](https://groups.google.com/forum/#!searchin/cartodb/liblwgeom%7Csort:relevance/cartodb/-7hdvF8KJzE/iHN1wvu2AQAJ), liblwgeom should be 2.2.2?  Using 2.1.8 causes a dependency error

### DIFF
--- a/doc/manual/source/install.rst
+++ b/doc/manual/source/install.rst
@@ -156,7 +156,7 @@ PostGIS
   .. code-block:: bash
 
     sudo apt-get install libxml2-dev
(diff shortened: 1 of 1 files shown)


Cartofante:

Cartofante:

javitonino:
Hey Chris,

I checked and indeed we need to use 2.2.2 since we upgraded to Postgis 2.2. I am merging and updating the docs right away.

Thanks for the patch!

Issue #10472, opened 2016-10-31
After editing a geometry, we should get the left panel back to the layer one (wherever the user would be before clicking on edit)


saleiva:

alonsogarciapablo:
Closed via https://github.com/CartoDB/cartodb/commit/d070f6bca46b5092648d249cb5311841e290c544

saleiva:
Not working for me in https://matatageom.carto-staging.com/builder/5b31d9b4-4bae-4fc3-874e-efc2cec65f21

alonsogarciapablo:
@saleiva it seems to be working "fine" for me (i.e: clicking back takes the left panel to latest panel the user was seeing when he/she clicked the Edit geometry button):

![back-to-the-future](https://cloud.githubusercontent.com/assets/390398/20052144/ee10d4fe-a4d2-11e6-9573-8be3fb8ab41a.gif)

In the example, if the user had clicked the layer before clicking the "Edit" and going back, it will show the Layer panel. But, if it clicked "Edit" and then Back, while viewing the list of layers, it will then see the list of layers again. That's how I understood `wherever the user would be before clicking on edit`...

Please check the gif and confirm if this is right or wrong. If it's wrong, I'd appreciate some Steps To Reproduce. Thanks @saleiva!

PR #10496, opened 2016-11-02
Method to go back to the previous layout

alonsogarciapablo:
Required for https://github.com/CartoDB/cartodb/issues/10472.

This PR implements a new method so that we can go back to the previous layout:

```
this.layoutModel.goBack();
```

Take a look at the tests for a better understanding of how it works...

I did some refactoring and `nextStep` and `prevStep` are now using `goToStep` internally.
Also:
- @matallo discovered that `nextStep` [was always going to the last step](https://github.com/CartoDB/cartodb/compare/go-back?expand=1#diff-7e62e2f46272af03a5083ad2923324a8L44) and we're fixing it here.
- `nextStep` always triggered a `positionChanged` event, even if the position didn't change. There was even [a test for this behaviour](https://github.com/CartoDB/cartodb/pull/10496/files#diff-67cd7945541b5e9eefbc127b2bc26b3cL78), but I don't quite get why (any idea @xavijam @javierarce?)

We could potentially use this in many "back" buttons that we display in the left side panel.

@xavijam would you please take look? Thanks 💋

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/stack-layout/stack-layout-model.js
+++ b/lib/assets/javascripts/cartodb3/components/stack-layout/stack-layout-model.js
@@ -1,4 +1,3 @@
-var _ = require('underscore');
 var Backbone = require('backbone');
 
 /**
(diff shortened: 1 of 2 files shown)


alonsogarciapablo:
cc: @javierarce

javierarce:
I'm not sure either :/  That trigger is captured in the `stack-layout-view.js`  

```  _onPositionChange: function (newPos, opts) {
    this._removeOldStackView();
    this._genNewStackView(_.flatten(opts));
  },
``` 

so I think that in that particular case, we just forgot to add a condition to avoid the triggering.

alonsogarciapablo:
We'll soon know if the event was indeed required... 😛 . Thanks guys!

PR #10510, opened 2016-11-02
9966 quote table names starting by number

### LINKED ISSUE #9966: Error granting SELECT to tiler user
I've spotted this exception

```
PG::Error: ERROR: syntax error at or near ".2016"
LINE 1: GRANT SELECT ON public.2016damagestest TO tileuser;
```

Error comes from [here](https://github.com/CartoDB/cartodb/blob/master/app/models/table.rb#L479)

I looks like comes from `Ghost tables: Error creating UserTable`

// @CartoDB/backend

javitonino:
Fixes #9966 and does a needed refactor of the default SQL query construction.


javitonino:
Please CR @xavijam

PR #10520, opened 2016-11-03
Migrator mapcap hofix

gfiorav:
Please CR @javitonino 

avoids migration when a layer belongs to a mapcap. https://rollbar.com/vizzuality/CartoDB/items/20903/


gfiorav:
Thanks :)

PR #10521, opened 2016-11-03
Show warning messages when there's the_geom_webmercator in the query

### LINKED ISSUE #10083: executing sql queries fail if there is no the_geom
@javisantana commented on [Fri Oct 07 2016](https://github.com/CartoDB/leapfrog/issues/7)

1) open a visualization and go to a layer
2) execute "select 1"

you look at network panel and you see sql api response error like this 

```
{"error":["column \"the_geom\" does not exist"]}
```

Form my POV when there is not the_geom or the_geom_webmercator the map tab should be hidden and show data one

For reference, the old editor in that case updates the map (shows nothing) and the legends and not changed

As side effect of this is:

```
SELECT 'SRID=3857;POINT(0 0)' as the_geom_webmercator
```

raises the same error (and in that case the map could be rendered)
#SMOKE_MG38

javierarce:
Closes #10083 

This PR:

- [x] Parses original error message and returns a more simple and informative one
- [x] Instead of showing an error notification, shows a warning one.
- [x] Shows the error in the layers view

I've also changed the name and moved the AnalysisNotificationErrorMessageHandler (uf). Now it's simply NotificationErrorMessageHandler, since it's used not only for analysis.

The error that I picked is `The column "the_geom_webmercator" doesn't exist` (notice that the screen captures are older and don't show exactly that message). Feel free to suggest a better alternative if you think is not good enough, please.

screen shot 2016-11-03 at 11 22 29
screen shot 2016-11-03 at 11 50 32
screen shot 2016-11-03 at 11 50 42


javierarce:
CR: @xavijam

xavijam:
I'd remove the loading bar when it has failed.

javierarce:
Ok, fixed [here](https://github.com/CartoDB/cartodb/pull/10521/commits/ba60582ce491d4ee9a3cffffef2a6e1aa373b72b)

javierarce:
Ok, releasing this thing.

xavijam:
Fine by me! :es:

Issue #10491, opened 2016-11-02
Empty Dataset workflow fails


saleiva:

alonsogarciapablo:
- Draw geometry buttons:
  - [ ] If QuerySchemaModel is not fetched/fetching... some sort of loader must be displayed.
  - [ ] Once the QuerySchemaModel is fetched, buttons should only be displayed if:
    - Layer doesn't has a `source` analysis
    - `source` analysis doesn't have a custom query applied.
- [ ] When the query of the source changes -> re-render buttons
- [ ] When the AnalysisDefinitionNodeModel the layer is pointing to changes (a new one is added, is removed, etc.) -> re-render buttons.

PR #10526, opened 2016-11-03
Re-render buttons to create geometries when "things" change

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb/issues/10491.

"Things" or events that cause the buttons to be re-rendered:

- The query of the source changes.
- The AnalysisDefinitionNodeModel the layer is pointing to changes (a new one is added, is removed, etc.).
- The type of geometry (queryGeometryModel.get('simple_geom')) changes.

Also, the layer panel tries to fetch the geometry type if not present after it's initialised.

I realised that there's a lot of logic in `lib/assets/javascripts/cartodb3/editor/layers/layer-content-view.js` that's related to the `change view buttons`, so perhaps it's worth extracting a new view or two for all these buttons (change view buttons + add geometry buttons). Thoughts?

@xavijam @matallo 👍  or 👎


lib/assets/javascripts/cartodb3/data/query-schema-model.js
```diff
@@ -59,7 +65,11 @@ module.exports = Backbone.Model.extend({
   },
 
   canFetch: function () {
-    return this.get('query') && !this.isFetched() && this.get('ready');
+    return this.hasQuery() && !this.isFetched() && this.get('ready');
+  },
+
+  hasQuery: function () {
+    return !!this.get('query');
```
xavijam: this.has('query') does the same

lib/assets/javascripts/cartodb3/data/query-geometry-model.js
```diff
@@ -75,7 +75,7 @@ module.exports = Backbone.Model.extend({
   },
 
   canFetch: function () {
-    return this.get('query') && !this.isFetched() && this.get('ready');
+    return this.get('query') && this.get('ready');
```
xavijam: Be careful with this change, I feel it will not break nothing currently, but it is important for the things we are implementing.

lib/assets/javascripts/cartodb3/data/query-schema-model.js
```diff
@@ -59,7 +65,11 @@ module.exports = Backbone.Model.extend({
   },
 
   canFetch: function () {
-    return this.get('query') && !this.isFetched() && this.get('ready');
+    return this.hasQuery() && !this.isFetched() && this.get('ready');
```
xavijam: If we do the change in the query-geometry-model, we should do the same here.

lib/assets/javascripts/cartodb3/editor/layers/layer-content-view.js
```diff
@@ -64,6 +64,14 @@ module.exports = CoreView.extend({
       this._firstNode.fetchTable();
     }
 
+    // Fetch geometry type if there isn't one yet
+    var queryGeometryModel = this._getQueryGeometryModel();
+    if (!queryGeometryModel.hasValue()) {
+      queryGeometryModel.once('change:simple_geom', this._renderContextButtons, this);
```
xavijam: If the simple_geom is the same, this bind will not be removed.

lib/assets/javascripts/cartodb3/data/query-schema-model.js
alonsogarciapablo: :raised_hands:

lib/assets/javascripts/cartodb3/data/query-geometry-model.js
alonsogarciapablo: Why?

lib/assets/javascripts/cartodb3/editor/layers/layer-content-view.js
alonsogarciapablo: I'm doing `this.add_related_model(queryGeometryModel);` in the following line... That'll take care of this binding when the view is cleaned... right?

lib/assets/javascripts/cartodb3/data/query-geometry-model.js
xavijam: Due to checks we need in order to reset styles or not when a new analysis is added, removed, or edited.

lib/assets/javascripts/cartodb3/editor/layers/layer-content-view.js
xavijam: That should do the trick yes.

matallo:
BOOM

PR #10527, opened 2016-11-03
Setup builder_enabled feature if environment BUILDER_ENABLED variable is set

santisaez:
It will set the `builder_enabled` attribute if the `BUILDER_ENABLED` environment variable is `true`.

I have tested and it works, but if I create a new user in the organization it doesn't work.

@javitonino can you review?

### DIFF
--- a/lib/tasks/db_maintenance.rake
+++ b/lib/tasks/db_maintenance.rake
@@ -865,6 +865,9 @@ namespace :cartodb do
         organization.display_name = ENV['ORGANIZATION_DISPLAY_NAME']
         organization.seats = ENV['ORGANIZATION_SEATS']
         organization.quota_in_bytes = ENV['ORGANIZATION_QUOTA']
+        if ENV['BUILDER_ENABLED'] == "true"
+          organization.builder_enabled = true
+        end
         organization.save
       end
       uo = CartoDB::UserOrganization.new(organization.id, user.id)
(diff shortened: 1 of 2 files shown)


javitonino:
I have been reading the linked ticket. Some thoughts:

 - We are going to change the default of builder_enabled to true before the onpremises freeze. So this change will not be too useful. Maybe we can improve this by allowing to force editor (disable builder_enabled) as well?
 - For organization users, they are created as builder_enabled = nil by default ([here](https://github.com/CartoDB/cartodb/blob/d50784b4b395d0e56ca5379bfe0286e634193195/app/models/user.rb#L235)). We should change the organization creation tasks in order to be able to set builder to true/false. This will also be changed to default to true soon.

I'd like to know why the change of plans to modify the tasks instead of just relying on the DB default.

rporres:
@javitonino: We don't want to depend on a change that is getting delayed.

As an aside note, it could be also interesting to be able to use this rake to create editor users in the near future.

luisbosque:
AFAIK and after a discussion with @juanignaciosl enabling the builder by default is something that will happen through the price plans in central. This means that a onpremise will still have the editor by default. Am I wrong?

santisaez:
I have made a small change to support `builder_enabled` management also on organizations.

javitonino:
:+1: this should work. I think it would be nice to have the possibility to also force the editor, but don't know if it would ever be used.

Issue #10351, opened 2016-10-25
Round number of nulls on DATA tab


saleiva:
We should round the number of nulls on the DATA tab of a layer in the same way we do in the widgets. Right now they show decimals so legibility is pretty bad

PR #10532, opened 2016-11-03
Null values rounded to 2 decimal positions

hacheka:
Now the NULL values on DATA tab get rounded to 2 decimal positions.

screen shot 2016-11-03 at 14 23 35

CR @nobuti


hacheka:
This PR fixes #10351

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/data/stat-view.js
```diff
@@ -79,7 +79,7 @@ module.exports = CoreView.extend({
         height: 10
       }));
 
-      this.$('.js-null').text(graph.getNullsPercentage() + _t('editor.data.stats.null'));
+      this.$('.js-null').text((graph.getNullsPercentage() * 100).toFixed(2) + _t('editor.data.stats.null'));
```
nobuti: For me, it's weird to make it fixed to `2`. For instance, according to this, if the number is 85, we would end having `85.00%`. I think we should follow the same rule as we do in deep-insights or cartodb.js:
```
var abs_v = Math.abs(value);

if (abs_v > 10) {
    p = 1;
} else if (abs_v > 0.01) {
    p = Math.min(Math.ceil(Math.abs(Math.log(abs_v) / Math.log(10))) + 2, 2);
}

value = value.toFixed(p);
var m = value.match(/(\.0+)$/);
if (m) {
  value = value.replace(m[0], '');
}
return value;
```

What do you think?

saleiva: THAT RULE

hacheka: The attached code has been added as a helper function in `utils.js` at the commit https://github.com/CartoDB/cartodb/commit/f88fdd254e6381c5827b2874240c59d06fe5a9bd
Then it has been applied to the three different figures that show up at the element.

- NULL values rounded up to 2 decimal positions. If finished with .00 or .0 we don't show decimals. Same happens with TOP CAT percentage. Before this PR, it showed 100.00%. Now, it's 100%

screen shot 2016-11-03 at 17 20 17

- For boolean columns, there was a bug. If we didn't have the percentage of `true`, we showed NaN%. Now, we don't show that if no data is available. Perhaps we should investigate why we don't have that number instead of hiding it.

screen shot 2016-11-03 at 17 20 08

PR #10536, opened 2016-11-03
Add autoStyle flag from owner to embeds

javitonino:


javitonino:
CR @xavijam @juanignaciosl

PR #10555, opened 2016-11-04
Migration to enable builder by default

javitonino:


javitonino:
Please take a look @juanignaciosl

javitonino:
I disconnected Central from my instance and tested without it for a better check. Normal users (via user creation scripts) default to true with this change, and organization users default to nil in both flows (unchanged), they inherit from organization.

Also, from Central, this works fine (the value from the request is taken, the DB default is not used).

PR #10557, opened 2016-11-04
Making nodes changes silently in order to avoid unnecessary changes in CARTO.js before all modifications have finished.

### LINKED ISSUE #10366: Deleting a layer incorrectly resets styles
# STR
1. Import [carto.zip](https://github.com/CartoDB/cartodb/files/552918/carto.zip). You get something like:

screenshot_20161026_115028
1. Drag A1 to a new layer, you will see the points + the centroids

screenshot_20161026_115044
1. Delete layer B
# Result

The recently deleted layer (points) is displayed

screenshot_20161026_115058

After refreshing the builder (or touching some styles) it goes back to working as expected.
# Expected

The original layer (centroids) is displayed

cc @xavijam

### LINKED ISSUE #10469: ghost node error
**`.carto` file to reproduce:** [ghost node (on 2016-10-31 at 17.02.56).carto.zip](https://github.com/CartoDB/cartodb/files/562430/ghost.node.on.2016-10-31.at.17.02.56.carto.zip)

### STR
- create a map with a point layer **[A]** + polygon layer **[B]**
- add a distance AOI analysis to **[A]** 

screen shot 2016-10-31 at 18 10 35

- add an Intersection analysis to **[B]**. Select **[A1]** as Intersect layer

screen shot 2016-10-31 at 18 09 35

- the layers panel should look like: 

screen shot 2016-10-31 at 18 00 27

**NOTE:** Use the `.carto` attached to get directly to this point

- Drag **[A1]** node out to get a different layer
- The UI 'breaks' and we cannot do anything

screen shot 2016-10-31 at 18 17 04

- Reload the page. The UI is lost, the map is not rendered and we cannot do anything else

- We can see several errors in the console: 
```
sourceId 'c1' doesn't exist
```

```
Error: Map instantiation failed(…)
```

```
Uncaught TypeError: Cannot read property 'queryGeometryModel' of undefined(…)
```

### LINKED ISSUE #10542: "Cannot read property 'cid' of undefined" after removing a layer created from another layer
**str**

1. add empty map
2. add layer
3. remove layer

screen shot 2016-11-04 at 10 40 21

xavijam:
When a layer is removed, If there is any node in that layer referenced in other layer, we try to move those nodes to that layer instead of removing them.

In that operation we change the source ids of the nodes and that was triggering changes in CARTO.js before all the stuff was correctly modified. That was the reason about the problem described in #10366, where the map was displaying something wrong (but not failing 💃 ).

So, the thing is we needed to avoid those "triggers".

Fixes #10366.
Fixes #10542.
Fixes #10469.

CR: @nobuti 
cc @alonsogarciapablo


xavijam:
_PS: I have to check if there is any side effect (I don't think so) and add some more tests_

xavijam:
Changed the solution. I've realized that CARTO.js analyses should be changed if id of a node has changed.

xavijam:
Another thing fixed here: https://github.com/CartoDB/cartodb/issues/10469 💃

nobuti:
Great job 💯

xavijam:
Being deployed in ded02.

xavijam:
Merging guys, already tested in staging and local. 🚀  💃

lib/assets/javascripts/cartodb3/data/query-geometry-model.js
```diff
@@ -78,6 +78,10 @@ module.exports = Backbone.Model.extend({
     return this.get('query') && this.get('ready');
   },
 
+  shouldFetch: function () {
```
nobuti: ✨

PR #10560, opened 2016-11-04
Update cartodb.js to get rid of old layer_selector.

nobuti:
This way we remove the old layer selector in the Builder. The migration is still pending of development.

cc @xavijam


xavijam:
Fine by me 🇳🇿  , but check it in staging please 💋 .

PR #10573, opened 2016-11-07
10482 fixes on sequential lines connect

### LINKED ISSUE #10482: Fixes on Sequential Connect with lines Analysis
- [ ] Order shouldn't be optional, it should be mandatory, as well as DESC or ASC.
- [ ] We should switch the order of the fields to make the GROUP BY one the last one since its optative.

hacheka:
Fixes #10482 

CR @javierarce


hacheka:
This is how it looks after this PR.

screen shot 2016-11-07 at 10 41 01

PR #10578, opened 2016-11-07
Send aggregation info in the Group points into polygons analysis

### LINKED ISSUE #10426: Points to polygon analysis always runs COUNT aggregation
In the point to polygon analysis, value aggregation always runs as _COUNT_, even if other method is specified.

convex_hulls_bug

javierarce:
Fixes #10426 

This PR fixes an issue where the Group points into polygon analysis wasn't sending the aggregation information. 

The reason of the problem was [the omission of the aggregate properties](https://github.com/CartoDB/cartodb/pull/10578/files#diff-d725fef39a6d28774892d28da2dcdad2L28) for each analysis. We were effectively getting rid of the `aggregate`, `aggregation`, `aggregation_column` properties, when we just needed to remove the first one (which is used internally by the form and is not expected by camshaft).


javierarce:
CR: @xavijam

xavijam:
🇪🇪

PR #10587, opened 2016-11-07
Disabling Backbone cached sync for visualizations endpoint (dashboard and editor)

### LINKED ISSUE #10305: Builder map dashboard stuck on Shared Maps
When looking at the Maps dashboard, some users are experiencing the inability to change tabs from Shared Maps to My Maps. 

When the user clicks on My Maps – Shared Maps is the tab it is stuck on – the interface flickers for .5 second and does nothing. This gives the user the impression that their maps 'went away'.

 Logging at the interface in superadmin shows a dashboard functioning as expected, with the ability to move between tabs. 

This behavior has been seen in the dashboards of users at NYU (using Chrome and on NYU's org account and on Builder) and in Carto's @bspiro. 

screen shot 2016-10-23 at 3 48 16 pm

xavijam:
CR: @javisantana
Fixes #10305

### DIFF
--- a/lib/assets/javascripts/cartodb/app.js
+++ b/lib/assets/javascripts/cartodb/app.js
@@ -30,7 +30,6 @@ var i18n = {
     throw new Error('A "url" property or function must be specified');
   };
 
-
   // backbone.sync replacement to control url prefix
   Backbone.originalSync = Backbone.sync;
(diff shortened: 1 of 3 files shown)


javisantana:
:es: :(

xavijam:
It will be back with more energies man.

PR #10588, opened 2016-11-07
Merge / Join columns from 2nd layer analysis doesn't store the right geometry source

### LINKED ISSUE #10377: Join layers analysis resets geometry source automatically
Better explained with a video (look at the "geometry from" option):

out11
1. Create a join analysis
2. Join by cartodb_id (probably not relevant)
3. Accept the default geometry (ec_kvy in this case, a point layer) and click apply

The default geometry switches to the second dataset (barrios, polygons), although the map shows the results as if the first one (ec_kvy, points) was selected.

cc @xavijam

javierarce:
Fixes #10377

This PR fixes a problem that prevented setting the right source for the source geometry in the merge analysis. 

[The condition that chose which source to parse](https://github.com/CartoDB/cartodb/compare/10377-geometry-source?expand=1#diff-3749339bf2595bbb1d42aeee87d37481) didn't worked for the 'left_source' value.

This PR also adds general specs for this particular analysis.


javierarce:
CR: @nobuti

PR #10595, opened 2016-11-07
Do not set feature definition attributes when row is saved

alonsogarciapablo:
This was preventing attributes from being marked as changed. Example:

Before:

```
feature.set('foo', 'bar');
feature.on('save', function () {
  console.log(feature.hasChanged('foo')); // -> false
});
feature.save();
```
After:

```
feature.set('foo', 'bar');
feature.on('save', function () {
  console.log(feature.hasChanged('foo')); // -> true
});
feature.save();
```

@matallo I tested in the UI and everything seems to fine okay. 👍 ?


matallo:
🚀

PR #10597, opened 2016-11-07
Modified editor attrs for the fill simple style

### LINKED ISSUE #10358: Minimum Symbol Size for Points > BY VALUE is too small
The default minimum symbol size for 'bubble' map ranges is too small.

If possible, please set the minimum to 5 px. 2px is too small to see on most maps.

Related discussion:
https://github.com/CartoDB/cartodb/issues/8853
# SMOKE_MG20
# SMOKE_MG21

hacheka:
Fixes #10358 

For the simple fill style, now the minimum size is 5 instead of 2.

This way, the bubbles are shown **now** like this ->

screen shot 2016-11-07 at 18 43 55

The former value (2px) showed the map like this ->

screen shot 2016-11-07 at 18 44 31

CR @javierarce

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/style/style-form/style-form-components-dictionary.js
+++ b/lib/assets/javascripts/cartodb3/editor/style/style-form/style-form-components-dictionary.js
@@ -191,6 +191,10 @@ module.exports = {
       animationType: params.animationType
     });
 
+    if (params.styleType === 'simple') {
+      editorAttrs.size.min = 5;
+    }
+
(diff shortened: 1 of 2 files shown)


makella:
Oh! one comment here

The minimum symbol size for simple styling should stay what it was (I believe it was 2 or 1.5).

This is only when someone styles by value with symbol size!

cc @saleiva

makella:
5 pts is NOT a good minimum symbol size for when a user adds points. Imagine if there are hundreds or thousands of points, the map will look really bad!!

PR #10608, opened 2016-11-08
Adds hover tip for tracts in the AOI analysis

### LINKED ISSUE #10352: ADD hover tip for tracts parameter on AOI Analysis
To explain what they are and how they work

javierarce:
Closes #10352

This PR adds an explanation about the tracts in the Area of Influence analysis.

screen shot 2016-11-08 at 09 31 37

As suggested by @ohasselblad, I've also swapped the `distance` and `units` labels to make the form more understandable and correct:

Before the change:

screen shot 2016-11-08 at 09 35 25

After the change:

screen shot 2016-11-08 at 09 35 01


javierarce:
CR: @xavijam

xavijam:
🇳🇿

Issue #10473, opened 2016-10-31
Numeric fields should accept any char, but show a validation error when having one that is not supported


saleiva:
In the same way we do everywhere else.
Right now is pretty hard to understand that only numbers are accepted with the actual approach

saleiva:
Not working correctly for me in https://matatageom.carto-staging.com/builder/5b31d9b4-4bae-4fc3-874e-efc2cec65f21

When entering letters on numeric fields, it does't give the error, and after saving it sets the value to null

saleiva:
With a different dataset, SAVE button only gets activated after typing a letter, and when saving a numeric input (even if it has letters on it) letters get removed and number is saved. We should show the error in the form like in the analysis ones.

PR #10619, opened 2016-11-08
edit feature: validate

matallo:
close https://github.com/CartoDB/cartodb/issues/10473


lib/assets/javascripts/cartodb3/editor/layers/edit-feature-content-views/edit-feature-action-view.js
```diff
@@ -35,10 +39,22 @@ module.exports = CoreView.extend({
       this.model.set('hasChanges', true);
     }, this);
     this.add_related_model(this._featureModel);
+    this._geometryFormModel.bind('validate', function (isValid) {
```
alonsogarciapablo: I'd make `this.model` a bit more "smart" and move these logic in there.

alonsogarciapablo: That model would hold references to the featureModel, the geometryFormModel, and the attributesFormModel, and listen to changes on those.

lib/assets/javascripts/cartodb3/components/form-components/editors/number.js
```diff
@@ -107,16 +107,12 @@ Backbone.Form.editors.Number = Backbone.Form.editors.Base.extend({
 
   _onInputKeyUp: function () {
     var value = this.$('.js-input').val();
-    var isFloatNotCompleted = new RegExp(/\d+\.$/);
```
alonsogarciapablo: I'm a little bit concerned with this change.... Inputs of type number will now be accepting digits and letters platform wise. Of course, [we're validating the input in this particular case](https://github.com/CartoDB/cartodb/pull/10619/files#diff-85b0501a50360453b160be78d0b4e399R56), but what about other forms in the app? Shouldn't all numeric fields get "number" validator by default?

@xavijam, I'd love to hear your thoughts about this...

lib/assets/javascripts/cartodb3/editor/layers/edit-feature-content-view.js
```diff
@@ -119,15 +122,47 @@ module.exports = CoreView.extend({
...
+      this._geometryFormModel = new EditFeatureGeometryPointFormModel({
+        lng: geojson && geojson.coordinates[0],
+        lat: geojson && geojson.coordinates[1]
+      }, {
+        model: this.model,
+        featureModel: this._featureModel
+      });
+    } else {
+      this._geometryFormModel = new EditFeatureGeometryFormModel({
+        the_geom: this._featureModel.get('the_geom')
```
alonsogarciapablo: I don't think we're using `the_geom` inside of `EditFeatureGeometryFormModel`, so I think we don't need to pass it here.

lib/assets/javascripts/cartodb3/components/form-components/editors/base.js
```diff
@@ -6,9 +6,21 @@ var ESC_KEY_CODE = 27;
 Backbone.Form.editors.Base = Backbone.Form.editors.Base.extend({
 
   _setOptions: function (opts) {
+    var validators = [];
+
+    if (this.options && this.options.validators) {
+      validators = validators.concat(this.options.validators);
+    }
+
+    if (opts.schema && opts.schema.validators) {
```
alonsogarciapablo: My only question here... do we want to concat the validators from the schema first and leave the default ones at the end? This way, if we have a numeric field with `validators: [ 'requited' ]`, 'required' validation will happen first.

lib/assets/javascripts/cartodb3/editor/layers/edit-feature-content-view.js
```diff
@@ -77,20 +80,39 @@ module.exports = CoreView.extend({
...
+        lat: geojson && geojson.coordinates[1]
+      };
+    } else {
+      attrs = {
+        the_geom: this._featureModel.get('the_geom')
+      };
+    }
+
+    this._geometryFormModel.set(attrs);
+    this._geometryFormModel.trigger('updateFeature', attrs);
```
alonsogarciapablo: GeometryFormModels hold a reference to the featureDefinitionModel, so could move this logic to those classes IMO.

lib/assets/javascripts/cartodb3/editor/layers/edit-feature-content-views/edit-feature-attributes-form-model.js
```diff
@@ -49,6 +49,14 @@ module.exports = Backbone.Model.extend({
         if (columnType === 'number') {
           this.schema[columnName].isNumber = true;
           this.schema[columnName].showSlider = false;
+          // backbone-forms 0.14.1 has a 'number' validator:
+          // https://github.com/powmedia/backbone-forms/blob/v0.14.1/src/validators.js#L64
+          // We're on 0.14.0 so we have to use a 'regexp' validator
+          this.schema[columnName].validators = ['required', {
+            type: 'regexp', // backbone-forms 0.14.1 has a 'number' validator
+            regexp: /^[0-9]*\.?[0-9]*$/,
```
javisantana: this regexp should be a little bit better, right? for example

```
/^[0-9]*\.?[0-9]*$/.test('-0.2') -> false
```

take a look at this http://www.regular-expressions.info/floatingpoint.html

Issue #9022, opened 2016-07-20
Write on boarding text for "Enrich from Data Observatory" analysis


javierarce:

ohasselblad:
{Enrich from the Data Observatory}

{A new column has} been included with your original data.

* {custom column name from user}
  {This column contains the data requested from the Data Observatory}

{The Data Observatory enrichment is the result of the geometry location and the measure requested.}

Read more in [documentation](https://carto.com/docs/carto-engine/data/).

PR #10620, opened 2016-11-08
Fixes local string in the DO analysis

### LINKED ISSUE #10609: Missing local string in the Enrich From the DO analysis
### Context
* In the Enrich From the DO analysis, the description for the `the_geom` label is not localized. 

screen shot 2016-11-08 at 09 57 17

### Steps to Reproduce

1. Create an Enrich From the DO

javierarce:
Fixes #10609

The original template was using a string (`data-observatory-measure.the-geom`) that wasn't defined. This PR fixes that and uses `data-observatory-measure.custom-column` instead.

Related to: #9022


javierarce:
CR: @MariaCheca

PR #10629, opened 2016-11-08
Set simple geom of geometry model as empty when query doesn't have any geometry

### LINKED ISSUE #10617: applying a query without the_geom_webmercator MG38
### Steps to Reproduce

1- create a ma with a table
2- apply a sql like "select count(1) from table"

- you get an error "the_geom_webmercator does not exist" which is good
- the map is shown (I think it should go to the tbale)
- if you reload you have access to style and so on but you shouln't since you don't have geometry

#SMOKE_MG38

xavijam:
- It removes the geometry from the query-geometry-model.
- Add loading state and no-geometry placeholder for legends panel (could you take a look @nobuti?).
- Add loading state and no-geometry placeholder for infowindows panel (could you take a look @matallo?).

Fixes #10617.
Sorry to add two reviewers but both of you are the sheriffs of them.


lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/legend/legends-view.js
```diff
@@ -48,8 +55,15 @@ module.exports = CoreView.extend({
     this.clearSubViews();
     this.$el.empty();
 
+    if (this._queryGeometryModel.isFetching()) {
```
nobuti: I'd rather only one return and a generic `if else`. It's the same thing as declaring vars inside if, sorry.

nobuti:
LGTM

xavijam:
Do the same for infowindows view please.

xavijam:
Our frontend tests have passed, merging @nobuti 🚀  thanks for your changes man ❤️

Issue #10626, opened 2016-11-08
Legend colors are wrong after updating style (bubble + colors)


rafatower:
### Context
Running tests:
> Perform a map with bubbles + colors. Change methods and check the results.

### Steps to Reproduce
1. Create a map with a points dataset
2. Modify style to get size by value and colors by value
3. Change the colors

### Current Result
![image](https://cloud.githubusercontent.com/assets/1914276/20097254/d7613664-a5ad-11e6-8d42-1e8d5913b983.png)

 

### Expected result
The legend is shown with the right colors (greenish)

### Browser and version
Chrome on Linux

 

### .carto file
[wrong legend (on 2016-11-08 at 11.18.36).zip](https://github.com/CartoDB/cartodb/files/577987/wrong.legend.on.2016-11-08.at.11.18.36.zip)

 

### Additional info
#SMOKE_MG21
#SMOKE_MG28 - changing colors do not work, same effect on the legend

xavijam:
Hey @rafatower, we can't apply the choropleth ramp over the bubble legend. It is a limitation, not a bug.

rafatower:
then I guess I misunderstood the description of the test. What about #SMOKE_MG28 - changing colors do not work, same effect on the legend before closing this ticket? Shall I create a different ticket?

xavijam:
No, the test is ok, and you did it properly, the thing is it is not a problem, we could consider it as a warning. I'd live it open or move it to design repo.

nobuti:
I'm fixing the issue that the bubble is not inheriting color from styles properly.

PR #10631, opened 2016-11-08
Parse fillColor only if layerDefModel is present as an option.

nobuti:
This PR avoid overwriting fill color when saving. Related to #10626.

cc @javierarce


lib/assets/javascripts/cartodb3/data/legends/legend-bubble-definition-model.js
```diff
@@ -18,9 +18,9 @@ module.exports = LegendBaseDefModel.extend({
   parse: function (r, opts) {
     var attrs = LegendBaseDefModel.prototype.parse.call(this, r);
 
-    if (r.definition) {
+    if (r.definition && opts.layerDefinitionModel) {
       attrs.fillColor = r.definition.color;
-    } else {
+    } else if (!r.definition && opts.layerDefinitionModel) {
```
javierarce: What about writing the condition like this instead? 

```js
if (opts.layerDefinitionModel) {
  if (r.definition) {

  } else  {

  }
}
```

PR #10641, opened 2016-11-08
10640 owner builder enabled check fix

### LINKED ISSUE #10640: Enabling builder for org whose owner is not builder yields builder for subs but not owner
### Steps to Reproduce

0. You have an org whose owner is editor
1. Enable builder for org from central

### Current Result

Owner is still editor although org users see builder.

### Expected result

Owner is builder, along with other org members.

cc/ @CartoDB/builder-backend

gfiorav:
Fixes #10640 

Basically if an owner was editor and his org gets builder, he will still be editor because of former `builder_enabled?`check. 

It looks like there was a good reason for this to be how it was, so please explain @javitonino :)

### DIFF
--- a/app/models/carto/user.rb
+++ b/app/models/carto/user.rb
@@ -510,7 +510,7 @@ class Carto::User < ActiveRecord::Base
   end
 
   def builder_enabled?
-    if has_organization? && builder_enabled.nil?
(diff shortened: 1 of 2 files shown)


gfiorav:
Feature, not bug!

PR #10646, opened 2016-11-08
Use direct connection to create table from sql query

ethervoid:
Please @CartoDB/dataservices take a look. What do you think? is safe?


rafatower:
I think so but I'd wait till Thursday morning before rolling this out.

ethervoid:
Sure, I was going to wait. There is no rush with this patch

xavijam:
Hey, what happened with this one? 💃

PR #10661, opened 2016-11-09
Reload map when polygons/lines are saved

alonsogarciapablo:
The problem was:

1. When the geometry changed in CartoDB.js, we're updating the `the_geom` attribute of feature definition model [here](https://github.com/CartoDB/cartodb/blob/4dc10ec49518253bf8dc9811023b1ff0b0cb9e73/lib/assets/javascripts/cartodb3/deep-insights-integrations.js#L209-L212).
2. We're then updating `the_geom` of the geometry form model [here](https://github.com/CartoDB/cartodb/blob/4dc10ec49518253bf8dc9811023b1ff0b0cb9e73/lib/assets/javascripts/cartodb3/editor/layers/edit-feature-content-view.js#L117).
3. This change, was causing the `the_geom` attribute of the feature definition model to be updated again [here](https://github.com/CartoDB/cartodb/blob/4dc10ec49518253bf8dc9811023b1ff0b0cb9e73/lib/assets/javascripts/cartodb3/editor/layers/edit-feature-content-views/edit-feature-geometry-form-model.js#L30). We're setting the same `the_geom` **again**.
4. Because we had set the same `the_geom` twice, when the form was Saved, we detected the "save" event, but `the_geom` had not changed, so the map wasn't being invalidated [here](https://github.com/CartoDB/cartodb/blob/4dc10ec49518253bf8dc9811023b1ff0b0cb9e73/lib/assets/javascripts/cartodb3/deep-insights-integrations.js#L188-L193).

My fix works now because `the_geom` is not editable from the UI, but it we made it editable we'll need to change this approach.

@matallo @xavijam 👍 ?

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/layers/edit-feature-content-views/edit-feature-geometry-form-model.js
+++ b/lib/assets/javascripts/cartodb3/editor/layers/edit-feature-content-views/edit-feature-geometry-form-model.js
@@ -1,5 +1,4 @@
 var Backbone = require('backbone');
-var _ = require('underscore');
 
 module.exports = Backbone.Model.extend({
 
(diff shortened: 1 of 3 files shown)


xavijam:
Fine by me, doesn't it break with the point stuff?

matallo:
LGTM, points use a different model

PR #10671, opened 2016-11-09
Shrinkwrap.

### LINKED ISSUE #10669: Migrated custom infowindow close button doesn't work

nobuti:
This PR fixes #10669.

cc @xavijam

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "cartodb-ui",
-  "version": "4.4.09",
+  "version": "4.5.0",
   "dependencies": {
     "backbone": {
       "version": "1.2.3",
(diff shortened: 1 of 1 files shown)


nobuti:
Merge in https://github.com/CartoDB/cartodb/pull/10670

PR #10678, opened 2016-11-10
Fixes "applying hexbins over layers with overlays" error

### LINKED ISSUE #10662: Applying hexbins over layers with overviews returns an error
### Context
We should wrap the error and check why it is failing
 

### Steps to Reproduce

1. Import the .carto added at the end.
2. Go to style and apply hexbins type.
3. See the error.

### Current Result
It fails and the error is huge.
 
### Expected result
We should check the error itself, but also we should wrap that error.

### .carto file
https://dl.dropboxusercontent.com/u/931536/Vasque%20Country%20Traffic%20Incidents%20%28on%202016-11-09%20at%2014.44.13%29.carto

javierarce:
Related to #10662

This PR adds error matching for general 'column doesn't exist' case

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/layers/notification-error-messages.json
+++ b/lib/assets/javascripts/cartodb3/editor/layers/notification-error-messages.json
@@ -18,5 +18,7 @@
     "match": "^.*?column \"the_geom_webmercator\" does not exist.*",
     "replaceWith": "without-geom-webmercator",
     "errorType": "warning"
+  }, {
+    "match": "ERROR: (column .*? does not exist)",
   }]
 }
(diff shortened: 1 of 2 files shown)


javierarce:
CR: @xavijam

xavijam:
Screenshot please? 💋

javierarce:
As I said in the original issue, I couldn't reproduce the error locally or in staging (I tried a couple of servers), so I can't really show you a screenshot.

xavijam:
Ok, so merge it then 🇪🇸

javierarce:
That didn't work :(

`"trafikoa: Postgis Plugin: ERROR:  column "agg_value" does not exist``

I didn't notice that before the word `column` there are two spaces and not one, so the regexp doesn't catch it. I'll submit another fix with a custom spec for this case. Sorry about that!

Issue #10672, opened 2016-11-09
When latitude or longitude are edited in the input fields, marker in the map is not updated with the new position


xavijam:

alonsogarciapablo:
Working for me right now:

![lat-lng](https://cloud.githubusercontent.com/assets/390398/20151005/cf3b19a8-a6b8-11e6-9d00-f7f0e8ea9e6b.gif)

matallo:
can't reproduce, either

xavijam:
Have you tried with a new point?

matallo:
yeppers, still no clue

xavijam:
## STR

1. Create a map with an empty dataset.
2. Click over the layer in order to add a new feature.
3. Choose the point.
4. Place it in the map.
5. Go to the lat field and change it.
6. It doesn't change in the map.

matallo:
![screen shot 2016-11-10 at 10 10 17](https://cloud.githubusercontent.com/assets/36676/20170784/f77d59d8-a72d-11e6-8b07-458adcdec84d.png)

The feature is updated ok for me (the 0 value was my change), but it's true the geometry editor still remains in the same place. I'll take a look with @alonsogarciapablo

xavijam:
Minor compared with the cid problem, so no rush.

PR #10685, opened 2016-11-10
Point edition

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb/issues/10672.

Geometry wasn't being updated because we're setting a new `cartodb_id` on the featureDefinitionModel after the new `the_geom` had been set, and `featureDefinitionModel.hasChanged('the_geom')` was returning `false`

I also did a little refactoring to make sure `the_geom` is only set once when both `lat` and `lng` change at the same time (eg: when the marker is dragged and dropped). Otherwise, `featureDefinitionModel.hasChanged('the_geom')` was returning `false` too.

@matallo 👍 ?


matallo:
OMG let's do this

matallo:
bump package.json, btw

Issue #10666, opened 2016-11-09
Cartodb_id set as null when adding new features


saleiva:
### Steps to Reproduce

1. Add a feature
2. Try to save or look at the cartodb_id field in the form

### Current Result
It is null, so saving the feature raises an error

### Expected result
Cartodb_id should be an incremental value automatically added

alonsogarciapablo:
![cartodb_id](https://cloud.githubusercontent.com/assets/390398/20144215/345342ba-a69c-11e6-8d1f-b6d8cb0f6f15.gif)

alonsogarciapablo:
I've been able to reproduce this :(

STR:

- Draw a new geometry.
- Update any of the fields.
- Click `Add`
- 💥 

Working on a fix.

PR #10691, opened 2016-11-10
Fix edition of new features with new data

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb/issues/10666.

The problem: We're passing a null `cartodb_id` attribute to queryRowModel, and this causing the following error (only when you edited any field in the feature form):

```
{"errors":["PG::Error: ERROR:  null value in column \"cartodb_id\" violates not-null constraint\nDETAIL:  Failing row contains (null, null, null, a, b).\n"]}
```

With these changes, `queryRowModel` doesn't get a `cartodb_id` when creating a new feature/row.

@matallo makes sense? Thanks! 💋


matallo:
LGTM

alonsogarciapablo:
Thanks @matallo!

PR #10695, opened 2016-11-10
Adding always infowindow and tooltip models for data layers

xavijam:
Basically, we have found a problem with null tooltips in #10688, so in order to prevent that problem, we have coded this solution in frontend too. What do you think @matallo @alonsogarciapablo?
Pending to be fully tested in staging. 

CR: @matallo


alonsogarciapablo:
This is @matallo's territory... :P

matallo:
👍 

take a look at the tests, tho No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

xavijam:
yeah! :(

xavijam:
Closing since there is no problem for the moment with that.

PR #10704, opened 2016-11-11
Use the new observatory function to get the metadata from observatory

ethervoid:
This PR changes how we get the metadata from observatory. Now we are going to use new functions that are going to be deployed in DS cartodb/dataservices-api#307

Please @xavijam could you CR this thx!


xavijam:
Check errors before merging, of course 💋 .

javisantana:
:es:

PR #10708, opened 2016-11-11
10694 geometry edition default query

### LINKED ISSUE #10694: Cannot edit geometry
### Context
In organization users, using the builder, you cannot edit geometries, and the table view in the builder is readonly, due to a conflict with query defaults.

### Steps to Reproduce
1. In an org user, create a map
2. Go to the layer, and try to edit geometry

### Current Result
Readonly (query not default) 

### Expected result
Be able to edit it

### Additional info
Likely related to #10510

javitonino:
Fixes #10694 

This is a messy fix, manipulating the tableName passed to models to always include the user name. A proposal to improve it lives in #10709.

### DIFF
--- a/lib/assets/javascripts/cartodb3/helpers/sql-utils.js
+++ b/lib/assets/javascripts/cartodb3/helpers/sql-utils.js
@@ -84,8 +84,12 @@ module.exports = {
     return name;
   },
 
-  getDefaultSQL: function (tableName, userName, inOrganization) {
+  getQualifiedTableName: function (tableName, userName, inOrganization) {
     var schemaPrefix = (inOrganization && userName) ? this.quoteIfNeeded(userName) + '.' : '';
-    return 'SELECT * FROM ' + schemaPrefix + this.quoteIfNeeded(this.getUnqualifiedName(tableName));
+    return schemaPrefix + this.quoteIfNeeded(this.getUnqualifiedName(tableName));
+  },
+
+  getDefaultSQL: function (tableName, userName, inOrganization) {
+    return 'SELECT * FROM ' + this.getQualifiedTableName(tableName, userName, inOrganization);
   }
 };
--- a/lib/assets/javascripts/cartodb3/helpers/sql-utils.js
+++ b/lib/assets/javascripts/cartodb3/helpers/sql-utils.js
@@ -2,6 +2,8 @@
  *  Checks properties from a SQL
  */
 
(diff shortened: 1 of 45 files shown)


javitonino:
Please CR @xavijam 

I tested for organizations/plain users, the following things:
 - Dataset view: own dataset and shared dataset (read-only and read+write)
 - Builder: data view and geometry edition, for layers from own and shared datasets, created from dashboard or from adding a layer

saleiva:
Cool, plz lets try to deploy tomorrow so we fix S/B: 10533494
❤️

lib/assets/javascripts/cartodb3/helpers/sql-utils.js
```diff
@@ -66,11 +66,20 @@ module.exports = {
...
+  getUsernameFromTableName: function (tablename) {
+    if (!tablename) return null;
+    var tk = tablename.split('.');
+    if (tk.length === 2) {
+      return this.getUnquotedName(tk[0]);
+    }
+    return '';
+  },
+
   getUnquotedName: function (tablename) {
```
xavijam: Same for this one

lib/assets/javascripts/cartodb3/helpers/sql-utils.js
```diff
@@ -66,11 +66,20 @@ module.exports = {
     if (!tablename) return null;
     var tk = tablename.split('.');
     if (tk.length === 2) {
-      return tk[1];
+      return this.getUnquotedName(tk[1]);
     }
     return this.getUnquotedName(tablename);
   },
 
+  getUsernameFromTableName: function (tablename) {
```
xavijam: In my opinion this function doesn't belong to this "utils", it is not related with SQL. So I'd add it the typical '_' in order to make it "private".

lib/assets/javascripts/cartodb3/helpers/sql-utils.js
```diff
@@ -84,8 +93,12 @@ module.exports = {
     return name;
   },
 
-  getDefaultSQL: function (tableName, userName, inOrganization) {
+  getQualifiedTableName: function (tableName, userName, inOrganization) {
```
xavijam: And this one too.

javitonino:
Moved some stuff to TableNameUtils from SQLUtils and Utils. I'm testing in staging and every combination I described above is working. Take a second look if you'd like @xavijam.

PR #10750, opened 2016-11-21
9869 sql clear view

### LINKED ISSUE #9869: About the behaviour of "Clear view" when you apply a query that gives an error
I have noticed a behaviour that could be confusing about how we show the "clear view" button:
#### Scenario 1: A query fails after another query has been applied
1. Go to a dataset. It has the typical `SELECT *...` query applied
   - There's no clear view button because it's using the default query, so there's no query applied
     - It will appear if the query is not 100% that specific `SELECT * FROM table`. For example, if you add a space/line feed at the end. I don't think this is important though. This is detected as a query applied, so you would not have access to edit the data either.
2. Apply a valid query, such as `SELECT * FROM table LIMIT 10`
   - The "clear view" button appears
3. Apply a query that returns an error. Silly example: `SELECT '{"type":"Point","coordinates":[0.15,40.1]}'::geometry`
   - As the "clear view" button was already there, it remains. You can comfortably click on "clear view" to  get back to your normal table. 
#### Scenario 2: A query fails without a query being applied previously
1. Go to a dataset with no query applied.
2. Apply a query that returns an error. Silly example: `SELECT '{"type":"Point","coordinates":[0.15,40.1]}'::geometry`
   - As the "clear view" button was not there because of the lack of a query applied beforehand, you can't clear your error with a button. To get back to your table, you must write the `SELECT * FROM table` query by yourself. If you're a not advanced user which doesn't know SQL, you might probably have some trouble here. The rows in the data view are removed:
   
   image
   
   Also, I'm seeing that with the SQL pane open you can't really see the SQL error message below.

At first sight I'd say that this perhaps is difficult to manage because in Scenario 2 there's clearly not a query applied, but we still hide the rows and show the "error" message so I assume we might be able to control it. I recognize it's a picky thing, but I thought it was worth sharing. :-)

hacheka:
Fixes #9869 

Now, if the query throws an error in the first edition, the 'Clear' button appears.

sql clear view

Furthermore, this PR has a refactor for creating a base view with common behaviours shared between the SQL views on the Builder and on the Dataset window.

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/dataset/dataset-base-view.js
+++ b/lib/assets/javascripts/cartodb3/components/dataset/dataset-base-view.js
@@ -92,33 +92,6 @@ module.exports = CoreView.extend({
     }
   },
 
-  _runQuery: function (query, callback) {
-    this._querySchemaModel.set({
-      query: query,
-      status: 'unfetched'
-    });
-
-    this._queryGeometryModel.set({
-      query: query
-    });
-
-    SQLNotifications.showNotification({
-      status: 'loading',
-      info: _t('notifications.sql.applying'),
-      closable: false
-    });
-
-    var saveSQL = _.after(2, callback);
-    this._querySchemaModel.fetch({ success: saveSQL,
-      complete: function (model, response, options) {
-        if (response === 'abort') {
-          SQLNotifications.removeNotification();
-        }
-      }
-    });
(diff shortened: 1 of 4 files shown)


hacheka:
CR @nobuti 

It's deployed at DED05, so please take a look on both Builder and Dataset SQL workflow.

xavijam:
BRAVO!

xavijam:
so?

PR #10763, opened 2016-11-21
Labels edition for legends

### LINKED ISSUE #10687: Be able to override legend labels
Designs in https://github.com/CartoDB/design/issues/284

nobuti:
This PR fixes #10687 on Builder's side.

custom-labels

CR frontend part @xavijam 
CR backend legend definition @gfiorav


lib/formats/legends/custom_choropleth.json
```diff
@@ -5,6 +5,8 @@
   "properties": {
     "prefix": { "type": "string" },
     "suffix": { "type": "string" },
+    "left_label": { "type": "string" },
+    "right_label": { "type": "string" },
```
gfiorav: Could you explain what the difference between a normal choropleth and a custom choropleth is? There's a lot of duplication so maybe we could rethink it

nobuti: custom choropleths is for pixel aggregated map. The right legend must be a choropleth but the maps api doesn't return the values we need to render it. So we fake it. A little more context https://github.com/CartoDB/cartodb/issues/10224

nobuti:
Tested in ded04

gfiorav:
🇪🇸 for the backend side!

nobuti:
Shrinkwrap brings some updates in deep-insights and torque.js, could you take a look @donflopez to make sure all is ok? cc @xavijam

xavijam:
Same in https://github.com/CartoDB/cartodb/pull/10787.

donflopez:
🇪🇸 torque side

PR #10764, opened 2016-11-21
Add options instead of keep the last one

### LINKED ISSUE #10756: Lost D.O. total measures
### Context
Measures like total people, total foreigners, etc. are lost from the combo box. At least in **Spain**.

### Steps to Reproduce
Just try to add a Population measure to a points dataset in Spain. The option is not there 

### Additional info
Looks like the values eventually tagged as potential denominators for normalizing are not being offered as measures in the combo. But then you have 'Total dwellings' within 'Housing' category, so this theory hace aguas.

cc @talos @danicarrion

ethervoid:
Closes #10756

Aggregate objects with the same id instead of use the last one

Please @xavijam check this PR :)

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/data-observatory-metadata.js
+++ b/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/data-observatory-metadata.js
@@ -48,7 +48,10 @@ module.exports = Backbone.Model.extend({
         _.each(s.f1.columns, function (c) {
           cols[c.f1.name] = c.f1.id;
         });
-        sub[s.f1.name] = cols;
+        if (!(s.f1.name in sub)) {
+          sub[s.f1.name] = {};
+        }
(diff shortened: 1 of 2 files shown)


lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/analyses/data-observatory-metadata.js
```diff
@@ -48,7 +48,10 @@ module.exports = Backbone.Model.extend({
         _.each(s.f1.columns, function (c) {
           cols[c.f1.name] = c.f1.id;
         });
-        sub[s.f1.name] = cols;
+        if (!(s.f1.name in sub)) {
```
xavijam: If you want to use underscore:

```
_.contains(sub, s.f1.name)
```

xavijam:
Code looks good, but I'm not sure if I'm the right person in order to check the change.

ethervoid:
Thx @xavijam we fixed it in the observatory extension :))

PR #10770, opened 2016-11-22
Fixing problem with table stats

### LINKED ISSUE #9737: Problem generating histogram
## STR
1. Import this file [street_name_labels_melbourne.zip](https://github.com/CartoDB/cartodb/files/467651/street_name_labels_melbourne.zip).
2. Create a map with it.
3. Once created, go to the layer created.
4. Check app is blocked for a while.

After some researching with @javisantana, looks like a problem generating the histogram within table stats.

### LINKED ISSUE #9975: No quota indicator doesnt appearing when doing an Enrich Analysis and not having quota assigned

xavijam:
Basically there is a problem with stats when there is only one value, where there histogram freezes the whole application. There is no reason to render a histogram preview with that.

Fixes #9737.
CR: @nobuti (although there is no much to review :S)


xavijam:
Tested in ded02, by the way.

PR #10779, opened 2016-11-22
Allow individually enabling salesforce/arcgis connectors

javitonino:


javitonino:
Please CR @gfiorav (I will deploy config changes before this PR)

app/helpers/application_helper.rb
```diff
@@ -85,6 +85,8 @@ def frontend_config
       oauth_gdrive:               Cartodb.get_config(:oauth, 'gdrive', 'client_id'),
       oauth_instagram:            Cartodb.get_config(:oauth, 'instagram', 'app_key'),
       oauth_mailchimp:            Cartodb.get_config(:oauth, 'mailchimp', 'app_key'),
+      arcgis_enabled:             Cartodb.get_config(:datasources, 'arcgis_enabled'),
+      salesforce_enabled:         Cartodb.get_config(:datasources, 'salesforce_enabled'),
```
javitonino: This automatically defaults to `nil/false` if unconfigured, which is the current default for anything but the SaaS.

PR #10781, opened 2016-11-22
M4723 account type

javitonino:
Fixes https://github.com/CartoDB/cartodb-management/issues/4723

The regex was matching "Mercator academic" plan as if it was "academic" and disabling the option to hide the map logo then. This is fixed by only matching starting from the beginning of the line. Checked all current plans, and the regex matches the plan settings.


javitonino:
Please CR @juanignaciosl

PR #10793, opened 2016-11-23
10790 duplicate with incomplete map options

### LINKED ISSUE #10790: Cannot duplicate map missing some map options
https://rollbar.com/vizzuality/CartoDB/items/21191/occurrences/18758558293/

javitonino:
Fixes #10790

### DIFF
--- a/app/models/carto/map.rb
+++ b/app/models/carto/map.rb
@@ -150,12 +150,11 @@ class Carto::Map < ActiveRecord::Base
   private
 
   def ensure_options
-    self.options ||= {
-      dashboard_menu: true,
-      layer_selector: false,
-      legends: legends,
-      scrollwheel: scrollwheel
-    }
+    self.options ||= {}
+    options[:dashboard_menu] = true if options[:dashboard_menu].nil?
+    options[:layer_selector] = false if options[:layer_selector].nil?
+    options[:legends] = legends if options[:legends].nil?
+    options[:scrollwheel] = scrollwheel if options[:scrollwheel].nil?
   end
 
   def validate_options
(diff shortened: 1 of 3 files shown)


javitonino:
CR'ing this is a mission for @gfiorav

app/models/carto/map.rb
```diff
@@ -150,12 +150,11 @@ def layer_selector
...
-      layer_selector: false,
-      legends: legends,
-      scrollwheel: scrollwheel
-    }
+    self.options ||= {}
+    options[:dashboard_menu] = true if options[:dashboard_menu].nil?
+    options[:layer_selector] = false if options[:layer_selector].nil?
+    options[:legends] = legends if options[:legends].nil?
+    options[:scrollwheel] = scrollwheel if options[:scrollwheel].nil?
   end
```
gfiorav: Not an issue but: It might be nice to return the options. Just because this might return null or scrollwheel depending on options. Its always best to have a consistent return IMHO.

spec/services/carto/visualizations_export_service_2_spec.rb
```diff
@@ -628,14 +628,26 @@ def verify_overlay_vs_export(overlay, overlay_export)
...
 
-          visualization.map.options.should be
+          it 'partial options' do
+            export_2_0_6 = export
+            export_2_0_6[:visualization][:map][:options].delete(:dashboard_menu)
+
+            service = Carto::VisualizationsExportService2.new
+            visualization = service.build_visualization_from_json_export(export_2_0_6.to_json)
+
+            visualization.map.options[:dashboard_menu].should be
```
gfiorav: I'm confronted with this `.should be`. I think it's better to future proof and go for `.should_not be_nil` which is difficult to imagine could change in the future. It seems to me that `.should be` is more change-prone

javitonino: I think it's fine, it is documented at the same level as `be_nil`, so I think they should be equally likely to change. I don't think we should start mistrusting the upgrade paths of libraries (unless we can predict the future :crystal_ball:)

https://www.relishapp.com/rspec/rspec-expectations/v/2-2/docs/matchers/be-matchers

javitonino:
Second look @gfiorav ?

PR #10804, opened 2016-11-23
Move rollbar to resque

javitonino:
Asynchronous Rollbar reporting

### DIFF
--- a/lib/cartodb/logger.rb
+++ b/lib/cartodb/logger.rb
@@ -17,7 +17,9 @@ module CartoDB
     rescue => rollbar_exception
       # Last chance to report error
       begin
-        Rollbar.error(rollbar_exception)
+        Rollbar.with_config(use_async: false) do
+          Rollbar.error(rollbar_exception)
+        end
       rescue
       end
     end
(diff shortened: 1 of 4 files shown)


javitonino:
A quick experiment but it works. Maybe it's worth trying it to avoid slowing the ruis when Rollbar goes down. Opinions from @CartoDB/builder-backend ?

lib/cartodb/logger.rb
```diff
@@ -17,7 +17,9 @@ def self.log(level, exception: nil, message: nil, user: nil, **additional_data)
     rescue => rollbar_exception
       # Last chance to report error
       begin
-        Rollbar.error(rollbar_exception)
+        Rollbar.with_config(use_async: false) do
```
javitonino: BTW: The new rollbar update is only needed for this (falling back to non-async mode in case of error). Maybe we can get by without upgrading.

xavijam:
Ping!?

javitonino:
Given that I would like to have some #systems feedback on this, I'd wait a couple of weeks. Archiving the PR for now, I'll revive it when we get the chance.

PR #10812, opened 2016-11-24
Return appropriate return code for non-existing organization

javitonino:
Fixes https://github.com/CartoDB/cartodb-central/issues/1614

### DIFF
--- a/app/controllers/superadmin/organizations_controller.rb
+++ b/app/controllers/superadmin/organizations_controller.rb
@@ -54,7 +54,7 @@ class Superadmin::OrganizationsController < Superadmin::SuperadminController
 
   def get_organization
     @organization = Organization[params[:id]]
-    raise RecordNotFound unless @organization
+    render json: { error: 'Organization not found' }, status: 404 unless @organization
   end # get_organization
 
 end


javitonino:
Please CR @gfiorav

app/controllers/superadmin/organizations_controller.rb
```diff
@@ -54,7 +54,7 @@ def destroy
 
   def get_organization
     @organization = Organization[params[:id]]
-    raise RecordNotFound unless @organization
+    render json: { error: 'Organization not found' }, status: 404 unless @organization
```
gfiorav: Can you raise our very own `Carto::LoadError`? I think that way we will stay "standard" across our api responses (even though this is only for super admin).

javitonino: It seems more complicated than needed. Superadmin controllers do not `rescue_from` that exception, so I would have to require that exception definition, raise it, add a rescue from, and add a helper method to render this json.

I thought about it and decided against since this is the only place in the controller that can return a 404, so the advantages of code deduplication are not there. Do you think it's so much better?

Issue #10771, opened 2016-11-22
Disable some analysis because of missing dataservices


luisbosque:
Some analysis depend on external dataservices like data observatory, geocoding, isolines, etc..
If the instance is not configured to have these services, analysis depending on them should not be available.

At this moment, there are two places in app_config.yml where we are setting the url of dataservices:

```yaml
geocoder:
  api:
    host:
    port:
    user:
    dname
```

and

```yaml
aggregation_tables:
  host:
  port:
  dbname:
  username:
  password:
```

Not totally sure if there is any other attribute with info about dataservices.

I think the best thing to do would be disable these analysis when the attributes are empty or not exist.

Any other suggestion?

xavijam:
We have to take a decision about if we should remove or disable them.

cc @rochoa @javisantana @saleiva

saleiva:
remove plz.

xavijam:
Roger roget!

xavijam:
About your second bullet @saleiva, how could we distinguish that? Is there anything I'm missing?
Should I ask data-services team?

saleiva:
We are already doing so AFAIK. Analysis that cannot be run with polygons for example, are disabled when launching the modal from a polygon layer

xavijam:
Ahhh, I thought you were mentioning something related with the data itself and not with the geometry 👍 .

PR #10813, opened 2016-11-24
Add analysis checker within analyses data entry point

xavijam:
This should be the way to check if an analysis should be visible or not. And we have move all the parameters related with analysis to the same point, the 'data/analyses' file. There we will have all the info we need to start working with any analysis in any place, form, add-analysis modal, onboarding, animations,...

Also, due to the fact that we have several forms where we use more than one analysis, we have added some checkers in:

**Area-of-influence**: time option shouldn't be enabled if there is no routing provider.
**Geocoding**: only lat-lon option should be available when there is no geocoder provider.

screen shot 2016-11-23 at 16 43 47

screen shot 2016-11-24 at 11 29 54

CR: @javierarce (when tests turns into green).
Also, I'll need help in order to test at length those changes.
@javierarce and @piensaenpixel can I count with you? 🙏  (maybe?).

Fixes https://github.com/CartoDB/cartodb/issues/10771.
cc @rporres @javisantana @santisaez @saleiva


xavijam:
So adding @CartoDB/dataservices team in order to check if we are validating "available" analyses properly. Also, those analyses we are checking are:

- data-observatory-measure: by custom install flag.
- georeference-city: by custom install flag.
- georeference-ip-address: by custom install flag.
- georeference-country: by custom install flag.
- georeference-postal-code: by custom install flag.
- georeference-street-address: by `geocoder_provider`.
- georeference-admin-region: by custom install flag.
- routing-sequential: by `routing_provider`.
- routing-to-layer-all-to-all: by `routing_provider`.
- routing-to-single-point: by `routing_provider`.
- trade-area: by `isolines_provider`.

jgoizueta:
These cases that rely on service providers:

* georeference-street-address: by geocoder_provider.
* trade-area: by isolines_provider.
* routing-sequential: by routing_provider.
* routing-to-layer-all-to-all: by routing_provider.
* routing-to-single-point: by routing_provider.

The problem is that now we can have default providers so. Using the quota has a similar problem because internal services can be used without a quota.

We should probably remove those defaults, at least for quota have non zero quotas for all usable services, but right now we better use another way of disabling services. A (hardcoded) configuration flag could be added for on-premises/saas/...

@ethervoid @rafatower will have a more informed opinion, please comment on this

ethervoid:
This config should be enough, what do you think @xavijam ?

captura de pantalla 2016-11-25 13 03 32

xavijam:
That's ❤️

ethervoid:
Deployed. Merge master and enjoy :)

xavijam:
I'm deploying changes in ded02. So we should check that:

- Available analyses appear, those not available should not appear.
- Analyses animations appear in the proper analysis and work correctly in the add-analysis-modal.
- Onboarding messages appear correctly after an analysis is applied.

cc @javierarce @piensaenpixel

xavijam:
Ooooooook, deployed again in ded02 after some small problems :).

xavijam:
Merging when green.

xavijam:
Thanks @javierarce for your help.

Issue #10792, opened 2016-11-23
Destroy visualizations before users to avoid state fk issue


gfiorav:
### Context

The state model has an FK relation with the visualisation model. This means that when removing a user, if visualizations with states are still present, an error is produced.

javitonino:
Fixed more or less by accident in https://github.com/CartoDB/cartodb/pull/10816, by deleting all visualizations before destorying the user.

gfiorav:
How does this work when a layer comes from a shared table? Is it deleted as well?

javitonino:
Yup, with the latest commit.

gfiorav:
Can this be closed @javitonino ?

javitonino:
Yup ✨

PR #10816, opened 2016-11-24
10635 delete layers on user deletion

### LINKED ISSUE #10635: Deleting an user doesn't delete its layers
When deleting an user, its layers do not get deleted:
 - In `User.before_destroy` the maps are destroyed, but that doesn't destroy the layers. Deleting a visualization works (because there is layer deletion on visualization destruction)
 - Custom basemaps (user layers) are disassociated but not deleted.

cc @CartoDB/builder-backend

javitonino:
Fixes #10635 and #10792 (by deleting visualizations before the user)

The trick is in `UserTable.after_destroy`. In this case, `super` triggers map destruction (via Sequel "FK"), and `service.after_destroy` triggers visualization destruction. If the map is destroyed first, the reference to the layers is lost.

~~Maybe layers should be destroyed from map (instead of from visualization), but I'm feeling cautious and I think this could be done as part of the move to AR model.~~
Also, layers are now destroyed from the map instead of the visualization. This forced to do some changed to how named maps are handled during deletions.


javitonino:
CR @gfiorav

javitonino:
ping!

spec/models/user_spec.rb
```diff
@@ -1161,25 +1161,24 @@ def overquota(delta = 0)
...
-      .returns(true)
+                    .with(".*#{uuid}:vizjson")
+                    .at_least_once
+                    .returns(true)
 
     doomed_user.destroy
 
-    DataImport.where(:user_id => doomed_user.id).count.should == 0
-    UserTable.where(:user_id => doomed_user.id).count.should == 0
+    DataImport.where(user_id: doomed_user.id).count.should == 0
```
gfiorav: is this better?

spec/models/user_spec.rb
```diff
@@ -2320,35 +2320,52 @@ def stub_and_check_version_pre_mu(version, is_pre_mu)
...
+      visualization = CartoDB::Visualization::Member.new(id: carto_visualization.id).fetch
+
+      # Force ORM to cache layers (to check if they are deleted later)
+      canonical_visualization.map.layers
+      visualization.map.layers
+
+      user_layer = Layer.create(kind: 'tiled')
+      user.add_layer(user_layer)
+
+      return user, table, [canonical_visualization, visualization], user_layer
```
gfiorav: +1

javitonino: It is not redundant here, it is needed for multi-return. Just removing the return gives an error. Maybe there is another way to do it?

spec/models/user_spec.rb
javitonino: This doesn't work with `should` :(

It seems it be used for normal comparisons though:
http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/NumericPredicate

spec/models/user_spec.rb
gfiorav: yeah! you have to:

```ruby
[user, table, [canonical_visualization, visualization], user_layer]
```

sorry I didn't realise in the first one

javitonino: You learn something new everyday. I thought it was special syntax, not just an array 🌴

javitonino:
This breaks visualization deletion, cannot believe the tests didn't catch it 😞

javitonino:
Ready for a second CR round @gfiorav. Please note that this is merged with #10837 so I can have proper tests, so a lot of the changes are being reviewed separately. If you prefer, we can wait until the other PR is in master so the changes of this PR are easier to read.

javitonino:
Merged with master

PR #10827, opened 2016-11-25
Tests for source analysis query

javitonino:
Tests for https://github.com/CartoDB/cartodb/pull/10825


javitonino:
Please CR @juanignaciosl

PR #10833, opened 2016-11-25
10116 centering layer

### LINKED ISSUE #10116: Options for centering map on filtered data / layer data
Relate to https://github.com/CartoDB/design/issues/59

hacheka:
This PR closes #10116 

We've added a new feature to allow the user to center the map view according to a particular layer, based on its content.

**NEW FEATURES**

Icon for centering map in layer edition

icon centering map

Icon also available in expert mode

icon centering map expert

Context menu option for centering map in layer list

menu centering map

Icon not available if layer has no geom

icon center disappears

Context menu option not available if layer has no geom

menu disappears

CR @nobuti 

CC @xavijam 

CC @urbanphes Please check that the design is on point

CC @brittanymicek @saleiva 
We've added this copy for the centering process:
- Loading: `Calculating coordinates...`
- Success: `Map centered`
- Error: `Error centering map`

Could you both check that these copies above are correct and good enough, please? 

Thx!!


xavijam:
cc @MariaCheca and please, let's open the needed ticket for the same behaviour in the widgets part. cc @saleiva

brittanymicek:
@hacheka LGTM!

lib/assets/javascripts/cartodb3/editor/layers/layer-header.tpl
```diff
@@ -26,6 +26,18 @@
     </div>
   </div>
 
+  <div class="Editor-HeaderInfo-zoom CDB-Shape js-zoom">
```
nobuti: This should be a button

app/assets/stylesheets/editor-3/header.css.scss
```diff
@@ -114,8 +114,12 @@
...
 }
 
-
-.Editor-HeaderInfo-inner + .CDB-Shape {
+.Editor-HeaderInfo-inner ~ .CDB-Shape {
   margin-top: 11px;
 }
 
+.Editor-HeaderInfo-zoom {
+  cursor: pointer;
```
nobuti: This should be a button, lets get rid of this.

lib/assets/javascripts/cartodb3/editor/layers/layer-header-view.js
```diff
@@ -147,6 +174,24 @@ module.exports = CoreView.extend({
     return source;
   },
 
+  _onSourceChanged: function () {
+    var nodeModel = this._layerDefinitionModel.getAnalysisDefinitionNodeModel();
+    if (this._topQueryGeometryModel !== null) {
+      this._topQueryGeometryModel.unbind('change:simple_geom');
+    }
+    this._topQueryGeometryModel = nodeModel.queryGeometryModel.bind('change:simple_geom', this._checkIfGeometryVisible, this);
```
nobuti: ❤️

lib/assets/javascripts/cartodb3/editor/layers/layer-header-view.js
```diff
@@ -155,6 +200,7 @@ module.exports = CoreView.extend({
     this._getLink().toggleClass('u-altTextColor', editing);
     this._getBack().toggleClass('u-whiteTextColor', editing);
     this._getToggleMenu().toggleClass('is-white', editing);
+    this._getZoomIcon().attr('fill', editing ? whiteColor : blueColor);
```
nobuti: I feel we can handle this at css level

lib/assets/javascripts/cartodb3/editor/layers/layer-header-view.js
```diff
@@ -181,6 +227,14 @@ module.exports = CoreView.extend({
     return this.$('.CDB-Shape-threePoints');
   },
 
+  _getZoom: function () {
+    return this.$('.js-zoom');
+  },
+
+  _getZoomIcon: function () {
```
nobuti: Once above achieved, we can get rid of this I think, right?

lib/assets/javascripts/cartodb3/editor/map-operations/zoom-to-data.js
```diff
@@ -0,0 +1,46 @@
...
+
+var ongoingQuery = false;
+
+module.exports = function zoomToData (configModel, stateModel, query) {
+  if (!configModel) throw new Error('configModel is required');
+  if (!stateModel) throw new Error('stateModel is required');
+  if (!query) throw new Error('query is required');
+
+  var queryLauncher = new cdb.SQL({
+    user: configModel.get('user_name'),
```
nobuti: {user, sql_api_template, api_key} = configModel.toJSON();
My kingdom for a leapfrog that prepares the frontend for es6.

nobuti:
Also, I've found a couple things to improve, not sure if they are possible, specially the second point:

1. If we press repeatedly the center map button, a new notification is shown. Not sure if we could compare the result of the center operation before to make the next one.

centering

2. In the called expert mode, the result is a bit confusing, because is actually centered but only when you exit that mode. While the editor is open, some points are hidden. I guess this is because the way we handle the map container css styles 

centering-expert

nobuti:
Not sure if this is right, if I resize down the window, the centering thing seems to not working right, some features left outside the map's viewport:

centering2

Not sure this is outside of scope and it belongs to cartodb.js.

app/assets/stylesheets/editor-3/header.css.scss
```diff
@@ -119,7 +120,10 @@
...
 
 .Editor-HeaderInfo-zoom {
-  cursor: pointer;
   margin-right: 5px;
+  fill: $cBlue;
 }
 
+.Editor-HeaderInfo-zoom.is-white {
+  fill: $cWhite;
+}
```
hacheka: - Changed HTML element to button.
- Fill color of SVG through CSS cascading.

@nobuti

lib/assets/javascripts/cartodb3/editor/layers/layer-header-view.js
```diff
@@ -200,7 +200,7 @@ module.exports = CoreView.extend({
     this._getLink().toggleClass('u-altTextColor', editing);
     this._getBack().toggleClass('u-whiteTextColor', editing);
     this._getToggleMenu().toggleClass('is-white', editing);
-    this._getZoomIcon().attr('fill', editing ? whiteColor : blueColor);
+    this._getZoom().toggleClass('is-white', editing);
```
hacheka: Changed SVG fill color to CSS cascade handling and got rid of auxiliar function to get the icon node.

@nobuti

lib/assets/javascripts/cartodb3/editor/layers/layer-header.tpl
```diff
@@ -26,17 +26,17 @@
     </div>
   </div>
 
-  <div class="Editor-HeaderInfo-zoom CDB-Shape js-zoom">
+  <button class="Editor-HeaderInfo-zoom CDB-Shape js-zoom">
     <svg width="16px" height="15px" viewBox="287 30 16 15" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
-      <g id="Icon" stroke="none" stroke-width="1" fill="#1785FB" fill-rule="evenodd" transform="translate(288.000000, 30.000000)">
+      <g id="Icon" stroke="none" stroke-width="1" fill-rule="evenodd" transform="translate(288.000000, 30.000000)">
```
hacheka: For cascading to work, we need to get rid of every affected fill attribute in the SVG code. If not, the fill attribute has priority over the inherited fill color.

@nobuti

app/assets/stylesheets/editor-3/header.css.scss
```diff
@@ -110,12 +110,20 @@
...
 }
 
-
-.Editor-HeaderInfo-inner + .CDB-Shape {
+.Editor-HeaderInfo-inner ~ .CDB-Shape {
   margin-top: 11px;
 }
 
+.Editor-HeaderInfo-zoom {
+  margin-right: 5px;
```
nobuti: I feel we can get rid of this, in favor of the helper u-rSpace class.

hacheka:
Regarding https://github.com/CartoDB/cartodb/pull/10833#issuecomment-263055762

1. The code actually prevents to make two simultaneous requests. The issue here is that you have a very fast finger. You can check it setting the throttling of Chrome to a very bad network and clicking several times: it doesn't trigger a new request if one is ongoing.

Perhaps we could set a delay time to prevent a query some time after a request has been done. What do you think @xavijam ?

2. Don't get the point. For me it seems to work the same way in expert and normal mode.

Regarding comment https://github.com/CartoDB/cartodb/pull/10833#issuecomment-263056108

I think it belongs to cartodb.js. We do nothing more than calling to its `setBounds()` through `DeepInsightsIntegration` It seems to center the map correctly according to the whole screen, but if you have a reduced viewport some points seems to be left outside.

nobuti:
What I meant in the point 2 is that some points stay outside the viewport in expert mode. Not a big deal, but it confused me up the first time.

xavijam:
1. We should not throw a new request or a new notification either if the first interaction is ongoing yet.

PR #10845, opened 2016-11-28
Reduce the maximum number of vertices of editable features

### LINKED ISSUE #10841: Reduce the max number of vertices for editable geometries
Since we're adding middle points, we should reduce the allowed number of vertices [here](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/deep-insights-integration/edit-feature-overlay.js#L9).

alonsogarciapablo:
Fixes #10841.

The new limit is 1000 vertices. Take into account that a geometry with 1000 vertices will also render ~1000 extra markers between each vertex so that users can expand the geometry (and that is the reason why we're changing this).

@matallo 👍 ?

cc: @xavijam


matallo:
🚀

PR #10856, opened 2016-11-28
HTML custom legends

### LINKED ISSUE #10436: Custom legend should allow free HTML.
In custom legend, since its not dynamic, we should let HTML to be customizable always. If not, there is no way of creating a custom legend actually...

nobuti:
This PR fixes #10436.

html-custom-legends

CR @xavijam 
cc @saleiva


nobuti:
Tests pending to be fixed, working on it.

xavijam:
When done (tests) ping me please 💋

nobuti:
Heading to ded08

nobuti:
Finally in ded03, sorry guys

xavijam:
Marvellous! Just two small things I've found:

- If html is used, we should show the editor panel instead of the forms one.
- When a filter is applied using a widget, panel is re-rendered, do we need to do that?.

html-legend

nobuti:
Working on it. I think point 2 is due to the analysis applied, not the filtering.

xavijam:
it is true that the second one it is a bit complicated, let's decide what to do later offline.

xavijam:
Both thing tested with new changes and looks great 💃 .

xavijam:
BTW, talking now about legends migration, are we setting this type when it is not possible to migrate it directly?

nobuti:
Yes, that's the idea

lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/legend/color/legend-custom-template.tpl
```diff
@@ -0,0 +1,12 @@
+<ul>
+  <% for(var i in items) { %>
+    <li class="Legend-categoryListItem u-flex u-justifySpace u-alignCenter">
+      <p class="Legend-categoryTitle CDB-Text CDB-Size-small u-upperCase u-ellipsis" title="<%= items[i].title %>"><%= items[i].title %></p>
+      <% if (items[i].color) { %>
+        <span class="Legend-categoryCircle" style="opacity:1; background: <%= items[i].color %>;"></span>
+      <% } else if (items[i].icon) { %>
+        <span class="Legend-categoryIcon" style="opacity:1; background-image: <%= items[i].icon %>;"></span>
```
xavijam: Ready for the future 👍

xavijam:
LOVELY!

xavijam:
cc @gfiorav due to migration.

lib/formats/legends/custom.json
```diff
@@ -30,6 +30,7 @@
           }
         }
       }
-    }
+    },
+    "html": { "type": "string" }
```
gfiorav: So all custom legends will have to have a `html`?

gfiorav: Do you think it could be better to not include the key at all if no html is present? That would make all current custom legends instantly compatible. You'd just have to remove `html` from the required attributes in this definition and make sure you're not including the key if no html is added.

So, in short: instead of `html: ''` -> nothing at all

What do you think?

nobuti: 👍  on it

xavijam:
Bravo! Merging tomorrow right? 💃

nobuti:
diosmediante

PR #10858, opened 2016-11-28
10116 hotfix bad margin

### LINKED ISSUE #10846: Improved text on center map button.
### Current Result
"Center map"

 ### Expected result
"Center map on layer"

hacheka:
Fixes bug in #10116 (see comment https://github.com/CartoDB/cartodb/issues/10116#issuecomment-263278517)

Also fixes #10846 

Icons now are on place even with long dataset names.

image

Menu option with proper copy 'Center map on layer'

image

Same copy applied to tooltip

image

CR @nobuti


piensaenpixel:
much better :)

PR #10867, opened 2016-11-28
editors: multiselect, tests, first commit

matallo:
added first test for multiselect editor ("should render properly"), still pending the rest of the spec

screen shot 2016-11-28 at 18 24 04

this is going to be tricky to test, because of the interpolation in _t, i.e.

```
  _getLabel: function () {
    var itemCount = _.compact(this.collection.pluck('selected')).length;
    return _t('components.backbone-forms.select.selected', { count: itemCount });
  },
```

CR @xavijam


xavijam:
It is a start so :+1:

PR #10874, opened 2016-11-29
change svg

piensaenpixel:

image

Fix https://github.com/CartoDB/design/issues/350

cc/ @ramiroaznar

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/legend/carousel-icons/category.tpl
+++ b/lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/legend/carousel-icons/category.tpl
@@ -5,7 +5,7 @@
         <use xlink:href="#path-1"></use>
     </mask>
   </defs>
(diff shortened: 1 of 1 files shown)


lib/assets/javascripts/cartodb3/editor/layers/layer-content-views/legend/carousel-icons/category.tpl
```diff
@@ -7,15 +7,15 @@
   </defs>
   <g id="category" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" transform="translate(529.000000, 429.000000)">
```
nobuti: I guess we can get rid of stroke, fill and fill-rule attributes, and save a few bytes along the way, right?

piensaenpixel: yeah, good catch :)

PR #10885, opened 2016-11-29
Correctly check for user error on creation task

javitonino:
CR @gfiorav 

cc @matallo

### DIFF
--- a/lib/tasks/setup.rake
+++ b/lib/tasks/setup.rake
@@ -68,7 +68,7 @@ DESC
         u.builder_enabled = true
       end
 
-      if !user.errors.empty?
+      unless user.valid?
         puts
         puts 'There are some problems with the info that you provided to create the new user:'
         user.errors.full_messages.each do |error|


matallo:
LGTM 😅

lib/tasks/setup.rake
```diff
@@ -68,7 +68,7 @@ DESC
         u.builder_enabled = true
       end
 
-      if !user.errors.empty?
+      unless user.valid?
```
gfiorav: What does .valid? source code do? I thought it checked `errors.empty?`

javitonino: It actually runs the validations. `errors` is empty until you call `.save` or `.valid?`.

javitonino: The problem here was that `errors` was being checked before calling `.save` or `.valid?`, so it is always empty.

gfiorav: Oh! Of course!! I didn't even consider that `.valid?` hadn't been run beforehand.

👌

Issue #10481, opened 2016-10-31
Editing the style of a layer reset's its legend's titles [DEPENDS ON #10904]


gfiorav:
**STR**

0. Have map with legends
1. Set a custom title for a legend
2. Change styles of that layer (color or whatever)

**Expected result:**

Styles change and maybe the legend has to change as well, but the title stays the same.

**Actual result:**

The legend is reset and the title is as well.

I guess it's debatable but it has given me some frustration today. What do you guys think? Bug or feature?

cc/ @xavijam

xavijam:
That's a good one, it opens the 🍉  about if we need states in the properties in order to store if the user took the decision of changing the title it was auto-generated due to whatever.

xavijam:
As @nobuti pointed in the PR, not merge until #10904.

ivanmalagon:
Brought back from `acceptance` to `development`.

The fix works perfectly when we modify the title of a gradient legend. If we modify the styling after changing the title, it saves the last title. OK

Nevertheless, the next scenario happens:

1. Choose a custom legend.
2. Check title and modify it.
3. Modify the style. For instance, select another color palette.
4. Builder changes the legend from `custom` to `gradient`
5. The given title is lost.

![legend-problem](https://cloud.githubusercontent.com/assets/1078228/20892557/02abc9bc-bb0f-11e6-8cf4-e29a613e0d9c.gif)

After discussing this with @xavijam we concluded that there are two issues here:

A. Should Builder change the custom legend to a gradient one after changing only the color palette?
B. Should Builder keep the title between changes of legend type?

A is not clear for us if it's a good approach. @saleiva has to tell us if it's the desired behaviour or it's a bug.

B is a bug and should be adressed in this issue. When changing between the two types of legends, the title has to be kept.

nobuti:
A bug? Really? It's not the same legend, not sure how to tackle.

xavijam:
Holidays means holidays @nobuti.

nobuti:
We need to define the specs, I think there will be more cases and we'll have to decide what to do.

nobuti:
Should B be applied when the user change between legends using the form? I guess we want this only by legend type (color or size) right?

nobuti:
Fixed and ready for another pass.

ivanmalagon:
Local acceptance 👍 
Staging acceptance at 08 :+1:

PR #10904, opened 2016-11-30
Add conf column to legends

### LINKED ISSUE #10877: Add configuration json column to Legend
It turns out it's getting a bit more complicated than was originally thought. 

The original spirit of this ticket is to have a separate place for the column name and the column that's being used for technical purposes. It's been agreed that the title of the legend should be held in the `title` field and to introduce a `conf` or `params` JSON that should contain technical details for the legend to work properly.

cc/ @nobuti

gfiorav:
**DON'T DEPLOY BEFORE https://github.com/CartoDB/cartodb/issues/10954**

Fixes #10877

The idea is to add a validated JSON column to legends called conf to store things related to how the legend works and that need to be persisted (column names, etc)

cc @nobuti


gfiorav:
Please CR @javitonino

db/migrate/20161130102840_add_conf_to_legends.rb
```diff
@@ -0,0 +1,9 @@
+Sequel.migration do
+  up do
+    add_column :legends, :conf, :json, null: false, default: '{}'
+  end
+
+  down do
+    drop_column :legends, :conf
+  end
+end
```
javitonino: Better to use the new migration wrapper to avoid locking issues.

Also, it goes without saying, split it to a different PR before deploying.

app/models/carto/legend.rb
```diff
@@ -32,6 +35,10 @@ def ensure_definition
       self.definition ||= Hash.new
     end
 
+    def ensure_conf
+      self.conf ||= Hash.new
```
javitonino: Is this needed even when the column is `NOT NULL` and has a default?

spec/models/carto/legend_spec.rb
```diff
@@ -113,6 +113,59 @@ module Carto
...
+          bad_conf = { not_column_name: 'not array' }
+          legend = Legend.new(layer_id: @layer.id,
+                              type: 'bubble',
+                              definition: {},
+                              conf: bad_conf)
+
+          legend.valid?
+
+          legend.errors[:conf].should_not be_empty
+          legend.errors[:conf].should_not include('could not be validated')
```
javitonino: Should not include that error message? Why?

app/models/carto/legend.rb
gfiorav: The JSON validator will swallow a `nil` on validation before db has the opportunity to add the default!

db/migrate/20161130102840_add_conf_to_legends.rb
gfiorav: Yeah, totally forgot, sorry.

spec/models/carto/legend_spec.rb
gfiorav: Careless copy paste, but still passes since that error only appears when something prevents the validation from happening (malformed json or whatever)

gfiorav:
Can I get a 👍 ?

db/migrate/20161125173248_simplify_states_table.rb
```diff
@@ -0,0 +1,26 @@
+Sequel.migration do
```
javitonino: 😱

db/migrate/20161125174333_create_snapshots.rb
```diff
@@ -0,0 +1,21 @@
+Sequel.migration do
```
javitonino: 😱

gfiorav:
Have a look now @javitonino

ivanmalagon:
Brought back from `acceptance` to `development`.

We found a bug in #10481 Perhaps some work in backend is needed. Until that frontend task is not fully OK, this current one must wait to pass acceptance and then be deployed.

For further information, please refer to the comment https://github.com/CartoDB/cartodb/issues/10481#issuecomment-264900788

ivanmalagon:
Deploying it to `ded05` to accept only this backend part.

If new backend code must be developed because of https://github.com/CartoDB/cartodb/issues/10481#issuecomment-264900788 a new ticket should be created.

javitonino:
Closing, merged via https://github.com/CartoDB/cartodb/pull/10980 due to Gemfile.lock conflicts

PR #10909, opened 2016-11-30
Fix html legends

nobuti:
This PR is a hotfix to 'migrate' the `html` legends to `custom` in the frontend. This will be done with a proper migration later.

CR @xavijam


xavijam:
Change version and deploy.

PR #10918, opened 2016-11-30
Upgrage cartodb extension to 0.18.5

jgoizueta:
New extension revision with some fixes and two new overview strategies.

### DIFF
--- a/app/models/user/db_service.rb
+++ b/app/models/user/db_service.rb
@@ -518,7 +518,7 @@ module CartoDB
       # Upgrade the cartodb postgresql extension
       def upgrade_cartodb_postgres_extension(statement_timeout = nil, cdb_extension_target_version = nil)
         if cdb_extension_target_version.nil?
-          cdb_extension_target_version = '0.18.4'
+          cdb_extension_target_version = '0.18.5'
         end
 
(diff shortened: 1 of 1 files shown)


app/models/user/db_service.rb
```diff
@@ -518,7 +518,7 @@ def reset_pooled_connections
       # Upgrade the cartodb postgresql extension
       def upgrade_cartodb_postgres_extension(statement_timeout = nil, cdb_extension_target_version = nil)
         if cdb_extension_target_version.nil?
-          cdb_extension_target_version = '0.18.4'
+          cdb_extension_target_version = '0.18.5'
```
ethervoid: why dont we have this as a constant?

ethervoid:
LGTM

rafatower:
:+1:

app/models/user/db_service.rb
jgoizueta: ...and why not have it (the default current version) in app_config.yml to avoid deploying just to increase it?

ethervoid: Because Chef doesn't progress all the changes immediately so we're not sure if the change is on all the servers.

PR #10923, opened 2016-12-01
10919 migrator migrates to custom instead of html

### LINKED ISSUE #10919: Make migrator migrate to custom instead of HTML

gfiorav:
Fixes #10919 

This RP changes how the legend migrator behaves: Instead of using an HTML type, it uses a custom type (that now supports an html field).


gfiorav:
Please CR @javitonino

PR #10925, opened 2016-12-01
10861 migrate html to custom

### LINKED ISSUE #10861: HTML Custom legend migrations.
To complete https://github.com/CartoDB/cartodb/pull/10856 we are going to remove the `html` legend type. The `html` will belong to the `custom` type from now on. So in this issue we need a migration that takes care of:
- changing the legend's type from `html` to `custom`
- setting the `html` definition attribute properly

gfiorav:
Fixes #10861 

Rake to change type from `html` to `custom`. This will be used once the migrator is updated to migrate to custom instead of html in https://github.com/CartoDB/cartodb/pull/10923. After that we'll drop the html type entirely.

### DIFF
--- a/lib/formats/legends/custom.json
+++ b/lib/formats/legends/custom.json
@@ -1,8 +1,5 @@
 {
   "type": "object",
-  "required": [
-    "categories"
-  ],
   "additionalProperties": false,
   "properties": {
     "categories": {
--- a/lib/tasks/legends.rake
+++ b/lib/tasks/legends.rake
@@ -1,0 +1,12 @@
+# encoding: utf-8
+
+namespace :carto do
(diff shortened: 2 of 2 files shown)


gfiorav:
@javitonino have a look please! This will be used once the migrator is updated to migrate to custom instead of html in https://github.com/CartoDB/cartodb/pull/10923. After that we'll drop the html type entirely.

javitonino:
Uhm, this doesn't change the legend itself, just the type. From the issue:
> setting the html definition attribute properly

Don't we need to do that?

gfiorav:
Yeah, the html definition is compatible with the custom definition, so only changing the type is enough.

PR #10934, opened 2016-12-01
Layers controller refactor [WAIT UNTIL SS AND SR DEPLOY]

javitonino:
Add more functionality to Layer, so it covers all Layer model tests

### DIFF
--- a/app/controllers/carto/api/vizjson3_presenter.rb
+++ b/app/controllers/carto/api/vizjson3_presenter.rb
@@ -322,7 +322,7 @@ module Carto
       def to_vizjson
         vizjson = { id: @layer.id }
 
-        if @layer.base?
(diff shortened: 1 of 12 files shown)


javitonino:
Ready for CR! @gfiorav

javitonino:
1. Yep, pretty sure. The old model called it `base_layer?` so the change only applies to usages of the Carto model (of which there are fewer).
2. `.carto?` ? I guess you meant `.data_layer?`. But that was fixed by Nacho some weeks ago: https://github.com/CartoDB/cartodb/commit/5393656309cd9c84d531bbe0c2c33d2f7c8a1250

matallo:
tested a combination of torque, aggr, and baselayers 🚀

PR #10935, opened 2016-12-01
10930 simplify states

### LINKED ISSUE #10930: Simplify states

gfiorav:
Fixes #10930 

Please CR @javitonino


app/models/carto/state.rb
```diff
@@ -1,21 +1,18 @@
-# encoding: UTF-8
+# encoding: uft-8
```
javitonino: Meeec (`uft`)

javitonino:
Oh, I thought we were going to make the relationship the other way (which makes FK usage much easier).

So instead of what you do we:
 - Delete State.user_id
 - Keep State.visualization_id, and set it as UNIQUE (or even PK)
 - Delete Visualization.state_id

This way, it's easier to handle visualization deletion via FK, and we ensure that there are no orphan states (via FK) or duplicates (via UNIQUE constraint/PK). Also, you save an update to visualizations when setting the state for the first time.

gfiorav:
That would imply a 'visualisation belongs_to state' type relation, right? Are you sure this is really a better approach

javitonino:
No, it would be
in Visualization: `has_one State`
in State: `belongs_to Visualization`

which I think makes sense.

gfiorav:
👍 so the belongs_to is on the model that has the FK, I was thinking it the other way around

PR #10940, opened 2016-12-01
SR (3/5) | add migration to drop states.user_id

gfiorav:
**DON'T DEPLOY BEFORE #10959**
**DON'T DEPLOY BEFORE #10957**

Please CR @javitonino


javitonino:
We could also drop `id` and make `visualization_id` the primary key, for great glory. But I don't know if we could have some problems with Rails.

Anyway, this cannot be the first PR to merge, we have to merge the columns.reject thing before this one.

gfiorav:
1. I'd leave that out. Let's not make it more complicated.

2. Yup, just realised. It has to be 4 steps.

javitonino:
1. :+1: We can always take it out later if we feel like it
2. I think you can leave all the migrations for last (this + visualizations.state_id), don't you?

db/migrate/20161201160541_drop_column_user_id_in_states.rb
```diff
@@ -0,0 +1,24 @@
+require 'carto/db/migration_helper'
+
+include Carto::Db::MigrationHelper
+
+migration(
+  Proc.new do
+    alter_table :states do
+      drop_column :user_id
+      add_index :visualization_id
```
javitonino: This should be UNIQUE. But we have to check that there are not duplicates currently in the DB (and clean them if there are).

javitonino:
CR @gfiorav

PR #10948, opened 2016-12-02
Return dates in consistent tz format

### LINKED ISSUE #10903: Each time you click on the date, the date decreases one day
ghost2

javitonino:
Fixes #10903 

Fixes the return format of records from the records controller. Instead of relying in Sequel to do the parsing of dates, we return a string and parse it using Ruby which does the right thing regarding timezones.

### DIFF
--- a/app/models/table.rb
+++ b/app/models/table.rb
@@ -1061,17 +1061,29 @@ class Table
   def record(identifier)
     row = nil
     owner.in_database do |user_database|
-      select = if schema.flatten.include?(THE_GEOM)
-        schema.select{|c| c[0] != THE_GEOM }.map{|c| %Q{"#{c[0]}"} }.join(',') + ',ST_AsGeoJSON(the_geom,8) as the_geom'
-      else
-        schema.map{|c| %Q{"#{c[0]}"} }.join(',')
-      end
+      select_sql = schema.map { |column|
+        name, type = column
+        if name == THE_GEOM
+          "ST_AsGeoJSON(the_geom,8) as the_geom"
+        elsif type == "date"
+          %{CAST("#{name}" AS text) AS "#{name}"}
+        else
+          %{"#{name}"}
+        end
+      }.join(',')
       # If we force to get the name from an schema, we avoid the problem of having as
       # table name a reserved word, such 'as'
-      row = user_database["SELECT #{select} FROM #{qualified_table_name} WHERE cartodb_id = #{identifier}"].first
+      row = user_database["SELECT #{select_sql} FROM #{qualified_table_name} WHERE cartodb_id = #{identifier}"].first
     end
     raise if row.nil?
-    row
+
+    db_schema = schema.map { |i| i.first(2) }.to_h
+    row.map { |name, value|
+      if db_schema[name] == 'date' && !value.nil?
+        value = DateTime.parse(value)
+      end
+      [name, value]
+    }.to_h
   end
 
   def run_query(query)
--- a/app/models/table.rb
+++ b/app/models/table.rb
@@ -27,8 +27,8 @@ class Table
 
   # TODO Part of a service along with schema
   # INFO: created_at and updated_at cannot be dropped from existing tables without dropping the triggers first
-  CARTODB_REQUIRED_COLUMNS = %{cartodb_id the_geom}.freeze
-  CARTODB_COLUMNS = %{cartodb_id created_at updated_at the_geom}.freeze
+  CARTODB_REQUIRED_COLUMNS = %w{cartodb_id the_geom}.freeze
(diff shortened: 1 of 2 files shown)


javitonino:
Please CR @gfiorav

app/models/table.rb
```diff
@@ -1061,17 +1061,29 @@ def records(options = {})
...
+      }.join(',')
       # If we force to get the name from an schema, we avoid the problem of having as
       # table name a reserved word, such 'as'
-      row = user_database["SELECT #{select} FROM #{qualified_table_name} WHERE cartodb_id = #{identifier}"].first
+      row = user_database["SELECT #{select_sql} FROM #{qualified_table_name} WHERE cartodb_id = #{identifier}"].first
     end
     raise if row.nil?
-    row
+
+    db_schema = schema.map { |i| i.first(2) }.to_h
```
gfiorav: It's not intuitive what you're doing here. Maybe using a variable other than `i` could help :)

app/models/table.rb
```diff
@@ -1061,17 +1061,29 @@ def records(options = {})
...
+    row.map { |name, value|
+      if db_schema[name] == 'date' && !value.nil?
```
gfiorav: I think `!value.nil?` is equivalent to `value` with `&&`

app/models/table.rb
```diff
@@ -1061,17 +1061,29 @@ def records(options = {})
...
+        value = DateTime.parse(value)
```
gfiorav: It's best not to overwrite params or iteration values. Maybe `parsed_value =`?

app/models/table.rb
```diff
@@ -1061,17 +1061,29 @@ def records(options = {})
...
-      select = if schema.flatten.include?(THE_GEOM)
-        schema.select{|c| c[0] != THE_GEOM }.map{|c| %Q{"#{c[0]}"} }.join(',') + ',ST_AsGeoJSON(the_geom,8) as the_geom'
-      else
-        schema.map{|c| %Q{"#{c[0]}"} }.join(',')
-      end
+      select_sql = schema.map { |column|
+        name, type = column
+        if name == THE_GEOM
+          "ST_AsGeoJSON(the_geom,8) as the_geom"
+        elsif type == "date"
```
gfiorav: Since you use this 'date' in other places: how about extracting it to a constant?

javitonino: The original "date" is defined here https://github.com/CartoDB/cartodb/blob/22c13caf35c3d3a4034fae530b0e87138741681a/config/initializers/string.rb#L144

Yeah, we are monkey-patching String to add a `convert_to_db_type` to all strings. Makes sense ¯\\\_(ツ)\_/¯. I won't add the constant there, not feeling like monkey-patching a constant into String is such a great idea.

Not opposed to adding it to Table and replacing the usages there. It won't cover everything, but better than nothing, though.

gfiorav: > Yeah, we are monkey-patching String to add a convert_to_db_type to all strings

I didn't need this today 🙈 

Ok, better something than nothing!

javitonino:
Ok, did some changes that should address all suggestions. Regarding the review comment, I added a test that should help clarify that.

For reference, postgresql returns:
```
type  |           psql         |  DateTime.parse (this PR) |      Sequel (master)
no_tz | 2016-12-01 06:53:35    | 2016-12-01T06:53:35+00:00 | 2016-12-01T06:53:35+01:00
tz    | 2016-12-02 10:53:37+01 | 2016-12-02T10:53:37+01:00 | 2016-12-02T10:53:37+01:00
date  | 2016-11-22             | 2016-11-22T00:00:00+00:00 | 2016-11-22
```

As you can see, this fixes the no timezone case, and returns a full date time for dates (which is the same behaviour as SQL API and what the frontend expects.

spec/models/table_spec.rb
```diff
@@ -1186,6 +1186,29 @@ def @data_import.data_source=(filepath)
       @user.in_database.run("CREATE TABLE cdb_importer.repeated_table (id integer)")
       expect { create_table(user_id: @user.id, name: 'repeated_table') }.to_not raise_error
     end
+
+    it 'correctly returns dates' do
+      table = create_table(:user_id => @user.id)
```
gfiorav: 👍

app/models/table.rb
```diff
@@ -32,6 +32,7 @@ class Table
   THE_GEOM_WEBMERCATOR = :the_geom_webmercator
   THE_GEOM = :the_geom
   CARTODB_ID = :cartodb_id
+  DATATYPE_DATE = 'date'
```
gfiorav: 👍

javitonino:
I ninja pushed a change fixing Hound, probably didn't refresh on your GH view due to lag, but it should be fine now.

PR #10956, opened 2016-12-02
Test fixes

javitonino:
Fixes a bunch of test issues from https://github.com/CartoDB/cartodb/issues/6754

Mostly, reusing the same username among tests


javitonino:
Please CR @gfiorav (some test fix goodness)

xavijam:
ping pong! should we add this one to the Board?

javitonino:
This one is from a leapfrog (reviewing some failing tests) so I don't think it makes much sense to add to the board. I was waiting for today's leapfrog to ask for review (but all backenders are out today).

PR #10957, opened 2016-12-02
SR (2/5) | Stop using user id in state

### LINKED ISSUE #10955: Stop using user_id in state

gfiorav:
Fixes #10955

Please review @javitonino

### DIFF
--- a/app/controllers/carto/api/state_presenter.rb
+++ b/app/controllers/carto/api/state_presenter.rb
@@ -8,10 +8,10 @@ module Carto
       def to_hash
         {
           id: @state.id,
+          visualization_id: @state.visualization_id,
           created_at: @state.created_at,
           updated_at: @state.updated_at,
-          json: @state.json,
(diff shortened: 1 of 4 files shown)


app/controllers/carto/api/state_presenter.rb
```diff
@@ -8,10 +8,10 @@ def initialize(state)
       def to_hash
         {
           id: @state.id,
+          visualization_id: @state.visualization_id,
```
javitonino: Needed? We always know the visualization_id from an state (it's 1:1), so I don't think it makes much sense to present it.

gfiorav: It's mostly oriented toward backbone models: It's easier for it to be reflected in a property. You're right that the property can be inferred on initialisation with a bit of front logic and that it can't be updated.

Whatever you prefer

javitonino: Hey, if it's being actively used in the frontend, 👍

Issue #10950, opened 2016-12-02
Alphamarker image necessary for heatmap (Pixel) style should be included in assets pipeline


xavijam:
Basically we are generating the current CartoCSS for this type of style:

```css
Map {
  -torque-frame-count: 1;
  -torque-animation-duration: 30;
  -torque-time-attribute: "cartodb_id";
  -torque-aggregation-function: "count(1)";
  -torque-resolution: 4;
  -torque-data-aggregation: linear;
}
#layer {
  marker-width: 45;
  marker-fill: white;
  marker-fill-opacity: 1;
  marker-file: url(http://s3.amazonaws.com/com.cartodb.assets.static/alphamarker.png);
  marker-allow-overlap: true;
  marker-line-width: 1;
  marker-line-color: #FFF;
  marker-line-opacity: 1;
  image-filters: colorize-alpha(#4b2991,#872ca2,#c0369d,#ea4f88,#fa7876,#f6a97a,#edd9a3);
}
```

In on premises will not work because the don't have access to external resources

matallo:
re: https://github.com/CartoDB/cartodb/pull/10960

ivanmalagon:
The fix doesn't pass acceptance. 👎 

The generated URL for the alphamarker gets built with double slash in the URL. It makes the staging server to deny access to the resource.

Fixing it myself.

ivanmalagon:
Deployed again to ded05.

Now the generated URL doesn't have a double backlash and Builder gets the proper .png file for hosted environments.

ivanmalagon:
Build is failing. @javitonino is working on it.

Deploy pending.

javitonino:
Merged master, tests should work this time, sorry for the inconvenience 🙇

xavijam:
💋

ivanmalagon:
Waiting for #10960 test to pass. Once they do, feel free to merge and deploy to Production.

xavijam:
So, things we have to change:

- For SASS, we should add `https` always. Protocol"less" doesn't work for purposes like image rendering in the server (previews).
- For on-premises, we should create the whole url, absolute one will not work as we have described in the previous bullet.

Thanks to @rochoa for those things ^^^.

PR #10960, opened 2016-12-02
Requesting alphamarker image (necessary for heatmaps) from local if needed

xavijam:
There are backend and frontend stuff.
Related with https://github.com/CartoDB/cartodb/issues/10950.


lib/assets/javascripts/cartodb3/editor/style/style-converter.js
```diff
@@ -2,9 +2,9 @@ var _ = require('underscore');
 var camshaftReference = require('../../data/camshaft-reference');
 var Utils = require('../../helpers/utils');
 var InputColorCategories = require('../../components/form-components/editors/fill/input-color/input-categories/input-color-categories.js');
+var S3_ASSET_URL = '//s3.amazonaws.com/com.cartodb.assets.static/';
```
matallo: what about moving it to `app_config.yml`?

lib/assets/javascripts/cartodb3/editor/style/style-converter.js
```diff
@@ -305,12 +305,19 @@ var cartocssFactory = {
...
   var ramp = style.fill.color.range;
   style.fill.size = style.fill.size || { fixed: 35 };
-  style.fill.image = CONFIG.HEATMAP_IMAGE;
+
+  var alphaMarkerBaseUrl = S3_ASSET_URL;
+  if (configModel.get('cartodb_com_hosted')) {
+    alphaMarkerBaseUrl = configModel.get('app_assets_base_url') + '/images/';
+  }
+
+  style.fill.image = 'url(' + alphaMarkerBaseUrl + 'alphamarker.png)';
```
matallo: could we have it as base64 and a template, instead of relying in external files?

matallo:
LGTM, besides the comments

lib/assets/javascripts/cartodb3/editor/style/style-converter.js
xavijam: I ❤️  the idea of using a base64 instead of a simple png. Bad thing is Maps API needs to access the file so if we used a base64 that premise will not be solved.

lib/assets/javascripts/cartodb3/editor/style/style-converter.js
xavijam: That's a good one, but take into account this is a constant and only for Sass, not configurable for any other option, so that's the reason I would not move it to any other place.

matallo: not really convinced, but I can disagree and commit

this is also happening https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/data/user-model.js#L19

if we moved it to the config there would be one single point of change for external assets

in any case, feel free to promote

xavijam: Yeah, I see... bad thing is we have both assets in different place :joy:, let me open a ticket for that in platform 💋 .

ivanmalagon:
Deployed to ded05.

xavijam:
Check that code is there, because I didn't change the version of the package.

ivanmalagon:
Related issue: https://github.com/CartoDB/cartodb/issues/10950

xavijam:
frontend and backend tests passed, thanks @javitonino 💋

PR #10963, opened 2016-12-05
10962 drop html type support

### LINKED ISSUE #10962: Drop html type support for legends
We need to stop accepting `html` types in the backend for legends. All migrations are performed to custom now and all existing html have been migrated (except 3 duplicated legends in 1 layer in prod, not worth it).

javitonino:
Fixes #10962 by dropping support for html legends.

HTML legends are no longer used, instead we use `custom` type with an `html` field.

### DIFF
--- a/app/models/carto/legend.rb
+++ b/app/models/carto/legend.rb
@@ -10,7 +10,7 @@ module Carto
 
     belongs_to :layer, class_name: Carto::Layer
 
-    VALID_LEGEND_TYPES = %(html category bubble choropleth custom custom_choropleth).freeze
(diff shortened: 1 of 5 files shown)


xavijam:
Acceptance needed?

xavijam:
Although we need to merge with master.

javitonino:
Merged master and fixed some tests. I'd say we skip acceptance on this one (there is no way to really test it at acceptance level). I vote for deploy as soon as some backend tests are 🍏

xavijam:
:+1:

PR #10971, opened 2016-12-05
Issue 10950

### LINKED ISSUE #10950: Alphamarker image necessary for heatmap (Pixel) style should be included in assets pipeline

javitonino:
Fixes #10950 (second try)


app/helpers/application_helper.rb
```diff
@@ -70,8 +70,15 @@ def maps_api_template(privacy="private")
   module_function :maps_api_template
   module_function :sql_api_template, :sql_api_url
 
+  def app_assets_base_url
+    asset_host = CartoDB.get_absolute_url(Cartodb.get_config(:app_assets, 'asset_host'))
+    base_url = asset_host.present? ? asset_host : CartoDB.base_url_from_request(request)
+    "#{base_url}/assets/unversioned"
```
xavijam: Not sure about adding `unversioned` string there, if it is the base url it should not be included because there are several assets in different folders... not sure, what do you think?

javitonino: Makes sense, it is also good not to include the version just in case in the future we need to use some versioned assets. Changing... 👍

javitonino:
Testing this in on-premise after I finish the official smokes

javitonino:
Acceptance done in development, opensource, staging and onpremise. Looks 👍 everywhere.

I'd give it one last CR and then (after tests are 🍏), deploy directly. Note that we have to merge this to then onpremises branch as well.

Issue #10805, opened 2016-11-23
Category widget filtering/styling behavior


ernesmb:
### Context
When using a Category widget to filter features, the category style changes dynamically, along with the legend bullets. It does so by assigning the first color in the ramp to the first category in the current selection, second color in the ramp to second selected category, etc

However, this leads to a situation where the map categories change drastically from what was designed in the first place (through `STYLE` tab), and for example red can become green or vice-versa, making the map difficult to read

Having the style adapt dynamically to widget filtering is great, but maybe it makes more sense for histogram widget/choropleth visualizations than for categories. 

### Steps to Reproduce
1. Create a map and assign a category style
2. Add a category widget for the same column
3. Filter data

### Current Result
Map features and legend bullets change dynamically to the current filtering, overriding the category style 

### Expected result
In category widgets, respect the color assigned to each category

### Additional info
I know this isn't an easy topic, as it has been discussed before and there are a lot of things to take into account, but maybe it's worth revisiting

saleiva:
Assigning to this sprint just in case we tackle the auto-style fixes

xavijam:
It is not related with auto-style things.

saleiva:
It should right?

saleiva:
Feel free to change the tags or reassign/whatever

rochoa:
So what we want is to have the old editor behavior when creating a category map, don't we?

saleiva:
yep yep

rochoa:
Does that mean we want to have _static_ legends for categories?

saleiva:
From my understanding, if we have a way to say windshaft render this color for this category, the legend would work, right?

rochoa:
But once your data changes you are trapped to the fixed colors you decided in the first place. You don't want that.

All countries:
https://team.carto.com/u/rochoa/builder/17bb226c-b592-11e6-92b3-0ecd1babdde5/embed

vs

All countries but top 3:
[...]

PR #10972, opened 2016-12-05
Changed way to set categories with turbo-carto in the Builder

xavijam:
We decided to use all the power of Turbo-carto :tm:, but we realized there are some exceptions described in #10805. So we have taken the conservative decision of setting the categories manually in the generation of the CartoCSS.

screen shot 2016-12-05 at 18 50 08

**What should we test?**

We should style per value over string columns:
- Using different geometries (point, line or polygon datasets).
- Setting the style over columns with less than 10 categories.
- Setting the style over columns with more than 10 categories.
- Setting the style over columns with `null` values.

### DIFF
--- a/lib/assets/javascripts/cartodb3/components/form-components/editors/fill/input-color/input-categories/input-color-categories.js
+++ b/lib/assets/javascripts/cartodb3/components/form-components/editors/fill/input-color/input-categories/input-color-categories.js
@@ -12,7 +12,7 @@ var MAX_VALUES = 10;
 // max values + "others" color
 var DEFAULT_COLORS = _.clone(rampList.Prism[MAX_VALUES + 1]);
 
-var queryTemplate = _.template('SELECT <%= column %>, count(<%= column %>) FROM (<%= sql %>) _table_sql GROUP BY <%= column %> ORDER BY count DESC, <%= column %> DESC LIMIT <%= max_values %>');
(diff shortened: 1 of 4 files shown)


nobuti:
Not sure if this is a colateral effect, but now when filtering the legend always shows an `Other` categories, even when there is no other. For example:

screen shot 2016-12-07 at 11 10 16

screen shot 2016-12-07 at 11 10 28

I couldn't spot any other issue, other than that it's working fine. Please let me know if I should deploy it.

xavijam:
Ping @rochoa and/or @dgaubert, because I'm not sure about that.

xavijam:
I've just checked it in production, with the same behaviour (applying a ramp using a color for each category), then filtering and the behaviour is happening too, so it is a current problem. Opening a ticket...

xavijam:
Ticket: https://github.com/CartoDB/Windshaft-cartodb/issues/600
@nobuti in my opinion it is ready to go, bug is present right now in production.

nobuti:
👍  I'll deploy it asap.

nobuti:
Frontend tests passing 🚀

saleiva:
Hi Guys, I don't think that having this legends behaving non dynamically is a good behaviour. I do think legends should be only showing the categories selected. If not they will be quite missleading

Issue #10975, opened 2016-12-06
Color Pop-Up styles are not kept when switching from Values to HTML


ztephm:
### Context
S/B 10521471

Enterprise client wants to use Color Pop-Ups. He sets styles in Values mode, but since he wants the infowindow to contain a clickable url he needs to switch to HTML mode. 

When you switch to HTML mode in a Color Pop-up, the mustache code contains ```{{true}}``` instead of column names. This makes it confusing for the user to know where to add his ```<a href>``` line. Unless there's another way we recommend doing this now?

Also, when you add code to the Pop-Up in HTML mode & click Apply, all styles set in Values mode are automatically reset to default.

### Steps to Reproduce

1. With any dataset click into a map layer and click on the Pop-Up tab (I used a point dataset).
2. In the Pop-Up tab's Click panel choose Color Pop-Up then change width & select more than one column to display. I used:
  ![screen shot 2016-12-06 at 1 18 30 pm](https://cloud.githubusercontent.com/assets/1779444/20937845/a30e8a1e-bbb6-11e6-80b7-c646ab160067.png)

  This infowindow looks like:
![screen shot 2016-12-06 at 1 20 43 pm](https://cloud.githubusercontent.com/assets/1779444/20937887/cf19d690-bbb6-11e6-9ebb-45e03d4be24f.png)

3. Toggle from Values to HTML b/c we want to make the carto.com link clickable.

### Current Result
* Instead of seeing mustache code with ```{{name}}``` and ```{{description}}``` we see ```{{true}}``:

  ![screen shot 2016-12-06 at 1 22 51 pm](https://cloud.githubusercontent.com/assets/1779444/20937969/1dc118b2-bbb7-11e6-811d-958b1a7479dd.png)

  How can the user tell which of these represents the value he wants to make a link out of?

* If we assume the second ```{{true}}``` is the value we want to make clickable and replace it with 
  ```
  <h4 class="CDB-infowindow-title">
  <a href="{{description}}">{{description}}</a>
  ```
  & click Apply, the original style/column choices are not maintained:

  ![screen shot 2016-12-06 at 1 25 22 pm](https://cloud.githubusercontent.com/assets/1779444/20938074/98097fc4-bbb7-11e6-8ba5-5bac4b60f835.png)

### Expected result
* Values panel settings are kept when clicking Apply in HTML view.
* Color Pop-Up mustache code contains column names, consistent with other Pop-Up types.

### Browser and version
[...]

PR #10985, opened 2016-12-07
Fixing problems with custom color and image infowindow templates

xavijam:
It will fix https://github.com/CartoDB/cartodb/issues/10975.

**Steps to test**
- Create a map of points with several columns.
- Add a click popup and select the color one.
- Select several columns.
- Open a popup in order to see it.
- Choose html and check that the html is representing the same thing you chose in the form (visit https://github.com/CartoDB/cartodb/issues/10975 in order to check differences).
- Same for image 💋 .


nobuti:
This work as expected. One thing I noticed (out of scope, happening in production) is that if you sort the selected fields, when the form is render again, the selected fields are put at the bottom. likely outside of the view because of the scroll:

popup-fields

Not sure if we should render those fields at the top.

xavijam:
Ouch, could you open a ticket with that please? Let's try to fix in another ticket. Thanks for that!

nobuti:
On premise failed because an `ECONNREFUSED` trying to download phantomjs. I'm deploying 🚀

PR #10989, opened 2016-12-08
Tested if defined when checking wether is custom template.

### LINKED ISSUE #10988: Hover Pop-Up UI is missing
### Context
S/B 10709224

When client tries to enable Hover Pop-Ups he's unable to see the UI panel with Style and Show items sections - I can replicate, looks like he's taken into the Custom/HTML view automatically.

### Steps to Reproduce

1. Import a points dataset with a string-type column that can be used for categories. Create map.
2. Click into layer & choose STYLE tab. 
3. Click on Fill color bar & choose BY VALUE. Specify the dataset's string-type category column. Leave the default category color palette.
4. Click on the POP-UP tab. Click on HOVER.

### Current Result
Looks like we're taken to the Custom code panel automatically:

screen shot 2016-12-07 at 5 52 04 pm

### Expected result
We should see a HOVER panel like this:

screen shot 2016-12-07 at 6 07 54 pm

### Browser and version
* me: Chrome 55.0.2883.75 on OSX
* client: Widows 10, OS Build 14393.447 | Google Chrome Version 55.0.2883.75 | Microsoft Edge 38.14393.0.0

### .carto file
[largecategories.zip](https://github.com/CartoDB/cartodb/files/638021/largecategories.zip)

* This is really hard to replicate consistently! In first test I started with a polygon layer, then added the points dataset with Categories & was able to replicate when trying the points layer hover Pop_Up...but then I deleted the polygon layer & didn't have the problem any more.

* I was able to replicate when I made a new map that started with a different polygon layer, and the same points dataset as 2nd layer - until I changed the datasets from Private to Public, then I didn't have the problem any more either.

### Additional info
When I have less than 11 categories the hover UI appears as expected:

screen shot 2016-12-07 at 6 16 08 pm

test w/few categories:
[small_test.zip](https://github.com/CartoDB/cartodb/files/637988/small_test.zip)

nobuti:
This PR fixes #10988.

@CartoDB/frontend I'm not an expert on infowindows, but I've observed that sometimes the infowindow model has no `template` attribute. Not sure if this an issue coming from anywhere else, but this makes `isCustomTemplate` method to fail:
```
isCustomTemplate: function () {
    return this.get('template_name') === '' && this.get('template') !== '';
  }
```
In this case, if `template` is not defined as it's happening in this case, this function returns `true`, so it tries to get the content of the custom template, that it doesn't exist. Because Builder thinks it has custom template, it loads the editor view, but it fails because content is undefined.

I feel maybe something else is going on here, and template can't be undefined, but this fixes the issue in the front.

In order to test it, follow the steps described in the issue's description.

### DIFF
--- a/lib/assets/javascripts/cartodb3/data/infowindow-definition-model.js
+++ b/lib/assets/javascripts/cartodb3/data/infowindow-definition-model.js
@@ -178,7 +178,7 @@ module.exports = Backbone.Model.extend({
   },
 
   isCustomTemplate: function () {
-    return this.get('template_name') !== undefined && this.get('template_name') === '' && this.get('template') !== undefined && this.get('template') !== '';
+    return this.get('template_name') === '' && this.get('template') !== '';
   }
 
 });
--- a/lib/assets/javascripts/cartodb3/data/infowindow-definition-model.js
+++ b/lib/assets/javascripts/cartodb3/data/infowindow-definition-model.js
@@ -178,7 +178,7 @@ module.exports = Backbone.Model.extend({
   },
 
   isCustomTemplate: function () {
-    return this.get('template_name') === '' && this.get('template') !== '';
+    return this.get('template_name') === undefined && this.get('template_name') === '' && this.get('template') !== undefined && this.get('template') !== '';
   }
 
 });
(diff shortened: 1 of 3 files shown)


nobuti:
Oops, totally right @javitonino, fixed. Is there anything wrong in having that attribute empty in the first place?

nobuti:
💥  Good catch @javitonino, right again. I revert the changes and default the template to an empty one for new layers.

alonsogarciapablo:
I suggest talking to @matallo about this if you haven't already.

matallo:
oooh, super good catch, accompanying this with a test would be golden

matallo:
👏 🚀

Issue #10724, opened 2016-11-16
being able to have pluggable frontends


javisantana:
We should be able to have different frontends in the builder so each user can use one.

- Each frontend should be able to change both styling (imagines and fonts included), logic and texts (translations). 

- There should be a main frontend (the one we have) and the others should inherit from it. A proposal could be to have a folder based thing
```
assets/carto_base-4.0.1/...
assets/VENDOR-VERSION/...
```

- each user can use different frontend vendor and version and can be changed at runtime
- it should be possible to develop with both at the same time (so grunt dev should take care of all the frontends)
- all the frontends should work with the available backend
- in the ideal world there should be possible to run it with a static html

matallo:
done, later work adding missing folders in https://github.com/CartoDB/cartodb/pull/11099

danicarrion:
A few caveats:

* Only the `cartodb3` folder is included in this process, but still part of the UI lives on the `cartodb` folder.
* Part of the UI lives in Ruby templates too.
* It's easy to add new items to a locale file, but there's no way to modify existing items without virtually copying the whole file.

I've worked around these problems, but the result is of course something that is not as easy to maintain as expected.

And also, less important, the `watch` task seems to hang after a while, at least for me, so I need to be restarting it manually quite often.

javisantana:
hey @danicarrion could you point to a ruby template where you'd need to change stuff?

matallo:
as per previous discussions `app/views/admin/shared/_footer.html.erb`

jorgesancha:
guys, I'm creating a new issue with the things that Dani mentions to prioritize it before it's too late... Otherwise BCG will easily go down the same road of not being able to merge the master branch

cc/ @saleiva @xavijam

danicarrion:
also: `app/views/admin/shared/_private_header.erb` and I guess there'll be others as we keep going.

thanks!

jorgesancha:
One comment @danicarrion :

> In respect to other assets, only CSS seems to be included, but not fonts or images for instance.

Not sure I understand this one. If you can change HTML and CSS, what prevents you from referring to other images and/or font-faces? Can you provide an example?

matallo:
fonts overrides are definitely included in the bundle

we left images out as Builder has no images in the stylesheets (it can be easily replicated, actually I implemented it and discarded it in the end)

danicarrion:
sorry fonts are indeed included, my bad

danicarrion:
[...]

PR #11022, opened 2016-12-12
pluggable frontends

matallo:
- [x] js
  - [x] watch
- [x] css
  - [x] watch
- [x] locale
  - [x] watch
- [x] tests
  - [x] watch

---

re: https://github.com/CartoDB/cartodb/issues/10724

after some research this is the proposed approach, and file structure:

**tl;dr** we move the current assets dirs to another dir, create a new one for client overrides, and generate the mix in the current assets dir (thus we don't change build tasks much)

I made it work for js in ded04, and are working in the rest of tasks

## inspiration

most of my inspiration come from newly released [Jekyll themes](https://jekyllrb.com/docs/themes/), where the assets (and even content files) come in the form of gem, and then the name of the theme is set in the config (see attached)

screen shot 2016-12-05 at 15 50 45

this way you are not exposing the underlying complexity of the UI, if you want to override any of those files you just have to recreate the file structure in the root path, and that will overwrite the files in the build

I'd like to do this using a npm package and a URL for the frontend, you can use any frontend, and also override the contents of that package

differences aside, I also like the approach of Calypso https://developer.wordpress.com/calypso/ where there is a REST API connected to every different client (it might be a browser URL, a desktop app, or an embedded view within a terminal)

---

## current status

taking into account we would need to break this into attainable milestones, the trailhead is **we have a folder where the assets are built**, being that folder an url, or compiled locally

currently the case is we have it under the versioned folder of the `asset_host` so, for future cases, the final produced assets will be, as they are now, accessible in `{{ asset_host }}/assets/{{ version }}/*`

## how to proceed

> There should be a main frontend (the one we have) and the others should inherit from it. A proposal could be to have a folder based thing

I'll call them:
- carto core frontend UI (or in the description `carto_base`)
- carto client frontend UI (or in the description `vendor`)

> it should be possible to develop with both at the same time (so grunt dev should take care of all the frontends)

as a first step, we'd copy the contents of the **carto core frontend UI** (from `lib/assets/core/javascripts/cartodb3/`) to the assets folder (`lib/assets/javascripts/cartodb3/`), then the **carto client frontend UI** (`lib/assets/client/javascripts/cartodb3/`) files will be copied there too, overwriting those with the same filename and making available the new ones, but having accessible the original core files to extend from

finally the resulting mix is _browserified_ into the destination folder

I'm adding a `watch` task when files in the **carto client frontend UI** are modified they are copied to the assets folder, and the currently existing watch task will detect those modifications and launch the bundles, I'm working on the watch task to do so with changes in the core files

> each user can use different frontend vendor and version and can be changed at runtime

This will be the sample setup, for example:

- `lib/assets/core/javascripts/cartodb3/editor.js` for the core frontend UI
  - copied to `lib/assets/javascripts/cartodb3/editor.js`
- file in `lib/client/assets/javascripts/cartodb3/editor.js` for the client frontend UI, extending from `lib/assets/core/javascripts/cartodb3/editor.js`
  - copied to `lib/assets/javascripts/cartodb3/editor.js`, overwriting the existing  

having the carto core frontend UI in the former, and the carto client frontend UI in the later (I really like _the rails way_ of having /lib and /app, wich really resonates with our use case, in the future we should remove these folders as we don't deal with assets internally in rails but we could replicate in our frontend)

- the bundle where that file is required will be built in `public/assets/{{ version }}/javascripts/editor3.js`
  - if `lib/core/assets/javascripts/cartodb3/editor.js` changes, it will be copied to `lib/assets/javascripts/cartodb3/editor.js` and that will trigger the bundle again 
  - if `lib/client/assets/javascripts/cartodb3/editor.js` changes, it will be copied to `lib/assets/javascripts/cartodb3/editor.js` and that will trigger the bundle again 

the bundled `public/assets/{{ version }}/javascripts/editor3.js` will be accessible on its own `asset_host` or simply be self hosted

in a nutshell we use the `asset_host` as the final frontend UI differenciator, and the `version` to update them. Both core frontend UI and client frontend UI can share versions number, or can totally diverge in the future

this is replicated for stylesheets, and locale folders

## further steps

ideally, and in the same line of thinking of

> in the ideal world there should be possible to run it with a static html 

as previously said, I'd like to see the **carto core frontend UI** being an npm package, with its own published dist and src files, ready to be consumed by the SaaS once they are published or available for other developers to build new client frontends (that eventually would be registered and published, too)

---

ps: I've intentionally tried to avoid the term theme, hence the longer term


matallo:
In my last changes I made the core/client strategy only for cartodb3 folder in `lib/assets/javascripts` as watch would be unbearably slow otherwise

we're hitting a wall against asset generation, is still very slow but is the best compromise I could find 😞 

also, regarding a question from @xavijam 

> why `lib/assets/core/javascripts` and not  `lib/assets/javascripts/core` for example?

as I explain in the introduction, we do so in order to "generate the mix in the current assets dir", so we don't have to change much the current task structure

xavijam:
Let's review that watch task later @matallo, don't worry, we will improve it little by little 💪 , great job.

matallo:
screen shot 2016-12-13 at 13 14 33

client stylesheets work like a charm, also watch

(note the favicon in the capture)

matallo:
done the locale (with the watch, which is quite slow but tolerable, as it doesn't change that much)

working on making the tests pass now

matallo:
🚀

Issue #11010, opened 2016-12-12
Map not refreshed after editing line or multipolygon feature.


ivanmalagon:
### Context
After editing manually the geometry of a line, the map is not refreshed. We still see the same line as before.

Same thing happened to me when editing a multipolygon.

![smoke line bug](https://cloud.githubusercontent.com/assets/1078228/21101085/c72bc1d0-c077-11e6-9e41-090f310da3fb.gif)

### Steps to Reproduce
Please watch the GIF above for a clip of the bug.

1. Edit a custom line.
2. Move the line points to get a different line.
3. Change the name field.
4. Save the line.
5. The drawn line is the previous one. You can check it editing the line again: the geometry is the modified one.
6. After refreshing the line is properly drawn.

### Additional info

#SMOKE_MG12
#SMOKE_MG14

CC @luisbosque @xavijam

luisbosque:
When you edit the line, the geometry appears updated. However if you edit the line and before saving you edit the form too (for example the Name field) the geometry doesn't get updated. You need to reload in order to see the updated geometry

xavijam:
PR: https://github.com/CartoDB/cartodb/pull/11029

xavijam:
Merge it for on-premises release branch please 💋 .

javitonino:
On premise PR: https://github.com/CartoDB/cartodb/pull/11035

Waiting for tests to pass, and doing acceptance on the on-premise branch, as the merge was a bit difficult and I may have broken it.

javitonino:
Merged to onpremise branch!

xavijam:
💋

PR #11029, opened 2016-12-13
Store all attributes changes when a feature is edited until it is saved

xavijam:
It will fix https://github.com/CartoDB/cartodb/issues/11010. Basically we store all attributes changed when it is edited before saving it. After that, changes history should be empty. 

**STTest**
1. Create a map of lines|points|polygons (take a dataset with several columns with data please).
2. Try to edit a feature.
3. Change the geometry (move the point, or any vertex of the line|polygon).
4. Edit any field of the feature form.
5. Save the changes.
6. Map should be refreshed.

cc @matallo


matallo:
👏 👏 👏

lib/assets/javascripts/cartodb3/data/feature-definition-model.js
```diff
@@ -125,10 +144,12 @@ var FeatureDefinitionModel = Backbone.Model.extend({
 
   fetch: function (options) {
     options = options || {};
+    this.off('change', null, null);
```
javierarce: Why the `null`, `null` params in the `this.off` function? Wouldn't it be better to write `this.off('change', this._onChange, this);`?

javierarce: 💃

xavijam:
Don't remove the branch when it is done.

Issue #11003, opened 2016-12-09
Misspelled & confusing message when trying to edit feature on layer with Analysis


talos:
### Context

Clicked on a feature in a layer that has an analysis applied

### Steps to Reproduce
*Please break down here below all the needed steps to reproduce the issue*

1. Create empty layer
2. Add a point
3. Add an analysis
4. Click the point

### Current Result

See picture below:

<img width="421" alt="screen shot 2016-12-09 at 2 12 23 pm" src="https://cloud.githubusercontent.com/assets/415831/21061363/081c177c-be1a-11e6-86a8-14f433367703.png">

### Expected result

* Message should read "Feature editing is disabled..." not "Feature edition is disabled..."

* According to https://github.com/CartoDB/cartodb/issues/8844, there is not yet an official way to flatten a layer.  The only way to do the edit is to export & download the dataset, then reimport it.  The current language seems to reflect a feature that is in the works, which is very confusing (and made me hunt around the interface for a while looking for a "flatten", which also makes the functionality of "collapse" very confusing.)  Some possible solutions:
  - Second sentence should just be removed.
  - Second sentence should say "To edit the data, export this layer and import it as a new layer."

### Browser and version

Firefox 50.0.2

### .carto file

[untitled_table_37 map (on 2016-12-09 at 19.19.03).zip](https://github.com/CartoDB/cartodb/files/642982/untitled_table_37.map.on.2016-12-09.at.19.19.03.zip)

xavijam:
It is true that we should not offer a feature that it is ongoing :(.

talos:
👍 we just need to make sure we change it again when #8844 is live.

PR #11043, opened 2016-12-14
11003 Edit feature confusing message 

ivanmalagon:
Related to issue #11003 

The text now says what it's 

screen shot 2016-12-14 at 10 25 15


lib/assets/test/spec/cartodb3/locale/en.json.spec.js
```diff
@@ -0,0 +1,10 @@
+var en = require('../../../../locale/en.json');
+
+describe('en.json', function () {
```
ivanmalagon: cc @CartoDB/frontend 

I've added test for the locale translation JSON.

Perhaps it's a bit overkilling but it allows us to check two things:

- A key exists.
- The text is the expected one.

At least, I think the first check it's an important one because if not present the UI can show only the key and not the text.

The second check is useful for critical texts or, like in this PR, to assure that a typo has been fixed.

xavijam:
Ooooooook!

xavijam:
Tests are not passing.

lib/assets/test/spec/cartodb3/locale/en.json.spec.js
ivanmalagon: I think that would be better so we can test also the string interpolation.

You're right. If in our tests we check that the locale gets translated there is no need for the above tests. Indeed, it'll be better.

can of worms

PR #11048, opened 2016-12-14
Shows a custom message when the longitude is out of bounds

### LINKED ISSUE #10881: Longitude is incorrect when adding a point
### Context
*Please explain here below what you were doing when the issue happened*
Adding a point to the map
 
### Steps to Reproduce

1. Add a point to San Francisco using the geometry point tool 
2. Check Longitude of point. 
3. Admire red warning box and extra character.

### Current Result
*Please describe here below the current result you got*
Longitude has an extra character, preventing it from being added
 

### Expected result
*Please describe here below what should be the expected behaviour*
Click 'Add' button to add point to map and point appears. 

### Browser and version
*What internet browser (Chrome, Firefox, etc) and version was you using and version* 
Chrome MacOS Version 54.0.2840.98 (64-bit)
 
### .carto file
*Don't forget to rename .carto file to .zip file to be able to upload to github*
I can't add the point, so ther is no .carto file
 

### Additional info
*Please add any information of interest here below*

point_add_bad_lon_start

@saleiva

javierarce:
### How to test this feature

point

1. Create a map
2. Make sure you are in a small zoom level (1 or 2, for example)
3. Place a point in one of the maps of the sides

a. Check that the error message in the form is not the default one (`Must be a valid longitude`) but   `Longitude is out of bounds [-180, 180]`
b. Check that the custom error still appears whenever you introduce a different erroneous value (eg: a letter).

Closes #10881

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/layers/edit-feature-content-views/edit-feature-geometry-point-form-model.js
+++ b/lib/assets/javascripts/cartodb3/editor/layers/edit-feature-content-views/edit-feature-geometry-point-form-model.js
@@ -17,6 +17,9 @@ var getCoordinatesFromGeoJSON = function (geoJSON) {
   }
 };
 
+var MIN_LNG = -180;
+var MAX_LNG = 180;
+
 module.exports = EditFeatureGeometryFormModel.extend({
 
   initialize: function () {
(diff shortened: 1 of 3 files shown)


javitonino:
Shouldn't we also try to automatically fix longitude when it comes from a click on the map? It seems strange to me that we display an error after what seems like a successful click on the map. Don't know if this is easy to do, but I think it would be nice to have.

matallo:
wow, excellent use of isNumber in the validation! LGTM

also, I believe we should do the same with latitude, I'm creating a new issue for that, as it fixed as it is and is a different one

@javitonino we do so for latitude, so it can be done (but out of the scope for this issue, feel free to create an new one)

PR #11056, opened 2016-12-14
11049 embed error instead of 404

### LINKED ISSUE #11049: Render private embed view when visualization does not exists (instead of redirecting to 404)
### Issue

We've found that when a map is removed by the user the embed present in an external site shows our generic 404 page. This is an issue because of two things:

- First an embed 404 page should have another design more adapted as it is present in an external source.
- Second, GA record doesn't specify what kind of object isn't available.

### Solution

- Render a different view when the error comes from the embeds controller.
- Set the GA code for embeds in that view, so metrics will follow the correct path.

cc/ @beloneys @arianaescobar @xavijam @juanignaciosl

javitonino:
Fixes #11049

### DIFF
--- a/app/controllers/admin/visualizations_controller.rb
+++ b/app/controllers/admin/visualizations_controller.rb
@@ -31,9 +31,11 @@ class Admin::VisualizationsController < Admin::AdminController
   before_filter :get_viewed_user, only: [:public_map, :public_table]
   before_filter :load_common_data, only: [:index]
 
-  before_filter :resolve_visualization_and_table, only: [:show, :public_table, :public_map,
-                                                         :show_organization_public_map, :show_organization_embed_map,
-                                                         :show_protected_public_map, :show_protected_embed_map]
+  before_filter :resolve_visualization_and_table,
+                :ensure_visualization_viewable,
+                only: [:show, :public_table, :public_map,
+                       :show_organization_public_map, :show_organization_embed_map,
+                       :show_protected_public_map, :show_protected_embed_map]
 
   before_filter :resolve_visualization_and_table_if_not_cached, only: [:embed_map]
   before_filter :redirect_to_builder_embed_if_v3, only: [:embed_map, :show_organization_public_map,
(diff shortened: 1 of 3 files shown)


app/controllers/admin/visualizations_controller.rb
```diff
@@ -434,6 +436,10 @@ def embed_forbidden
     render 'embed_map_error', layout: false, status: :forbidden
   end
 
+  def embed_not_found
+    render 'embed_map_error', layout: false, status: :not_found
```
gfiorav: do you think there is a way to avoid all the `layout: false`?

app/controllers/admin/visualizations_controller.rb
```diff
@@ -517,6 +526,7 @@ def resolve_visualization_and_table_if_not_cached
     @cached_embed = embed_redis_cache.get(@table_id, is_https)
     if !@cached_embed
       resolve_visualization_and_table
+      embed_not_found if disallowed_type?(@visualization)
```
gfiorav: why did you abstract this method if it's only used here?

app/controllers/carto/builder/public/embeds_controller.rb
```diff
@@ -61,22 +61,26 @@ def load_state
...
-              return(render 'admin/visualizations/embed_map_error', status: 403)
+              render_embed_error(403)
             else
-              return(render 'admin/visualizations/embed_map_error', status: 404)
+              render_embed_error(404)
             end
           end
         end
 
+        def render_embed_error(status)
```
gfiorav: minor: it's pretty trendy to use keyword arguments for render stuff these days. 

```ruby
def render_embed_error(status:)
  render('admin/visualizations/embed_map_error', status: status)
end
```

so you can call 
```ruby
render_embed_error(status: 404)
```

up to you

app/controllers/admin/visualizations_controller.rb
javitonino: I refactored prior to finding usages, I'll defactor it!

app/controllers/admin/visualizations_controller.rb
javitonino: Can't remove it, or this happens:

screenshot_20161214_184711

javitonino:
Round 2!

Issue #11018, opened 2016-12-12
The icon carousel of the fill component shows a very long horizontal scroll + minor CSS issue


javierarce:
### Context

The icon carousel of the fill component shows a very long horizontal scroll.

![infinite](https://cloud.githubusercontent.com/assets/4933/21104038/45126ec0-c085-11e6-8220-5200bf10e343.gif)
 
### Steps to Reproduce
*Please break down here below all the needed steps to reproduce the issue*

1. Enable the icon-styling feature flag.
2. Go to the style tab.
3. Open the fill component.
4. Go to the icon tab and have a look to the icon carousel.

Watch the animation above to understand what's the problem. 
PS: This same problem is occurring with the basemp carousel, BTW.

Once that is done, there's a minor style issue with the underline of the image (probably due to the height of the SVG icon):

![](https://cloud.githubusercontent.com/assets/2141690/21105342/278575b4-c08a-11e6-9b62-589e87d96ede.png)

xavijam:
Working like a charm, acceptance done!

xavijam:
Tested in different browsers and like a charm as I said above ^^^.

xavijam:
BTW, tested in ded02.

PR #11061, opened 2016-12-14
fix perfect scrollbar width in carousel

matallo:
re: https://github.com/CartoDB/cartodb/issues/11018

better set the min width in the perfect scrollbar initialization than in the CSS

scroll


xavijam:
Hey, could we fix the basemaps list scroll? It is the same bug...

matallo:
fixed in the same PR 🙌 

scroll

xavijam:
Grande!

xavijam:
Running on-premises tests again, acceptance passed.

PR #11079, opened 2016-12-16
Analysis notification

### LINKED ISSUE #10944: Analysis queue is not showing states correctly
Analysis verbose in Builder is not correctly showing states, as is reading "enqueued" when its actually running as per the api. 

You can download the .carto here https://drive.google.com/open?id=0B0nG__Cq6i-7elRjenRZazZoZW8

This happens when solving node D3

nobuti:
This PR fixes #10944.

### How to test it

- [ ] Check that creating an analysis, the notification is right when the api is retunrning `running` state.

In ded02.

### DIFF
--- a/lib/assets/javascripts/cartodb3/editor/layers/analysis-views/analysis-notifications.js
+++ b/lib/assets/javascripts/cartodb3/editor/layers/analysis-views/analysis-notifications.js
@@ -39,11 +39,11 @@ var AnalysisNotifications = {
 
   _addStatusChangedNotification: function (analysisNode) {
     var status = analysisNode.get('status');
-    if (status === STATUS_WAITING) {
+    if (status === STATUS_WAITING || status === STATUS_PENDING) {
       this._addAnalysisWaitingNotification(analysisNode);
     }
-    if (status === STATUS_PENDING || status === STATUS_RUNNING) {
-      this._addAnalysisPendingNotification(analysisNode);
+    if (status === STATUS_RUNNING) {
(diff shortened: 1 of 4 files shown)


nobuti:
I did the same question back in the day, and if I don't remember wrong, it's a matter of perception, `pending` transmit a negative sense. But I could be mistaken. @saleiva could you confirm is't ok?

javitonino:
Well, I'm just suggesting to display the same message (`"waiting": "Analysis %{nodeId} enqueued...",`) for both pending and waiting analysis. And use `"pending": "Analysis %{nodeId} running...",` for running analysis only.

But anyway, it's a minor detail. Also, I'm not sure what the difference between `pending` and `waiting` is, so maybe I'm saying something very wrong.

matallo:
LGTM, I agree on having the discussion (but out of the scope for this issue, tho)

I wouldn't group waiting and pending if we don't want it to have a negative connotation, but expose it in the UI with a different message

what is the difference between pending and running, then?

javitonino:
Talked with out friendly engine overlords.
`waiting` is waiting for a dependent analysis to finish, `pending` is waiting for start (queued), `running` is running.

So I guess we should display the queued message for both `waiting` and `pending`, and just use running for the case where it's actually running.

nobuti:
Done, ready for another pass.

javitonino:
Tests are broken, moving back to dev

nobuti:
Fixed.

matallo:
LGTM

PR #11088, opened 2016-12-16
Recreate layer including legends as attributes.

### LINKED ISSUE #11037: Reordering layers hide legends
legend reorder

### Steps to Reproduce
1. Edit a map with several layers.
2. Go to the top layer.
3. Create a legend. In the GIF above I created one for color and another one for size.
4. Check that the legend is visible.
5. Go back to the layer list.
6. Go to the bottom one.
7. Create a legend for that layer.
8. Check that now we have a legend for three different values.
9. Go back to the layer list.
10. Reorder the layers. Grab the top one and move it one position down, for instance (like in the GIF above).
11. The two legends of the top layer disappear!!
12. If you edit that layer, you can notice that the legends are still there, but they're not visible anymore.

nobuti:
This PR fixes #11037.

When moving a layer, we silently remove the layer and insert it again in the layers collection. When this happens, we were copying the all the layers attributes, but not the legends of this layer. So cartodb.js thinks it has no legends to show. This PR solves it, adding the layer's legends to the attributes when the new layer is created in cartodb.js.

### DIFF
--- a/lib/assets/javascripts/cartodb3/deep-insights-integrations.js
+++ b/lib/assets/javascripts/cartodb3/deep-insights-integrations.js
@@ -47,6 +47,7 @@ var F = function (opts) {
   this._widgetDefinitionsCollection = opts.widgetDefinitionsCollection;
   this._editorModel = opts.editorModel;
   this._editFeatureOverlay = opts.editFeatureOverlay;
+  this._legendsDefinitionCollection = opts.legendDefinitionsCollection;
 
   this._layerDefinitionsCollection.each(this._linkLayerErrors, this);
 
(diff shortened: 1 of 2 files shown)


lib/assets/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -47,6 +47,7 @@ var F = function (opts) {
   this._widgetDefinitionsCollection = opts.widgetDefinitionsCollection;
   this._editorModel = opts.editorModel;
   this._editFeatureOverlay = opts.editFeatureOverlay;
+  this._legendsDefinitionCollection = opts.legendDefinitionsCollection;
```
ivanmalagon: Only a comment. Nothing to do with the review of the task.

There's a module `required-opts.js` [https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/helpers/required-opts.js](url) that does all the required options stuff and leaves the code cleaner.

An example of use is in `layer-content-view.js` [https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/editor/layers/layer-content-view.js#L50](url)

It'd be great to apply it along the way of touching files that could benefit from it.

nobuti: Yep, I know and I'm glad you mention it, I'm using it a lot on auto-styles fixes PR, so go ahead and do a CR :)

nobuti: I'll add the awesome require-opts helper, but please, keep going with the acceptance.

ivanmalagon: It was just a reminder for next tasks. This issue is already being deployed to production. 😆

PR #11122, opened 2016-12-22
Autostyle fine tune

nobuti:
This PR fixes:
- [x] It keeps the autostyle button on screen with viewport minor than 1200px.
- [x] It fix an issue for histogram widgets, when they are created, the styles for custom ramps are not exist unless you try to edit the widget. 

### How to test it
- [x] Check that autostyle button remains visible for small screens (less than 1200px actually).
- [x] Check that options button remains visible for small screens (less than 1200px actually) being in the private builder.
- [x] Create an histogram widget and autostyle next creation.


xavijam:
Just two small details for small screens, auto-style tooltip for applying and removing appears over the button, it lets click over it but looks weird.

screen shot 2016-12-27 at 11 04 21
screen shot 2016-12-27 at 11 04 37

xavijam:
Also options button doesn't appear, is it on purpose?

app/assets/stylesheets/editor-3/editor.css.scss
```diff
@@ -129,9 +129,12 @@
   }
 }
 
-@media (max-width: 1280px) {
+@media (max-width: 1200px) {
   .Editor {
     .CDB-Widget-options {
+      display: block;
+    }
+    .CDB-Widget-actions {
```
xavijam: I feel we should enable widget actions (menu button) when screen is smaller than 1200px, shouldn't we @saleiva @piensaenpixel ?

screen shot 2016-12-27 at 11 14 03

xavijam:
It closes https://github.com/CartoDB/deep-insights.js/issues/500

xavijam:
Hey @MariaCheca @javierarce and @rochoa, these changes will take last changes from carto repository, all ok with that?

xavijam:
Ok, let me fix the menu thing, why not?? :D

xavijam:
@saleiva 

Small dropdown, with 180px of width, 11px of font size, and 8px-12px of padding.

screen shot 2017-01-05 at 11 48 18

yay or nay?

xavijam:
So, toggle button out and stretch dropdown?

xavijam:
Bad thing about removing toggle button is, we should not remove it because user could use it in a big screen, and then when the screen is made smaller, you could not change that widget state.

saleiva:
But we only remove from THAT layout, right? It makes no sense the way it works on when being in the horizontal strip

javitonino:
Still deploying, but a quick comment on the old version (which probably still applies and is not directly related to this PR). The horizontal widget bar in the builder has infinite scroll, I can scroll to the right forever. It works fine in the embed.

xavijam:
Firefox?

javitonino:
Both Firefox and Chrome

javitonino:
I am testing with a window of size 1162 x 1155

xavijam:
About toggle button -> https://github.com/CartoDB/cartodb/issues/11192.
And finally about scroll -> https://github.com/CartoDB/cartodb/issues/11194.

And deploying to ded03 finally.

javitonino:
Looking good ✅ 

One final nitpick I noticed is that the alignment of the remove autostyle tooltip for the last widget is still a bit weird, but is very minor.

screenshot_20170105_133833

Issue #11127, opened 2016-12-23
StrongPasswordValidator escapes characters unnecessarily 


gfiorav:
The development is done in: https://github.com/CartoDB/cartodb/pull/11128

This issue can go straight to CR.

### Context

Since some characters are escaped unnecessarily, they're companion `\` is also taken as a valid character. Also, the space character wasn't properly escaped and not counted as a symbol

NOTE: This was fixed as part of another PR, but was extracted to it's own.

### Steps to Reproduce

1. Try to signup for an org with Strong passwords activated
2. Choose a password containing `\` and complying with all other requirements.

### Current Result

The password is accepted.

### Expected result

The password shouldn't be accepted.

### Acceptance:

Create an org with strong passwords and test a few different ones to make sure that requirements are met (length, number of numbers or symbols, etc).

javitonino:
We are finally not merging this. The reported issue was only a problem when suggesting password (which is not merged), but password verification works fine.

PR #11128, opened 2016-12-23
11127 strong password fix

gfiorav:
Fixes https://github.com/CartoDB/cartodb/issues/11127

### DIFF
--- a/lib/carto/strong_password_validator.rb
+++ b/lib/carto/strong_password_validator.rb
@@ -8,10 +8,10 @@ module Carto
     DEFAULT_MIN_SYMBOLS = 1
     DEFAULT_MIN_LETTERS = 1
 
-    SYMBOLS = %w({ } [ ] , . < > ; : ‘ “ \? \/ \| \ ` ~ ! @ # $ % ^ & \* ( ) _ - \+ =).freeze
+    SYMBOLS = %w({ } [ ] , . < > ; : ‘ “ ? / | \  ` ~ ! @ # $ % ^ & * ( ) _ - + =).freeze
 
     def initialize(min_length: DEFAULT_MIN_LENGTH,
-                   max_length: DEFAULT_MAX_LENGTH,
+                   max_length: min_length > DEFAULT_MAX_LENGTH ? min_length : DEFAULT_MAX_LENGTH,
                    min_letters: DEFAULT_MIN_NUMBERS,
                    min_symbols: DEFAULT_MIN_SYMBOLS,
                    min_numbers: DEFAULT_MIN_LETTERS)
@@ -56,6 +56,19 @@ module Carto
       message << " and #{errors.last}"
 
       message
+    end
+
+    def suggest
+      chars = [*'a'..'z', *'A'..'Z'].sample(@min_letters) +
+              [*0..9].sample(@min_numbers)                +
+              SYMBOLS.sample(@min_symbols)
+
+      padding = [@max_length - chars.length, 0].max
+      padding.times do
+        chars << [*'a'..'z', *'A'..'Z', *0..9, *SYMBOLS].sample
+      end
+
+      chars.shuffle.join
     end
 
     private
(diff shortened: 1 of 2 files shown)


lib/carto/strong_password_validator.rb
```diff
@@ -8,10 +8,10 @@ class StrongPasswordValidator
     DEFAULT_MIN_SYMBOLS = 1
     DEFAULT_MIN_LETTERS = 1
 
-    SYMBOLS = %w({ } [ ] , . < > ; : ‘ “ \? \/ \| \ ` ~ ! @ # $ % ^ & \* ( ) _ - \+ =).freeze
+    SYMBOLS = %w({ } [ ] , . < > ; : ‘ “ ? / | \  ` ~ ! @ # $ % ^ & * ( ) _ - + =).freeze
 
     def initialize(min_length: DEFAULT_MIN_LENGTH,
-                   max_length: DEFAULT_MAX_LENGTH,
+                   max_length: min_length > DEFAULT_MAX_LENGTH ? min_length : DEFAULT_MAX_LENGTH,
```
javitonino: Suggestion: `[min_length, DEFAULT_MAX_LENGTH].max`

lib/carto/strong_password_validator.rb
```diff
@@ -58,6 +58,19 @@ def formatted_error_message(errors)
...
 
+    def suggest
+      chars = [*'a'..'z', *'A'..'Z'].sample(@min_letters) +
+              [*0..9].sample(@min_numbers)                +
+              SYMBOLS.sample(@min_symbols)
+
+      padding = [@max_length - chars.length, 0].max
+      padding.times do
+        chars << [*'a'..'z', *'A'..'Z', *0..9, *SYMBOLS].sample
+      end
```
javitonino: You can do `chars += [*'a'..'z', *'A'..'Z', *0..9, *SYMBOLS].sample(padding)` instead of looping. The behaviour is not the same (my option does not generate repeated characters), so yours is a little bit more random.

spec/lib/carto/strong_password_validator_spec.rb
```diff
@@ -92,9 +92,35 @@ module Carto
...
+    it 'for min letters' do
+      @validator = Carto::StrongPasswordValidator.new(max_length: 20)
+    end
+
+    it 'for min numbers' do
+      @validator = Carto::StrongPasswordValidator.new(max_length: 20)
+    end
+
+    after(:each) do
+      @validator.validate(@validator.suggest).should be_empty
```
javitonino: I think this should check `@validator.errors` instead of relying on `validate` returning it. Just for consistency with other tests. Maybe `validate` should return `true`/`false`?

Something here feels a bit off, but it's far from major, so as you wish.

lib/carto/strong_password_validator.rb
gfiorav: The promblem with passing a number to `sample` is that if that number is bigger than the amount of elements in the array it returns the all the elements of the array since it doesn't repeat.

javitonino:
Well, we should not dump it, we need at least the special characters part, right?

About suggest itself, I don't know if we will have a use case for it, as we use activation links instead of one-use password. So I guess we can remove it via git revert, so the code is still in the history if we need it in the future. What do you think?

gfiorav:
I'm looking at it now (after vacay) and I'm realising that the escape issue was probably only affects when creating a suggested password, not when reading it. Oh well, whatever: stay or leave. Sorry for making you review this on a false premise

lib/carto/strong_password_validator.rb
javitonino: 😆 For those times when you need an 80-char password (longer than a line of code :trollface:)

gfiorav: hahaha hey, let's not be limited if there's no restriction!

lib/carto/strong_password_validator.rb
gfiorav: totally

javitonino:
Double-checked that verification works, archiving this, we can always necrobump it if we feel we need it.

PR #11165, opened 2017-01-02
Get the table_name the right way

javitonino:
The `options attribute` only exists upon synchronization with backend, as the in-memory structures are flattened (and `table_name` is a root attribute).

Fixes https://github.com/CartoDB/cartodb-management/issues/4758


lib/assets/core/javascripts/cartodb3/components/onboardings/analysis/analyses/filter-by-node-column.tpl
```diff
@@ -1,3 +1,3 @@
 <p class="CDB-Text Onboarding-description">
-  <%- _t('analyses-onboarding.filter-by-node-column.description', { name_of_the_layer: source.attributes.options.table_name }) %>
+  <%- _t('analyses-onboarding.filter-by-node-column.description', { name_of_the_layer: source.get('table_name') }) %>
```
xavijam: There is always a table_name?

xavijam:
In my opinion, if we can prevent this problem with a test would be awesome.

javitonino:
I've been trying to write a test but it is getting messy and I prefer to do a sanity check before continuing.

I think the issue here is that the model passed to the analysis onboarding view comes directly from cartodb.js (via deep-insight-integration), so it doesn't have exactly the same attributes as the builder model, depending on when the model gets synced. So that's why in the original issue, just refreshing the builder fixes the problem, as the cartodb.js model gets populated with `options.table_name`, which doesn't happen on layer creation.

Maybe a more correct fix would be to ensure that the model is synced after creating the layer/analysis, so the cartodb.js copy also has that attributes.

Anyway, after taking a step back and looking at the whole problem, I have a different suggestion on how to fix this. This is the onboarding text (emphasis mine):
> Filter by layer allows you to filter the data in your layer by propagating the filters in the **'table_name'** layer. Now if you filter on widgets added to the attached layer, the results in this layer would be also filtered.

My suggestion is changing the table_name for the layer letter (e.g: 'C'), as I think it makes it more clear, specially as this analysis is often used when having two layers based on the same dataset, so I think the layer letter makes it easier to know where the filter will be applied (as the table name may be shared between two layers). It is also more generic, as the table_name idea only works when applying this analysis on top of a source node, while the layer letter is more generic. What do you think? @xavijam

xavijam:
I like your proposal, and it is true we should not rely on datasets anymore, because an analysis can be performed over another analysis or a source. 👍

javitonino:
screenshot_20170103_110932

PR #11173, opened 2017-01-03
M4768 fullstory

javitonino:
Closes https://github.com/CartoDB/cartodb-management/issues/4768

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -217,6 +217,12 @@ module ApplicationHelper
     end
   end
 
+  def insert_fullstory
+    if Cartodb.get_config(:fullstory, 'org').present?
+      render(partial: 'shared/fullstory', locals: { org: Cartodb.get_config(:fullstory, 'org') })
+    end
+  end
+
   ##
   # Checks that the precompile list contains this file or raises an error, in dev only
   # Note: You will need to move config.assets.precompile to application.rb from production.rb
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -218,7 +218,7 @@ module ApplicationHelper
   end
 
   def insert_fullstory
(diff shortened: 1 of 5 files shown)


xavijam:
No api key needed?

javitonino:
No, just the org ID, which must be added to app_config. I'll document it before deploying.

app/helpers/application_helper.rb
```diff
@@ -217,6 +217,12 @@ def insert_hubspot_form(form = 'newsletter')
     end
   end
 
+  def insert_fullstory
+    if Cartodb.get_config(:fullstory, 'org').present? && current_user && current_user.account_type.casecmp('FREE') == 0
```
xavijam: Hound rulez!

config/app_config.yml.sample
```diff
@@ -68,6 +68,8 @@ defaults: &defaults
     app_keys:
       editor: ''
       embeds: ''
+  fullstory:
+    org: ''
```
gfiorav: Very late shot, but it serves as an example. I think we should start commenting or documenting options that are "off by default" and stop including them equal to `''`. I've recently modified the `fetch_config` method to only return if present, but it feels a bit weird to have this here. Just for the record, anyway.

xavijam: I like your idea gfiorav 👍

javitonino:
Some notes for acceptance:
 - Test that the snippet is included and sessions recorded on fullstory panel
 - The snippet should be included only for FREE users
 - Please check that logging in as different users in the same browser (do not use private browsing) works as expected (each sessions is correctly matched to user)
 - Only builder sessions should be logged (no public pages, dashboard, account settings)

config/app_config.yml.sample
javitonino: It's an easy change, I'll do it along with the NEWS (which I wrote but forgot to commit).

Issue #11177, opened 2017-01-04
Fallback mechanism for Locales in "pluggable frontends" functionality


jorgesancha:
Continuing with the "pluggable frontends" work implemented in #10724, we should improve how locale literals are overwritten.

Currently it's easy to add new items to a locale file, but there's no way to modify existing items without virtually copying the whole file.

It should be possible to update just a few literals without modifying the core file and without needing to duplicate the whole core file. It should be a fallback mechanism, such that, if a literal is present in the customized version, it gets picked up; otherwise, it falls back to the core locale file to look for the literal.

Please review and update issue if necessary @danicarrion @matallo @xavijam @javisantana

matallo:
it's true that it was easy to _append_ new i18n strings, but not that easy to replace (without losing siblings)

I modified a bit the example, and replaced the underscore's `_.defaults` with jquery's `$.extend` which can do a deep extend

find the implementation in https://github.com/CartoDB/cartodb/pull/11180

PR #11180, opened 2017-01-04
locales example, deep extend [DO NOT MERGE]

matallo:


lib/assets/client/locale/index.js
```diff
@@ -0,0 +1,9 @@
+var Locale = require('../core/locale/index');
+var _ = require('underscore');
+var $ = require('jquery');
+
+module.exports = _.extend(Locale, {
+  sw: require('../core/locale/sw.json'),
+  es: require('../core/locale/es.json'),
+  en: $.extend(true, Locale.en, require('./en.json'))
```
javierarce: Probably silly question but we are using two different extends here, one from jQuery and other one from underscore… couldn't we use the underscore one in the two cases?

matallo: it's a pity not, underscore can't deep extend the object, (we can with jquery as commented in https://github.com/CartoDB/cartodb/issues/11177#issuecomment-270345988)

javierarce: Oh, true! Then maybe we don't need to load the whole jQuery library for one method. I'm thinking we could have a little custom function to handle the deep extension:

Since our json files aren't very complex, maybe the first accepted answer could be useful in our case? http://stackoverflow.com/questions/14843815/recursive-deep-extend-assign-in-underscore-js

What do you think? Maybe too much hassle?

matallo: seems legit, also removed the other core locales, if we are extending that `index.js` object there's no need to do that

matallo:
actually this branch is not intended to be merged, but as an example for client locales, we'll worry when some work is done with locales first

/cc @danicarrion

gfiorav:
@matallo so we can move this back out of acceptance?

matallo:
yeah, as long as it works it could be closed, no need to deploy (just leave the branch for reference)

PR #11183, opened 2017-01-04
Check for custom view templates before loading default ones

### LINKED ISSUE #11178: Fallback mechanism for erb templates in "pluggable frontends" functionality
Continuing with the "pluggable frontends" work implemented in #10724, we should enable also customizing in a similar fashion rails .erb templates, because part of the UI also leaves there.

These are some of the templates that @danicarrion has needed to update so far:

`app/views/admin/shared/_footer.html.erb`
`app/views/admin/shared/_private_header.erb`

As part of this issue, we should:

1. make inventory of all the templates that have literals
2. implement a fallback mechanism for templates such that it looks first for the template in a designated "client" folder and falls back to the regular template if it is not present

gfiorav:
Fixes #11178 

## Acceptance 

Test

1. In staging, make sure view from builder load as normal
2. In local:

change app_config.yml to read:
```yaml
custom_paths:
  views: ['app/a/path', 'app/anohter/path']
```

Grab a file under `app/views` and replicate it's structure under both directories specified in the `app_config.yml`. Make distinct changes to both files so you can recognise them when rendered (i.e. one file contains on the word "manolo" and the other "escobar"). Now reboot your instance and access the view. 

You should see the file of the leftmost path first. If you remove that file and reboot, the second file should show. Lastly, if you remove all custom files but leave the config, you should see the original unaltered file render. Lastly, check that removing the config also works as normal.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -22,7 +22,11 @@ Development
     * The core frontend is in `lib/assets/core`
     * The customizations are in `lib/assets/client`
     * The end result are in `lib/assets/`
-  * You may also plug backend view templates in `app/views/custom`
+  * You may also plug backend view templates by specifying alternative paths in `app_config.yml`:
+```yaml
+  custom_paths:
+    views: [] # an array of paths were alternate view templates are found
+```
 * Snapshots (backend: #10928) allow to save and share map state.
 * Icon styling through in component (#11005)
 * Allow to set opacity for color ramps (#10952)
--- a/NEWS.md
+++ b/NEWS.md
@@ -23,9 +23,10 @@ Development
     * The customizations are in `lib/assets/client`
     * The end result are in `lib/assets/`
   * You may also plug backend view templates by specifying alternative paths in `app_config.yml`:
+    * Paths are inspected in the order supplied. First valid template is used. Default path is always inspected last.
 ```yaml
   custom_paths:
-    views: [] # an array of paths were alternate view templates are found
+    views: [] # an array of paths were alternate view templates are located.
 ```
 * Snapshots (backend: #10928) allow to save and share map state.
 * Icon styling through in component (#11005)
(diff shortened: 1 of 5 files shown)


javitonino:
Code looks good ✅. Some comments though:
 - Do you think it is worth it to add tests for this? I am unsure myself, as they give little value and they are probably messy to do. So I am thinking no.
 - Maybe we should move all views to `app/views/core`, and set the views array to `['app/views/client', 'app/views/core']` to be consistent with the JS asset structure. We could even move it to `lib/assets/core/templates` so everything that can be overridden is in the same place, but maybe that is too much.

gfiorav:
1. I though about it. I didn't see a clear way of doing it without testing a stub o writing a file. Not worth retiring a file IMHO.

2. Uhm no, I strongly oppose :D. The things is that in front end there is no default way of doing things. I'm back we do, and fighting the framework over this is a no-no for me. Plus the added complexity of dealing with "core" in the standard case which is the principal one. I'm all for "adding stuff" but all against "changing defaults" to suit the needs.

javitonino:
2. I'm not a fan of creating `core` either, but though it might be worth it. Oh well, nothing that can't be solved with docs.

And a new one. What about we put the custom path(s) in the app_config? This way, for normal use it can be an empty array (so it only looks in `app/views`) and if you want to customize, you can set the directory you want. A boolean toggle might be enough too, but I think some kind of switch is good to save the bit of performance of having to look up templates in two places.

gfiorav:
That would be nice indeed! I tested something like that in an earlier version. Couldn't decide on the scheme or if it was worth it. How about:

```yaml
custom_paths:
  views: []
```

It's a bit over engineered to be honest, and it's not like we expect any more templatable custom things, but we could use this for any future multipurpose RAILS default path overwrites.

javitonino:
I will pre-ask the question since I'm seeing lots of activity around it. Answer when you ask for CR again, as you know it is coming anyway 😆 Why not `get_config`? Doesn't it work here?

gfiorav:
Fixed!

PR #11206, opened 2017-01-10
Start using GeoJSON for ArcGIS datasource

### LINKED ISSUE #4722: update icons sprite

ethervoid:
We face a problem with multiline geometries with multiples features,
they get merged in one multiline feature instead of remain as multiple
features in the same multiline geometry.

The problem seems to be the use of a JSON structure that is not well
recognized by ogr2ogr when is uploaded to Postgres

The solution is to make the ArcGIS query requesting GeoJSON as the output
format that is perfectly recognized by ogr2ogr


ethervoid:
Not useful because it's only supported by >= 10.4.x ArcGIS servers.

PR #11208, opened 2017-01-10
Fix js tests

ivanmalagon:
We lost both the npm-tests and Builder frontend tests.

### npm-tests

After the pluggable frontend shift, no proper paths in grunt copy path were done. Check for the fix in  `lib/build/tasks/copy.js` file

### frontend tests

After a refactor, we introduced a circular dependency in `legends-metadata` to use it in `legend-factory`. That provoked that Jasmine stopped test run when referencing any legend model.

The fix has been reverting to old `legend-factory`


javierarce:
Nice detective work!

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/legend/legend-factory.js
```diff
@@ -105,6 +117,11 @@ var manager = (function () {
       }, this);
     },
 
+    hasMigratedLegend: function (layerDefModel) {
```
nobuti: This is an old code, there is no `html` type anymore, so no need for this method.

Issue #10726, opened 2016-11-16
Icons on custom legends [v0.1]


urbanphes:
Possibility to add an icon for custom legends.

It would not be possible to implement it until #10746 is implemented.
Design ticket: https://github.com/CartoDB/design/issues/304

xavijam:
This will be done after https://github.com/CartoDB/cartodb/issues/10452.

matallo:
support for cartodb.js already done https://github.com/CartoDB/cartodb.js/pull/1545

nobuti:
PR seems good, but I need more context to understand it fully. I left a comment, waiting for response.

nobuti:
@matallo I'm putting it back to development, I left some comments it would be nice to address, but feel free to obvious them if you consider it's not worth.

nobuti:
Works like a charm 🚀

xavijam:
Tested with and without the feature flag enabled?

nobuti:
Yes, working great in both scenarios.

xavijam:
:+1:

PR #11221, opened 2017-01-11
icons on custom legends

matallo:
close https://github.com/CartoDB/cartodb/issues/10726

**steps to test this:**

- create a map, add a custom legend, and pick an icon in any of the items
- check the icon (and the color) changes both in Builder and the map

screen shot 2017-01-10 at 14 59 41

this needs to be reviewed and merged first:

- [x] cartodb.js support https://github.com/CartoDB/cartodb.js/pull/1545


lib/assets/core/javascripts/cartodb3/components/form-components/editors/legend/category-item.js
```diff
@@ -1,3 +1,47 @@
 var Backbone = require('backbone');
 
-Backbone.Form.editors.List.CategoryModel = Backbone.Form.editors.NestedModel;
+Backbone.Form.editors.List.CategoryModel = Backbone.Form.editors.NestedModel.extend({
```
nobuti: Could you explain a bit the logic here? I'm not an expert in Backbone.Form, at all, I need to understand it. The construction seems a bit weird, rendering a form inside an form editor.

matallo: yeah, it was hard for me to create a mental model of this too, but it is a nested form, I'm just extending it in order to have access and pass `usermodel`, `configmodel`, and `modals` options

the original model comes from backbone-forms https://github.com/powmedia/backbone-forms/blob/master/src/editors/nestedmodel.js

nobuti: I feel this is too complicated. Why not passing those params as model options as we do in https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/editor/widgets/widgets-form/widgets-form-factory.js#L48-L55? This options should go in the form model we are using https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/legend/form/legend-nested-category-definition-form-model.js

nobuti: After a deeper look, you are absolutely right, there is no way to pass options to the nested model, and this is the way to go.

lib/assets/core/javascripts/cartodb3/components/form-components/editors/legend/category-item.js
```diff
@@ -1,3 +1,47 @@
...
+
+  initialize: function (options) {
+    Backbone.Form.editors.Base.prototype.initialize.call(this, options);
+    Backbone.Form.editors.Base.prototype._setOptions.call(this, options); // Options
+
+    if (!this.form) throw new Error('Missing required option "form"');
+    if (!options.schema.model) throw new Error('Missing required "schema.model" option for NestedModel editor');
+
+    this._userModel = this.options.userModel;
```
nobuti: I think is too specific, I'm thinking in a case when we need other/more dependencies.

lib/assets/core/javascripts/cartodb3/components/form-components/editors/legend/category-item.js
```diff
@@ -1,3 +1,47 @@
...
+
+  render: function () {
+    // Get the constructor for creating the nested form; i.e. the same constructor as used by the parent form
+    var NestedForm = this.form.constructor;
+
+    var data = this.value || {};
+    var NestedModel = this.schema.model;
+
+    // Wrap the data in a model if it isn't already a model instance
+    var modelInstance = (data.constructor === NestedModel) ? data : new NestedModel(data, {
```
nobuti: Related to comment above, I'd go with something like this, if possible:
```
var modelInstance = (data.constructor === NestedModel) ? data : new NestedModel(data, this.options));
```

nobuti:
Let's fix the conflicts.

matallo:
YAY

nobuti:
Frontend test passing 🚀

PR #11228, opened 2017-01-11
Category widget aggregation other than count

### LINKED ISSUE #11198: Category widget is failing when aggregating not by COUNT
1. Simple map
2. Add category widget 
image
3. change the aggregation to any other than COUNT, over a numeric column
image
4.The widget is now like
image

Looks like the reason is that the NULL values in the column fuck up the aggregations. We should set null values aside before aggregating

cc @saleiva

nobuti:
This PR fixes #11198. It updates dependencies to bring https://github.com/CartoDB/cartodb.js/pull/1546 once it's merged.


javitonino:
Looks nice! Acceptance is 🍏 👏 

Instructions for deploy?

I guess we need at least to reset the package.json, but not sure if we need to update any dependencies.

nobuti:
Yes, at least cartodb.js needs to be updated, I'm going to update the files.

nobuti:
🚀

PR #11241, opened 2017-01-12
11239 auto creation and strong passwords

### LINKED ISSUE #11239: Strong password validation blocks google signup
### Context
If an organization has google signup and strong password enabled, user creation fails.

### Steps to Reproduce
1. Create an organization with google signup an strong password validation
2. Try to sign up

### Current Result
Error because the auto-generated password is not strong enough. 

### Expected result
User is created

javitonino:
Fixes #11239 

Skips checking password strength when the password is automatically generated.

Acceptance: Create an organization with google/github signup and strong password validation. Try to signup for the organization using google/github. Also check for regressions: that a user/password signup still checks password strength.

### DIFF
--- a/lib/user_account_creator.rb
+++ b/lib/user_account_creator.rb
@@ -137,7 +137,7 @@ module CartoDB
           validate_organization_soft_limits
         end
 
-        if requires_strong_password_validation
+        if requires_strong_password_validation?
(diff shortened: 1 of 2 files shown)


javitonino:
Back to 🍏 (hopefully). Damn those dependencies between tests.

PR #11242, opened 2017-01-12
Display signup errors as flash message if form is not displayed

### LINKED ISSUE #11236: No feedback about signup errors when org has only Google enabled
### Context
Related to #11238

When logging in or signing up into an organization which does not have User/pass login enabled, there is no feedback about errors. This is because the errors are shown as field validation errors, but those fields are not rendered if not needed.

We need an alternate way to display errors when user/password is not used.

### Steps to Reproduce
1. Create an organization and disable all login forms except Google Plus
2. Try to create an account with something that will fail (e.g: email not in whitelisted domains)

### Current Result
Signup fails (correctly), but no user feedback 

### Expected result
User feedback

We may need @CartoDB/design here to provide a way to display errors for orgs with just the button. Current page looks like:
screenshot_20170112_114532

Errors are only displayed in the fields, like this:
screenshot_20170112_114421

So when user/password is disabled, we don't have a way to see the errors.

We already display some generic errors as a flash message:

screenshot_20170112_114604

javitonino:
Fixes #11236 

During signup, if an error occurs, it is displayed in the form. However, this form is hidden when user/pass auth is disabled, so in that case, we display it in the flash message.

Acceptance: Follow STR in the ticket :)

screenshot_20170112_163833


app/views/shared/_flash_message.html.erb
```diff
@@ -1,7 +1,11 @@
-<% if cookies[:flash] || flash[:success] || flash[:error] %>
-  <div class="CDB-Text FlashMessage <%= 'FlashMessage--success' if (flash[:success]) %><%= 'FlashMessage--error' if flash[:error] %>">
+<%
```
MariaCheca: I'm just being picky, but I would prefer to have the code in separated blocks of `<% whatever %>`to make it more legible :)

app/views/shared/_flash_message.html.erb
```diff
@@ -1,7 +1,11 @@
...
+  error_message = local_assigns[:error_message]
+  if cookies[:flash] || flash[:success] || error_message || flash[:error] %>
+  <div class="CDB-Text FlashMessage <%= 'FlashMessage--success' if (flash[:success]) %><%= 'FlashMessage--error' if error_message || flash[:error] %>">
     <div class="u-inner">
-      <p class="FlashMessage-info <%= 'success' if (flash[:success]) %><%= 'error' if flash[:error] %>"><%= raw cookies[:flash] || flash[:success] || flash[:error] %></p>
```
MariaCheca: Are those parenthesis above the `flash[:success]` necessary? Same thing in the line below.

MariaCheca:
❤️

MariaCheca:
Acceptance passing 🇩🇰

PR #11256, opened 2017-01-16
11238 choose alternate name google signup

### LINKED ISSUE #11238: Cannot choose an alternate name in google signup if only google is used
### Context
Similar to #11236

When using google signup and only this signup method is enabled, there is now way to choose an alternate username if it fails. This only affects org signups (Central works).
 

### Steps to Reproduce
1. Create an organization with only google auth enabled
2. Try to sign in with a google account that will conflict with an existing username (but not email)

### Current Result
No way to choose alternative username:
screenshot_20170112_120632

 ### Expected result
Display username field to be able to change it

javitonino:
Fixes #11238 and some related errors:
- Signing up into an org with Google/GitHub should ask for username in case of conflict.
- Markup errors (div nesting was incorrect)
- JS errors when loading Google (spinner didn't stop, and it only worked the first time you pressed the button, further times did nothing)
- JS error when email/password fields where hidden (username/password is disabled)

**Acceptance**
In an organization with username/password auth disabled, should test all combinations of:
 - Login and signup, via organization login/signup page (no need to test central), with username conflict
 - Google and Github

Signups should be done with username conflicts. To achieve this, create a user with the same username as your Google/GitHub account, but a different email. I suggest signing up using your google/GH account, and then changing the email associated to that account in Central, but you can do this in anyway you see fit.

Detail checks:
 - Opening signup and login page should require clicking on Google button.

**Deploy**
Don't forget to bump package.json

### DIFF
--- a/app/helpers/signup_helper.rb
+++ b/app/helpers/signup_helper.rb
@@ -1,7 +1,13 @@
 module SignupHelper
   def duplicated_username_prompt?
     # Ask for a different username for Google or GitHub signups with username errors
-    @user.try(:errors).try(:[], :username).present? && (google_sign_in? || github_sign_in?)
+    only_username_errors? && (google_sign_in? || github_sign_in?)
+  end
+
+  def only_username_errors?
+    errors = @user.try(:errors)
+    return false unless errors.present?
+    @user.errors.select { |_, v| v.present? }.keys == [:username]
   end
 
   def google_sign_in?
--- a/app/helpers/signup_helper.rb
+++ b/app/helpers/signup_helper.rb
@@ -1,0 +1,14 @@
+module SignupHelper
+  def duplicated_username_prompt?
+    # Ask for a different username for Google or GitHub signups with username errors
+    @user.try(:errors).try(:[], :username).present? && (google_sign_in? || github_sign_in?)
+  end
+
+  def google_sign_in?
+    @user.google_sign_in && @organization.auth_google_enabled
+  end
+
+  def github_sign_in?
+    @user.github_user_id && @organization.auth_github_enabled
+  end
+end
--- a/Makefile
+++ b/Makefile
@@ -299,6 +299,7 @@ SPEC_HELPER_MIN_SPECS = \
 	spec/requests/carto/api/organization_assets_controller_spec.rb \
 	spec/lib/carto/assets_service_spec.rb \
 	spec/lib/carto/organization_assets_service_spec.rb \
+	spec/lib/carto/storage_options/local_spec.rb \
 	spec/models/carto/username_proposer_spec.rb \
 	$(NULL)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -45,6 +45,7 @@ Development
 * Fixed a bug with vizjson invalidation (#11092). It was introduced in #10934
 * Refactor Layer model (#10934)
 * Fix bugs where legends where being hidden by reordering layers (#11088)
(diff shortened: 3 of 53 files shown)


app/helpers/signup_helper.rb
```diff
@@ -0,0 +1,20 @@
+module SignupHelper
+  def duplicated_username_prompt?
+    # Ask for a different username for Google or GitHub signups with username errors
+    only_username_errors? && (google_sign_in? || github_sign_in?)
+  end
+
+  def only_username_errors?
```
juanignaciosl: Minor: `?` is for boolean methods, but this returns username errors. Renaming it to `username_errors` would be more accurate.

app/helpers/signup_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+    errors = @user.try(:errors)
+    return false unless errors.present?
+    @user.errors.select { |_, v| v.present? }.keys == [:username]
+  end
+
+  def google_sign_in?
+    @user.google_sign_in && @organization.auth_google_enabled
```
juanignaciosl: Can `@user.google_sign_in` and `@user.github_user_id` be trusted for new signups?

app/helpers/signup_helper.rb
javitonino: But this returns `true` or `false`. It's true if there were errors during creation, but only related to username. It's false if there were no errors, or there were errors with other things (e.g: duplicated email, or organization is out of seats).

app/helpers/signup_helper.rb
javitonino: These are `nil` the first time you are loading the `/signup` page, and are only set when doing a POST to signup. In that case, they contain the token for the service (which is re-checked if the user retries with a different username).

Not sure if that answers your question.

app/helpers/signup_helper.rb
juanignaciosl: You are right. My bad. 1€ punishment.

app/helpers/signup_helper.rb
juanignaciosl: It does! 👍

MariaCheca:
After doing the acceptance in staging I've seen some things that I think we should improve:

- Error messages in the signup/login form are not rendered properly:
screen shot 2017-01-16 at 15 50 36
screen shot 2017-01-16 at 16 01 35

- These are two weird error messages after an organization owner removes a member of the org from the administration panel. The first one appears only when the user has been introduced in the org from the signup form, no matter the way (user+password, Github or Google). The second one if the user has been created from the admin:
screen shot 2017-01-16 at 15 58 56
screen shot 2017-01-16 at 16 07 51

- In the "your account is active" view the logo is not the right one (it is a spinner, which doesn't make sense there because it shouldn't be giving any "wait for it" feedback to the user, so it should be a static image), we need the `carto-logo-darkblue.svg` (located in the same route as the other one). In case this design is used in both normal and org session views (dark and light design respectively) it would be better to tell the template when to render what resource (for the dark design `carto-logo.svg`). Plus, it would be nice to have the text lines aligned to the center:
screen shot 2017-01-16 at 15 54 23

Anyhow the rest looks ok 👍

javitonino:
Oh shit, I thought I got that fixed. On it. Thanks 💋

javitonino:
Fixed first and third ones. I'm looking at and reporting the second one as a different ticket (will link here afterwards).

Issue #11252, opened 2017-01-16
Add an analysis will fail if autostyle is applied


ramiroaznar:
### Context

* This bug was found doing the "Risk Analysis Demo".

### Steps to Reproduce

1. Create a map with two layers (points - A, and polygons - B).
2. Add Intersect analysis to B layer.
3. Add histogram widget from "sum_value".
4. Autostyle.
5. Add Outliers & Clusters analysis.

### Current Result

* An error related to CartoCSS pops up and the analysis fails.

`US Counties: style0:18:2 Invalid code: polygon-fill: ;style0:19:2 Invalid code: polygon-opacity: ;`

![error](https://cloud.githubusercontent.com/assets/5215798/21978723/1889ac70-dbdc-11e6-8ad9-145c15260757.png)

### Expected result

* Analysis should run OK.

### Browser and version

* Google Chrome v54.

### .carto file

* [Risk Analysis Demo (on 2017-01-16 at 10.04.16).carto.zip](https://github.com/CartoDB/cartodb/files/707973/Risk.Analysis.Demo.on.2017-01-16.at.10.04.16.carto.zip)

xavijam:
In fact, after a deep look, if autostyle is selected and we apply a new analysis on top of the layer where the autostyle is applied, the map instantiation will fail. Changing the title @ramiroaznar.

nobuti:
Back to development, some errors with lines raised.

xavijam:
:+1:, taking care of it, thanks man 💋

MariaCheca:
Merged https://github.com/CartoDB/cartodb/pull/11261

PR #11261, opened 2017-01-16
Fixing problem when an auto-styled polygon layer adds a new analysis

xavijam:
Ticket: #11252.

**THINGS TO BE TESTED:**
First, steps to be done before:
- Create a map with a layer of each geometry (points, lines and polygons).
- Add a widget with auto-style for each layer.

I would try to make the thing fail with each type of geometry. So, in short:
- Add the layer.
- Create a widget with autostyle powers.
- Autostyle it.
- Add a new analysis, no matter the type.

Also, I'd test when user re-uses the autostyle styles, in order to check everything works as expected:
- Create a map with one layer (no matter which geometry).
- Add a widget with autostyle powers.
- Autostyle it.
- Go to the layer style.
- Change something in the style.
- Check that autostyle style now is removed.

Check that the autostyle state disappears and a non-desired legend appears in the map.

cc @nobuti

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/data/layer-definition-model.js
+++ b/lib/assets/core/javascripts/cartodb3/data/layer-definition-model.js
@@ -106,7 +106,7 @@ module.exports = Backbone.Model.extend({
   save: function (attrs, options) {
     // If the layer is saved by any circunstance, we should disable
     // autostyle if it was enabled and remove any style from it
-    this.styleModel.resetPropertiesFromAutoStyle();
+    this.styleModel && this.styleModel.resetPropertiesFromAutoStyle();
 
     attrs = _.extend(
       {},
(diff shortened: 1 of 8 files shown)


xavijam:
Fixing them right now, thanks man.

xavijam:
Tests passing 💋

nobuti:
I found an error while testing with lines:

screen shot 2017-01-16 at 18 55 39

STR the error:
- Add a layer with lines.
- Add a category widget, and apply autostyle
- Add an analysis (in my case, a subsample)
- 💥

xavijam:
Found, probably, the source of the problem, doing:

- Add a points layer.
- Add a histogram widget with autostyle powers.
- Go to style tab.
- Apply autostyle from the widget.
- Go to analysis tab.
- Apply an analysis.
- Thing fails.

xavijam:
Deploying to ded03 again, seems like issues are gone, hopefully.

xavijam:
Last problem reported (above ) was related to a problem with the object referencing.

MariaCheca:
Acceptance passing 🌈

PR #11276, opened 2017-01-17
11214 GitHub requires signup confirmation

### LINKED ISSUE #11214: Github login requires email confirmation
When signing up to an organization with Github, we require the email to be confirmed (through CARTO). 

That shouldn't be needed, as GH already verifies emails, so we could skip it.

We should also check if this is related https://github.com/CartoDB/cartodb/issues/10489. It may be because the account creation does the autologin but then requires email activation, and the email activation code used to fail if a session was already present.

javitonino:
Fixes #11214 

Acceptance: Signup into an organization with GitHub. You should be automatically logged in, and you should not receive an email telling you to activate your account (you will receive a welcome one, but it shouldn't ask for activation).


MariaCheca:
Acceptance passing 🇩🇰

Issue #11006, opened 2016-12-12
Implement icons preview list in organization profile


xavijam:
Images for context:

![screen shot 2016-12-12 at 12 31 42](https://cloud.githubusercontent.com/assets/132146/21097984/733eb394-c067-11e6-8e12-35d4092ae13b.png)

![screen shot 2016-12-12 at 12 32 01](https://cloud.githubusercontent.com/assets/132146/21097983/733ccb2e-c067-11e6-85ed-96c6cb96e312.png)

We should show the list of icons that the organization offers to its users. We should manage several things:

- List should show last (20?) uploaded icons.
- If there isn't any icon we should show something, like a hint (state empty).
- Possibility to select and remove one or several icons at the same time: this behaviour should be *implemented independently*, because it will be used in the assets modal (check design ticket).
- When an icon is removed, we should fetch more icons to complete the list.

Things related to design: https://github.com/CartoDB/design/issues/352

ivanmalagon:
Talked with @piensaenpixel and he can help with CSS part in this issue. Ping him when this issue gets started.

MariaCheca:
Hey mates, I would like to do this ticket. Ping me if at some point it gets really urgent and it can't wait more :)

ivanmalagon:
@MariaCheca I'll take it if you don't mind. Icon styling needs some pushing, I think.

ivanmalagon:
One remark / question.

According to the UI, it seems that the user can make any changes and they're not saved until she clicks on `Save changes`.

In the case of icons, if we upload icons and delete them through the list, those changes will be effective even if `Save changes` button is not clicked.

That is, icons are set independently of clicking on `Save`or not. Is this the desired behaviour?

nobuti:
@urbanphes could you please throw some light?

ivanmalagon:
Another implication.

This UI allows to delete a marker file that could be referenced in a number of maps. Since the CartoCSS style saved with the map refers directly to the file, the system won't be able to render it.

I've made a test about this and the user will see this error and won't see any marker in the map.

![image](https://cloud.githubusercontent.com/assets/1078228/21894115/4cbd1b7c-d8de-11e6-8ec8-742b35996d86.png)

Just for you to know. Allowing to delete markers outside the own map that it's using it can break visualizations.

@saleiva @xavijam @jorgesancha

saleiva:
[...]

PR #11292, opened 2017-01-18
11006 org icons preview

ivanmalagon:
Ticket: #11006

IMPORTANT: To test this, a new section should be included in `app_config.yml`. Add `organizations` in `assets`. See the `app_config.yml.sample` in this PR for an example.

preview-icons


app/assets/stylesheets/organization/assets.css.scss
```diff
@@ -0,0 +1,85 @@
+@import '../variables/mixins';
```
xavijam: assets is a bit generic, what about icon-list?

xavijam: assets is a bit generic, what about icon-list?

xavijam: And I feel @piensaenpixel has reviewed/(helped with) this.

config/app_config.yml.sample
```diff
@@ -276,6 +276,10 @@ defaults: &defaults
   assets:
     s3_bucket_name: "tests"
     max_file_size: 5242880 # 5.megabytes
+    organization:
```
xavijam: :+1:

lib/assets/javascripts/cartodb/organization/entry.js
```diff
@@ -110,6 +111,16 @@ $(function() {
       }, this);
     }
 
+    // Icon picker
+    if (this.$('.js-iconPicker').length > 0) {
+      new IconPickerView({
+        el: this.$('.js-iconPicker'),
+        orgId: currentUser.organization.get('id')
```
xavijam: Why do we need the same value in different context?

lib/assets/javascripts/cartodb/organization/icons/delete_icons_modal.jst.ejs
```diff
@@ -0,0 +1,39 @@
+<div class="CDB-Text Dialog-header u-inner">
+  <div class="Dialog-headerIcon Dialog-headerIcon--negative">
+    <i class="CDB-IconFont CDB-IconFont-trash"></i>
+    <% if (numOfIcons > 0) { %>
+    <span class="Badge Badge--negative Dialog-headerIconBadge CDB-Text CDB-Size-small"><%- numOfIcons %></span>
+    <% } %>
+  </div>
+  <h4 class="CDB-Text CDB-Size-large u-mainTextColor u-secondaryTextColor u-bSpace--m u-tSpace-xl">
+    You are about to delete
+    <% if (numOfIcons <= 0) { %>
```
xavijam: Is it possible to delete 0 icons?

lib/assets/javascripts/cartodb/organization/icons/delete_icons_modal.jst.ejs
```diff
@@ -0,0 +1,39 @@
...
+    from your organization.
+  </h4>
+  <p class="CDB-Text CDB-Size-medium u-altTextColor">
+    The deleted 
+    <% if (numOfIcons === 1) { %>
+      icon
+    <% } else { %>
+      icons 
+    <% } %>
+    cannot be recovered, be sure before proceeding.
```
xavijam: I'd warn the user about the possibility to break maps where the icons are being used.

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
```diff
@@ -0,0 +1,372 @@
...
+  },
+
+  render: function () {
+    this.$el.html(this.template());
+
+    return this;
+  },
+
+  _initBinds: function () {
+    this._iconCollection.on('change:selected', this._refreshActions, this);
```
xavijam: this.add_related_model(this._iconCollection);

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
```diff
@@ -0,0 +1,372 @@
...
+    }
+  },
+
+  _addIconElement: function (iconModel) {
+    iconModel.set('index', this._getIconIndex(iconModel));
+    this._renderIcon(iconModel);
+    this._refreshActions();
+  },
+
+  _onAddIconClicked: function (evt) {
```
xavijam: I haven't seen this kind of variable anywhere. I encourage you to use ev for events, e for errors. That's what I do 👅 .

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
```diff
@@ -0,0 +1,372 @@
...
+    this._hideErrorMessage();
+    this.$('#iconfile').trigger('click');
```
xavijam: Why do we need this?, being an id 😱 .

app/views/admin/organizations/settings.html.erb
```diff
@@ -73,6 +73,11 @@
         </div>
       </div>
 
+      <% if current_user.has_feature_flag?('icon-styling') %>
```
xavijam: 👍

lib/assets/javascripts/cartodb/organization/icons/organization_icon_model.js
```diff
@@ -0,0 +1,28 @@
...
+  fileAttribute: 'resource',
+
+  save: function (attrs, options) {
+    options || (options = {});
+    attrs || (attrs = _.clone(this.attributes));
+
+    // Filter the data to send to the server
+    delete attrs.selected;
+    delete attrs.visible;
+    delete attrs.deleted;
```
xavijam: _.omit?

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
```diff
@@ -0,0 +1,372 @@
...
+
+  _onIconUploadComplete: function () {
+    this._numOfUploadingProcesses--;
+    if (this._numOfUploadingProcesses <= 0) {
+      this.model.set('isProcessRunning', false);
+    }
+  },
+
+  _resetFileSelection: function () {
+    this.$('#iconfile').val('');
```
xavijam: Again, why an id and not the .js-class?

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
```diff
@@ -0,0 +1,372 @@
...
+      return;
+    }
+    var limit = Math.min(this._maxIcons, this._iconCollection.length);
+    var numOfSelectedIcons = this._getNumberOfSelectedIcons();
+    var iconText = (numOfSelectedIcons === 1 ? '1 icon selected' : '' + numOfSelectedIcons + ' icons selected');
+    this.$('.js-iconMainLabel').text(iconText);
+
+    if (numOfSelectedIcons === 0) {
+      this.$('.js-iconMainLabel').text('Icons');
+      this._hide('.js-selectAllIcons');
```
xavijam: What about to create a function that let to pass an array or a single value in order to perform the same behaviour instead of running it several times in a row...

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
```diff
@@ -0,0 +1,372 @@
...
+      this._hide('.js-deselectAllIcons');
+      this._hide('.js-deleteIcons');
+      this._show('.js-iconsInfo');
+    } else if (numOfSelectedIcons < limit) {
+      this._show('.js-selectAllIcons');
+      this._hide('.js-deselectAllIcons');
+      this._show('.js-deleteIcons');
+      this._hide('.js-iconsInfo');
+    } else {
+      this._hide('.js-selectAllIcons');
```
xavijam: Dixit ^^^

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
```diff
@@ -0,0 +1,372 @@
...
+    this.$('.js-iconMainLabel').text(iconText);
+
+    if (numOfSelectedIcons === 0) {
+      this.$('.js-iconMainLabel').text('Icons');
+      this._hide('.js-selectAllIcons');
+      this._hide('.js-deselectAllIcons');
+      this._hide('.js-deleteIcons');
+      this._show('.js-iconsInfo');
+    } else if (numOfSelectedIcons < limit) {
+      this._show('.js-selectAllIcons');
```
xavijam: Dixit

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
```diff
@@ -0,0 +1,372 @@
...
+    }
+
+    if (this._iconCollection.length > this._maxIcons) {
+      this._show('.js-viewAllIcons');
+    } else {
+      this._hide('.js-viewAllIcons');
+    }
+  },
+
+  _onDeselectAllIconsClicked: function (evt) {
```
xavijam: Already mentioned the evt thing (this is the last one, not repeating myself).

lib/assets/javascripts/cartodb/organization/icons/organization_icon_view.js
```diff
@@ -0,0 +1,50 @@
...
+    this._template = cdb.templates.getTemplate('organization/icons/organization_icon');
+    this._initBinds();
+  },
+
+  render: function () {
+    this.$el.html(this._template({
+      url: this.model.get('public_url')
+    }));
+
+    this.model.set('visible', true);
```
xavijam: There should be a `return this;` in render method always.
Also, modify a model within a render function is not convenient.

app/assets/stylesheets/organization/assets.css.scss
ivanmalagon: It was named `assets.css.scss` because is a mirror of the file with the same name in `editor3` area. Not exactly a mirror but you get the idea.

You're right in the sense that inside the file all the styles are named `Icon-list whatever` but when we migrate we should refactor looking at the original `assets.css.scss` I would keep the name just for ease of finding it.

Does that make sense?

ivanmalagon: @piensaenpixel reviewed it with me, yes.

lib/assets/javascripts/cartodb/organization/entry.js
ivanmalagon: Do you mean why we're using twice `this.$('.js-iconPicker')` ?

No reason in particular. We could refactor it and assigning the element to a variable but it's not a big deal IMO.

lib/assets/javascripts/cartodb/organization/icons/delete_icons_modal.jst.ejs
ivanmalagon: No. Just a paranoid check in case someone instantiates the modal without assigning a proper value. Just a protection.

But it's not instantiated with 0 icons in this particular code.

lib/assets/javascripts/cartodb/organization/entry.js
xavijam: Forget it because you have already fixed with the last commit.

lib/assets/javascripts/cartodb/organization/icons/delete_icons_modal.jst.ejs
ivanmalagon: You're right.

There's an ongoing conversation about soft deleting the assets. But in the meantime, the behaviour could break some maps so we must warn the user.

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
ivanmalagon: Added.

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
ivanmalagon: Well. I like `ev` as well but @matallo uses `e` for events so we must set a common ground.

I'd go for full names, `event` for events and `error` for errors. What do you think?

xavijam: I like it!

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
ivanmalagon: Changed to `js-inputFile`

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
ivanmalagon: Changed to `js-inputFile`

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
ivanmalagon: Refactored to use comma-separated values (multiple jQuery selectors)

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
ivanmalagon: Refactored to use comma-separated values (multiple jQuery selectors)

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
ivanmalagon: Refactored to use comma-separated values (multiple jQuery selectors)

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
ivanmalagon: Refactored to `event`

lib/assets/javascripts/cartodb/organization/icons/organization_icon_model.js
ivanmalagon: :+1:

lib/assets/javascripts/cartodb/organization/icons/organization_icon_view.js
ivanmalagon: this returned.

Model updated in the render calling.

ivanmalagon:
I forgot about warning the user on the possibility of breaking a map.

Should we go for it or wait until something about soft deleting is decided?

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
```diff
@@ -0,0 +1,364 @@
...
+      iconView.render();
+      iconModel.set('visible', true);
+      this.$('.js-items').append(iconView.$el);
+    }
+  },
+
+  _addIconElement: function (iconModel) {
+    iconModel.set('index', this._getIconIndex(iconModel));
+    this._renderIcon(iconModel);
+    this._refreshActions();
```
nobuti: Is it necessary to make this call in each iteration? We are listening `change:isProcessRunning` and when it stops running, it calls it. And the first action of `_refreshActions` is a `return` if `isProcessRunning`.

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
```diff
@@ -0,0 +1,364 @@
...
+  },
+
+  _getNumberOfSelectedIcons: function () {
+    return this._iconCollection.where({ selected: true }).length;
+  },
+
+  _beforeIconDelete: function () {
+    this._numOfDeletingProcesses++;
+    if (this._numOfDeletingProcesses > 0) {
+      this._runningMessage = 'Deleting icons...';
```
nobuti: Locale?

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
```diff
@@ -0,0 +1,364 @@
...
+  _refreshActions: function () {
+    if (this.model.get('isProcessRunning')) {
+      return;
+    }
+    var limit = Math.min(this._maxIcons, this._iconCollection.length);
+    var numOfSelectedIcons = this._getNumberOfSelectedIcons();
+    var iconText = (numOfSelectedIcons === 1 ? '1 icon selected' : '' + numOfSelectedIcons + ' icons selected');
+    this.$('.js-iconMainLabel').text(iconText);
+
+    if (numOfSelectedIcons === 0) {
```
nobuti: IMO all these show/hide calls should be part of the template, which it should be the result of the state. The result should not contain any data that it's not necessary for the current state.

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
```diff
@@ -0,0 +1,364 @@
...
+
+    if (this._iconCollection.length > this._maxIcons) {
+      this._show('.js-viewAllIcons');
+    } else {
+      this._hide('.js-viewAllIcons');
+    }
+  },
+
+  _onDeselectAllIconsClicked: function (event) {
+    event.preventDefault();
```
nobuti: Why all these `preventDefault`? Those controls should be `button`'s.

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
```diff
@@ -0,0 +1,364 @@
...
+    });
+  },
+
+  _onDeleteIconsClicked: function (event) {
+    event.preventDefault();
+    this._openDeleteIconsDialog();
+  },
+
+  _openDeleteIconsDialog: function () {
+    var numOfIconsToDelete = _.filter(this._iconCollection.models, function (icon) {
```
nobuti: I guess the Backbone version we are using here doesn't have the `filter`method itself, right?

nobuti: But I guess we could use `this._iconCollection.where({ selected: true }).length;` or even better `this. _getNumberOfSelectedIcons()`.

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
```diff
@@ -0,0 +1,364 @@
...
+      numOfIcons: numOfIconsToDelete,
+      okCallback: this._deleteIcons.bind(this)
+    });
+
+    dialog.appendToBody();
+  },
+
+  _deleteIcons: function () {
+    this._hideErrorMessage();
+    var iconsToDelete = _.filter(this._iconCollection.models, function (icon) {
```
nobuti: Same as above.

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
```diff
@@ -0,0 +1,364 @@
+var _ = require('underscore-cdb-v3');
```
nobuti: IMO there is too much logic here, we should break it into subviews when refactor.

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
nobuti: Confirmed, no local :)

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
ivanmalagon: The object is big. I agree. Perhaps we could separate UI logic from uploading / deleting logic somehow.

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
ivanmalagon: Taking care of this in the template would mean to re-render the view for each operation.

I wanted to avoid re-rendering the whole component at all costs. That would mean to download again the `IMG` resources for every rendering and even in the case of getting them from cache the user would see a repainting for big images.

Re-rendering the `IMG` tag was a real pain when changing the icon color in the fill component. That's why all the operations done in this component are done treating directly with DOM instead of relying on the template. That's another reason for getting a lot of logic in this JS file.

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
ivanmalagon: I went for `anchor` instead of `button` because that's the way it's done in that organization dashboard page. I know in Builder it's done with buttons but I wanted to maintain consistency in the organization page.

Like the old spanish saying says: "Wherever you go, do what you see".

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
nobuti: I agree, that's why I suggested to break it into subviews, to render every piece on demand. I understand this is legacy code, so take it for the migration party.

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
ivanmalagon: Unfortunately, there's no locale support in the old codebase. Once we migrate, we should turn these magic strings into templates in `locale.json`

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
nobuti: Boo, when it's not right, it's not right.

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
ivanmalagon: Well spotted. 👀 

Got rid of it here and in other two places that were unnecessary.

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
ivanmalagon: Fixed.

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
ivanmalagon: Fixed.

ivanmalagon:
Added a warning for the message.

image

The copy about this operation affecting maps is taken from deleting layers operations.

Remark that this warning should be modified if we finally go for the 'soft delete' operation. cc @saleiva  @jorgesancha

lib/assets/javascripts/cartodb/organization/icons/organization_icon_picker_view.js
```diff
@@ -228,9 +222,7 @@ module.exports = cdb.core.View.extend({
 
   _deleteIcons: function () {
     this._hideErrorMessage();
-    var iconsToDelete = _.filter(this._iconCollection.models, function (icon) {
-      return (icon.get('selected'));
-    });
+    var iconsToDelete = this._iconCollection.where({ selected: true });
```
nobuti: why not `this._getNumberOfSelectedIcons()`?

nobuti: Dumb, it's not the number.

xavijam:
Moving to acceptance then?

juanignaciosl:
@ivanmalagon if `app_config.yml` should change in production, configuration changes must be deployed, is this already planned?

ivanmalagon:
@juanignaciosl No. Indeed this is my first issue with changes in `config` so I don't know the protocol to follow. Any help on that is appreciated.

juanignaciosl:
@ivanmalagon without configuration, this crashes. Is that ok, or should icon upload be disabled?

juanignaciosl:
PS: I guess that, without a bucket, this works with local storage, isn't it? How come an specific organization bucket is needed? Why not using the same asset bucket as other asset uploads?

ivanmalagon:
This is a question for @gfiorav The endpoint used by this feature was developed in https://github.com/CartoDB/cartodb/pull/11034 There you can find a better explanation about the new config data needed.

Related to buckets, in my machine I was using the same bucket but I had it duplicated. I guess it was developed for having an option for storing this kind of assets in another bucket if needed.

ivanmalagon:
Comments added to `app_config.yml.sample` to give more information about the config possibilities.

PR #11295, opened 2017-01-18
Fixed problem rendering size legends panel

xavijam:
Basically we need all those pending attributes in order to render the panel correctly. Nothing else.
cc @CartoDB/frontend


xavijam:
Acceptance done by @nobuti and me.

PR #11297, opened 2017-01-18
Quota precheck

nobuti:
This PR implements https://github.com/CartoDB/camshaft/issues/238.

### How to test
- [ ] Try to add an analysis with quota, when the user has quota enough to run it. Check that the infoboxes are right.
- [ ] Try to add an analysis with quota, when the user has quota, but the quota isn't enough to run it. In this case, the infobox should show an error message saying that.

Also, buttons in the infoboxes should be vertical aligned.


nobuti:
Ping @CartoDB/frontend here to not pollute the cramshaft's issue. Ready for CR.

xavijam:
Reviewing the implementation and the workflow, I've noticed a JS problem, I feel we should tackle it before keep going with the CR:

- I've deployed changes in RUI servers.
- I've used `jmedina-rui` user for testing the stuff.
- Opened "Failing map with quotas".
- Added a "Enrich from Data Observatory" analysis.
- JS error appears.

Could you take a look before continuing?

nobuti:
The error is due to `User must have user configuration in the config table`, and we handled using a `console.error`. I'm opening the debate about how we should manage this things.

xavijam:
If there is enough quota to run the analysis, we should not say that the operation could cost money:

screen shot 2017-01-19 at 17 11 37
screen shot 2017-01-19 at 17 12 01

xavijam:
When there is no enough credits:

screen shot 2017-01-19 at 17 18 49

xavijam:
hey @rafatower and @rochoa, if there is no enough quota to run the whole analysis, could it be run it or we should not let them to start it?

rafatower:
I wouldn't let it run, cause it will be half-done.

My 2 cents

xavijam:
But, it is possible, isn't it? @saleiva, sorry to disturb you but I feel you should give us your opinion.

saleiva:
I do think we have to go with the yellow option and say something like:

Your current quota might not be enough to enrich all your data. Rows will be set to null when running out of quota.

lib/assets/core/javascripts/cartodb3/data/analyses-services-quota.js
```diff
@@ -0,0 +1,67 @@
+var _ = require('underscore');
```
MariaCheca: @nobuti I think you should answer this question as you took that decision :)

lib/assets/core/javascripts/cartodb3/data/analyses-services-quota.js
```diff
@@ -0,0 +1,67 @@
...
+    this.configModel = options.configModel;
+    this.SQL = new cdb.SQL({
+      user: this.configModel.get('user_name'),
+      sql_api_template: this.configModel.get('sql_api_template'),
+      api_key: this.configModel.get('api_key')
+    });
+
+    this.serviceQuotasCollection = new Backbone.Collection();
+
+    return this;
```
MariaCheca: This is necessary as in the initialisation of the module in `editor.js` it is followed by a `.getAllQuotas()` call, which will need the `this` to be returned to use it.

brittanymicek:
@nobuti @saleiva @xavijam 
Sorry for the delay guys! <3

COPY FOR CONFIRMATION OF ANALYSIS FEE MESSAGE:
**PLEASE CONFIRM YOUR ANALYSIS**
Using this analysis component may incur extra fees.[1] 

COPY FOR QUOTA MESSAGE:
**INSUFFICIENT QUOTA**
We're sorry the current quota is insufficient to enrich your data. Rows will be set to null and analysis may not complete. Please contact us to extend your quota for this function.

NOTES:
[1] Questions: Why the indecisive language around this? Will this analysis result in an extra fee or not? If we aren't sure about that, then we should ADD: "depending on XXX." EX: "Using this analysis component may incur extra fees depending on your accounts quota." (Feel free to use this sentence if it is dependent on the account quota). This helps justify whether or not the fee will occur (we don't know because it varies for all accounts) or the decision to use "may" which is indecisive.

**THE BEST REVISION would be:** 
"Using this analysis component may incur extra fees depending on your account quota. Please review your current quota and/or request a larger allocation.

I agree we shouldn't list how much an account might be charged. Why? It is scary and may keep them from performing the analysis. Also, I imagine this price might change or be dependent on the account type. Additionally, I think by giving them the option to "apply" the analysis after being warned means they have accepted the blind risk of being charged. 

We should also add where they can go to increase the account quota or to view their current account quota.

Thanks!

saleiva:
Hey britt, can we prepare a version of the text that says when the quota
will be automtaically refresh? "Tomorrow"?, "Next 1st, October, 2017"?

brittanymicek:
Sure thing! :) @saleiva When would this notification appear and where?

lib/assets/core/javascripts/cartodb3/data/analyses-services-quota.js
```diff
@@ -0,0 +1,67 @@
...
+      }.bind(this),
+      error: function (error) {
+        var userName = this.configModel.get('user_name');
+        console.error('An error ocurred trying to get the quotas for user ' + userName + ': ' + error);
+      }.bind(this)
+    });
+  },
+
+  getServiceQuota: function (service) {
+    return this.serviceQuotasCollection.findWhere({service: service});
```
nobuti: We are calling the init method when the visualisation is initialised to assure in this point everything is in place.

lib/assets/core/javascripts/cartodb3/data/analyses-services-quota.js
nobuti: IMO the advantage of using collection + fetch is when you use a REST api. So in these kind of scenarios, I tend to make the request out the context of the Backbone object.

lib/assets/core/javascripts/cartodb3/data/analyses-services-quota.js
nobuti: That, to chain calls on the same instance.

brittanymicek:
@saleiva ???

nobuti:
I'm closing this in favor of https://github.com/CartoDB/cartodb/pull/11480

PR #11298, opened 2017-01-18
URI encode more problematic characters

javitonino:
Closes https://github.com/CartoDB/cartodb-management/issues/4756

Encodes more characters. URI::Encode has an unsafe parameter which takes a regex on which characters to encode. By default, it takes al URL safe characters, this changes it to all special characters including things like comma or " which have special meaning in headers (but not in URLs)

**Acceptance**
Rename a map with some special characters and try to export it as .carto. Test at least the following characters: quotes (" and '), space, comma/semicolon (, ;), star (*), spanish characters, chinese characters, emoji.

It's important to test with different browsers. At least Chrome (as it was the one giving issues), but it would be great to cover the major browsers as well.

### DIFF
--- a/app/models/carto/visualization_export.rb
+++ b/app/models/carto/visualization_export.rb
@@ -44,7 +44,10 @@ module Carto
       file = CartoDB::FileUploadFile.new(filepath)
 
       s3_config = Cartodb.config[:exporter]['s3'] || {}
-      s3_config['content-disposition'] = %{attachment;filename="#{URI::encode(file.original_filename.force_encoding('iso-8859-1'))}";filename*=utf-8''#{URI::encode(file.original_filename)}}
+
+      plain_name = header_encode(file.original_filename.force_encoding('iso-8859-1'))
+      utf_name = header_encode(file.original_filename)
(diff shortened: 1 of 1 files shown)


MariaCheca:
The test I made in Chrome for OS X/Windows and Opera for Windows made struggles in the filename.

1. Created a new map and renamed it to **Untitled "map'**
2. Exported the map as a .carto
3. The file was downloaded succesfully but the name is a bit weird: `Untitled _map' (on 2017-01-19 at 10.08.45).carto;filename-=utf-8''Untitled _map' (on 2017-01-19 at 10.08.45).carto`

The rest of the browsers (tested in Chrome, Firefox, Safari, IE (only Windows, the rest in OS X) and Opera) where ok.

javitonino:
Good catch :). Some quotes were missing. I added them back and tested with Firefox + Chrome on Linux and it works fine now. Thanks 💋

MariaCheca:
Yay!

MariaCheca:
Now acceptance is fully passing 🇩🇰

Issue #11308, opened 2017-01-19
Previous style is applied after reopening a map and applying a new style


ramiroaznar:
### Steps to Reproduce
*Please break down here below all the needed steps to reproduce the issue*

1. Create a map selecting `ne_10m_railroads` dataset
2. Style color layer based upon `cartodb_id` field from the UI
3. Back to the main menu
4. Open the map again
5. Change style using the CartoCSS panel:

```css
#layer {
  line-width: 1.5;
  line-color: ramp([cartodb_id], (#00e6a4,#a7e369, #e6de00,#ed8d36,#e60042), jenks);
  line-opacity: 0.9;
}
```

6. Apply (clicking or typing `CTRL+S`)

### Current Result

* A previous CartoCSS appears in the panel and the style associated with it is applied.

```css
#layer {
  line-width: 1.5;
  line-color: ramp([cartodb_id], (#00e6a4,#a7e369, #e6de00,#ed8d36,#e60042), jenks);
}
```

![cartocss](https://cloud.githubusercontent.com/assets/5215798/22115255/90e3c092-de6c-11e6-9414-f3d2899221ee.png)

* Surprisingly, moving with the arrow you can go back and then forward to the "last" CartoCSS and apply again.

### Expected result

* CartoCSS and style should be applied correctly.

### Browser and version

* Google Chrome v54.

### .carto file

* [Untitled Map (on 2017-01-19 at 16.24.09).carto.zip](https://github.com/CartoDB/cartodb/files/717173/Untitled.Map.on.2017-01-19.at.16.24.09.carto.zip)

xavijam:
Thanks @ramiroaznar for your report, already fixed (not deployed yet) with https://github.com/CartoDB/cartodb/commit/6d8fecf6e9cbbd2cdf45fb0fdd8132e5fcdb12e3.

ramiroaznar:
This is still happening. For instance, when changing a popup, it resets the style.

xavijam:
We are still on it @ramiroaznar, we haven't deployed anything yet (check status in the builder project board).

makella:
this is happening to me too.

ivanmalagon:
Deploying to ded02 for acceptance. Build number 1153.

saleiva:
@ramiroaznar fancy passing acceptance?

ivanmalagon:
Deployed in production.

@ramiroaznar 's life will be better from now on.

PR #11315, opened 2017-01-20
Fixing #11308 problem

xavijam:
Ticket: #11308.

The problem is great described about: "Previous style is applied after reopening a map and applying a new style". We were resetting the styles whenever the layer was saved, and it was not correct.

**Things to test:**
- Check the steps mentioned in the ticket above.
- Check if after applying an autostyle, if user makes any cartocss change, autostyle is reseted correctly.


matallo:
LGTM

javitonino:
This appears to work, but I find the behaviour confusing.

If I am editing CartoCSS, apply autostyle and go to the CSS panel, I see the autostyle code. I can edit it, but after saving, my changes are lost and the previous CSS is applied.

out

This looks very confusing to me. Another issue is if you apply a custom CSS, then apply autostyle and go to the style panel (switch to another panel and back) you see the style_properties panel (not CSS). If you disable the autostyle, the custom CSS is applied on the map, but the panel is stuck displaying the style_properties style.

I don't know, this feels very confusing to use, but I cannot point exactly to what's wrong.

xavijam:
Let me take a look buddy!

xavijam:
I've reviewed your concerns @javitonino, I feel the result is 👌 .

lib/assets/core/javascripts/cartodb3/data/layer-definition-model.js
```diff
@@ -104,17 +104,19 @@ module.exports = Backbone.Model.extend({
...
-
-    attrs = _.extend(
-      {},
-      {
-        autoStyle: false
-      },
-      attrs
-    );
+    if (this.get('autoStyle')) {
+      // If the layer is saved by any circunstance, we should disable
```
ivanmalagon: It's `circumstance`, not `circunstance`.

lib/assets/core/javascripts/cartodb3/editor/style/style-view.js
```diff
@@ -77,6 +77,7 @@ module.exports = CoreView.extend({
   _initBinds: function () {
     this.listenTo(this._layerDefinitionModel, 'change:error', this._appendMapsAPIError);
     this.listenTo(this._layerDefinitionModel, 'change:visible', this._onChangeLayerVisibility);
+    this.listenTo(this._layerDefinitionModel, 'change:autoStyle', this._infoboxState);
```
ivanmalagon: Why do we use `listenTo` in this case when usually we use `on` on views listening to model changes?

Not a suggestion to do another thing, just asking if there's an actual reason for this.

lib/assets/core/test/spec/cartodb3/editor/style/style-view.spec.js
```diff
@@ -62,6 +62,7 @@ describe('editor/style/style-view', function () {
 
     this.promise = $.Deferred();
     spyOn(this.userActions, 'saveLayer').and.returnValue(this.promise);
+    spyOn(StyleView.prototype, '_infoboxState').and.callThrough();
```
ivanmalagon: Why spying on the prototype instead of the actual method?

Just like above. Only asking for the reason for that to see if it's a common way to do it or not. (First time I run into this in the codebase.)

lib/assets/core/javascripts/cartodb3/data/layer-definition-model.js
xavijam: 👍 , changing it 💋 .

lib/assets/core/javascripts/cartodb3/editor/style/style-view.js
xavijam: I've just followed the stuff was already there. We don't have a rule but I fully agree we should make an statement for that. cc @CartoDB/frontend .

lib/assets/core/test/spec/cartodb3/editor/style/style-view.spec.js
xavijam: It is a trick. If the method is already used in a bind in the code, that spy will not work in the test file. So we usually do is bind the method from the View in order to spy it from the beginning (jasmine tips and tricks).

ivanmalagon: Great. When I ran into that problem before I always did some workaround if possible. I didn't think on this solution. 👍

PR #11319, opened 2017-01-20
Fixing bad argument passing for data observatory metadata

### LINKED ISSUE #11318: DO measures for points only show US and France
If you add a DO analysis to a **points** node, you can only choose US or France

image

And once there, you can only select a subset of all the available measures:

image

image

**This behavior doesn't affect polygon nodes.**

cc @ethervoid

ethervoid:
Closes #11318 

Lecacybuildermatada function should receive no argument when is not a polygon geometry type

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/data-observatory-metadata.js
+++ b/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/data-observatory-metadata.js
@@ -3,7 +3,11 @@ var Backbone = require('backbone');
 var cdb = require('cartodb.js');
 
 function getQueryForGeometryType (geometryType) {
-  return 'select * from obs_legacybuildermetadata(\'' + (geometryType === 'polygon' ? 'sum' : '') + '\')';
+  if (geometryType === 'polygon') {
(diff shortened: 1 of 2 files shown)


lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/data-observatory-metadata.js
```diff
@@ -3,7 +3,14 @@ var Backbone = require('backbone');
 var cdb = require('cartodb.js');
 
 function getQueryForGeometryType (geometryType) {
-  return 'select * from obs_legacybuildermetadata(\'' + (geometryType === 'polygon' ? 'sum' : '') + '\')';
+  var QUERY = 'select * from obs_legacybuildermetadata({{aggregate_type}})';
+
```
rafatower: the linter is complaining about trailing spaces

ethervoid: Trailing spaces removed

ethervoid:
Changed it to make it clearer

javisantana:
could we merge this as soon as we can and then add a automated test

Issue #11328, opened 2017-01-20
Edit feature panel doesn't render columns with number[] type


xavijam:
So we have found that datasets where it contains number[] column type makes the Builder fail when user tries to edit a feature. Looks like a problem with the backbone form and the way it renders that kind of "field".

PR #11329, opened 2017-01-20
forms: make Text default type in schema

matallo:
re: https://github.com/CartoDB/cartodb/issues/11328

**to reproduce in acceptance**

- import a dataset with culumn type `number[]`

https://data.rennesmetropole.fr/explore/dataset/topologie-des-stations-le-velo-star0/download?format=geojson&timezone=Europe/Berlin&use_labels_for_header=true

- go to edit feature
- column should appear as Text

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/editor/layers/edit-feature-content-views/edit-feature-attributes-form-model.js
+++ b/lib/assets/core/javascripts/cartodb3/editor/layers/edit-feature-content-views/edit-feature-attributes-form-model.js
@@ -49,7 +49,7 @@ module.exports = Backbone.Model.extend({
       return fieldType === columnType;
     });
 
-    return supportedColumnType || 'string'
+    return supportedColumnType || 'string';
   },
 
   _generateSchema: function () {
--- a/lib/assets/core/javascripts/cartodb3/editor/layers/edit-feature-content-views/edit-feature-attributes-form-model.js
+++ b/lib/assets/core/javascripts/cartodb3/editor/layers/edit-feature-content-views/edit-feature-attributes-form-model.js
@@ -44,6 +44,14 @@ module.exports = Backbone.Model.extend({
     return rows && rows.length ? 'Suggest' : 'Text';
   },
 
+  _getSupportedColumnType: function (columnType) {
+    var supportedColumnType = _.find(_.keys(FIELDTYPE_TO_FORMTYPE), function (fieldType) {
+      return fieldType === columnType;
+    });
+
+    return supportedColumnType || 'string'
+  },
+
   _generateSchema: function () {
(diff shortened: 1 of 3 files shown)


lib/assets/core/javascripts/cartodb3/editor/layers/edit-feature-content-views/edit-feature-attributes-form-model.js
```diff
@@ -44,14 +44,22 @@ module.exports = Backbone.Model.extend({
     return rows && rows.length ? 'Suggest' : 'Text';
   },
 
+  _getSupportedColumnType: function (columnType) {
+    var supportedColumnType = _.find(_.keys(FIELDTYPE_TO_FORMTYPE), function (fieldType) {
+      return fieldType === columnType;
+    });
+
+    return supportedColumnType || 'string'
```
xavijam: I feel lint will fail, ; is needed 💃 .

matallo:
effing linter, long live standardjs

javitonino:
Ok, tested it buf found two issues during edition:
 1. The fields are sent as strings like "23,45", but the correct syntax is "{23,45}". This makes the edition fail with message `row identified with 1 not found`
 2. If you edit any other field in the row, the whole row is sent, and the `number[]` column is sent as an array. If you edit the array field, it is sent as an string as described.

Maybe we can help address this problems by touching the backend controller if needed, but we should be getting a consistent input all the time (see 2).

xavijam:
Back to development then, thanks man 💋 .

MariaCheca:
Hey @matallo, I've tried to reproduce the environment with the dataset you provided, but it returns the same error message when I try to edit a field: 
```
Error editing adressevoie with cartodb_id 1: PG::Error: ERROR: column "coordonnees" is of type double precision[] but expression is of type record
```

This happens in local, staging and production, so looks like it doesn't have anything to be with the changes in this branch.

Do you know the cause (maybe this is due to changes in the file to reproduce the original bug) or provide another dataset? Thanks!

javitonino:
This happened to me when the frontend sends an array to the records controller, instead of a string formatted like `{1,2}'. Either we modify the frontend to always send the string, or the backend to accept the array.

MariaCheca:
Looks like those problems @javitonino has detected should be solved in another issue, as they are from other part of the application and it's better to solve it separated of the one we are handling here. So, after making the acceptance and confirm the fix is done and it's working, I'm moving it to deploy.

MariaCheca:
Here it is -> https://github.com/CartoDB/cartodb/issues/11348

javitonino:
Waiting for one last test run after merging master, and deploying right after it finishes.

PR #11344, opened 2017-01-25
11335 delete layer ternaries

### LINKED ISSUE #11335: Delete ternary relations on layer deletion
### Context
When deleting a layer, we are not deleting the ternary relations in the tables `layers_maps` and `layers_user_tables`. It is probably related to the layer refactor.

### Steps to Reproduce
1. Create a map, add a second layer
2. Note the ID of the layer
3. Delete the layer from Builder

### Current Result
The layer doesn't exist in the 'layers` table, but is still present in `layers_maps` and `layers_user_tables`

### Expected result
Complete deletion

### Additional info
This can be solved by code, but also by a foreign key with a `ON DELETE CASCADE`. We should consider both options, and can try to use the chance to add the FK.

javitonino:
Fixes #11335 on Rails side.

I would have liked to do this from the DB (foreign keys), but I ran into some blockers (lack of indices), so I'm launching this for now to unblock #11303. I split the DB issues to https://github.com/CartoDB/cartodb-management/issues/4817

~~I would suggest doing acceptance for both at once (combined branch: `10939-11335-merged`), since they are related and the tests for them have a large overlap.~~

### DIFF
--- a/app/models/carto/layer.rb
+++ b/app/models/carto/layer.rb
@@ -59,13 +59,13 @@ module Carto
     serialize :infowindow, CartoJsonSerializer
     serialize :tooltip, CartoJsonSerializer
 
-    has_many :layers_maps
+    has_many :layers_maps, dependent: :destroy
     has_many :maps, through: :layers_maps, after_add: :set_default_order
 
-    has_many :layers_user
+    has_many :layers_user, dependent: :destroy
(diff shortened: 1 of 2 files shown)


MariaCheca:
Individual acceptance for this PR is passing.

PR #11347, opened 2017-01-25
11324 migration of simple style with property set

### LINKED ISSUE #11324: Error migrating map with wizard polygon and property set
From https://github.com/CartoDB/cartodb-management/issues/4814

The map in question had the following wizard properties:
```
*****{"type"=>"polygon",*****
 "properties"=>
  {"marker-width"=>1,
   "marker-fill"=>"#FF6600",
   "marker-opacity"=>0.9,
   "marker-allow-overlap"=>true,
   "marker-placement"=>"point",
   "marker-type"=>"ellipse",
   "marker-line-width"=>1,
   "marker-line-color"=>"#FFF",
   "marker-line-opacity"=>1,
   "marker-comp-op"=>"none",
   "text-name"=>"km",
   "text-face-name"=>"DejaVu Sans Book",
   "text-size"=>10,
   "text-fill"=>"#F11810",
   "text-halo-fill"=>"#FFF",
   "text-halo-radius"=>1.5,
   "text-dy"=>-1,
   "text-allow-overlap"=>"true",
   "text-placement-type"=>"dummy",
   "text-label-position-tolerance"=>0,
   "text-placement"=>"interior",
   *****"property"=>"wadus",*****
   "zoom"=>18,
   "geometry_type"=>"point"},
 "property"=>nil}
```

It shouldn't be possible for a `polygon` type to have a `property` set, but it happened and the backend crashed:
https://rollbar.com/vizzuality/CartoDB/items/18924/occurrences/20993712360/

We should make this more robust, to avoid crashing at least.

javitonino:
Fixes #11324 

There are some old wizard properties of type polygon which have a `property` set, even though they shouldn't. This caused the migrator to crash. This avoids the crash and does not try to set the property in those cases.

**Acceptance**
- [ ] Create an editor map, apply a polygon wizard, and manually edit the wizard properties of the layer to add a `property` field, using a column name as the value (see the ticket for an example).
- [ ] Then switch to builder and open the map. The style should be migrated without crashing.

### DIFF
--- a/app/controllers/carto/api/layer_presenter.rb
+++ b/app/controllers/carto/api/layer_presenter.rb
@@ -475,7 +475,8 @@ module Carto
                       when 'density'
                         spp['aggregation']['value']
                       else
-                        raise "Unsupported source type: #{@source_type}"
+                        # Ignore some malformed wizards that have a property set even when the type does not support it
+                        return
                       end
 
         destination['attribute'] = wpp['property']
--- a/spec/requests/carto/api/layer_presenter_spec.rb
+++ b/spec/requests/carto/api/layer_presenter_spec.rb
@@ -229,16 +229,21 @@ describe Carto::Api::LayerPresenter do
             }
           end
 
-          before(:each) do
+          it 'sets fill size from marker-width' do
             layer = build_layer_with_wizard_properties(polygon_wizard_properties)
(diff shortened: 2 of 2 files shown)


MariaCheca:
Acceptance passing 👍

PR #11364, opened 2017-01-26
Provide date columns in "merge" analysis

xavijam:
cc @AbelVM


xavijam:
Enabling only date column for the moment, let's check what we decided to filter only by those columns later.

xavijam:
When green I'll merge, we are in a hurry, cc @CartoDB/frontend, already checked in staging with @AbelVM.

xavijam:
Frontend green, merging!

Issue #9501, opened 2016-08-22
Fields of type date on JOIN analysis


jsanz:
### Steps to reproduce
- Create a map with a dataset with fields of type `date` (a twitter import for example)
- Add a new join analysis with any other table

![2016-08-22_14 45 41_selection](https://cloud.githubusercontent.com/assets/188264/17855123/2a641eb8-6877-11e6-9eb7-abb0498bb9e7.png)
### Expected result
- Date fields should appear with the rest of the dataset fields
### Current result
- Date fields are not available

![2016-08-22_14 41 45_selection](https://cloud.githubusercontent.com/assets/188264/17855027/9df04592-6876-11e6-92d1-d9a5a16091da.png)

Attaching a sample exported map: 
[test_join_date_fields map (on 2016-08-22 at 12.42.40).carto.zip](https://github.com/CartoDB/cartodb/files/430120/test_join_date_fields.map.on.2016-08-22.at.12.42.40.carto.zip)

saleiva:
Again this is something that never got implemented, but that was there since the beginning

For the by value dropdown
![screen shot 2016-08-22 at 16 47 02](https://cloud.githubusercontent.com/assets/132149/17858964/0ce11a60-6888-11e6-8833-99e4912e4182.png)

For the other places
![screen shot 2016-08-22 at 16 46 12](https://cloud.githubusercontent.com/assets/132149/17858976/1a64fd64-6888-11e6-9407-9ff6f41fccf0.png)

ramiroaznar:
Hey @xavijam, this is still happening. I do not get what @saleiva closed it.

saleiva:
wait, this looks wrong. That comment has nothing to do with the ticket... :/

xavijam:
Yeah, that comment is for other thing, but it doesn't mind man 💋 .

xavijam:
BTW, "hotfix" deployed, now evaluating why we don't offer any kind of column type within the analysis.

javitonino:
Forgot to close it, but this is deployed.

PR #11367, opened 2017-01-26
Provide all column types for merge analysis

xavijam:
Ticket: #9501.

Basically, after a checking with @javierarce, we don't know why we decided to filter the columns by type of the dropdowns in this analysis. So, we should check if:

- merge analyses with date columns works.
- merge analysis with boolean columns works.
- merge analysis with geometry columns works.


xavijam:
And cc @AbelVM in this one too, sorry mate.

javitonino:
I have tested everything I could think of, including incorrect mixes of different column types, and it seems to work pretty well. I'd say this is a :+1:.

PR #11383, opened 2017-01-27
fix for table header menu item visibility bug

mbektas:
When a filter is applied, rename, delete and some other menu items should be hidden. They are not hidden until page is reloaded. This is a fix for the issue.


Cartofante:

Cartofante:

Issue #11371, opened 2017-01-27
Legend migrator should remove `url(` from icon legends


javitonino:
### Context
From investigation of https://github.com/CartoDB/cartodb-management/issues/4815#event-939084200

If you migrate an editor map with icon legend to Builder, the legends will not render correctly. The migrated version sets `"icon": "url(http......)", and Builder expects `"icon": "http...."`. This behaviour changed here: https://github.com/CartoDB/cartodb.js/commit/1a01cbe0cc73c01ed45357449be56d8b558de1c4#diff-30d6f484d6f47d6e1d8ba3ea13211260

### Steps to Reproduce
1. Create a map in editor
2. Style by icons
3. Add a legend
4. Migrate to Builder

### Current Result
Icons in legend do not load. It tries to load an incorrect URl including `url(` as part of the URL 

### Expected result
The legend renders

### Additional info
The migration code is around here: https://github.com/CartoDB/cartodb/blob/b4bb9e41e3e6692c37bddc2dfb1fa537ff749fb3/lib/carto/legend_migrator.rb#L73

We should also write a script to fix affected maps in production. Maybe use `MapcappedVisualizationUpdater`.

cc @xavijam

@matallo has asked to take this one if possible.

matallo:
on hold waiting for https://github.com/CartoDB/cartodb.js/pull/1562

xavijam:
CartoDB.js thing checked, tested and approved.

gfiorav:
Closed via https://github.com/CartoDB/cartodb/issues/11391

PR #11391, opened 2017-01-30
legends: custom: add support icons beyond svg

matallo:
re: https://github.com/CartoDB/cartodb/issues/11371

currently only svg icons in custom legends will display properly, thus we had to update cartodb.js for this purpose

also, custom legends with icons from old editor won't appear properly, as they were saved with `url()` in the property value

a rake task to replace those icons in the custom legends for currently migrated legends as well as the migrator from editor to builder has been done

**acceptance test**

- create a custom legend with icon, both svg and png in editor, promote the user to builder and check the legends look OK
- set an icon with `url()` for a builder user, run the rake task and check the icon look properly now
- Test the rake task:
  - Create a builder map and add a legend. Manually send a request to set the legend icon to be wrapped inside `url(...)` instead of a plain URL. The image should be broken.
   - Publish the map
   - Run the rake `carto:legends:migration:migrate_custom_image`
   - Now, refreshing the builder and the embed, you should be able to see the images without needing to republish the map. Be careful with caches.
- In all cases, see that the icon renders correctly (see https://github.com/CartoDB/cartodb/pull/11391#issuecomment-276625838)

**Deploy**
After the deploy finishes, run the `carto:legends:migration:migrate_custom_image` rake in all clouds.

### DIFF
--- a/lib/build/tasks/watch.js
+++ b/lib/build/tasks/watch.js
@@ -49,7 +49,7 @@ exports.task = function () {
       files: [
         'lib/assets/core/test/spec/cartodb3/**/*'
       ],
-      tasks: ['copy:js_test_spec_core_cartodb3', 'copy:js_test_spec_client'],
+      tasks: ['copy:js_test_spec_core_cartodb3', 'copy:js_test_spec_client_cartodb3'],
       options: {
         spawn: false,
         atBegin: false
--- a/lib/build/tasks/watch.js
+++ b/lib/build/tasks/watch.js
@@ -111,6 +111,8 @@ exports.task = function () {
         'app/assets/stylesheets/organization/**/*.css.scss',
         // cartodb3
         'app/assets/stylesheets/editor-3/**/*.css.scss',
+        // Uncomment next line to watch Dashboard and Editor SCSS files
+        // 'app/assets/stylesheets/common/**/*.css.scss',
         'app/assets/client/stylesheets/**/*.css.scss'
       ],
       tasks: ['css'],
--- a/lib/build/tasks/watch.js
+++ b/lib/build/tasks/watch.js
@@ -69,7 +69,7 @@ exports.task = function () {
       files: [
         'lib/assets/core/test/jasmine/cartodb3/**/*'
       ],
-      tasks: ['copy:js_test_jasmine_core_cartodb3', 'copy:js_test_jasmine_client'],
+      tasks: ['copy:js_test_jasmine_core_cartodb3', 'copy:js_test_jasmine_client_cartodb3'],
       options: {
         spawn: false,
         atBegin: false
--- a/lib/carto/legend_migrator.rb
+++ b/lib/carto/legend_migrator.rb
@@ -68,14 +68,13 @@ module Carto
         category_definition = { title: title }
 
         if value
-          if value =~ COLOR_REGEXP
-            category_definition[:color] = value
-          else
-            byebug
-            match = CSS_URL_REGEX.match(value)
+          match = CSS_URL_REGEX.match(value)
 
-            if match
-              category_definition[:icon] = match[2]
+          if match
+            category_definition[:icon] = match[2]
+          else
+            if value =~ COLOR_REGEXP
+              category_definition[:color] = value
             else
               category_definition[:icon] = value
             end
--- a/lib/carto/legend_migrator.rb
+++ b/lib/carto/legend_migrator.rb
@@ -72,12 +72,10 @@ module Carto
 
           if match
             category_definition[:icon] = match[2]
+          elsif value =~ COLOR_REGEXP
+            category_definition[:color] = value
           else
-            if value =~ COLOR_REGEXP
-              category_definition[:color] = value
-            else
-              category_definition[:icon] = value
-            end
+            category_definition[:icon] = value
           end
         end
 
--- a/lib/carto/legend_migrator.rb
+++ b/lib/carto/legend_migrator.rb
@@ -58,6 +58,7 @@ module Carto
     end
 
     COLOR_REGEXP = /^#(?:[0-9a-fA-F]{3}){1,2}$/
+    CSS_URL_REGEX = /^(?:url\(('|")?)(.*?)(?:('|")?\))$/
 
     def build_custom_definition_from_custom_type
       categories = items.each_with_index.map do |item, index|
@@ -70,7 +71,14 @@ module Carto
           if value =~ COLOR_REGEXP
             category_definition[:color] = value
           else
-            category_definition[:icon] = value
+            byebug
+            match = CSS_URL_REGEX.match(value)
+
+            if match
+              category_definition[:icon] = match[2]
+            else
+              category_definition[:icon] = value
+            end
           end
         end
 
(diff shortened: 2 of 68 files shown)


matallo:
do I really need to use `MapcappedVisualizationUpdater` as in https://github.com/CartoDB/cartodb/blob/master/lib/tasks/layers.rake#L87

or is it OK with the proposed task? no prob to take the mapcapped change, but I'd like to have a bit more of background

xavijam:
We need to support the change made in CartoDB.js in BUILDER, so we need to modify this template https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/legend/color/legend-custom-template.tpl and add the necessary changes. Also, we can't forget the update of the dependencies (shrinkwrap). 💋

lib/carto/legend_migrator.rb
```diff
@@ -67,7 +68,11 @@ def build_custom_definition_from_custom_type
         category_definition = { title: title }
 
         if value
-          if value =~ COLOR_REGEXP
+          match = CSS_URL_REGEX.match(value)
+
+          if match
+            category_definition[:icon] = match[2]
```
javitonino: A suggestion: you are capturing the quotes in capture groups. You could fix it by changing the regex to `/^(?:url\(['"]?)(.*?)(?:['"]?\))$/` and then using `match[1]`.

spec/lib/carto/legend_migrator_spec.rb
```diff
@@ -142,7 +142,7 @@ module Carto
               {
                 "name" => "Untitled",
                 "visible" => true,
-                "value" => "superduper.png",
+                "value" => "url(superduper.png)",
```
javitonino: You should add a spec later that checks that this is correctly interpreted as `superduper.png` (without the `url(`) part). As it is, I don't think this change to the test would have created a failure without the new code, does it?

lib/tasks/legends.rake
```diff
@@ -27,6 +27,44 @@ namespace :carto do
...
+      desc 'Fix custom legends with images'
+      task migrate_custom_image: :environment do
+        CSS_URL_REGEX = /^(?:url\(('|")?)(.*?)(?:('|")?\))$/
+
+        custom_image_legends = Carto::Legend.where(type: 'custom')
+        custom_image_legends_count = custom_image_legends.count
+
+        errored = []
+        puts "#{custom_image_legends_count} custom image legends..."
+        custom_image_legends.each_with_index do |legend, index|
```
javitonino: `each_with_index` will load all legends into memory. For rakes, we tend to use `find_each` instead, which works in batches so it's nicer on the memory. Sadly, `find_each` doesn't have an index, so you'd have to add it manually (via a counting variable).

matallo:
pushed the change in the template and doing the dependencies update now (it was done with the dev branch of cartodb.js)

lib/tasks/legends.rake
```diff
@@ -27,6 +27,44 @@ namespace :carto do
...
+          if Carto::Layer.exists?(legend.layer_id)
+            legend[:definition][:categories].each_with_index do |cat, cat_index|
+              icon = cat[:icon]
+              match = icon && CSS_URL_REGEX.match(icon)
+
+              if match
+                legend.definition[:categories][cat_index][:icon] = match[2]
+
+                if !legend.save
```
javitonino: Maybe you can move the save outside this loop to avoid multiple saves if changing multiple category items in the same legend.

xavijam:
💋

javitonino:
BTW, @matallo: Welcome to backend CR's! 🎉

package.json
```diff
@@ -24,7 +24,7 @@
     "carto": "cartodb/carto#master",
     "cartodb-deep-insights.js": "cartodb/deep-insights.js#master",
     "cartodb-pecan": "0.2.x",
-    "cartodb.js": "CartoDB/cartodb.js#v4",
+    "cartodb.js": "CartoDB/cartodb.js#4815-legendicons",
```
xavijam: You should point to v4 again, because it is merged.

xavijam: And run the dependencies update of course :D.

matallo:
mapcapped, suggestions done, and test added, although I'm not sure it is the best approach (I'm too used to frontend tests), let's discuss tomorrow!

lib/build/tasks/watch.js
```diff
@@ -69,7 +69,7 @@ exports.task = function () {
       files: [
         'lib/assets/core/test/jasmine/cartodb3/**/*'
       ],
-      tasks: ['copy:js_test_jasmine_core_cartodb3', 'copy:js_test_jasmine_client'],
+      tasks: ['copy:js_test_jasmine_core_cartodb3', 'copy:js_test_jasmine_client_cartodb3'],
```
javitonino: This seems unrelated to this PR, but I trust you know what you are doing :+1:

lib/tasks/legends.rake
```diff
@@ -30,40 +30,62 @@ namespace :carto do
...
+          acc += 1
+
+          puts "#{acc} / #{total}" if (acc % 100).zero?
+
+          begin
+            visualization = layer.visualization
+            next unless visualization
+
+            success = update_visualization_and_mapcap(visualization) do |vis, persisted|
+              vis.layers.each do |l|
```
javitonino: Small improvement, you can use `data_layers` here instead of `layers` in order to skip basemaps (which won't have legends). Very minor.

lib/tasks/legends.rake
```diff
@@ -30,40 +30,62 @@ namespace :carto do
...
+                l.legends.each do |legend|
+                  acc_category = 0
+                  categories = legend[:definition][:categories]
+
+                  categories && categories.each do |category|
+                    icon = category[:icon]
+                    match = icon && CSS_URL_REGEX.match(icon)
+
+                    if match
+                      l.legends[acc_legend][:definition][:categories][acc_category][:icon] = match[1]
```
javitonino: Can you not do `category[:icon] = match[1]`? Just asking, it seems like it should work, but I can believe it doesn't.

matallo:
thanks for the review! 🙇

juanignaciosl:
Added "do not merge" because of package.json version for acceptance.

juanignaciosl:
Acceptance issue:

1.- New map in Editor. Enable custom legend and set a SVG icon from "pin gallery".
2.- Open in Builder. Icon won't load:

captura de pantalla 2017-02-01 a las 11 40 48

juanignaciosl:
Back to development after talking to @matallo

matallo:
added migration for old assets libraries, ready for review and acceptance again, well spotted!

matallo:
in any case I tried with non-library assets, too but I couldn't find a way to check both library and custom assets in the tests

gfiorav:
Acceptance is 🍏 

Tested:

- Migrate map with pin, svg and png in master. Issue occurs.
- Deploy new code. Problem is still there
- Run rake, problem disappears for all 3 formats.
- Make map in Editor with new code, `url()` is wrapping pin, png and svg
- Migrate map to Builder with new code, the migration is performed correctly and `url()` removed.

PR #11396, opened 2017-01-30
fixed grammar in delete layer dialog, when an analysis is applied

csobier:
@xavijam , I found this dialog when deleting a map layer with an analysis applied. That doesn't sound right, so I edited the text. (I'm not sure if if affects other auto-generated dialogs, so please confirm).

image


xavijam:
That text could have other strings, like:

Deleting this layer affects one analysis, two widgets, and three legends.

Sounds ok?

csobier:
Yes, sounds good.

xavijam:
:+1:

Issue #11339, opened 2017-01-24
Two time series widgets added when styling a layer by Torque


arianaescobar:
### Context

I styled a layer of tweets with Torque and it created two time-series widgets.

### Steps to Reproduce

1. Searched tweets from #womensmach and created a map with it 
2. Entered the only layer (with the tweets) who where styled as points and changed it to torque
3. It gave me an error, nothing showed on the map, and it created a time-series widget that didn't load a thing:

![1](https://cloud.githubusercontent.com/assets/16247628/22251228/4b5beb94-e24a-11e6-8ce6-96edb34f745a.png)

4. I refreshed the tab and the torque animation appeared on the map and a second time-series widget appeared but none showed data:

![2](https://cloud.githubusercontent.com/assets/16247628/22251244/5899487e-e24a-11e6-8ba7-f89207e56ff4.png)

5. I refreshed a second time and both widgets finally showed their data:

![3](https://cloud.githubusercontent.com/assets/16247628/22251308/91c1a894-e24a-11e6-8826-d579fb475775.png)

### Current Result

https://team.carto.com/u/arianaescobar/builder/20034d80-e221-11e6-ae82-0ef7f98ade21

### Expected result

Torque animation showing up and just one time-series added with loaded data once the style got switched to Torque without needing to refresh the tab. 

### Browser and version

Chrome Version 55.0.2883.95 (64-bit)

### .carto file

The .carto file is 46,5MB, I can't upload it here. If whoever is working on this needs it, let me know and I'll send it through slack or something.

AbelVM:
I can confirm

saleiva:
Adding to next since this is quite explicit

nobuti:
Picking this.

nobuti:
Ready for CR

xavijam:
Thanks @nobuti for the fix 💋

saleiva:
@arianaescobar check it!

arianaescobar:
Works perfectly now 👌 ✨

PR #11398, opened 2017-01-30
Wait server response to add the model.

nobuti:
This PR implements #11339.

Not waiting caused a race condition and two creation requests were made. The first one was aborted by the sync abort thing, but the model was already added to the collection. Then the style
definition model changed, causing the widgets to persist.

So two time-series were created when this race condition met.


xavijam:
Could we consider a test avoiding the issue again?

xavijam:
Working like a charm, acceptance accepted! 💃 .

xavijam:
Frontend tests passed, merging.

PR #11404, opened 2017-01-31
Removed auto-style flag logic

xavijam:
Related ticket: #11392 

Auto-style should be enabled by default for everybody, it doesn't matter if the user has enabled the feature flag or not. So, for testing:

- If user has the `auto-style` feature flag enabled, auto style behaviour should be **enabled**, and the widget form should show the color options.
- If user doesn't have the `auto-style` feature flag enabled, auto style behaviour should be **enabled** too, and the widget form should show the color options.


xavijam:
👍  👏

ivanmalagon:
Tested a user with and without the `auto-style` feature flag. Widgets behave same way in both cases.

:+1:

Passing to deploy.

Issue #11283, opened 2017-01-17
If image is not selected in the category list, we should not add marker-file attribute in the generated cartocss code


xavijam:
![screen shot 2017-01-17 at 16 30 35](https://cloud.githubusercontent.com/assets/132146/22026910/6b15cbb2-dcd2-11e6-8796-7fe0b7f4e6ad.png)

nobuti:
Ready for CR.

ivanmalagon:
Deployed to production.

PR #11408, opened 2017-01-31
Ramp for only filled images.

nobuti:
This PR implements #11283.

### How to tests

Check that the ramp is only generated for those markers with images, leaving off the ramp those with images equal to ''.

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/editor/style/style-converter.js
+++ b/lib/assets/core/javascripts/cartodb3/editor/style/style-converter.js
@@ -139,7 +139,7 @@ function pointFill (props, animationType) {
 
     css['marker-fill-opacity'] = markerFillOpacity != null ? markerFillOpacity : 1;
 
-    if ((props.images && _.isArray(props.images)) || (color.images && _.isArray(color.images))) {
+    if (hasImagesSelected(props.images) || hasImagesSelected(color.images)) {
       markerFill(props.images || color.images, color, css);
     } else if (props.image || color.image) {
       var url = props.image;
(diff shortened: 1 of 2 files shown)


lib/assets/core/javascripts/cartodb3/editor/style/style-converter.js
```diff
@@ -166,14 +175,13 @@ function markerFill (images, color, css) {
...
-  }
-
-  for (var d in color.domain) {
-    categoryNames.push(color.domain[d]);
-  }
+  _.each(images, function (image, index) {
+    if (image !== '') {
+      var urlFormat = 'url(\'' + image + '\')';
+      filesUrls.push(urlFormat);
+      color.domain[index] && categoryNames.push(color.domain[index]);
```
ivanmalagon: I've a concern here.

The generated cartocss renders two arrays, one with the images and another one with the corresponding categories. I see two problems here:

1. There are no categories (`color.domain`) at all: in this case, the images array will be interpreted sequentially for the first X categories. I think this is not going to happen because we only call to `marker fill`in category stiling, but if there's any case for it to happen, this won't work properly.
2. There are categories, but one of them is falsy. Legit but falsy. In this case, we'd render a marker file but not its category. I've tested this and it's happening before, not related only with icons. In case a category returns 0, the empty string or null it appears as a category with that value but the generated cartocss is broken. You can modify a color for a category and another category gets modified. I'm opening a general ticket for this but in this particular case, if a category is falsy and we apply an image to it, all subsequent categories get the same icon.

ivanmalagon:
Made a minor fix to take into account falsy categories as null, '' and 0. Those can be legit categories.

Previous to this, setting an icon to a falsy category assigned the same icon to all the subsequent categories. An example of this issue is shown in the next screenshot.

image

ivanmalagon:
Fixed @nobuti Nice catch.

ivanmalagon:
Accepted.

Tested with numeric and text columns. Only setting icons for a few categories generates a CartoCSS with only urls for those markers.

Tested also with falsy categories.

Issue #11063, opened 2016-12-14
Style with icons - some design issues


urbanphes:
It looks good but I found some designs issues:

- [x] When we have selected an image and pass the mouse over the icon appear arrow cursor and it should appear hand cursor.
<img width="178" alt="captura de pantalla 2016-12-14 a las 17 36 32" src="https://cloud.githubusercontent.com/assets/14546701/21204153/e4fa1f42-c223-11e6-9d31-2667468d315a.png">

- [x] When we have selected an image, on the input the margin-left should be different. If we work with 16x16 (px) for the image the margin-left should be 8px.
<img width="675" alt="captura de pantalla 2016-12-14 a las 17 37 55" src="https://cloud.githubusercontent.com/assets/14546701/21204190/180ee552-c224-11e6-9d28-fc5c2c4b8538.png">

- [x] When showing the dropdown the thumbnails should be size 40x40 (px)
<img width="271" alt="captura de pantalla 2016-12-14 a las 17 38 39" src="https://cloud.githubusercontent.com/assets/14546701/21204204/2de620e8-c224-11e6-85fd-67df00b9f451.png">

- [x] When selecting an image there is a jump in this second nav bar.
<img width="147" alt="captura de pantalla 2016-12-14 a las 17 39 11" src="https://cloud.githubusercontent.com/assets/14546701/21204222/49f06cbc-c224-11e6-8507-ca35e6d0f256.png">

- [x] When in a category map you select images, the input gets wider than the max limit.
<img width="197" alt="screen shot 2017-01-18 at 15 39 18" src="https://cloud.githubusercontent.com/assets/132149/22082118/7675f564-dd94-11e6-862d-94860b5c62a4.png">

jorgesancha:
![](https://github.trello.services/images/mini-trello-icon.png) [Style With Icons](https://trello.com/c/ehXyg2nE/1-style-with-icons)

xavijam:
Last point is done.

nobuti:
Picking this.

nobuti:
Ready for CR.

xavijam:
Taking it!

xavijam:
Deployed!

saleiva:
wow!

PR #11417, opened 2017-01-31
Fine tune design

nobuti:
This PR implements #11063

This should be easy to test, just read the issue and check all issues are fixed in every browser your machine has.

### DIFF
--- a/app/assets/stylesheets/editor-3/assets.css.scss
+++ b/app/assets/stylesheets/editor-3/assets.css.scss
@@ -13,6 +13,11 @@
   @include display-flex;
   position: relative;
 }
+.AssetListItem.Carousel-item {
+  width: 40px;
+  height: 40px;
+  min-width: 40px;
+}
 .AssetsList-inner {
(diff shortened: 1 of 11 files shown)


nobuti:
pinging @piensaenpixel for CR about styles.

xavijam:
I've checked that:

- Fill input is bigger than rest.

screen shot 2017-02-01 at 12 54 52

- Icons are not aligned in the center of the box.

screen shot 2017-02-01 at 12 39 15

Reviewing that myself.

xavijam:
About first thing, it turns out that select field is smaller than the rest, fixed in https://github.com/CartoDB/cartodb/pull/11417/commits/6fe04636aea89ee39137ffd402d9a3c8ac99e8cd.
And further to the second point, the problem is due to the SVG, not about the position of the  itself (as @nobuti pointed offile), so 👍 .

xavijam:
Tested in Chrome, Safari and Firefox, :+1:

PR #11423, opened 2017-02-01
Setting autostyle always true in public views

ivanmalagon:
Related to #11392

This PR sets autostyle to true in public views.

### Acceptance

1. Delete the `auto-style` FF of your user.
2. Set a map with autostyle widgets. That is, select a map with category and histogram widgets.
3. In private and public views, autostyle icons should show up in both widgets.

### Deploy
Remember to purge caches


javitonino:
Deleted the feature flag in staging. Tested category and histogram widgets in Builder and embed. Everything is ✅ 

Waiting for backend tests and deploying.

PR #11424, opened 2017-02-01
Return null instead empty string if no date.

nobuti:
This PR implements #11145.

The problem is when we insert a value, the form grabs all the data from the fields, returning '' for date time fields. In the first scenario, saving directly without adding new attribute values, the initial value is `null`. In order to fix this, we need to return `null` if the value is not a formatted date. Otherwise, Postgres will complain.


xavijam:
👏

javitonino:
Acceptance looks OK to me. I tried editing number/string/date fields in the feature edition and table views. I could not find anything that break and adding a new feature makes the date default to NULL. 📦 'int it!

Issue #11244, opened 2017-01-12
Persistent Notifications in Builder UI


ernesmb:
### Context
There are certain actions that trigger notifications in Builder UI. Sometimes these notifications are impossible to close by clicking on *X* button, with no errors appearing in the console. 

This bug [has already been reported](https://github.com/CartoDB/cartodb/issues/10734), but without clear STR. 

I've found some STR that work, but bear in mind that this bug may still appear in different situations. 

### Steps to Reproduce

1. Create a map with one layer *A*
2. Style that layer *BY VALUE* in order to make a legend appear
3. Open *Map Options*
4. Turn *LEGENDS* **off**
5. Turn *LEGENDS* **on** again (this may look silly, but I couldn't reproduce the issue without this step)
6. Add a second layer *B*
7. From the layer list, hide *A* by clicking on the 'eye' icon
8. Open *Map Options* and enable the *LAYER SELECTOR* 
9. Use the layer selector to **activate** layer *A*
10. Use the layer selector to **deactivate** layer *A*
11. Open *LAYERS* pane, see notification below
12. Click *X* to close notification

### Current Result
The notification doesn't close

 

### Expected result
The notification should close

### Browser and version
Tested in Chrome v55

cc/ @saleiva

saleiva:
I have seen this same problem in other random situations. I think it would be worth to take a look at this logic

saleiva:
spotted this a few times more

nobuti:
Picking this.

nobuti:
Ready for CR

ivanmalagon:
Deployed to produ.

PR #11431, opened 2017-02-02
Rebind events.

nobuti:
This PR implements #11244.

Notifier view is a singleton view. This means, the view is created once, then added and removed.
When removed, because of the internal management of our views, we destroy the bindings. This happens when we change to the options view. The notifier parent's view, editor-map-view, is deleted, so it performs a clearSubViews, deleting the bindings. In order to fix this, every time we append the notifier view, we need to make sure bindings are working properly.

In order to test this, make sure notifications are working properly when back from the options panel. It's in this scenario where the re-binding should get back to listen to every event again.

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/components/notifier/notifier-view.js
+++ b/lib/assets/core/javascripts/cartodb3/components/notifier/notifier-view.js
@@ -22,8 +22,16 @@ module.exports = CoreView.extend({
     return this;
   },
 
+  rebindEvents: function () {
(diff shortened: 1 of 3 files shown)


ivanmalagon:
Tested this staging.

Switching to options panel back and forth and nofitication behaves correctly. It closes clicking on X. In production, where this is not deployed yet, it's quite easy to break notifications doing this tecnique of toggling options panel so I feel confident that this PR fixes the issue.

PR #11432, opened 2017-02-02
11409 inconsistent deletion

### LINKED ISSUE #11409: Inconsistent state if deleting a dataset fails
### Context
From multiple support tickets, but finally managed to consistently reproduce. When deleting a dataset, if something fails, the database will be in an inconsistent state, where the user table and map are deleted, but the visualization is not, so it shows as a ghost table in the dashboard. 

### Steps to Reproduce
1. Create a dataset
2. Create a view over the dataset (`CREATE VIEW wadus AS SELECT * FROM table`)
3. Delete the dataset

### Current Result
Error, but the user table is deleted and the visualization is kept. 

### Expected result
Error, but all metadata objects are kept.

javitonino:
Fixes #11409 

There are two changes in this PR:
- Execute visualization deletion inside a transaction in order to avoid half-deletions
- More detailed errors in the UI in case something fails

screenshot_20170202_124904

**Acceptance**
First, we need to test that the change this PR addresses is fixed.
Create a dataset. Then execute `CREATE VIEW something AS SELECT * FROM <table_name>`. Then, try to delete the dataset:

- [x] An error telling you about the view should be displayed.
- [x] The main error should be displayed in the notifier.
- [x] The dataset should not be deleted.
- [x] Any maps that have a layer using that dataset should not be deleted.
- [x] The dataset should be accessible (try opening the dataset view, it should work).

We need to do regression testing of all kinds of visualizations deletions, that is:
- [x] Delete a map
- [x] Delete a dataset that is not used anywhere
- [x] Delete a dataset that is used in a map with a single layer (map should be deleted)
- [x] Delete a dataset that is used in a map with multiple layers (only the affected layer will be deleted)
- [x] Delete a map with widgets, analyses and published. Check that widgets, analyses and mapcaps are deleted, at DB level.


lib/assets/javascripts/cartodb/common/dialogs/delete_items_view_model.js
```diff
@@ -106,8 +115,16 @@ module.exports = Backbone.Collection.extend({
...
-        callback('something failed');
+      .fail(function(response) {
+        var errorMessage = 'something failed';
+        try {
+          var responseJson = JSON.parse(response.responseText);
+          var responseMessage = responseJson.errors.join('; ');
+          if (responseMessage) {
+            errorMessage = responseMessage;
+          }
+        } catch (e) {}
```
juanignaciosl: I wouldn't do a `catch (e) {}`, but I'm not frontender... 🙊

juanignaciosl:
@matallo to acceptance, then? How come `errorMessage` doesn't contain error info? It might be useful.

javitonino:
@juanignaciosl I think you are referring to the JS catch (which btw, it cannot catch specific exception, there is only catch-all). That only triggers when the response cannot be parsed as JSON (before this, we were throwing 500 errors with no body when this happened). We keep "something failed" for compatibility with the old code, since that part of the code is generic and called from 4 or 5 times, so we didn't want to alter the behaviour everywhere.

matallo:
thanks for the explanation @javitonino

xavijam:
We should add the error message when the user tries to delete the dataset and it fails within the new dataset view.

xavijam:
Acceptance passed, all the points above are covered. Added one more and changing frontend to be passed.

xavijam:
screen shot 2017-02-06 at 14 58 37

Issue #10875, opened 2016-11-29
Increase the spec coverage of the icon styling part of the Fill component


javierarce:
Right now that area of the fill component lacks of a proper spec coverage. We should add some tests to ensure nothing gets broken as we continue developing and adding more features.

jorgesancha:
@javierarce When you have a mo, please can you provide some more info for this ticket?

javierarce:
Done! This issue was more of a note to self that I created before we introduced the new workflow.

xavijam:
Done!

xavijam:
Ouch, it should be closed.

PR #11433, opened 2017-02-02
10875 increase spec coverage

ivanmalagon:
Implements #10875 

Added missing tests for several objects related to icon markers.


javitonino:
Does not look like there should be any acceptance for this (it seems there are only tests). Can you confirm @ivanmalagon? Or in case we need acceptance, suggest something to try?

ivanmalagon:
I would just select one marker icon in fill styling from the modal. So instead choosing one from the preview list, pick one from the icons modal. I inserted one check that shouldn't break anything (indeed, it does the contrary).

Being paranoic here but a test doesn't hurt anybody.

xavijam:
Let me accept it :D

xavijam:
Checked in local, working like a charm. Not need to be test it in staging because it is super tiny thing.

Issue #11419, opened 2017-01-31
Map creation and deletion are extremely slow in my team account


saleiva:
### Context
I've been noticing lately the slow speed when creating or deleting a map in my account. Some map creation with already imported data is taking more than 20 seconds. I think there is a relevant difference between this in my team account and another one I have in a free server.

cc/ @juanignaciosl

matallo:
happening in staging enviroinments when creating a new map with new datasets (requests taking minutes for creation/deletion with no data)

javitonino:
I just update the local env to master and started suffering this. A suspect is this query block which is taking long, doing the same thing over and over again:
```
(0.001961s) SELECT COUNT(*) AS "count" FROM "visualizations" WHERE ("user_id" = '06b974db-de0d-49b7-ae9b-76c63af8c122') LIMIT 1
(0.005179s) SELECT * FROM "visualizations" WHERE ("user_id" = '06b974db-de0d-49b7-ae9b-76c63af8c122') LIMIT 300 OFFSET 0
(0.003452s) SELECT COUNT(*) AS "count" FROM "visualizations" WHERE ("user_id" = '06b974db-de0d-49b7-ae9b-76c63af8c122') LIMIT 1
(0.008552s) SELECT * FROM "visualizations" WHERE ("user_id" = '06b974db-de0d-49b7-ae9b-76c63af8c122') LIMIT 300 OFFSET 0
(0.000680s) SELECT COUNT(*) AS "count" FROM "visualizations" WHERE ("user_id" = '06b974db-de0d-49b7-ae9b-76c63af8c122') LIMIT 1
(0.007542s) SELECT * FROM "visualizations" WHERE ("user_id" = '06b974db-de0d-49b7-ae9b-76c63af8c122') LIMIT 300 OFFSET 0
(0.002731s) SELECT COUNT(*) AS "count" FROM "visualizations" WHERE ("user_id" = '06b974db-de0d-49b7-ae9b-76c63af8c122') LIMIT 1
(0.006689s) SELECT * FROM "visualizations" WHERE ("user_id" = '06b974db-de0d-49b7-ae9b-76c63af8c122') LIMIT 300 OFFSET 0
(0.002182s) SELECT COUNT(*) AS "count" FROM "visualizations" WHERE ("user_id" = '06b974db-de0d-49b7-ae9b-76c63af8c122') LIMIT 1
[...]

PR #11435, opened 2017-02-02
Memoize visualization names taken by user in order to check name candidates

javitonino:
This should improve the situation in #11419 by memoizing temporary results. In map creation, locally, it reduces the time to create a visualization from 2600ms to 60ms, in an account with 300 visualizations (half of that are the data library), 40 of which are unnamed maps.

**Acceptance**
If you have a staging account with many maps, you can take measurements on map creation before and after this patch. Creating several maps named like "Untitled Map XX" (the default) makes the problem more obvious. This also affects visualizations renaming.

- [x] Create map, check that the generated map doesn't conflicts with any others
- [x] Try to rename a map to an existing name -> it should fail
- [x] Try to rename a map to a non-existent name -> it should suceed
- [x] All the previous operations should be quick (a couple of seconds).
- [x] Import a dataset.
- [x] Import the same dataset again.
- [x] Import a .carto
- [x] Import the same .carto again.
- [x] Create a dataset from a query.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -56,6 +56,7 @@ Development
 * Fix color for "Other" category (#11078)
 * Custom errors for latitude/longitude out of bounds (#11060, #11048)
 * Fix timeseries widget height (#11077)
+* Improve speed of map name availability check, improves map creation and renaming times (#11435)
 * Fix redirection after logout for subdomainless URLs (#11361)
 * Fix scrollbar in carousel (#11061)
 * Restrict login from organization pages to organization users, and redirect to Central otherwise
--- a/app/models/visualization/name_checker.rb
+++ b/app/models/visualization/name_checker.rb
@@ -2,26 +2,23 @@
 require_relative './collection'
 
 module CartoDB
-  module Visualization 
+  module Visualization
     class NameChecker
       def initialize(user)
         @user = user
-      end #initialize
+      end
 
       def available?(candidate)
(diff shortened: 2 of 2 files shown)


javierarce:
Finished acceptance. In general it seems to work ok (in term of speed, I noticed that creating an empty layer is quicker by a couple of seconds to production)

The only problem I've found is that importing a file (a .carto or a CSV, for example) creates maps with the same name:

screen shot 2017-02-06 at 18 12 03

Is this a deal breaker?

juanignaciosl:
@javierarce how is it different from current behavior?

captura de pantalla 2017-02-07 a las 9 34 02

javitonino:
Yeah, I would consider that a different ticket, maybe we are not checking for duplicate names in all places we should be (it appears we are missing the `.carto` importer). I'd deploy this and open a new ticket, agreed?

juanignaciosl:
My screenshot is a consecutive .csv importing, unrelated to .carto then.

javierarce:
@juanignaciosl yeah, it's like the current behaviour, I mentioned it because the PR talked about renaming and consecutive file upload and wanted to warn you before proceed with the merge.

javierarce:
I open a ticket for the consecutive name issue: https://github.com/CartoDB/cartodb/issues/11472 Feel free to correct me or update the issue if you find is not accurate.

Issue #11390, opened 2017-01-30
[Google Maps] Add support for Google Maps basemaps


alonsogarciapablo:
- [x] Google Maps library should be loaded, using the user's API_KEY.
- [x] Upon clicking on a Google Maps basemap, BUILDER should talk to cartodb.js (this is being implemented in https://github.com/CartoDB/cartodb.js/pull/1561) and:
  - Update the baselayer.
  - Create/update the labels layer.
  - Switch the map provider.
- [x] If user/org (not sure how this is implemented) has support for Google Maps, Google Maps basemaps should be an option in the basemap selector.

xavijam:
Something to take into account @saleiva @noguerol 

![screen shot 2017-02-20 at 14 44 39](https://cloud.githubusercontent.com/assets/132146/23127388/2f24018c-f77b-11e6-9413-017d3f1673c6.png)

PR #11447, opened 2017-02-03
Google Maps basemaps

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb/issues/11390.

**WARNING: Review changes in package.json before merging! Make sure cartodb.js points to v4 and the following issues are closed:**

- https://github.com/CartoDB/cartodb.js/issues/1527
- https://github.com/CartoDB/cartodb.js/issues/1528
- https://github.com/CartoDB/cartodb.js/issues/1529


app/views/carto/builder/visualizations/show.html.erb
```diff
@@ -17,6 +17,11 @@
   var dashboardNotifications = <%= safe_js_object @dashboard_notifications.to_json %>;
   var ACTIVE_LOCALE = 'en';
 </script>
+
+<% if current_user.google_maps_api_key %>
```
alonsogarciapablo: Do we need to add this in other templates?

lib/assets/core/javascripts/cartodb3/editor/layers/basemap-content-view.js
```diff
@@ -41,106 +38,52 @@ module.exports = CoreView.extend({
...
 
-    // google maps basemaps not supported yet
-    var basemapList = _.clone(basemaps);
-    delete basemapList['GMaps'];
-    delete basemapList['Google'];
-
-    // basemaps defined in app_config.yml
-    _(basemapList).each(function (categoryBasemaps, category) {
+    // Basemaps defined in app_config.yml
+    _(basemaps).each(function (categoryBasemaps, category) {
```
alonsogarciapablo: Do we want to check if user has a Google Maps API Key here and filter Google Maps basemaps if he/she doesn't have one?

app/views/carto/builder/visualizations/show.html.erb
alonsogarciapablo: cc: @xavijam @juanignaciosl

lib/assets/core/javascripts/cartodb3/editor/layers/basemap-content-view.js
alonsogarciapablo: cc: @xavijam @matallo

xavijam: @noguerol @saleiva, what do you think? IMHO we should display google-maps basemaps only when the user or the user organization has google-maps enabled, that means a google-maps API key.

alonsogarciapablo: To give @noguerol and @saleiva a bit more context, we're adding support for Google Maps basemaps in this PR. Only work for users that have setup their Google Maps API Key will be able to use Google Maps basemaps (otherwise the Google Maps library can't be loaded). The thing is that, as of now, we're always showing the Google Maps category + basemaps:

screen shot 2017-02-14 at 16 08 04

And, when the user selects a Google Maps basemap and doesn't have a valid API Key, he/she will see the following message (from Google):

screen shot 2017-02-14 at 16 11 07

Good enough? Should we be more explicit (eg: disable Google Maps category)? Anyway, I think we should do it as an improvement as this pull request is already big enough....

app/views/carto/builder/public/embeds/show.html.erb
```diff
@@ -29,6 +29,11 @@
     <%= javascript_include_tag 'tangram.min.js' %>
   <% end %>
   
+  <% if current_user.google_maps_api_key %>
```
alonsogarciapablo: We have a problem with this ☝️ . There's (obviously) no `current_user` in the templates for our embeds so this will not work (tests are failing for this reason). I guess we need to use the Google Maps API key of the map's owner. @juanignaciosl what would be the best way to find the owner of the map and use it here? Thanks!

xavijam: @vizjson.owner or something like that?

lib/assets/core/javascripts/cartodb3/editor/layers/basemap-content-view.js
noguerol: The feedback is not clear enough and this will turn confusing, as a lot of people will try to use Google basemaps without knowing they need an api key, so I propose to disable them when no key is present: https://github.com/CartoDB/cartodb.js/issues/1577

app/views/carto/builder/public/embeds/show.html.erb
juanignaciosl: Uhm. Vizjson contains user, but it doesn't expose Google Maps API.

I was wondering about dumping more user information at the embed page, but for cartodb.js to work you need it in the vizjson, isn't it? If you confirm this we'll do it.

alonsogarciapablo: CartoDB.js currently expect the Google Maps API library to have been loaded before it's loaded. That's why it's enough for now to just do this ☝️ 👇  or [this](https://github.com/CartoDB/cartodb/pull/11447/files#diff-88095d79eb397b7ebbf5ddbe214e53d7R21). 

We could do something more sophisticated in the future and load the Google Maps API library dynamically from within CartoDB.js... but that's a separate thing. For now, I think it's ok if we just load the script in this template...

I've seen we have access to `@visualization` here... Would we be able to get the owner from there? What about organizations? Thanks!

juanignaciosl: @alonsogarciapablo I don't think so, because ERBs should only access data available through presenters, and currently there's no presenter that exposes Google API Key. If it's not needed for vizjson we could dump it in some presenter, probably in Visualization presenter, optionally.

alonsogarciapablo: I was referring to https://github.com/CartoDB/cartodb/blob/master/app/controllers/carto/builder/public/embeds_controller.rb#L43. @juanignaciosl whatever you do in these cases will work as long as the script is loaded...

juanignaciosl: @alonsogarciapablo Well, you _could_, but you shouldn't use that in a ERB...

alonsogarciapablo: ok! Will you guys take care of this 🙏 ?

juanignaciosl: Issue: #11541

app/views/shared/visualizations/_google_maps.html.erb
```diff
@@ -0,0 +1,3 @@
+<% if google_maps_qs.present? %>
+  <script type='text/javascript' src="//maps.google.com/maps/api/js?<%= google_maps_qs %>&v=3.26"></script>
```
xavijam: Attaching to 3.26 everywhere @alonsogarciapablo?

config/app_config.yml.sample
```diff
@@ -515,6 +515,115 @@ defaults: &defaults
         name: 'Watercolor'
         className: 'watercolor_stamen'
         attribution: 'Map tiles by <a href="http://stamen.com">Stamen Design</a>, under <a href="http://creativecommons.org/licenses/by/3.0">CC BY 3.0</a>. Data by <a href="http://openstreetmap.org">OpenStreetMap</a>, under <a href="http://www.openstreetmap.org/copyright">ODbL</a>.'
+    GMaps:
```
javitonino: Should google maps be here (in the default `app_config`)?

lib/assets/core/javascripts/cartodb3/data/basemap-providers-and-categories.js
```diff
@@ -0,0 +1,28 @@
+var PROVIDER_LEAFLET = 'leaflet';
```
xavijam: This is another smell we need to separate layer-definition-model in two, basemap and data layer models.

app/views/shared/visualizations/_google_maps.html.erb
xavijam: Why don't we use v3.27 (current stable one)?

alonsogarciapablo: Yes, that's the version we should use now.

config/app_config.yml.sample
alonsogarciapablo: I added it here so that we all have this feature "enabled" in our local environments. As discussed offline, feature won't be available unless the user has a Google Maps query string setup.

lib/assets/core/javascripts/cartodb3/editor/layers/basemap-content-view.js
xavijam: So, no changes needed then right?

config/app_config.yml.sample
xavijam: :+1:

xavijam:
I feel we have google-maps icons duplicated, is there any reason for that?

lib/assets/core/javascripts/cartodb3/data/layer-definitions-collection.js
```diff
@@ -253,130 +257,100 @@ module.exports = Backbone.Collection.extend({
...
     if (this.isBaseLayerAdded(newBaseLayerAttrs)) {
       this.trigger('baseLayerChanged');
-
       return false;
     }
 
     var newBaseLayerHasLabels = newBaseLayerAttrs.labels && newBaseLayerAttrs.labels.url;
+
+    var labelsLayerOptions = {
+      wait: true, // do not add/remove the layer unless it's created/removed/updated successfully
```
xavijam: Great, good decision.

lib/assets/core/javascripts/cartodb3/data/layer-types-and-kinds.js
```diff
@@ -44,6 +44,10 @@ module.exports = {
```
xavijam: In my opinion, we should move this to a layers helper or something like that (already pointed that in a previous file).

lib/assets/core/javascripts/cartodb3/data/map-definition-model.js
```diff
@@ -39,12 +41,15 @@ module.exports = Backbone.Model.extend({
...
+    this._layerDefinitionsCollection.bind('baseLayerChanged', this._onBaseLayerChanged, this);
   },
 
-  _saveMaxMinZoom: function () {
+  _onBaseLayerChanged: function () {
     var baseLayer = this._layerDefinitionsCollection.getBaseLayer();
+    var newProvider = basemapProvidersAndCategories.getProvider(baseLayer.get('category'));
+
     this.save({
+      provider: newProvider,
```
xavijam: Why do we need to know the provider in the map options?

lib/assets/core/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -580,6 +583,11 @@ F.prototype._formatWidgetAttrs = function (changedAttrs, widgetDefModel) {
 };
 
 F.prototype._onLayerDefinitionAdded = function (m, c, opts) {
+  // Base and labels layers are synced in a separate method
```
xavijam: Arg, I feel a bit resposible of that (and I don't like it). Let's try to avoid those problems in https://github.com/CartoDB/cartodb/issues/11581.

lib/assets/core/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -589,47 +597,117 @@ F.prototype._onLayerDefinitionAdded = function (m, c, opts) {
...
+  // decide if delegate or not events
+  this._editFeatureOverlay.render();
+};
+
 var CARTODBJS_TO_CARTODB_ATTRIBUTE_MAPPINGS = {
   'layer_name': ['table_name_alias', 'table_name']
 };
 
-F.prototype._adaptAttrsToCDBjs = function (attrs) {
+var BLACKLISTED_LAYER_DEFINITION_ATTRS = {
```
xavijam: Could we move those "static" cons at the beginning?

lib/assets/core/javascripts/cartodb3/editor/layers/basemap-content-views/basemap-model-factory.js
```diff
@@ -0,0 +1,198 @@
+var _ = require('underscore');
```
xavijam: Great factory

app/views/shared/visualizations/_google_maps.html.erb
alonsogarciapablo: Bumped to v3.27!

lib/assets/core/javascripts/cartodb3/data/layer-definition-model.js
```diff
@@ -339,6 +339,28 @@ module.exports = Backbone.Model.extend({
       }
```
alonsogarciapablo: Totally!

lib/assets/core/javascripts/cartodb3/data/layer-types-and-kinds.js
alonsogarciapablo: The fact that we have to check the type of layer models (using this helper) is also a smell. If we had a layer definition model for each type of layer, we would hopefully not need this helper and we could replace usages of it (conditionals) by polymorphism. I suggest we try to get rid of this in https://github.com/CartoDB/cartodb/issues/11582.

lib/assets/core/javascripts/cartodb3/data/map-definition-model.js
alonsogarciapablo: Map provider is persisted in the backend and should be stored as part of the map "model".

lib/assets/core/javascripts/cartodb3/deep-insights-integrations.js
alonsogarciapablo: We could, but file is sooooo big, that I thought it'd be better to leave them here, closer to where they are being used. We're splitting this file anyway, so we'll have to move them anyway. Unless you feel super strong about this, I'm okay with leaving them here.

lib/assets/core/javascripts/cartodb3/editor/layers/basemap-content-views/basemap-model-factory.js
alonsogarciapablo: Thanks!

alonsogarciapablo:
Thanks for the CR @xavijam! I've removed the duplicated icon files.

> have you "touched" something that could break something in the editor part?

Not that I'm aware of :crossed_fingers:.

xavijam:
> Thanks for the CR @xavijam! I've removed the duplicated icon files.
> 
> > have you "touched" something that could break something in the editor part?
> 
> Not that I'm aware of :crossed_fingers:.

Pefect then!!

lib/assets/core/javascripts/cartodb3/deep-insights-integrations.js
xavijam: oooook

lib/assets/core/javascripts/cartodb3/data/map-definition-model.js
xavijam: But do we use it?

lib/assets/core/javascripts/cartodb3/data/basemap-providers-and-categories.js
alonsogarciapablo: Not sure what's the convention there. If you thing there's a better place for this, I'm cool with moving it.

lib/assets/core/javascripts/cartodb3/deep-insights-integrations.js
alonsogarciapablo: Well, this is due to the fact that basemap changes have multiple actions (changes to the base layer, changes to the labels layer, and changes to the map provider) that need to be propagated to CartoDB.js all at once. That's the reason we handle them separately.

alonsogarciapablo: But I don't like it much either.

lib/assets/core/javascripts/cartodb3/data/map-definition-model.js
alonsogarciapablo: Yes. It's then included in the viz.json file:

```
{
  ...,
  'map_provider': 'leaflet',
  ...
}
```

Issue #11442, opened 2017-02-02
Don't let add a new geometry (point, line or polygon) when a query is applied over a source


xavijam:
## STR
- Create a map of points.
- Apply a query over that layer.

## Current behaviour
- Check that "add new feature" button is still there.
- 💥 .

## Expected behaviour
- We should disable the "add new feature" button, we can't provide that.

Problem found thanks to https://github.com/CartoDB/cartodb/issues/11356.

nobuti:
Picking this.

xavijam:
I'm adding all the checks we need before starting the acceptance (just in case).

nobuti:
Already in production.

PR #11448, opened 2017-02-03
Fixing problem related to custom query applied and displaying geometry edit buttons

### LINKED ISSUE #11053: Cannot edit geometry in newly added layer from import
### Steps to Reproduce

1. Create a map
2. Add a new layer, and in the dialog to choose dataset, select "Add an empty layer" (or connect a dataset)
3. Select the new layer

### Current Result
You cannot edit geometry/data. If you refresh the page, you can.

### Expected result
You can edit geometry

### Additional info
It appears the issue is related to the return values from the modal. When choosing an existing layer, it returns a table model with the proper data. When importing a new dataset/empty layer, it instead returns a table model, which has an attribute `table_visualization` which has the correct things.

The add layer model should always set a consistent table model.

cc @xavijam

xavijam:
Related ticket: #11442.
Also fixes #11053

---
**THIS IS AN IMPORTANT CHANGE**

From now on, we should check if a node is read-only, not within its related table-model. Check files changed in order to know what has changed.

```js
Analysis-definition-node-source-model
├── Query-Schema-Model
├── Query-Geometry-Model
└── Table-Model
…
analysisDefNodeSource.isReadOnly();
analysisDefNodeSource.isCustomQueryApplied();
```

cc @javierarce @nobuti @ivanmalagon @matallo 

---
Also, related to the ticket, we can't provide a way to edit or add new features when:
- a query is applied.
- if the source node is synced.
- if the node type is not source type.
- if user doesn't have write permission over that "table".

So we have to test these possibilities, if any of those (above list) are true, edit buttons should not appear. Added tests for render within `layer-content-view`.

So, in short, we should test (in an organization account):

· Starting with a map with a layer of points.
- [x] It should let you add a new feature (line, polygon, point, whatever).
- [x] It should let you edit a feature (line, polygon, point, whatever).
- [x] It should let you edit or add rows in the table view.

· Apply a custom query over that layer.
- [x] It should NOT let you edit a feature.
- [x] It should NOT let you add a feature.
- [x] It should NOT let you edit or add rows in the table view.

· Add a new empty dataset.
- [x] It should let you add a new feature (whatever you decide).
- [x] It should let you edit a feature (line, polygon, point, whatever).
- [x] It should let you edit or add rows in the table view.

· Remove the query and add an analysis over the layer.
- [x] It should NOT let you edit a feature.
- [x] It should NOT let you add a feature.
- [x] It should NOT let you edit or add rows in the table view.

· Add a sync table.
- [x] It should NOT let the user edit a feature in that sync layer.
- [x] It should NOT let the user add a feature in that sync layer.
- [x] It should NOT let you edit or add rows in the table view.

· Add a shared table (with read permissions).
- [x] It should NOT let the user edit a feature in that shared layer.
- [x] It should NOT let the user add a feature in that shared layer.
- [x] It should NOT let you edit or add rows in the table view of that shared table.

· Add a shared table (with write permissions but synced).
- [x] It should NOT let the user edit a feature in that shared layer.
- [x] It should NOT let the user add a feature in that shared layer.
- [x] It should NOT let you edit or add rows in the table view of that shared table.

· Add a shared table (with write permissions).
- [x] It should let the user edit a feature in that shared layer.
- [x] It should let the user add a feature in that shared layer.
- [x] It should let you edit or add rows in the table view of that shared table.


xavijam:
Sorry but I don't get you ^^^

xavijam:
From my point of view, it tests the problem we have found, buttons should not be visible when any of the circunstancies happen. It should not cover the functionality of the button, it is not related (here).

xavijam:
But, I'm totally open to test whatever we agree, no problem at all ("que conste!").

matallo:
basically we're testing the different variables sent to the template to show or hide the buttons depending on the different cases 
https://github.com/CartoDB/cartodb/blob/20ddd23dbe9f53ecba23b6aa2ded592f0c05742b/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-view.js#L306-L331

so, instead of testing `this.view._renderContextButtons();` every time, with different attributes
https://github.com/CartoDB/cartodb/pull/11448/files#diff-0a3dc81da17fb283203db02183608c43R129

I'd test

- isThereOtherWidgets: `this._widgetDefinitionsCollection.isThereOtherWidgets()`
- isThereTimeSeries: `this._widgetDefinitionsCollection.isThereTimeSeries()`
- isThereAnimatedTimeSeries: `this._widgetDefinitionsCollection.isThereTimeSeries({ animated: true })`

...

If those tests pass, there's no reason the template would fail, that's my point from doing truly unit tests rather than more integration-oriented tests

matallo:
my concern here, the scope of testing the render method should be it renders properly and the element is appended to the dom, if then it is visible or not is a different responsibility which maybe can be extracted to another method and then test that method

I'm trying to escape the strategy I myself follow of testing the different results of the same view/method/whatever from changing attributes 

sorry to be this nitpicking, as already said for me the code is 👍 it is just a comment for further approaches

matallo:
finally, we can set 🔥 on this line https://github.com/CartoDB/cartodb/blob/20ddd23dbe9f53ecba23b6aa2ded592f0c05742b/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-view.js#L316

xavijam:
Finally we are using this new logic (cc @javitonino) in Builder side, but we have to change this logic in the dataset view (https://github.com/CartoDB/cartodb/issues/11475).

lib/assets/core/javascripts/cartodb3/data/table-model.js
```diff
@@ -34,6 +34,13 @@ module.exports = Backbone.Model.extend({
   parse: function (r, opts) {
     var configModel = (opts && opts.configModel) || this._configModel;
 
+    // "Sometimes" table presenter returns the needed data within table_visualization
```
xavijam: Hey @CartoDB/builder-backend, this is a problem we (@javitonino and I) checked and we should fix it at some point. When a new table is created and added to a map, the returned attributes don't contain the real table data, it provides all the visualization info instead. So we have to trick it a bit taking the table_visualization.

javitonino: Some more info: The presenter for tables is a little insane, with information duplicated at different nesting levels, and inconsistent with the index presentation). I don't think we can fix it by changing the current presenter (as the current format is being used by the old editor). But we could create a /v2/ tables API that returns something more sensible, or fix it when we deprecate the editor.

juanignaciosl: Could you open an issue with the STR and the expected outcome? Depending on the changes, maybe we can tackle it now, or in a _likely_ API rewriting to come.

javitonino: I took a second look trying to reproduce, and it all starts making more sense now. It seems more like a confusion in frontend, but it's not easy to see:
- `table-model.js` actually represents a canonical visualization
- `table-model.js` uses the `/tables` endpoint to load data. This returns a table model with the visualization wrapped inside `table_visualization`. This is what this patch fixes
- `tables-collection.js` is a collection of TableModels. However, it is loaded with the `/viz` endpoint, which returns the visualization at top level.

So that's why TableModel worked when choosing a dataset from a list, but it didn't load properly when creating a new table. Table model represents a canonical visualization (not a table!) and uses the tables endpoint. TableCollection contains uses the visualizations endpoint. So depending on how the TableModel is initialized, it works or not.

matallo:
so much ❤️ I'm sorry as I feel responsible for the further changes in how we make the check, but I feel much more confident now

nobuti:
I found an issue. If we applied an analysis, the button to add a new feature is hidden. But if we apply a custom query, it remains visible. In fact, the user can add new features, at least, to the dataset:

feature-with-query

If adding the feature doesn't represent a real issue, at least it should be nice to follow the same behaviour we have for analyses.

nobuti:
I can't add features to an empty layer/dataset:

feature-empty

xavijam:
Checking both 💋

xavijam:
Checked that the staging server didn¡t have the proper changed files 🃏 .

nobuti:
This works 💯  (once deployed)

xavijam:
yay!

nobuti:
Waiting green after merged master. Then 🚀 .

nobuti:
Frontend tests passing.

PR #11457, opened 2017-02-06
split browserify tasks

matallo:
- as a first step to make assets compilation faster

compare running times:

```
( grunt && grunt dev; )  83.30s user 15.05s system 0% cpu 3:04:00.06 total

( grunt && grunt dev; )  208.18s user 56.56s system 109% cpu 4:01.48 total
```


matallo:
this is being done in https://github.com/CartoDB/cartodb/issues/11962

closing, then

Issue #11465, opened 2017-02-06
Uglify task slowness


rochoa:
Review the recent degradation of uglify task. It seems related to tangram integration and some problems with the uglify version we have.

See https://github.com/mishoo/UglifyJS2/issues/321

cc @xavijam @luisbosque @jorgesancha @javisantana

xavijam:
cc @javitonino

javitonino:
Created a PR for my little experiment updaing uglifyJS -> https://github.com/CartoDB/cartodb/pull/11471

donflopez:
Thanks @javitonino! I'll try it in staging :)

donflopez:
It seems to be working properly, closing!

PR #11471, opened 2017-02-07
Update uglify

javitonino:
#11465 

This brings back uglify time for vendor_editor 3 from 25 min to 2 min:
http://deploy.int.cartodb.net/user/javitonino/my-views/view/CartoDB+Central/job/Carto%20-%20deploy%20to%20staging%20(carto.com)/1242/

The only concerning thing I could find in the [changelog](https://github.com/gruntjs/grunt-contrib-uglify#release-history) is "screwIE8 is enabled by default." (but we do not support IE8 already, so it should not be a problem)

If this is a concern, grunt-contrib 1.0.2 + uglifyjs 2.6.4 should also work, but I haven't tested those versions.


donflopez:
Thank you @javitonino!!

PR #11480, opened 2017-02-07
Quota precheck

nobuti:
This PR implements https://github.com/CartoDB/camshaft/issues/238. This is a complete revamp of https://github.com/CartoDB/cartodb/pull/11297

We are using a dataservices api to: 
- obtain the user's quota
- check the quota before to run an analysis. 

Whenever we enter in (or changes) the analysis form view, we request the user quota in order to update the data. Take in mind that the quota could have shared under the organisation, so other users could change the remaining quota at the same time. So at least, one request is made. Also, when the analysis needs a quota precheck, two more requests need to be made:
- one for getting an estimation to get the credits needed to run the analysis
- another to check if the quota is enough to run the analysis with the estimation above.

So, if the analysis needs to check the quota, 3 requests are made.

### How to test
- Check the scenarios above, making sure that the request matches.
- You should check that the information we are showing to the user is right in these cases:
    - The user has soft limit but no quota assigned.
    - The user has soft limit but no quota remained.
    - The user has soft limit but the quota is not enough to run the whole analysis.
    - The user has soft limit and the quota is enough.
    - The same as above but with hard limit.
- Also, if dataservices api is not ready, you should check that analyses that require quota are not available. For instance, data observatory should not be part of the list of analysis in the modal.

Take this messages as a guidance:
3f5888f4-e8a1-11e6-83a0-1f110c7b1ea8

Test it in ded05, all has been setup properly there. Ask for user too.

cc @xavijam


lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-quota/analyses-quota-info.js
```diff
@@ -0,0 +1,62 @@
...
+      this._state = 'fetching';
+      this.SQL.execute(SERVICE_QUOTA_QUERY, null, {
+        success: function (data) {
+          this._state = 'fetched';
+          this._onFetchSuccess(data);
+          this._success && this._success();
+        }.bind(this),
+        error: function (err) {
+          this._state = 'errored';
+          var error = JSON.parse(err.responseText);
```
xavijam: Could we use `error-parser.js`?.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-quota/analyses-quota-info.js
```diff
@@ -0,0 +1,62 @@
...
+          this._error && this._error(error.error[0]);
+        }.bind(this)
+      });
+    }
+  },
+
+  _onFetchSuccess: function (data) {
+    var models = data.rows;
```
xavijam: Great :+1:.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-quota/analyses-quota-info.js
```diff
@@ -0,0 +1,62 @@
...
+    _.each(models, function (model) {
+      var m = this.findWhere({service: model.service});
```
xavijam: `this.getService(model.service)`?

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-quota/analyses-quota-info.js
```diff
@@ -0,0 +1,62 @@
...
+    }, this);
+  },
+
+  getService: function (service) {
+    return this.findWhere({service: service});
+  },
+
+  isFetching: function () {
+    var state = this._state;
+    return state === 'fetching';
```
xavijam: `this.getState() === 'fetching'`?

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-quota/analyses-quota-info.js
```diff
@@ -0,0 +1,62 @@
...
+  },
+
+  getSate: function () {
```
xavijam: typo, `getState`?

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-quota/analyses-quota-options.js
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+var unity = function (input) {
+  return input;
+};
+
+var routingAnalysisQuota = {
+  name: _t('editor.layers.analysis-form.quota.analysis-type.routing'),
+  api: 'routing',
+  user: 'mapzen_routing',
```
xavijam: What do we mean with user? provider?

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-quota/analyses-quota-options.js
```diff
@@ -0,0 +1,99 @@
...
+  needsQuota: true,
+  transform: unity
+};
+
+var ANALYSES_WITH_QUOTA_MAP = {
+  'trade-area': {
+    name: _t('editor.layers.analysis-form.quota.analysis-type.trade-area'),
+    api: 'isolines',
+    user: 'here_isolines',
+    transform: function (input, formModel) {
```
xavijam: 👍 , like it.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-quota/analyses-quota-options.js
```diff
@@ -0,0 +1,99 @@
...
+      // the isoline analysis consume N rows x M tracts
+      return input * formModel.get('isolines');
```
xavijam: cc @rafatower.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-quota/analyses-quota-presenter.js
```diff
@@ -0,0 +1,102 @@
+var _ = require('underscore');
+var AnalysesQuotaOptions = require('./analyses-quota-options');
+var CONTACT_LINK_TEMPLATE = _.template("<a href='mailto:<%- mail %>'><%- contact %></a>");
+
+var getNoCreditMessage = function (quota, type) {
+  var link;
+  var message;
+  if (quota.totalQuota === 0) {
+    link = CONTACT_LINK_TEMPLATE({
+      mail: _t('editor.layers.analysis-form.quota.no-credits-mail'),
```
xavijam: There are 5 possibilities with this email:

- If it is a CARTO account in the cloud (standard one) -> sales@carto.com :+1:.
- If it is a CARTO account in the cloud and user belongs to an organization (he/she is NOT the owner) -> organization owner email.
- If it is a CARTO account in the cloud and user belongs to an organization (he/she IS the owner) -> support@carto.com.
- If it is an open source installation -> no idea, maybe support@carto.com.
- If it is an on-premise installation -> onpremise-support@carto.com.

That's all!

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-quota/analyses-quota-presenter.js
```diff
@@ -0,0 +1,102 @@
...
+      contact: _t('editor.layers.analysis-form.quota.no-quota-assigned-contact')
+    });
+    message = _t('editor.layers.analysis-form.quota.no-quota-assigned-body', {
+      analysis: AnalysesQuotaOptions.getAnalysisName(type),
+      contact: link
+    });
+  } else {
+    link = CONTACT_LINK_TEMPLATE({
+      mail: _t('editor.layers.analysis-form.quota.no-credits-mail'),
```
xavijam: same as above I feel.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-quota/analyses-quota-presenter.js
```diff
@@ -0,0 +1,102 @@
...
+        if (data.quotaInfo.enoughQuota) {
+          // You need to use approximately xxx of your credits.
+          body = _t('editor.layers.analysis-form.quota.enough-quota', {
+            credits: estimation
+          });
+          type = 'success';
+        } else {
+          // We're sorry the current quota is insufficient to enrich your data. Rows will be set to null and analysis may not complete. Please contact us to extend your quota for this function.
+          link = CONTACT_LINK_TEMPLATE({
+            mail: _t('editor.layers.analysis-form.quota.no-credits-mail'),
```
xavijam: dixit

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-controls-view.js
```diff
@@ -25,31 +27,52 @@ module.exports = CoreView.extend({
...
     this._analysisNode = opts.analysisNode;
+    this._quotaInfo = opts.quotaInfo;
+    this._querySchemaModel = opts.querySchemaModel;
+
+    AnalysesQuotaEstimation.init(opts.configModel);
+    AnalysesQuotaEnough.init(opts.configModel);
 
     this._viewModel = new Backbone.Model({
-      userFetchModelState: 'idle',
+      userFetchModelState: opts.quotaInfo.getSate(),
```
xavijam: getState I guess.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-quota/analyses-quota-options.js
nobuti: the key in the userModel, to get the block price and size of the current analysis.

xavijam: Ahhhhh 👍

saleiva:
cc/ @brittanymicek there you have the mockups!

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-quota/analyses-quota-presenter.js
nobuti: how do we know the last two?

xavijam: Open source: with `cartodb_com_hosted` attribute within `configModel`.
On premise: ... arg, I don't remember, checking...

xavijam:
:+1:, let's go to next step, acceptance!

javitonino:
Merged master so we can deploy (DB migrations...). No conflicts ✨

javitonino:
Deployed to ded07, didn't work (no DS API there). For reference: we need to test this in the staging cloud.

javitonino:
Some issues:
 - When geocoding with mapzen, no quota message is displayed.
 - Sometimes, an undefined credits message is shown:
screenshot_20170209_150632
To reproduce, I added a geocoding analysis, by street address, and clicked rapidly in the city checkbox. Looks like a race condition (notification displaying before explain finishes).
- Not pixel perfect (notification is 1 pixel to the left)
screenshot_20170209_152700

- I'm getting this in a single user with soft geocoding enabled and not enough quota (30 rows for 5 quota):
screenshot_20170209_152931
  - The EXPLAIN is correct (30 rows), the quota response is `monthly_quota: 5, soft_limit: true`, `cdb_enough_quota: true`. Not sure if this situation is possible in real code.
  - If I run the analysis, it works fine, and then, trying a second one is also incorrect (I don't know where that 290 comes from):
screenshot_20170209_153503

nobuti:
Taking a look.

nobuti:
I found a problem with the estimation when we are editing an analysis already run. In this scenario, we need to calculate the estimation using the right query. Working on it.

nobuti:
All fixed but the last two points. In fact, I'm a little worried about it. I feel I need a deeper understanding of how the `EXPLAIN` is working using analysis query. For example, take in mind the case above:

- The EXPLAIN is correct (30 rows), the quota response is monthly_quota: 5, soft_limit: true, cdb_enough_quota: true. Not sure if this situation is possible in real code.
- If I run the analysis, it works fine, and then, trying a second one the estimation using the `EXPLAIN` query returns 290. The query used is something like `api/v2/sql?q=EXPLAIN%20select%20*%20from%20analysis_1ea6dec9f3_b9d7df9713f3778f2f3bb9babb5019eb4077549f`

I think this is pretty critical because we are telling the user a second analysis is going to take 290 credits over a dataset with 30 rows. 

cc @xavijam @javisantana @rafatower.

rafatower:
@nobuti et al.

What happens with the `EXPLAIN` is the following:

* The first analysis is applied on a freshly imported table. They meet two interesting conditions:
  - indices are created on them as part of the cartodbfy'ation.
  - a `VACUUM ANALYZE` is executed on them as well.

either of these cause the statistics of the table to be updated, so that next time a `EXPLAIN` is issued it can return an accurate value.

* When running the `EXPLAIN` on an analysis table, as opposed to freshly imported tables, they do not have indices on them nor went through an `ANALYZE`. Therefore the statistics are not up to date and the `EXPLAIN` (or the query planner for what matters) can only return an estimation of the real size of the table.

Here is a couple of links explaining the row estimation:

https://www.postgresql.org/docs/9.5/static/sql-explain.html
https://www.postgresql.org/docs/9.5/static/row-estimation-examples.html

There are several solutions to this, ranging from assuming it's just an estimate, to `ANALYZE` the resulting table as part of the analysis to using a `SELECT COUNT` instead.

Let me explore those possible solutions a little and then decide what to do with this.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-quota/analyses-quota-estimation-input.js
```diff
@@ -0,0 +1,49 @@
...
+      query: query
+    });
+
+    if (!this.deferred || this.deferred.state() !== 'pending') {
+      var deferred = $.Deferred();
+      this.deferred = deferred;
+
+      // The response is like {"rows":[{"QUERY PLAN":"Seq Scan on paradas_metro_madrid  (cost=0.00..6.97 rows=325 width=108)"}],"time":0.001,"fields":{"QUERY PLAN":{"type":"string"}},"total_rows":1}
+      var parseResponse = function (response) {
+        var regex = /rows=(.+)\s/;
```
rafatower: In order to avoid the regex you could instead use `EXPLAIN (FORMAT JSON) <%= query %>;` in the query template above. It'll return something like this:
```js
{
  "total_rows": 1,
  "fields": {
    "QUERY PLAN": {
      "type": "JSON"
    }
  },
  "time": 0.002,
  "rows": [
    {
      "QUERY PLAN": [
        {
          "Plan": {
            "Plan Width": 66,
            "Plan Rows": 5,
            "Total Cost": 1.01,
            "Startup Cost": 0,
            "Alias": "analysis_1ea6dec9f3_361cd4c17ab0221f98485b0d05284164523f59d4",
            "Relation Name": "analysis_1ea6dec9f3_361cd4c17ab0221f98485b0d05284164523f59d4",
            "Node Type": "Seq Scan"
          }
        }
      ]
    }
  ]
}
```

HTH

rafatower:
There's a background process to keep stats updated. In an ideal word we shouldn't worry much about stats. See https://www.postgresql.org/docs/9.5/static/routine-vacuuming.html#VACUUM-FOR-STATISTICS 

But there's another important aspect that I did not realize and kept me puzzled for a while: analysis tables are created with a `CREATE TABLE AS SELECT...LIMIT 0`, then populated in a batch query.

There are some interesting effects because of that, that fully apply to the case tested above:

* When a table is empty, its stats are kept "blank", producing very bad estimations. E.g:
```
tests=# CREATE TABLE test_row_estimation AS SELECT id, random() AS value FROM generate_series(1,5) AS id LIMIT 0;
SELECT 0
tests=# EXPLAIN SELECT * FROM test_row_estimation ;
                               QUERY PLAN                               
------------------------------------------------------------------------
 Seq Scan on test_row_estimation  (cost=0.00..30.40 rows=2040 width=12)
(1 row)

tests=# SELECT count(*) FROM test_row_estimation ;
 count 
-------
     0
(1 row)
```

funny enough, the row estimation is wrong even after forcing the update:
```
tests=# ANALYZE test_row_estimation ;
ANALYZE
tests=# EXPLAIN SELECT * FROM test_row_estimation ;
                               QUERY PLAN                               
------------------------------------------------------------------------
 Seq Scan on test_row_estimation  (cost=0.00..30.40 rows=2040 width=12)
(1 row)
```

I think this is due to postgres considering "zero" as "I don't have stats yet":
```
tests=# SELECT relpages, reltuples FROM pg_class WHERE relname = 'test_row_estimation';
 relpages | reltuples 
----------+-----------
        0 |         0
(1 row)
```

* Eventually the table is populated, but depending on the config and the number of changes/transactions in the table, the `autovacuum` process may or may be not triggered. See https://www.postgresql.org/docs/9.5/static/runtime-config-autovacuum.html

> autovacuum_analyze_threshold (integer)
> Specifies the minimum number of inserted, updated or deleted tuples needed to trigger an ANALYZE in any one table. The default is 50 tuples. This parameter can only be set in the postgresql.conf file or on the server command line; but the setting can be overridden for individual tables by changing table storage parameters.

So if I insert a few rows and then run `ANALYZE` the stats will report pretty accurate results:
```
tests=# insert into test_row_estimation values (1,1), (2,2), (3,3), (4,4);
INSERT 0 4
tests=# EXPLAIN SELECT * FROM test_row_estimation ;
                               QUERY PLAN                               
------------------------------------------------------------------------
 Seq Scan on test_row_estimation  (cost=0.00..30.40 rows=2040 width=12)
(1 row)

tests=# ANALYZE test_row_estimation ;
ANALYZE
tests=# EXPLAIN SELECT * FROM test_row_estimation ;
                             QUERY PLAN                             
--------------------------------------------------------------------
 Seq Scan on test_row_estimation  (cost=0.00..1.04 rows=4 width=12)
(1 row)
```

but the automatic ANALYZE will never happen for less than 50 rows.

I'll check how reasonable would be to run the `ANALYZE` on analysis cache tables.

rafatower:
I investigated a little further and ran a [benchmark](https://gist.github.com/rafatower/efbaa2924b3a1cc7b7ea0698adafe3b9).

@rochoa given the numbers and for the general case I'd say we may be interested in running `ANALYZE` on cache tables. I'd add it around these lines:

https://github.com/CartoDB/camshaft/blob/44bfe26100a49c394c8cca2fc51e42050ffdd15a/lib/service/database.js#L323

It is anyways what the `autovacuum` process is meant to do in the background and the analyses queries are already executed asynchronously.

nobuti:
@rafatower in order to test this again, do we need to install something in staging? We did some things in ded05 before.

nobuti:
@rafatower all set up, @ethervoid took care 💯

rafatower:
Something I noticed while testing (as DS API is not installed in all stagings as you probably know by now): this is going to require the DS API to be configured in order to work. I think we should do a couple of things about that: add it to the NEWS.md as a required dependency for Builder (at least to run analyses), and maybe add a feature flag (?) to bypass the checks. This is kind of important for the FOSS.

brittanymicek:
Hey guys! Sorry for the delay here! <3
@nobuti @saleiva 

CONFIRM YOUR ANALYSIS
To complete this request[1] your account will be charged $14 per 1,000 additional analysis credits.[2]


CONFIRM YOUR ANALYSIS
The requested action will only geocode as many locations as your available credits allow.[3]

CONFIRM YOUR ANALYSIS
Please use your credits for this action[4].



NOTES:
[1] Is "request" the correct action being taken? As in, the user is requesting to run an analysis function?
[2] I would not include the estimated cost "max. $26 extra cost." It does not match the equation of $14 per 1,000 credits. While I'm sure $26 is correct (presuming the user doesn't use a full 2,000 credits) it requires the user to do some mental math, to confirm that they aren't being overcharged or taken advantage of. Also, the exact price should appear on the monthly or annual bill as opposed to here. I think the goal here is not so much to say it will cost you X amount of dollars to perform a function, but to raise awareness that there is a cost for extra credits and that the requested analysis requires these extra credits. I have also deleted the ambiguity from the original copy ("might incur"). I don't think we should introduce an element of surprise when it comes to billing/paying. Either the user will be charged $14 per 1,000 credits or not.
[3]What is the goal here? To communicate the objective we do not need to draw attention to the fact that they aren't being charged.  
[4] Is this just a general "confirmation"? I would try to specify what the "this" is, if possible. What are they using some of their credits for?

nobuti:
@brittanymicek the idea is to give an estimation of extra credits needed to run the analysis, not money because it's not accurate at the end. 

- It should be great if we include this estimation in [2]. Right now we are showing this: 
  

- [3] should be like [2] but without the extra credits part, because the user's config doesn't allow to over charge. The goal is to let the user run the analysis up to the limit of her/his credit. The analysis' result will be uncompleted. Right now we are saying this: `We're sorry the current quota is insufficient to enrich your data. Rows will be set to null and analysis may not complete. Please contact us to extend your quota for this function.`
- In [4] happens pretty much the same thing as [1]. It should be great to include the estimation:

nobuti:
Ready for another round.

brittanymicek:
@nobuti 
Is this better? Clearer?

CONFIRM YOUR ANALYSIS
To complete this request an estimated {%number.credits} additional credits are needed. Additional fees will incur when applicable.


CONFIRM YOUR ANALYSIS
The requested action will only perform geocoding analysis of your location data up to the limit of your available credits. The remaining, non-geocoded, rows will be set to null and analysis may not complete. Please contact us to extend your quota for this function.

CONFIRM YOUR ANALYSIS
Please use an estimated {%number.credits} credits for this action.

lib/assets/core/javascripts/cartodb3/data/analyses.js
```diff
@@ -78,7 +79,8 @@ var MAP = {
     modalCategory: 'create_clean',
     checkIfValid: function (configModel) {
       if (!configModel) throw new Error('configModel is required');
-      return configModel.dataServiceEnabled('data_observatory');
+      var dsReady = DataServicesApiCheck.get().isReady();
```
xavijam: I'd not mix if the analysis is valid with the data-services check. Why not make that check in the add-analysis-view for each analysis that requires if it is valid? what do you think?

xavijam: Ahh, wait, we talked about moving this logic inside the dataServiceEnabled method, using that model within the configModel. Do you feel more comfortable with this idea?

nobuti: At view level, we need to create a hash of analyses to compare, pretty much the same we are doing already in this file. Doing it as part of dataServiceEnabled would lead us to a circular dependency, data-services-api-check.js already requires configModel to be able to use the SQL query api. We need to change the signature of the constructor at least, to get access to those attributes without accessing to the model, right?
On the other hand, checkIfValid is a terrible name IMO, I can't see the validity being checked :trollface:

xavijam: :joy:, ok, agree about the circular dependency, so :+1:. About the name, what about `checkIfEnabled`?

nobuti: It makes more sense to me 👍

xavijam:
I've spotted a problem checking the quotas with DO analysis. When user types the new column name, quotas fetch is triggered while user is typing. We should wait until the form is valid at least.

nobuti:
In order to check if quota is required we need the quota request to get the analyses with quota. So I guess this should be a global object, created when Builder is initialised, not an object created in that view. We need to change the approach then.

nobuti:
Ready for another round :crossed_fingers:

saleiva:
@noguerol be sure you test this or comment the issue you found yesterday. It might be fixed with this

xavijam:
We have found a problem related with a cache of the `cdb_dataservices_client.cdb_service_quota_info()` service. Already checked with @rochoa and seems like it is already known (@javitonino @rafatower). Thing:

- Take a random user, and set an isolines quota (whatever).
- Go to Builder, spend all isolines quota, enjoy it :D.
- If you try to run again isolines, an infobox saying you can't run it because your user doesn't have enough credits should appear.
- Then, change the user quota and add 1000 more, for example.
- Go back to Builder and run another isolines analysis. Check that the previous infobox keeps appearing, because the service quota info keeps saying the quota is the previous one, not the new one.

Is there a ticket for this?

rafatower:
Let's discuss it here: https://github.com/CartoDB/dataservices-api/issues/349

noguerol:
So far I've found some friends concerning credit estimation:

1) Running a 1 tract time based isoline over 10 points returns 120 credit consumption estimation, while it should be 10.
2) Running a 1 measurement DO augmentation over 10 points returns 340 credits estimation (over the 10 credit quota I previously assigned)
3) Changing geocoder to Here, reducing quota to 5 and running the analysis on 10 points returned a 95 credit estimation. On the same scenario but fixing the quota on 200, a correct 10 credit consumption is calculated, but only 110 free credits seem to be remaining (maybe because they've been consumed anywhere in the org?). Continuing with the same scenario, changing provider back to Mapzen resets the credit count to the 200 limit.

noguerol:
I've also found an inconsistency in estimations of remaining credits: Running isolines with Here with a 500 credit limit estimates a correct 10 credit consumption (for 1 tract , 10 points) and 370 remaining credits. But pushing down softlimit to 400 shows an incorrect estimation of 40 credits and no remaining credits available. In that same scenario, changing provider to Mapzen (with the same softlimit) shows a correct estimation of 10 credits, now with 250 available.

noguerol:
**Incorrect messaging**: when applying a 1 tract isoline analysis over 536 points a 122 credit consumption is calculated, but even there are free credits remaining and the analysis could be applied an "We're sorry the current quota is insufficient to enrich your data" message appears

captura de pantalla 2017-03-07 a las 18 00 45

nobuti:
I've fixed an issue in the frontend, we were showing `estimation - creditsLeft` instead of just `estimation`. This solves some issues above. @noguerol I could not reproduce all of them, so it should be great to be able to do the acceptance together in order see requests for every steps in the road.

On the other hand, I've found some weird stuff that it could explain the case changing the isoline provider. It turns out that the consumed quota it's not the same for every provider, so changing provider shows different consumed quota. This is an edge case because it's not the usual scenario to have to change the provider due to contract constraints.

And for mapzen and isoline specifically, I found additional issues: 
- Mapzen supports 4 tracts max, and we are not controlling this in the frontend. As far I could see, if you try to run an analysis with mapzen and  6 tracts, the analysis seems to be inconsistent, not sure it is related to the next point.
- Mapzen analysis returns minutes for `data_range` when it should be seconds. This could cause that the numbers returned by the `EXPLAIN` query we are using to estimate the consume to have a huge different (estimation of 120 vs final consume of 22, in some tests).

Also, we have the handicap of dealing with cache when changing the user's data (https://github.com/CartoDB/dataservices-api/issues/349).

So, IMO there are too many caveats in order to get a reliable result right now. If we can't get an understandable result as product's owners, for the final user it could be frustrating.

ethervoid:
### Quota usage

Yes, we have a different key to compute the usage of quota for different providers. So when we check the usage for a user and provider we don't aggregate the consumption of the different providers.

### Mapzen isolines tracts support

Mapzen isolines support, by now, a max of 4 contours (tracts), you can check it with this request""

```
http://matrix.mapzen.com//isochrone?json={"locations": [{"lat": 40.42028, "lon": -3.70568}], "costing": "pedestrian", "polygon": true, "contours": [{"color": "tbd", "time": 5}, {"color": "tbd", "time": 15}, {"color": "tbd", "time": 20}, {"color": "tbd", "time": 25}, {"color": "tbd", "time": 30}]}&api_key={{matrix_api_key}}
```

the result for that request is:

```
{
  "error_code": 152,
  "error": "Exceeded max contours:4",
  "status_code": 400,
  "status": "Bad Request"
}
```

### Mapzen isolines range

I've created an issue for the data range part: https://github.com/CartoDB/dataservices-api/issues/350

As @nobuti said if we can't split the tract, for example 100 seconds in 4 tracts, we can get weird results. Take into account that 100 seconds in 4 tracts are 25 seconds per tract and Mapzen accepts minutes instead of seconds so are passing 0 or 0.X. If we pass 0.X (which is what I'm pretty sure we're doing) we get a 4xx error:

```
{
  "error_code": 111,
  "error": "Insufficiently specified required parameter 'time'",
  "status_code": 400,
  "status": "Bad Request"
}
```

and if we pass 0 is ignored.

### Cache issues

The cache issue is IMHO not really a killer problem because we, as admins, can change the quota for a user and wait until is propagated before we communicate them the change.

nobuti:
I've opened an issue to address the max tracts for mapzen in the UI: https://github.com/CartoDB/cartodb/issues/11723

noguerol:
So, everything except the over quota but applicable analysis seem to be solved: 

image

nobuti:
Does the user have soft limit?

noguerol:
Yes, the 10 available credits in that case were the limit.

nobuti:
Fixed the case when the user couldn't run the analysis even having soft limit. In this case, the user had no credits left, but the button was shown disabled.

screen shot 2017-03-10 at 10 53 03

PR #11488, opened 2017-02-08
Ds client upgrade

ethervoid:
Deploy it after the deploy the client version in all user DBs

// I've changed Gruntfile to check for 2.14 and 2.15 npm version because nvm install the 2.15 for the 0.10.x version of node. What do you think @xavijam ?

### DIFF
--- a/app/models/user/db_service.rb
+++ b/app/models/user/db_service.rb
@@ -22,7 +22,7 @@ module CartoDB
       SCHEMA_GEOCODING = 'cdb'.freeze
       SCHEMA_CDB_DATASERVICES_API = 'cdb_dataservices_client'.freeze
       SCHEMA_AGGREGATION_TABLES = 'aggregation'.freeze
-      CDB_DATASERVICES_CLIENT_VERSION = '0.14.1'.freeze
+      CDB_DATASERVICES_CLIENT_VERSION = '0.15.0'.freeze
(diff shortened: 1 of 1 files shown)


Gruntfile.js
```diff
@@ -3,7 +3,7 @@
  var jasmineCfg = require('./lib/build/tasks/jasmine.js');
  var duplicatedDependencies = require('./lib/build/tasks/shrinkwrap-duplicated-dependencies.js');
 
- var REQUIRED_NPM_VERSION = /2.14.[0-9]+/;
+ var REQUIRED_NPM_VERSION = /2.1[4-5].[0-9]+/;
```
xavijam: Uhm, I don't remember exactly why we set this specific version, but there was a reason behind. Do you remember it @rochoa ?

ethervoid:
Removed the Gruntfile part to avoid problems

Gruntfile.js
rochoa: We want to keep npm-shrinkwrap as similar as possible. Using different versions of npm 2.x series was creating different npm-shrinkwrap files, making the review process a bit painful.

@ethervoid, you can force a version by doing `npm install -g npm@2.14.x`, that way you don't have to keep switching it with nvm.

ethervoid: Understood, makes sense!

Thank you for the tip to for version 2.14 :) :)

PR #11491, opened 2017-02-08
Exclude FREE users from overquota calculations

javitonino:
Fixes https://github.com/CartoDB/cartodb-platform/issues/3073

Acceptance: Run this manually in staging. You may need to do some usage of quota services beforehand in order to test that this works.


app/models/user.rb
```diff
@@ -517,28 +517,28 @@ def password_confirmation=(password_confirmation)
...
-        over_mapzen_routing = u.get_mapzen_routing_calls > limit
+      limit = u.mapzen_routing_quota.to_i - (u.mapzen_routing_quota.to_i * delta)
+      over_mapzen_routing = u.get_mapzen_routing_calls > limit
 
-        if over_geocodings || over_twitter_imports || over_here_isolines || over_mapzen_routing || over_obs_snapshot || over_obs_general
-          users_overquota.push(u)
-        end
+      if over_geocodings || over_twitter_imports || over_here_isolines || over_mapzen_routing || over_obs_snapshot || over_obs_general
+        users_overquota.push(u)
+      end
```
gfiorav: Maybe we can take the opportunity to create a `Carto::QuotaService` and get this out of the user model!

javitonino:
Hah, tests are failing, back to dev.

app/models/user.rb
javitonino: Not a bad idea, can be worth it to remove code that is pretty much only used by a rake (and a superadmin controller) from the model.

app/services/carto/overquota_users_service.rb
```diff
@@ -0,0 +1,54 @@
...
+      ::User.where(enabled: true, organization_id: nil).exclude(account_type: 'FREE').use_cursor.each do |u|
+        overquota_users << u if services_overquota(u, delta)
+      end
+      overquota_users
+    end
+
+    SERVICES = %w(geocoding twitter_datasource here_isolines obs_snapshot obs_general mapzen_routing).freeze
+    def services_overquota(user, delta)
+      SERVICES.any? do |service|
+        user.send("get_#{service}_calls") > user.send("#{service}_quota").to_i * (1 - delta)
```
gfiorav: Oh boy the meta!

I'll forever be divided: this or have an method array?

Since you already have an array `SERVICES` maybe you could just have an explicit array. I mean, it's not full on META where you're looking for methods that start with `get_` and end in `_calls` so why not?

javitonino: We would have to make two arrays (or array of hashes) since we are calling two different methods per service, wouldn't we?

javitonino:
Rerunning tests, but it only failed in that random shp_spec that sometimes decides to randomly explode.

gfiorav:
Acceptance is 🍏 , but let's wait until the discussions are over before deploy.

Issue #11462, opened 2017-02-06
Enable the possibility to select an icon previously uploaded by the organization admin


xavijam:
So the use case is:

- The organization admin uploads an icon, in order to be used in any map by any organization user.
- Any organization user should start using it from its map view, selecting it from the assets modal.

We are enabling the possibility to upload any kind of asset in order to be used like a marker (https://github.com/CartoDB/cartodb/pull/11450), but we have realized that I didn't specify the need of adding the `Organization icons` panel in that new modal view.

So instead of making the PR even bigger, I'm opening this ticket.

cc @saleiva @jorgesancha

xavijam:
Great @matallo we have this in CR 💋

saleiva:
@noguerol can you also participate on this one?

noguerol:
Is this on staging already?

javierarce:
@noguerol Yes, I'll answer you in private.

PR #11492, opened 2017-02-08
image picker: select an icon previously uploaded by the organization admin

matallo:
fix https://github.com/CartoDB/cartodb/issues/11462

added the organization uploads tab to the image picker

a7f44640-9c81-44b9-8f12-5e2810304d55

I didn't cover the case of icon upload by the organization admin, as I think it fails out of the scope of the issue, but it could be extended afterwards to support this (and better resemble the "Your icons" tabs)

---

**acceptance**

- add an icon as an organization admin
- and use it in the map
- if the user doesn't belong to an organization, it shouldn't appear at all


matallo:
quite advanced on this, do we have an state for the case when the organization doesn't have any icon uploaded yet? @xavijam @javierarce 

b5e6ff68-4a4f-4d6e-aa99-ee8c53eb9858

the first thing that came to my mind was not showing the tab at all, but it is after the modal is rendered that the organization icons are fetched :/

xavijam:
That's a good one, don't we have a 0 results state? have you seen it in the designs @javierarce?

javierarce:
No, I'm afraid we don't have it / I haven't seen it.

xavijam:
Let's go to the source of truth, design 🍡 .

javierarce:
I guess hiding the tab in case the user didn't have any assets would be the best solution, but that would require an extra call.

matallo:
if we can afford that extra call I'm down to follow that path of thinking

xavijam:
Fine by me, go ahead!

matallo:
now the tab would only appear if the User is an Organization user and the organization icons collection is not empty

we need to make two extra calls (the organization icons collection, and the user icons collection, which we had to extract) but further requests are cached

lib/assets/core/javascripts/cartodb3/components/form-components/editors/fill/input-color/assets-picker/assets-view.js
```diff
@@ -352,8 +404,22 @@ module.exports = CoreView.extend({
...
+  _onFetchOrganizationAssetsError: function (model, response) {
+    this._setError(response);
+  },
+
+  _onFetchUserAssetsError: function (model, response) {
+    this._setError(response);
+  },
+
+  _onFetchOrganizationAssets: function () {
+    this._stateModel.unset('status');
```
nobuti: I feel this weird, I understand it but it's not natural to me. Also, this led to having two render calls, if I'm not wrong.

matallo: Indeed, let's open for discussion, there's a listen with the model change which renders the view.

https://github.com/CartoDB/cartodb/blob/b7409461881839b582d170ca608fe568317dee58/lib/assets/core/javascripts/cartodb3/components/form-components/editors/fill/input-color/assets-picker/assets-view.js#L114

I just tried to follow the current flow. Any suggestion to improve this?

xavijam:
I encourage you how, for example, we code the add widget modal, where several requests are needed before render the view displaying a new state. For me it should work in the same way, if possible (take into account this modal will be the main modal in the future, so better to make it properly than other thing).

matallo:
roger!

matallo:
I tried to go full widgets implementation, but there's no model related to fetched collections to rely on, and after replicating the `bodyView` and that stuff I trimmed back to something quite similar

anyways, I took into account the `$.when` suggestion (and fixed the tests)

https://github.com/CartoDB/cartodb/pull/11492/files#diff-f6ad32d8aa3da552f4ab34d6b38853a1R114

xavijam:
Hey, I suggested you to not follow the same solution, just take it as an example :).
Question, if the user doesn't belong to an organization, organization icons fetch fails or it returns an empty array? I guess we don't show the error state in that case right?

matallo:
no prob! my bad for trying too far. If the user doesn't belong to the organization, the array of collections to `apply` fetch will only contain`userAssetCollection`, thus no errors and only the "your uploads" tab will be shown

(solving the conflicts now to move on)

xavijam:
:+1:

javierarce:
I went ahed and included the change for the `findWhere`. Feel free to act on the other two comments :)

xavijam:
Nice one, let's go to acceptance?

javierarce:
### Acceptance

I've detected a (probably small) bug:

1. An organization uploads a bunch of icons.
2. A member of the organization selects an icon to style a marker.
3. The organization removes that particular icon from the ones it offers.

If that happens and the user goes back to the icon modal and to the organization tab, the "set button" will be enabled, but there won't be any icon selected.

There's also the issue of having a map with an icon that no longer exists (only in the cache).

xavijam:
About the icon removed and how the fill component will look like, it is a good question. @CartoDB/design what do you think?
Further to the last "question", it is something already discussed and there is a big and great warning in the organization side. We decided to not do it nothing else for the moment (https://github.com/CartoDB/cartodb/pull/11292#issuecomment-273751733).

saleiva:
@noguerol's call

noguerol:
Can't we just go back to a default bubble whenever a map with missing icon is loaded?

javierarce:
@noguerol I don't think that's an easy thing to do right now, since we'd need to store the previous size somewhere.

javierarce:
### Acceptance 

I've discovered a big issue with this PR. It seems that uploading a custom icon as a regular user, doesn't refresh the list of icons in the modal :(

matallo:
d'oh, very well spotted as I think I only tested this as owner :((( move to development then andI'll take as soon as I can

javierarce:
Oh, don't worry, I'll fix it and ask you for help if I need it.

javierarce:
Back to CR! I've addressed the problems ([1](https://github.com/CartoDB/cartodb/pull/11492#issuecomment-279968180) and [2](https://github.com/CartoDB/cartodb/pull/11492#issuecomment-280019341)) found in Acceptance.

Props to @matallo for the help me debugging one of the problems.

xavijam:
Taking a look @javierarce 💋

noguerol:
OK, so if found one little issue and one thing that concerns me:

a, the issue: on the organizational settings, the copy for deleting an icon is in plural ("delete icons"), while it only deletes one icon, so is confusing as it seems it will erase them all

b. the concern: I understand the dont-break-maps policy when removing an organizational because is very difficult to know how many maps use them, but I can't find a reason to still render a map with an icon that has been deleted when its opened on Builder. On that situation I think we can afford to do it and will be better whatever the use case is.

xavijam:
@noguerol 

a. It is not related with this PR (check description of this pull request), if you consider it is confusing, please open a new ticket.
b. we already discuss that and we took a decission. Weren't you in that thread?. What's your proposal?

noguerol:
Understood. Opening two new tickets, lets move this to production.

xavijam:
💋

matallo:
merging and bumping assets in master

Issue #11474, opened 2017-02-07
[blocked by spec errors] [Google Map Support] Edit feature button no longer works after switching from Google Maps basemap to Leaflet basemap (or viceversa)


alonsogarciapablo:
**STR:**

- Open a map.
- Click on a feature and click the edit button.
- Editing should work like a charm.
- Change the basemap (Leaflet -> Google Maps or viceversa).
- Click on a feature again. You should see the edit button.
- Click the edit button.

**Expected results:**

- You should be able to edit the feature, as you did before.

**Actual results:**

- Edit feature overlay is closed and feature is not editable.

I triaged this and it looks like, for some reason, [this event handler](https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/deep-insights-integration/edit-feature-overlay.js#L160-L184) is no longer invoked after a provider change, so clicking the button has no effect.

xavijam:
Maybe, after https://github.com/CartoDB/cartodb/pull/11448, this is gonna work again, but not sure 100%.

matallo:
did a quick test (and built the dependencies already for this) and when merging both branches I can confirm this is still happening

javierarce:
I'm unassigning myself from this issue, since the CR passed but there are some specs failing [in the PR](https://github.com/CartoDB/cartodb/pull/11494). Once the build turns green, this ticket could be moved to Acceptance.

xavijam:
Taking myself, in order to test it, once more. Then It will be merged into `google-maps` branch.

xavijam:
Merged in `google-maps` branch.

PR #11494, opened 2017-02-08
Fixing problem with edit-feature-overlay button when basemap has changed

xavijam:
Releated: #11474.

There is no new code, basically because it was working ok previously, but when a new baselayer is set, we render the view again in order to decide if events should keep working or not. Good thing is we enable/disable events if the feature is editable or not, so we avoid to check if the proper element has a class, which is a bit unstable. Added a notification when the feature data is requested (It was a TODO so...).

**I made changes over google-maps branch so there is one commit not present in the diff, please take a look [here](https://github.com/CartoDB/cartodb/commit/612be006a1b9596b90cfbcc41ae9e161d4734667). It should be part of this PR**.

**Things to test**
- [ ] When a new base layer (with a different provider) is set, edit a feature button should keep working.
- [ ] Try editing a feature and check a loading notification appears, and when row data is loaded, it should disappear.
- [ ] If user applies a custom query over a layer, edit feature button should be disabled.


xavijam:
Let's change in this PR, I totally agree, it is a bit weird and probably complex.

javierarce:
@alonsogarciapablo the build is showing some errors in the DI specs. Should we do something about them?

alonsogarciapablo:
@xavijam I suggest you merge `google-maps` branch into this branch (I fixed some tests there yesterday). On the other hand, it looks like failures might actually be related to the EditFeatureOverlay thing... not sure if it has something to do with your changes, but:

```
`TypeError: 'undefined' is not a function (evaluating 'this._editFeatureOverlay.render()') in file:///home/ubuntu/workspace/CartoDB-PR-testing/.grunt/cartodb3-specs.js (line 47636)`
```

xavijam:
I did that, but still failing, let me try again.

xavijam:
They keep failing :(

javierarce:
It looks like now there's just one failing spec: 
 
screen shot 2017-02-13 at 09 46 08

Issue #11477, opened 2017-02-07
Icon preview overflow in styling


noguerol:
### Context
Styling with an icon through the modal window makes some of the icons preview to overflow. It seems to happen only to some icons in the "Pin icons" collection.
 

### Steps to Reproduce

1. Style your point layer with icons, opening the modal window
2. Select the five pointed star icon under the "pin icons" collection and apply

### Current Result
Icon preview in Fill options renders overflowed
![image](https://cloud.githubusercontent.com/assets/70603/22705382/9b14e18c-ed6b-11e6-9abd-111b3b5c2975.png)

### Expected result
Icon should rendered entirely on preview
![image](https://cloud.githubusercontent.com/assets/70603/22705419/ae9f2186-ed6b-11e6-8f4e-802d653fa68b.png)

### Browser and version
Chrome 55.0.2883.87
 

### .carto file
[icon_overflow (on 2017-02-07 at 18.27.36).zip](https://github.com/CartoDB/cartodb/files/758613/icon_overflow.on.2017-02-07.at.18.27.36.zip)

nobuti:
Picking this

nobuti:
Ready for CR.

xavijam:
Deployed!

PR #11500, opened 2017-02-09
Set viewBox in order to make it responsive.

nobuti:
This PR implements #11477.


lib/assets/core/javascripts/cartodb3/components/img-loader-view.js
```diff
@@ -32,13 +34,30 @@ module.exports = CoreView.extend({
         $svg = $svg.removeAttr('xmlns:a');
         $svg.attr('class', self._imageClass + ' js-image');
 
+        var bbox = {
```
xavijam: If it is always 18, let's move it to a "const" var.

javierarce:
Acceptance: ✅

Issue #11478, opened 2017-02-07
Activate for all and remove icon-styling feature flag


saleiva:

xavijam:
Icon styling right?

saleiva:
style by icon, sorry

nobuti:
Picking this.

xavijam:
Are we ready for this @saleiva?
Probably we want to block it until we merge all the pending things... or am I wrong?

nobuti:
Adding the `don't deploy` to put it on hold once developed.

nobuti:
Already in production. Closed by #11502.

PR #11502, opened 2017-02-09
FF deleted.

nobuti:
This PR implements #11478.


nobuti:
Fixed.

javitonino:
Tested icon styling in Builder, applying the style to the entire dataset and by value (categories).

I also checked that the organization assets upload form still appears. So 👍, but not deploying until we have green light.

saleiva:
BOOM

Issue #11498, opened 2017-02-09
Styling with icon should change the point size if it's smaller than 20


saleiva:
### Context
Right now, if styling with icons without having a big marker size, you can't even see the change.

### Steps to Reproduce

1. Enter to a points map with default styling
2. Select an icon to style with

### Current Result
You can barely appreciate the changes

### Expected result
I'd expect my markers to get bigger.

matallo:
In case you edit the style "per value", you can have mixed markers and images, and currently I think we only support a common marker width/height for them all (see attached).

Shall we have bigger markers in this case?

![screen shot 2017-02-13 at 11 02 13](https://cloud.githubusercontent.com/assets/36676/22878928/f075cd98-f1db-11e6-8e73-6ef0c97a3ba2.png)

saleiva:
I think so...

saleiva:
I can do this one, assigning to me :)

xavijam:
💃

PR #11530, opened 2017-02-14
Adjusts mininum icon size to 20px

javierarce:
Original issue: https://github.com/CartoDB/cartodb/issues/11498

This PR solves the problem of setting icons to markers with a very small size. The default size is ok for the default marker, but too small for icons (for instance, it's a bit difficult to see if they've changed when the size is small than 10 pixels). To avoid that, it was proposed to use a default base size of 20px just for icons.

The way I've solved the issue is the following:

- If a user sets an icon for the first time and the current size is less than 20 pixels, the size gets enlarged to 20 px.
- If a user had previously selected an icon, the size won't change regardless its current value (this way users can set an icon, give it a size of say 5px and when they pick a different icon, the fill component will respect their previous selection of 5px)

As noted by @matallo, this solution should also work for categories.

### Acceptance

1. Enable the `icon-styling` flag.
2. Create a map with points.
3. Style the points.

- If the size that's set before selecting an image for the first time is less than 20 pixels, the size should get updated to 20px (otherwise, it should respect the size).
- Replacing an icon with another one won't change the size (even if it's less than 20 pixels).

Please, do the acceptance for categories with icons too. The behaviour should be similar, since there's just one global size for all the categories with icons.


NEWS.md
```diff
@@ -45,6 +45,7 @@ Development
   * `sudo apt-get install gdal2.1-static-bin`
 * Style with icons
   * Removed Pin, and Simple icons (#11479)
+  * Sets the default initial size for icons to 20px (#11498)
```
xavijam: ❤️

lib/assets/core/javascripts/cartodb3/components/form-components/editors/fill.js
```diff
@@ -133,9 +135,23 @@ Backbone.Form.editors.Fill = Backbone.Form.editors.Base.extend({
   },
 
   _onInputChanged: function (mdl) {
+    this._adjustImageSize(mdl);
     this.trigger('change', this);
   },
 
+  _adjustImageSize: function (mdl) {
+    var changedImage = mdl.get('image') && mdl.changed.image && !mdl.previous('image');
```
xavijam: We could use [hasChanged](http://backbonejs.org/#Model-hasChanged) instead of accessing to the info through the object.

javierarce: Ups, true.

javierarce:
I'll fix that real quick. Thanks!

xavijam:
Thanks!, moving it to acceptance.

xavijam:
I like the approach and how it works, but not sure if it could be a bit weird for the user.
Because when you select style by value, size of the "markers" changes automatically to 20, no matter if you haven't selected a category. It is true the user didn't select any other size, but not sure, what do you think @noguerol @saleiva?

thing

javierarce:
That behaviour looks like a bug, @xavijam  :/

xavijam:
Ahhhh, ok. Moving to development then 💋 .

javierarce:
Pushed a fix + spec, merged with master + bump front.

xavijam:
Yay, working as described now 👍 .

xavijam:
@noguerol, let's check this tomorrow please, ping me when you arrive 💋 .

saleiva:
Looks good to me. As an enhancement I'd make the default style to revert to the one we had before selecting an icon, but not for now. Let's merge and deploy!

Issue #11426, opened 2017-02-01
UI updates for the Georeference analysis


arianaescobar:
## Context

We'd like to make a few adjustments to the Georeference analysis UI to avoid confusions in our users.

Design issue: https://github.com/CartoDB/design/issues/534

### General

- [x] On the second step: change description to "Select data type"
- [x] ~Change the required notifications color to black (like the normal tooltip) and keep the input in its default style~ -> Bug fixed.

### If "Lon/Lat Columns" type is selected, change the following placeholder texts:

- [x] _Lon/Lat Columns_ to **Longitude and Latitude**
- [x] _Select a latitude_ to **Select a latitude column**
- [x] [In the dropdown] _Search by latitude_ to **Search by column name**
- [x] _Select a longitude_ to **Select a longitude column**
- [x] [In the dropdown] _Search by longitude_ to **Search by column name**

### If "City" type is selected, change the following placeholder texts:

- [x] _City names_ to **Cities**
- [x] _Select a city_column_ to **Select a city column**
- [x] [In the dropdown] _Search by city_column_ to **Search by column name**
- [x] _Select a admin_region_ to **Select an admin. region column**
- [x] [In the A. Regions's dropdown] _Search by admin_region_ to **Or enter region name**
- [x] [In the A. Regions's tooltip] _In which column are your country names stored?_ to **Specify an admin. region or select a column with region names**
- [x] _Select a country_ to **Select country column**
- [x] [In the Country's dropdown] _Search by country_ to **Or enter country name**
- [x] [In the Country's tooltip] _In which column are your country names stored?_ to **Specify a country or select a column with country names**

### If "Country names" type is selected, change the following placeholder texts:

- [x] _Country names_ to **Countries**
- [x] _Select a country_column_ to **Select a country column**
- [x] [In the dropdown] _Search by country_column_ to **Search by column name**

### If "Admin Region" type is selected, change the following placeholder texts:

- [x] _Select a admin_region_column_ to **Select a admin. region column**
- [x] [In the dropdown] _Search by admin_region_column_ to **Search by column name**
- [x] _Select a country_ to **Select country column**
- [x] [In the Country's dropdown] _Search by column_ to **Or enter country name**
- [x] [In the Country's tooltip] _In which column are your country names stored?_ to **Specify a country or select a column with country names**

### If "Postal Codes" type is selected, change the following placeholder texts:

- [x] _Select a postal_code_column_ to **Select a postal code column**
[...]

PR #11548, opened 2017-02-15
11426 UI fixes for georeference

ivanmalagon:
Implements #11426 

#### Acceptance

1. Check that every copy change in https://github.com/CartoDB/cartodb/issues/11426 is present.

2. Instead of changing the look & feel of the required notification, we've fixed the problem of the validation getting triggered when changing georeference type. (check GIF below)

3. Check also that `suggest` and `merge from` dropdowns work as always. `suggest` dropdowns appears in _geometry edition_ and `merge from` in _predict trends_ analysis. Just check that every dropdown in those places work. These two components share a template that has been modified in this PR, hence this check.

georeference


matallo:
When doing CR I like to setup the branch locally so I can better follow/understand the code. I hope you don't mind, but I just found the placeholder is missing in the template, for the first Georeference analysis I tried.

screen shot 2017-02-17 at 10 47 27

matallo:
I don't usually do consecutive CR and acceptance but ...

I found the text is not right in **Admin. Regions**

- "Select a admin. region column", the text says "Select a region column"

screen shot 2017-02-20 at 09 27 16

matallo:
Actually I'd better say "Select an admin. region column"

matallo:
Some minor literal details that maybe have already been discussed out of the PR, feel free to move on to deploy but state those changes somewhere in the issue or the PR

nobuti:
Works like a charm ✨

nobuti:
Waiting green and 🚀

nobuti:
Frontend passing, vamos.

Issue #11451, opened 2017-02-03
Long column name gets render outside the box in the popup component


saleiva:
### Steps to Reproduce

1. Style a layer by value
2. Select a very long column name
3. 

### Current Result
![screen shot 2017-02-03 at 16 35 47](https://cloud.githubusercontent.com/assets/132149/22597204/f56670fa-ea2e-11e6-903b-74c121ffeaa8.png)

### Expected result
We should add a ellipsis on this text when being so long.

xavijam:
Added to UI fixes, 👍  or 👎  ?

jorgesancha:
IMO, I think this is Nice To Have™️, but not a blocker for Free users migration. What do you think @noguerol ?

saleiva:
It is not

noguerol:
let's push that migration asap

javierarce:
This appears in acceptance but the PR is already merged? Should we move it to the deploy column?

nobuti:
It was @piensaenpixel doing the cowboy, I think so.

javierarce:
Haha… where's the kanban police when you need them? :_

saleiva:
:sanchismo:

PR #11563, opened 2017-02-17
Ellipsis.

nobuti:
This PR implements #11451.


matallo:
🤔 I think just one `u-ellipsis` in the element to cut is enough

nobuti:
we still need `overflow: hidden` in the container, do we have a helper to do that? it's the easy way I found to achieve it without creating a new selector.

matallo:
GO then! If this is a recurrent case I'd consider adding an utility for the overflow, or if these two are usually accompanied to have the ellipsis parent, and ellipsis child

nobuti:
Getting @piensaenpixel into the loop, I feel we need his expertise.

piensaenpixel:
+1 to @matallo's comment, i think we can merge this one and I can prepare this helper in cartoAssets.

nobuti:
Are you guys saying I'm too lazy?!?

piensaenpixel:
hahahahaah yes. Nah merge this fix and i'll take care.

Issue #11138, opened 2016-12-27
Remove second back button from fill by category


MariaCheca:
## Current status

We have two back actions in the fill by category view to edit the color: the first one to the list of categories, the second one to the list of columns.

<img width="316" alt="screen shot 2016-12-27 at 16 19 43" src="https://cloud.githubusercontent.com/assets/2141690/21502470/4e9ec1c4-cc50-11e6-8b47-4593e9da5b51.png">

## Steps to reproduce

1. In Builder enter a layer with point geometries and go to the style tab.
2. Style with no aggregation and in the 'Style' step, introduce a fill by value.
3. Select a string type column to filter by categories, and once the list of categories appear select one of them.
4. You should see something similar to the image above. 

## Expected result

There should be only one back action to the list of categories.

Updated by @javierarce: as mentioned by @saleiva, we should hide the complete title row.

cc @CartoDB/frontend

javierarce:
Back to CR!

PR #11564, opened 2017-02-17
Hides header back arrow  

javierarce:
Original issue: #11138 

This ticket removes an arrow from the fill component form when styling by value:

screen shot 2017-02-17 at 09 54 15

With this change we'll prevent users to inadvertently go to the column list instead of going back to the list of categories.


saleiva:
The idea was hiding the complete "row" so not only the arrow, but also the name of the column

javierarce:
@saleiva maybe I'm missing something, but the original ticket says: "There should be only one back action to the list of categories." It's a bit ambiguous to me.

javierarce:
@matallo I think in this case is not worth it.

saleiva:
agree on that, I checked with the designs and yeah, the whole row should be removed. (sorry!)

javierarce:
No problem!

Issue #10654, opened 2016-11-08
Switching between qualitative and quantitative attributes uses category ramps not sequential


makella:
### Context
If you symbolize by a qualitative attribute you get a category ramp. If you want to switch to a quantitative attribute, you get the option for the classification method and class breaks, but only the CartoColor category ramps (that are not available anywhere else in Builder) show up.

### Steps to Reproduce
1. Style by `complaint_` (qualitative)
2. Switch to style by `count` (quantitative) 
3. Change the number of classes to `4`

### Current Result
Basically, if you switch from a qualitative attribute to a quantitative one, we ramp the category palette and only give you the option to choose another category palette instead of a sequential one:

![screen shot 2016-11-08 at 12 05 53 pm](https://cloud.githubusercontent.com/assets/1566273/20113001/ca457dd4-a5ab-11e6-9fe3-c1570526c194.png)

The way to get to the sequential palettes is to change the classification method or the number of classes, but we still hang on to the original category ramp as the default:

![screen shot 2016-11-08 at 12 07 00 pm](https://cloud.githubusercontent.com/assets/1566273/20113054/fd86bb86-a5ab-11e6-8ae4-9f7705016036.png)
 

### Expected result
When you switch from a qualitative to quantitative attribute, we should automatically restyle the map with a sequential (not categorical) color ramp.

### Browser and version
Chrome Version 54.0.2840.71 (64-bit)

### .carto file
[Symbolizing categories and quantities (on 2016-11-08 at 19.13.52).zip](https://github.com/CartoDB/cartodb/files/579159/Symbolizing.categories.and.quantities.on.2016-11-08.at.19.13.52.zip)

### Additional info
I have reported about the category ramps not being available in the past, I'm wondering if we are reading them in, why they aren't being assigned to the right attribute type.

https://github.com/CartoDB/cartodb/issues/9002

The only way to see any of the other category ramps is to do this workflow which is completely wrong.

makella:
@xavijam @saleiva just checking in on this ticket this is a pretty important workflow that we aren't supporting. 

with this bug we are also enabling people to make maps that are totally wrong. for example, by allowing a user to create a choropleth map using a qualitative scheme.

if there is any additional information i can provide, just let me know.

related also is this ticket: https://github.com/CartoDB/cartodb/issues/9002
thanks!

saleiva:
Moved to the TODO column

nobuti:
Picking this.

saleiva:
[...]

PR #11570, opened 2017-02-17
Revamp ramps when attribute_types changes.

nobuti:
This PR implements #10654.

### How to test

- [ ] Change between string and number attributes. The ramps should be revamped and the selected one should be the first.
- [ ] On change, no custom ramp should be present.


xavijam:
Acceptance done, working like a charm changing from category ramp and then to diverging ramp (and viceversa). Tested in local because it is not necessary to test it in staging, it is located.

Issue #11586, opened 2017-02-20
Marking "Do not show me this again" on the Onboarding has to affect all tabs


arianaescobar:
By watching our users react with the new Layer Onboarding, we're seeing a lot of frustration with the onboarding on every tab. 

Let's make it so that when the user marks the "Don't show me this again" checkbox, it affects the onboarding of every tab. 

PR: https://github.com/CartoDB/cartodb/pull/11596

PR #11596, opened 2017-02-21
11586 global do not show [Don't deploy until script]

ivanmalagon:
Implements #11586

I've reduced the three different keys that we had for every tab onboarding to only one `layer-onboarding`. It makes no sense to set three different keys for the same behaviour.

The problem is that the current users that chose to skip the onboarding will see the onboarding again (since this is a new notification key).

We have to run a simple script to set the new key `layer-onboarding` to every user that has currently one of these `layer-data-onboarding`, `layer-analysis-onboarding`, `layer-style-onboarding`. Only 315 users will be affected.

Summoning backenders here to add this script or, at least, to get them in the same page regarding this. Already discussed with @javitonino @CartoDB/builder-backend


javitonino:
Please, assign me when you move to deploy and I'll take care of running the script and deploying in unison :)

javierarce:
### Acceptance

I've found a small problem: in the onboarding for analyses, if I select the checkbox and click on the button to add an analysis, I'm correctly redirected to the add analysis modal… but if I close it and navigate to another tab (style, for example) the onboarding gets open again.

analysis

And I have a doubt: if I check the box of one of the 3 new onboardings (data, style, or analysis) and then reload the page, the main onboarding window still appears. Is that an OK behaviour?

javierarce:
I'm moving it back to development.

ivanmalagon:
@javierarce 

Answering the second question. This ticket refers only to the three different onboardings on layer edition. The principal one is separated from these ones and should remain as is, according to the ticket description.

ivanmalagon:
🇯🇲

ivanmalagon:
Assigned @javitonino to deploy and run the script.

Issue #11316, opened 2017-01-20
Invite to georeference proactively when detecting non-georeferenced data.


carlostallon:
### Context
When you enter non georeferenced data, we are adding a new enhancement to avoid a user could end up not seeing anything on the map and not knowing he must georeference the data, which
would end up in a blocker

### Step by step mockups
![01 initial state](https://cloud.githubusercontent.com/assets/1712650/22145563/3671b66e-df02-11e6-9090-5a0c5ae22c80.png)
![02 added new layer - take action](https://cloud.githubusercontent.com/assets/1712650/22145567/36761894-df02-11e6-9e42-9234ae5d0109.png)
![03 new layer added](https://cloud.githubusercontent.com/assets/1712650/22145566/36729070-df02-11e6-8672-8fcf64c355e7.png)
![04 any georeferenced layer](https://cloud.githubusercontent.com/assets/1712650/22145562/36717c30-df02-11e6-8933-72cd7479664d.png)
![05 not georeferenced layer hover](https://cloud.githubusercontent.com/assets/1712650/22145565/36726afa-df02-11e6-86d0-5a78ef8c83f1.png)
![06 layer style](https://cloud.githubusercontent.com/assets/1712650/22145564/36720704-df02-11e6-84e7-da6bb5fa5811.png)
![07 layer legend](https://cloud.githubusercontent.com/assets/1712650/22145569/36877990-df02-11e6-90fe-554448447ad6.png)
![08 layer pop-up](https://cloud.githubusercontent.com/assets/1712650/22145568/3685c974-df02-11e6-98e5-c4a17706df81.png)

### Design ready
Sketch archive here: `Dropbox / CartoDB / Projects / 01-webapp / 03-Map-editor / 02-mockups / 3.x / _issues / 465 proactive georeference `

### Related design issue
https://github.com/CartoDB/design/issues/465

xavijam:
Oido cocina!

saleiva:
While fixing this, we should also fix another issue which is that when not having georeferenced data, we shouldn't let people to add widgets from that layer.

xavijam:
But it is possible that the user wants to connect a widget to that "node", where at the same time, it is connected to another node by an analysis. Make sense?

saleiva:
I don't think widgets work without georef data...

xavijam:
Arg, you are right, I was blind by the data stats.

[...]

PR #11604, opened 2017-02-21
onboardings: georeference

matallo:
re: https://github.com/CartoDB/cartodb/issues/11316

---

**acceptance**

take into account latests comments https://github.com/CartoDB/cartodb/pull/11604#issuecomment-286700157 for acceptance

- a georeference button should appear for non-georeferenced layers, also a georeference infobox must appear in data, and analyses unless the layer is hidden
- test thoroughly the table view

try these, and the trickier cases you might think of


matallo:
first draft, this needs some refactor and tests, but looking good so far

- when builder onboarding is enabled AND georeference onboarding is enabled both will appear consequtively
- when just georeference is enabled then it will appear
- if none is enabled none will appear
- if style onboarding is enabled it will suggest to style the map and georeference FROM the style panel

some pending stuff:

- show the onboarding after import layer from map view
- show the georeference style panel when adding an analysis (i.e. sample of 100%)

(I'm adding these instructions for acceptance in the PR description)

matallo:
moved to a single use of `queryRowsCollection` with a `statusModel`, to check if it should be fetched, the problem now is the requests collide, so we'd better move the table view to the new `queryRowsCollection`

I'm moving on having the `tableView` `initialize` commented, but this is something to eventually tackle.

@xavijam has more info

nobuti:
Before anything, a 💯  job. I agree the constructor is a bit overkilled, these tests belong to requiredopts, and should be considered "duplicated". Nevertheless tests are always welcome :)

lib/assets/core/javascripts/cartodb3/editor/style/style-content-view.js
```diff
@@ -119,25 +109,37 @@ module.exports = CoreView.extend({
...
-    this.add_related_model(this._styleModel);
-
-    this._querySchemaModel.bind('change:status', this._onStatusChanged, this);
-    this._queryGeometryModel.bind('change:status', this._onStatusChanged, this);
-    this.add_related_model(this._querySchemaModel);
-    this.add_related_model(this._queryGeometryModel);
+    this.listenTo(this._styleModel, 'undo redo', this.render);
+    this.listenTo(this._styleModel, 'change', this._onStyleChanged);
+
+    this.listenTo(this._querySchemaModel, 'change:status', _.debounce(this._onStatusChanged.bind(this), 100));
```
xavijam: Why do we need a debounce here?

lib/assets/core/javascripts/cartodb3/editor/style/style-content-view.js
```diff
@@ -154,8 +156,9 @@ module.exports = CoreView.extend({
   },
 
   _onStatusChanged: function () {
-    if (this._querySchemaModel.get('status') === 'fetched' &&
-        this._queryGeometryModel.get('status') === 'fetched') {
+    if (this._querySchemaModel.isFetched() &&
```
xavijam: :+1:

lib/assets/core/javascripts/cartodb3/editor/style/style-content-view.js
```diff
@@ -94,12 +76,20 @@ module.exports = CoreView.extend({
         if (this._queryGeometryModel.get('simple_geom') === 'point') {
           this._renderCarousel();
         }
+
         this._renderForm();
         this._renderOverlay();
       } else {
-        this.$el.html(styleFormPlaceholderTemplate());
+        if (!this._layerDefinitionModel.hasAnalyses() &&
+            !this._layerDefinitionModel.isCustomQueryApplied() &&
```
xavijam: Remember, `isCustomQueryApplied` should be checked from the node model (analysis definition model).

xavijam: Already checked that `layerDefinitionModel` is using that method :+1:, it is like an alias. I'd like to have the direct answer instead of the alias but :+1:.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/legend/georeference-placeholder.tpl
```diff
@@ -0,0 +1,11 @@
+<h2 class="CDB-Text CDB-Size-huge is-light u-secondaryTextColor"><%- _t('editor.legend.no-geometry-data') %> <%- _t('editor.style.style-form.properties.manually-add') %></h2>
```
xavijam: Looks like we are using the same text in several places, could we move it to a "generic" position? -> Further to `editor.style.style-form.properties.manually-add` for example.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindows-view.js
```diff
@@ -78,19 +98,29 @@ module.exports = CoreView.extend({
...
-
     if (this._layerInfowindowModel && !this._hasColumns()) {
-      placeholder = _t('editor.layers.infowindow.placeholder-columns-text');
+      this.$el.append(popupPlaceholderTemplate({
+        placeholder: _t('editor.layers.infowindow.placeholder-columns-text')
+      }));
     } else if (!this._queryGeometryModel.hasValue()) {
-      placeholder = _t('editor.layers.infowindow.placeholder-geometry');
+      if (!this._layerDefinitionModel.hasAnalyses() &&
+          !this._layerDefinitionModel.isCustomQueryApplied() &&
```
xavijam: Already dixit 💋

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindows-view.js
```diff
@@ -51,23 +59,35 @@ module.exports = CoreView.extend({
...
+      edition: false
+    });
+  },
+
   _initBinds: function () {
-    this._queryGeometryModel.bind('change:status', this.render, this);
-    this.add_related_model(this._queryGeometryModel);
-    this._editorModel.on('change:edition', this._changeStyle, this);
-    this.add_related_model(this._editorModel);
+    this.listenTo(this._querySchemaModel, 'change:status', _.debounce(this._onStatusChanged.bind(this), 100));
```
xavijam: Why do we need a debounce here?

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-service.js
```diff
@@ -33,6 +34,21 @@ module.exports = (function () {
       this._stackLayoutView = stackLayoutView;
     },
 
+    addGeoreferenceAnalysis: function () {
```
xavijam: Ha!, that's a good one, I feel you have found the right place for this. 👍

lib/assets/core/javascripts/cartodb3/editor/layers/basemap-content-views/basemap-select-view.js
```diff
@@ -40,19 +40,14 @@ module.exports = CoreView.extend({
...
-        break;
       case 'mapbox':
-        this._renderCustomMosaic(category);
-        break;
       case 'wms':
         this._renderCustomMosaic(category);
         break;
       default:
         this._renderMosaic();
+        break;
```
xavijam: break in default case doesn't matter, so no need. Why have we done this change?

xavijam: And the previous ones :D.

lib/assets/core/javascripts/cartodb3/editor/editor-view.js
```diff
@@ -175,18 +211,115 @@ module.exports = CoreView.extend({
...
+    this.listenTo(this._editorModel, 'change:edition', this._changeStyle, this);
+    this.listenTo(this._layerDefinitionsCollection, 'reset remove add', function () {
+      this._nonGeoreferencedLayer = null;
+      this._infoboxModel.unset('state');
+
+      this._findNonGeoreferencedLayer();
+
+      this._updateAddButtonState();
+    }, this);
+    this.listenTo(this._layerDefinitionsCollection, 'add', function (model) {
```
xavijam: I'd move all this code to a function, definitely.

lib/assets/core/javascripts/cartodb3/editor/editor-view.js
```diff
@@ -175,18 +211,115 @@ module.exports = CoreView.extend({
...
+
+        if (queryGeometryModel.shouldFetch()) {
+          queryGeometryModel.fetch({
+            success: this._findNonGeoreferencedLayer.bind(this)
+          });
+        }
+
+        if (querySchemaModel.shouldFetch()) {
+          querySchemaModel.fetch({
+            success: _.debounce(this._findNonGeoreferencedLayer.bind(this), 100)
```
xavijam: Why do we need a debounce here?

lib/assets/core/javascripts/cartodb3/data/layer-definition-model.js
```diff
@@ -242,16 +242,24 @@ module.exports = Backbone.Model.extend({
     return this.getNumberOfAnalyses() > 0;
   },
 
+  isCustomQueryApplied: function () {
```
xavijam: As I've said, I'd try to not make an alias for that. But only if it is possible.

lib/assets/core/javascripts/cartodb3/editor/style/style-content-view.js
matallo: honestly, I don't have an answer, when rearranging the `query-rows-collection` I noticed this debounce in the query schema model change, so I populated this whenever I'm listening to changes to that model https://github.com/CartoDB/cartodb/blob/8bebfbd03887a854f3f35126193d8b9444a52bb4/lib/assets/core/javascripts/cartodb3/components/table/table-view.js#L74

we can try without the debounce

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/legend/georeference-placeholder.tpl
matallo: 👍

lib/assets/core/javascripts/cartodb3/editor/layers/basemap-content-views/basemap-select-view.js
matallo: following @ivanmalagon coding style ¯\_(ツ)_/¯

lib/assets/core/javascripts/cartodb3/editor/editor-view.js
matallo: I tried removing it but we definitely need it for the infobox to succesfully appear, we can get rid of the rest, tho

xavijam:
Let's not forget ⚠️  button @matallo  💋 .

xavijam:
So, pending things before passing to acceptance:

- [ ] Add ⚠️  icon when a layer is not georeferenced or it doesn't have any data (cc @noguerol @saleiva).
- [ ] Fix broken tests after new changes (ongoing).
- [ ] Fix small things with the georeferencing onboarding template.
  - [ ] It appears in the whole page, but it should appear only on top of the map.
  - [ ] Title should be bold.
- [ ] Check that georeference onboarding appears after adding a new row (moved note from surface @matallo 😛 ).

matallo:
missing tests added, this can be tested in ded04 @noguerol

noguerol:
So, some things I found so far:

1) Infowindow doesn't make a direct reference to layer name

image

2) The style walkthrough should not pop up if layer is not georeferenced

image

3) When two non-georeferenced layers are added, only the top one in the list seems to trigger the walkthrough. The infowindow refers to "layer", in singular. We should list the non-georeferenced layer names both on a single walkthrough and infowindow notification.

<3

matallo:
> Infowindow doesn't make a direct reference to layer name

there seems to be something wrong, let me take a look at this, and come back to you for further info if I don't spot this

> The style walkthrough should not pop up if layer is not georeferenced

in my opinion, out of the scope of this PR, also I suggested adding an "add geometry onboarding" instead

> When two non-georeferenced layers are added, only the top one in the list seems to trigger the walkthrough.

I suppose you refer to infobox, "only the top one in the list seems to trigger the walkthrough" comes directly from the requirements and designs, once we work out the layer name issue I'd stick to it, but I would really reconsider listing all the layers

xavijam:
Hey @noguerol, if you don't mind, let's call that box as infobox 💋 .

1. Agree, let's make reference, no prob.
2. Uhm... I agree with @noguerol, @matallo. We should review that.
3. Yes, that's a design requirement as @matallo has pointed.

matallo:
🙇 looking into it

matallo:
1. and 2. addressed (and deployed to ded04 again for acceptance)

noguerol:
Ok, so after reviewing this with @CartoDB/design seems some behavior criteria was not clear enough, sorry for that.

3. Walkthroughs and infoboxes were designed for coming up only for the first non-georeferenced layer, as is the one preventing the user for visualizing data as a map. Other non-geo layers could be added afterwards for analysis purposes and there is no need to warn the user repeatedly, except for the layer icon and the buttons inside. In other words: **Walkthrough and infobox should only pop-up for the first non georeferenced layer added in the map, and when dismissed they don't appear ever again. Layer warning icon and georeference buttons inside the layer still appear on every non-geo layer.**

4. Infobox seems to not pop-up when opening a map with a non georeferenced layer that has not been dismissed before. They appear tho if you get inside the layer and back to the layers view.

5. Currently you cannot permanently dismiss infoboxes, as you close them and they pop-up again when returning to the layer view. **User should be able to dismiss infoboxes permanently, same way that with the notification**

6. Walkthroughs and infoboxes could only be dismissed per map, not as a general rule. Currently if you dismiss one of this walkthroughs in one map and create another map with a non-geo layer, the walkthrough will not appear. **In other words, this onboarding is dismissable per map and not in a general basis as the rest of the onboardings.**

noguerol:
So, forget everything about my last comment. In order to support every use case and in the sake of simplicity, this is what we're going to do:

1) **Infoboxes will not pop-up in the layers view but inside the layer pane.** Doing this we give support to analysis with non geo layers without bothering the users.

2) **We get rid of the semitransparent onboarding layer**. As radical as this may sound, showing a warning as an onboarding layer was both redundant with the infobox and counterproductive, as conflicts with the other parts of the onboarding in timing and behavior.

Thank you for your effort, buds!

xavijam:
It is :+1: @noguerol, but about **1**, have you considered that there are other infoboxes inside the layer panels. For example, the visibility one. 

screen shot 2017-03-10 at 17 18 56

What should be display before? Should we display both?

noguerol:
It doesn't make any point to show the invisible layer notification with a non-geo layer, right?

xavijam:
Agree, but you can hide it in any case 💃 . So, layer without geometry infobox has preference over rest 👍 .

noguerol:
What im saying is we should never show the invisible layer warning in non-geo layers? I better open a ticket for this?

xavijam:
But the option is there, that's another thing. But my question is about more related with the possibility to have several infoboxes in the same context. Is it possible show more than one?

noguerol:
Not imo. We'll solve that if we meet a use case that needs it.

matallo:
reverted the points above as to the new specifications, but I think this needs some proper thought:

- **warning in style, popups, and legends is redundant**: two georeference actions and almost the same content.
- **warning in analysis makes no sense at all**, if I click on the georeference action elsewhere, as it detects it can be georeferenced the warnign appears even if I'm at the georeference step

find the captures below

screen shot 2017-03-13 at 17 49 15
screen shot 2017-03-13 at 17 53 18
screen shot 2017-03-13 at 17 53 27
screen shot 2017-03-13 at 18 24 00
screen shot 2017-03-13 at 18 28 51

noguerol:
You're completely right, my friend. While the infobox is redundant in the geo tabs: style, pop-up and legends, is **still needed in the non-necessarily-geo ones: data and analysis**. On the analysis one I would make the infobox disappear when adding an analysis: if its the geocoder, the warning is no longer needed and if its a non-geo one, its most probable that the layer is about to be used for different purposes.

matallo:
working as expected, and fixed tests, stumbled upon a bug that is currently occuring https://github.com/CartoDB/cartodb/issues/11759

matallo:
sorry to raise this again, but I really _really_ think we should consider hidden infobox taking preference over georeference infobox

(I know it makes no sense to show hidden layer notification with a non-geo layer, but let me explain my point)

I tried showing the georeference infobox, indeed. And then when trying edge cases you still can make the layer hidden through the context menu, so you end up with a non-georeference, hidden layer.

If you go to the analyses tab, and click the georeference action you end up in the next screen (see attached).

e8ce2f39-946f-4e5c-9a13-56d6fa557206

functionally it can be correct (it will work), but in terms of form I don't see it is correct, because we are showing the infobox below the actions, as opposed to other panels (see attached) where the infobox is _on top_ of the actions.

124b47c4-ec06-40ec-9d6a-1b79287c55d6

this is due to how the analyses panel is coded in the `panelswithoptionsview`, with the `controlview` inside the `contentview` https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-view.js#L77-L97 as opposed to the `controlview` outside the `contentview` https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/data/data-view.js#L344-L366

furthermore this brought some further points to take into account, that infobox which lives in the analyses actions menu has its own infobox that deals with quotas, so we can end up with quotas infobox properly on top of the analyses actions, and a hidden layer infobox below the actions, weird.

thus my proposal: in any case, show hidden layer infobox and discard it to be able to unlock further steps. This is working in ded01.

let's get together and discuss this for 5 minutes before making any further decisions

xavijam:
@noguerol ^^^

noguerol:
+1 @matallo

noguerol:
Two things: 

- we should show the georeference message and button in the geo-tabs (style, popup and legends) regardless there is an analysis node or not in the chain (not including georreference, of course)
- Style onboarding pop-ups even when a non-geo analysis is chained in a non-geo layer and you go to the geo-tabs

xavijam:
Easy change 💋

Issue #11598, opened 2017-02-21
Dropdowns are missing their gradients and scroll bars


arianaescobar:
### Context

All the dropdowns are missing their gradients and scroll bars so it's not clear when there's more content on them. 

### Current Result

Looks like there's nothing else there,

<img width="460" alt="screen shot 2017-02-21 at 16 41 47" src="https://cloud.githubusercontent.com/assets/16247628/23172122/0310d7d4-f855-11e6-9482-b15c9e3bcbe7.png">

but...

<img width="445" alt="screen shot 2017-02-21 at 16 41 56" src="https://cloud.githubusercontent.com/assets/16247628/23172130/08314492-f855-11e6-8455-da35a73a2355.png">

 (Notice that the scroll bar is only appearing when you scroll...)

### Expected result

Dropdowns with more content than what they can show, need a visible scroll bar at all times, as well as a gradient on the bottom part.

xavijam:
I opened mine first 😛  -> https://github.com/CartoDB/cartodb/issues/11597

xavijam:
Hey, what about if we follow the same behaviour we already have in the infowindows? No gradient, but scroll bar visible always.

xavijam:
If we start adding gradients also in dropdowns, we could have performance problems due to we should listen all scroll movements in all dropdowns.

javierarce:
Where's the PR for this ticket?

javierarce:
Found it: https://github.com/CartoDB/cartodb/pull/11609

xavijam:
Ouch sorry, already referenced :(.

PR #11609, opened 2017-02-22
Fixed problems with custom scroll in generic dropdown

xavijam:
Related: https://github.com/CartoDB/cartodb/issues/11598.

So, discovered that until the element is not rendered in the DOM, scroll is not calculated correctly and it was not displayed. That was the first problem, solved with a 0 timeout. Then I've realized there was several CSS in the `custom-list` class that were not needed, like the min-height, because the perfect-scrollbar already provide a way to set the min length of the scroll.
Also, `_checkScroll` function was not used and it is not needed, because scroll is only displayed when it is needed.
And finally, bonus thing, operators dropdown was not being displayed correctly, fixed too.

screen shot 2017-02-22 at 11 55 30

**THINGS TO CHECK:**
- [x] Check dropdowns display custom scrolls correctly. For example, map options, table view menu options,... (there are tons in the application).
- [x] Apply an aggregation style (squares, hexabins,...), then change the aggregation operation and check that the element is rendered correctly with a different operation (SUM for example).

cc @nobuti @piensaenpixel

### DIFF
--- a/app/assets/stylesheets/editor-3/custom-list.css.scss
+++ b/app/assets/stylesheets/editor-3/custom-list.css.scss
@@ -112,7 +112,7 @@
   display: block;
   background: none;
 }
-.CustomList-list.ps-container > .ps-scrollbar-y
+.CustomList-list.ps-container > .ps-scrollbar-y,
 .CustomList-list.ps-container > .ps-scrollbar-y-rail {
   width: 4px;
 }
--- a/app/assets/stylesheets/editor-3/custom-list.css.scss
+++ b/app/assets/stylesheets/editor-3/custom-list.css.scss
@@ -33,7 +33,7 @@
 .CustomList-list {
   position: relative;
   width: 100%;
-  max-height: 167px;
+  max-height: 168px;
   overflow: hidden;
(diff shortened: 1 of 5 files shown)


javierarce:
### Acceptance: :1st_place_medal: 👌

javierarce:
screen shot 2017-02-22 at 17 18 20

Mmm, there seems to be a failing spec :/ Maybe after merging with master?

xavijam:
Yep, that comes from master :(. Also there are tests failing from the importer (backend) :(x2.

rafatower:
The shp test has been identified as one of the flaky ones when ran on parallel (see https://github.com/CartoDB/cartodb/issues/6754)

I triggered the [BE tests build again](https://ci.cartodb.net/job/CartoDB-PR-testing/8629/)

javierarce:
Front and backend tests are ok now, but I still see the error I mentioned before in the Travis CI build. I don't know what to make of that :|

xavijam:
@javierarce, tests passed finally, probably a cache issue in Travis, so 👍 .

Issue #10921, opened 2016-12-01
Consolidate close actions


urbanphes:
### Context

Related to https://github.com/CartoDB/design/issues/372
(This feature has been worked on CartoDB/design)

nobuti:
Picking this

nobuti:
@urbanphes how about the quota infoboxes?

![3f5888f4-e8a1-11e6-83a0-1f110c7b1ea8](https://cloud.githubusercontent.com/assets/1366843/23157466/ad901a2e-f81b-11e6-83cb-7afa5f7b00b9.png)

nobuti:
also in this infobox:

![screen shot 2017-02-22 at 06 59 07](https://cloud.githubusercontent.com/assets/1366843/23199086/78d37054-f8cc-11e6-8c77-c9f23ea9696b.png)

should we add the close button? does it replace it the confirm?

xavijam:
@CartoDB/design ^^^??

urbanphes:
@nobuti the infobox now have the action to close/cancel on an icon "x" on the right top corner.

The normal inboxes are like this:
![9 infoboxes](https://cloud.githubusercontent.com/assets/14546701/23204766/1f435768-f8e8-11e6-981f-af4849ef8a14.png)

The quota infoboxes are like this:
![11 quota warnings](https://cloud.githubusercontent.com/assets/14546701/23204796/328a91d8-f8e8-11e6-86a6-4e7c5b5a0b3e.png)

nobuti:
@urbanphes so the cartocss one, should we keep two actions plus the close button?

piensaenpixel:
@nobuti continue is replace by close they have the same action

xavijam:
In production!

PR #11615, opened 2017-02-22
Consolidate close

nobuti:
This PR implements #10921.

In order to test it, check all infoboxes play nice with the issue's description.


xavijam:
Tested with show/hide layer infobox, cartocss applied (no changes there) and quota info. 👍 .

Issue #11592, opened 2017-02-21
Icon legend UI options and map icon legend are not syncronized


ramiroaznar:
### Steps to Reproduce

1. Create a map with a points layer
2. Style points using icons
3. Click on legend
4. Activate legend

### Current Result

* Map legend is not showing the icon (but the legend UI option does):

![legend](https://cloud.githubusercontent.com/assets/5215798/23161898/332129ac-f82c-11e6-8edc-6fe0976c40b3.png)

* Changing the icon solve the issue.
* Clicking on another layer tab and going back to the legend options. The legend option does not show the icon anymore:

![legend2](https://cloud.githubusercontent.com/assets/5215798/23161996/95ec7550-f82c-11e6-9e2e-c5e8ad74816e.png)

### Expected result

* Legend options UI and map legend should be syncronized.

### Browser and version

* Google Chrome v56.

### .carto file

[Green Area Acces in Madrid (on 2017-02-21 at 10.47.39).carto.zip](https://github.com/CartoDB/cartodb/files/789860/Green.Area.Acces.in.Madrid.on.2017-02-21.at.10.47.39.carto.zip)

nobuti:
Closed by #11622.

PR #11622, opened 2017-02-23
11592 legend icons

ivanmalagon:
Implements #11592.

**What was happening?**

The backend expects the icon to come in an `icon` property, but the fill component and the style definition has this property as `image`. So, depending on the origin of data to create the form, we are one or the other.

icon custom legend


javierarce:
### Acceptance: :1st_place_medal:

PR #11633, opened 2017-02-24
Fixes basemap switching right after a new map from dataset is created

### LINKED ISSUE #11625: Basemap does not update
### Steps to Reproduce

1. Create a map from a dataset
2. Change basemap from Positron to Dark Matter

### Current Result

* The new basemap is not updated, so it shows the former.

basemap

### Expected result

* The new basemap should show after changing from one to another.

### Browser and version

* Google Chrome v56.

### .carto file

[Untitled Map 7 (on 2017-02-23 at 14.59.45).carto.zip](https://github.com/CartoDB/cartodb/files/796788/Untitled.Map.7.on.2017-02-23.at.14.59.45.carto.zip)

### Additional info

* The basemap is changed after refreshing the browser. And after that, if you change again the basemap, it works correctly.

alonsogarciapablo:
Fixes #11625.

The problem was that we're using `category` attribute to determine the provider of the base layer, but `category` attribute is not always present. And, is not present when creating a map from an existing dataset, so updates to cartodb.js where being done silently and so, the base layer was not being changed.

I changed things so that we use the `type` attribute instead of the `category`.


nobuti:
Works like a charm ✨

xavijam:
Special K!

nobuti:
Frontend tests passing 🚀

Issue #11174, opened 2017-01-03
Time Series Widget Shows Wrong Date Range - Enterprise Client


michellechandra:
### Context
This is an issue for an Enterprise client that I have replicated locally in my team account. The time series widget is showing the wrong date range.

**User:** chandra
**Map:** [chicago_crimes_1](https://team.carto.com/u/chandra/builder/85156658-d1e2-11e6-9621-0ecd1babdde5/embed)
**Dataset:** [chicago_crimes_1](https://team.carto.com/u/chandra/dataset/chandra.chicago_crimes_1)

### Steps to Reproduce
Using a dataset with date values, create a map.

Select the "animated" aggregation option. 

Choose the date column for the aggregation.
![screen shot 2017-01-03 at 10 46 32](https://cloud.githubusercontent.com/assets/5423288/21618860/eb25a5de-d1a1-11e6-8d66-72c4fd3be535.png)

Select the beginning of the time series widget onward using the manual selector. The wrong dates will be shown. The time series widget says the dataset starts December 1, 2011. 
![screen shot 2017-01-03 at 10 30 14](https://cloud.githubusercontent.com/assets/5423288/21618813/a4e0b852-d1a1-11e6-8888-169c97517600.png)

However, the earliest date values in the dataset are December 31, 2012. 
![screen shot 2017-01-03 at 10 31 38](https://cloud.githubusercontent.com/assets/5423288/21618834/bb5273e6-d1a1-11e6-9470-0ff92eb251c9.png)

### Current Result
When manually selecting to show a certain range of data, the wrong date value is shown in the time series widget if you select data from the beginning of the dataset onward. 

### Expected result
The time series widget date range should show the correct values in the dataset for the selected time period.

S/B: 10970892
cc @CartoDB/support

zingbot:
I have a second map like this, if needed. The time series widget is off, and by almost the same amount of time – about two years. I verified the discrepancy looking at the data, which shows 'pre_filing_date' values outside of what is being shown. 

Account: jferzoco. 
Table: Renovation Dashboard
Embed Link: https://team.carto.com/u/jferzoco/builder/c710be34-d939-47f0-b5e2-758089711095/embed

ramiroaznar:
I have already reported the same issue here https://github.com/CartoDB/cartodb/issues/9746 I think it is related to `deep-insights`: https://github.com/CartoDB/deep-insights.js/issues/476

dracodoc:
[...]

PR #11634, opened 2017-02-24
Dates timeseries filtering

nobuti:
This PR implements #11174


nobuti:
Frontend tests passing 🚀

PR #11636, opened 2017-02-24
Zeus ci

### LINKED ISSUE #11643: Speed up backend testing
This a tracking issue for the merge process of my latest leapfrog, speeding up tests using Zeus. The idea is to use Zeus to avoid reloading the code between rspec executions. There are also some improvements that can be made to the CI scripts, avoiding creation of so many redis and postgresql databases, which can be handled in the code.

Steps:
1. Can be done in parallel:
  a. Create a script in the repo to run the tests in parallel with a single command, so it is easier to modify the CI process. Basically, move part of the code from chef to the repo. #11644
  b. Merge Zeus / Stellar #11645
2. Wait for a bit after merging 1a, so everyone can get the script in their branches. Then change chef configuration to use it.
3. Merge a modified version of Zeus CI with a runner script #11646 (THIS ISSUE)

Expected speed-up, about 5 minutes for backend tests (15-20%)

javitonino:
Closes #11643

Saves a bit over 5 minutes in CI. Also saves a couple of minutes creating and destroying databases.

Changes:
- Uses Zeus to avoid reloading the environment between specs. Very helpful for small specs
- Uses one database per worker, instead of one database per spec
- Massive cleanup of CI scripts. 
- All the CI logs/tempfiles are now contained in the `parallel_tests` directory. This makes it easier to run this locally without polluting the root directory.
- Fixes some specs
- Adds timing information to parallel test runner
- A recent version of the `parallel` tool is needed. I deployed the needed version to CI nodes already.

This bring down the CI time to ~15 minutes (from ~20), which is limited by `user_spec` which takes that long to run. It is at the top of the Makefile so it starts at the beginning, but still is the last one to finish. In further work, we should focus on that spec (or even considering splitting) in order to make this even faster.


javitonino:
Tests took 23 min instead of current average of 29. Looking ✅ -> 🚀

PR #11638, opened 2017-02-24
Fixed problems found with input-fill

xavijam:
Related ticket: https://github.com/CartoDB/cartodb/issues/11648

Lately I've been suffering several weird things with the fill component. For example, the category list rendering null strings instead of the top 10 values for that column, slow performance between tabs, ...

**The main problem was the way we have been rendering several views and not adding them as subviews to be cleaned when the main view is removed (`view.addView(subview)`). That caused several ghost views and unexpected behaviours.**

Fixed that, I've reviewed the whole component looking for that pattern, and I've applied the Boy Scout law (:tm:), trying to change several things like:

(1) Add those created subviews to the main view (the most important thing of this PR) using `addView` method.
(2) Start using `listenTo` instead of on/bind due to the fact that it avoids the use of `add_related_model` (for "storing" listeners to external models or collections).
(3) Created template instead of using inline HTML.
(4) Generate components (`_initViews`) when the view is rendered, not in the initialize (and of course applying (1) ).
(5) Following the Boy Scout (:tm:), I've been using the `checkAndBuildOpts` (_ya puestos..._).
(6) And a bonus track, as @nobuti pointed, there was a problem with the style-content-view, where it was rendered once a change in the styles was done. It was caused by a not initialization of the `autoStyle` layer definition attribute.

**THINGS TO TEST (:scream):**
- [x] Test deeply the fill component (several times) in the style form:
  - Changing size.
  - Changing size by value.
  - Change fixed color.
  - Apply icons.
  - Change color by value.
  - Change ramps.
  - Change category colors and images.
  - …
- [x] Check that after changing any value of the style (ramp to categories, stroke color,...), the dropdown is not closed (compare behaviour with a production account).
- [x] Check the fill component in the custom legends form.
- [x] Check the fill component in the widget form (change fill color and custom ramp).
- [x] Check the fill component in any other context not already mentioned ^^^.

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/components/form-components/editors/fill/color-picker/color-picker-template.tpl
+++ b/lib/assets/core/javascripts/cartodb3/components/form-components/editors/fill/color-picker/color-picker-template.tpl
@@ -1,0 +1,14 @@
+<div class="colorpicker dropdown-menu">
+  <div class="colorpicker-saturation">
+    <i><b></b></i>
+  </div>
+  <div class="colorpicker-hue">
+    <i class="ColorPicker-ball"></i>
+  </div>
+  <div class="colorpicker-alpha js-alpha">
+    <i class="ColorPicker-ball"></i>
+  </div>
+  <div class="ColorPicker-colorWrapper">
+    <div class="ColorPicker-color colorpicker-color js-color"></div>
+  </div>
+</div>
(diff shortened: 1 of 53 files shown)


xavijam:
:dancer: moved to acceptance then! Do you want me to remove it for doing that or you will follow it?

javierarce:
I'll take it again! :)

javierarce:
### Acceptance: fast & furious! 👌

javierarce:
I've done a quick test this morning and I found a minor problem. Currently whenever a user selects an icon from the modal window, the fill component gets closed. This PR solves this issue (now, clicking in an icon won't close the fill component), but I've just found that if the user clicks in the modal window itself (not in an icon) the component gets closed :/

javierarce:
… after a talk with @xavijam we've decided that the best strategy is to create an new ticket for this issue.

xavijam:
Hey, but share the new ticket :D.

Issue #11583, opened 2017-02-20
Layer onboarding minor detail when layer is synced


xavijam:
We forgot in the matrix of possible cases when the layer is synced.

![screen shot 2017-02-20 at 12 59 30](https://cloud.githubusercontent.com/assets/132146/23124298/8a525f40-f76c-11e6-80bd-a3de4d0874eb.png)

arianaescobar:
Also: 

<img width="532" alt="screen shot 2017-02-20 at 14 34 46" src="https://cloud.githubusercontent.com/assets/16247628/23127179/5a338c54-f77a-11e6-9e9a-ef2eb2c2f949.png">

ivanmalagon:
Ready for code review.

📄  👀

xavijam:
NICE!

xavijam:
Done! 💋 . Great job @ivanmalagon 👏 .

PR #11532, opened 2017-02-14
10905 layer onboarding

ivanmalagon:
Implements #10905 

This PR enables a user onboarding flow when entering the layer edition.

onboarding

### Minimum Viable Acceptance

There are three distinct onboardings when entering layer edition:
1. Data tab.
1. Analysis tab.
1. Style tab.

There are plenty of combinations per each one.

#### Data tab

There are four steps in the onboarding process. The last step is the tricky one since its location depends of the map state.

The different cases to tests are:
- No widgets.
- With side widgets.
- With time series widget.
- With animated time series widget.

And per each one there are three responsive breakpoints that modify page elements sizes.
- More than 1600px width. (1680x1050, for instance).
- More than 1400px width. (1440x900, for instance).
- Less than 1400px width.  (1024x768, for instance).

**Known issues**
The buttons for geometry edition and dataset switching doesn't place themselves in the right position in the smaller breakpoint. The onboarding highlights the right place where they should be placed once this issue is fixed.

#### Analysis tab

Only one step. Test that the button triggers an analysis creation.

#### Style tab

Two different scenarios.

1. Points geometry: in this scenario test all different aggregations (simple, heatmap, hexabins, ...) since there are different sizes for the menu options that reflect on this onboarding.

2. Lines or polygons geometry: one less step than in points onboarding.

The onboarding was already checked and accepted by the one and only @arianaescobar and @noguerol will take a look also, as far as I know.

#### Warning

If you check `Don't show this again` you won't see the onboarding (that's how it works 😉 ) so test those checks as the last step.

If you accidentally check them and you need to trigger the onboarding againg, ping me.

PR #11657, opened 2017-02-28
11583 synced onboarding

ivanmalagon:
Implements #11583 

I've done a big refactor on the layer onboardings. This is because for every corner case that we run into, the amount of CSS and custom code is exploding.

The new way of highlighting panels in these onboardings is via selector: we tell the DOM element that we want to focus on and the onboarding goes there programatically. As you can check in the code, this opens a way for having generic onboardings that are very fast to develop and the code has been cut dramatically, especially CSS.

I've also included the closing of the onboarding when clicking on the highlighted panel.

## Acceptance

Since the onboardings have suffered a complete refactor we should pass:

- Original acceptance for layer onboardings -> [#11532](https://github.com/CartoDB/cartodb/pull/11532)

- Check synced layers. The onboarding should highlight correctly the different steps besides having a different height than non synced layers.

- Check all possible status of feature edition:
    - No edition available: in layers with an applied analysis, for instance.
    - Edit one type of geometry: in layers with geometry, points for instance.
    - Add any geometry: in empty layers, for instance, you can see three buttons instead of one.

- Check that the onboarding gets closed when clicking the highlighted panel in every possible step. 

That's all, folks!


lib/assets/core/javascripts/cartodb3/components/onboardings/generic/generic-onboarding-view.js
```diff
@@ -0,0 +1,222 @@
...
+    }
+  },
+
+  _forget: function () {
+    this._onboardingNotification.setKey(this._notificationKey, true);
+    this._onboardingNotification.save();
+  },
+
+  clean: function () {
+    $(document).off('keydown', this._keyDown);
```
xavijam: This is a mania, put `clean` method at the end of the class, because it is the last method to be used. Don't take it into account if you don't want :).

xavijam:
@matallo, this will affect us with the geocoding onboarding 💋 .

xavijam:
Already done the acceptance, perfect man!. Smoother than ever, and working in cases. Checked in rui account:

- [x] Onboarding for each type of geometry (lines, polygons, points).
- [x] Onboarding for a synced layer.
- [x] Onboarding for a read-only layer.
- [x] Clicking in the highlighed stuff, close the onboarding.
- [x] Checked onboarding with different widgets (time-series, categories,...).

Issue #11666, opened 2017-02-28
Having a bubbles legend, if you apply an analysis, JS error is thrown


xavijam:
**STR**
- Create a map of points (I've used `paradas_metro_madrid` one).
- Style size by value, it should create a bubbles legend automatically.
- Style color by value, it should create a categories legend automatically.
- Reload the page.
- Open the browser console.
- Go to the points layer and add a new analysis (subsample for example 💃 ).
- Apply the analysis.
- 💥 .

It fails because it needs a color for the bubble legend but in this moment the style model doesn't have one because `none` type is applied, so everything fails, and the worst thing, this is one of the reasons of a ghost node.

javitonino:
Deployed 🚀

PR #11667, opened 2017-02-28
Fixed problem with Bubbles legend when a new analysis is applied

xavijam:
Related: https://github.com/CartoDB/cartodb/issues/11666.

Already described the problem in the ticket ^^^. In order to check it, just follow the steps described there.

- [x] Check that steps described by @javitonino below are working.
- [x] Check that if a style has changed, bubble legend should react to that change (if it implies the fill or stroke color).
- [x] Check that, when a bubble legend is visible, if you apply a widget autostyle, that bubble legend doesn't change the color.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -56,7 +56,8 @@ Development
 * Migrate to use GNIP v2 for twitter search connector (#10051, #11595)
 
 ### Bug fixes
-* Fixed missing metadata option in header when dataset is sync
+* Fixed problem with Bubbles legend when a new analysis is applied (#11666)
+* Fixed missing metadata option in header when dataset is sync (#11458)
 * Fixed problem with dates when filtering time series widget
 * Fixed problem switching between qualitative and quantitative attributes (#10654)
 * Fixed problem found in Surfaces related with map panning and widgets filtering
--- a/lib/assets/core/javascripts/cartodb3/data/legends/legend-bubble-definition-model.js
+++ b/lib/assets/core/javascripts/cartodb3/data/legends/legend-bubble-definition-model.js
@@ -1,6 +1,7 @@
 var _ = require('underscore');
 var LegendBaseDefModel = require('./legend-base-definition-model');
 var LegendColorHelper = require('../../editor/layers/layer-content-views/legend/form/legend-color-helper');
+var DEFAULT_BUBBLES_COLOR = '#999999';
 
 module.exports = LegendBaseDefModel.extend({
   defaults: _.extend({}, LegendBaseDefModel.prototype.defaults,
@@ -38,8 +39,9 @@ module.exports = LegendBaseDefModel.extend({
   },
 
   _onChangeStyle: function () {
-    if (!this.layerDefinitionModel.get('autoStyle')) {
-      this.set('fillColor', this._inheritStyleColor(this.layerDefinitionModel.styleModel));
+    var styleModel = this.layerDefinitionModel.styleModel;
+    if (!this.layerDefinitionModel.get('autoStyle') && !styleModel.hasNoneStyles()) {
+      this.set('fillColor', this._inheritStyleColor(styleModel));
     }
   },
 
@@ -69,7 +71,7 @@ module.exports = LegendBaseDefModel.extend({
   _inheritStyleColor: function (styleModel) {
     var fill = styleModel.get('fill');
     var stroke = styleModel.get('stroke');
-    var color = fill.color || stroke.color;
-    return LegendColorHelper.getBubbles(color).color.fixed;
+    var color = (fill && fill.color) || (stroke && stroke.color);
+    return (color && LegendColorHelper.getBubbles(color).color.fixed) || DEFAULT_BUBBLES_COLOR;
   }
 });
--- a/lib/assets/core/javascripts/cartodb3/data/legends/legend-bubble-definition-model.js
+++ b/lib/assets/core/javascripts/cartodb3/data/legends/legend-bubble-definition-model.js
@@ -41,7 +41,7 @@ module.exports = LegendBaseDefModel.extend({
   _onChangeStyle: function () {
     var styleModel = this.layerDefinitionModel.styleModel;
     if (!this.layerDefinitionModel.get('autoStyle') && !styleModel.hasNoneStyles()) {
-      this.set('fillColor', this._inheritStyleColor(styleModel));
+      this.save({ fillColor: this._inheritStyleColor(styleModel) });
     }
   },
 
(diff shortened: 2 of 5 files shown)


javitonino:
Found an issue.

STR:
1. Create points map
2. Style size by value, and then color by value
3. Add subsample analysis, and apply
4. Style size by value, and then color by value
5. Remove analysis

The style is reset to the default yellow instead of persisting the state (layer style node cache stuff).

Another one:

STR:
1. Create points map
2. Style size by value, and then color by value
3. Refresh the map

The bubble legend will be yellow (it should be the primary color of the color ramp).

Finally, I managed to create a ghost node while playing with analysis editions. But I haven't been able to reproduce, and I tried several times. 😕 Maybe you guys have some clue? My approximate steps were:
STR:
1. Create points map
2. Style size by value, and then color by value
3. Add subsample analysis, and apply
4. Refresh
5. Apply some styles
6. Edit the analysis
7. Delete the analysis

As I said, not sure how it happened, so feel free to ignore. I just point this out in case you can think of something that might be missing in this process with the new code.

xavijam:
About case 1 and 3, this is not related with styles cache stuff or other problems, so we should not stop this release. About 2, taking a look....

javitonino:
@xavijam go home!

1, Happens in this PR but not production, so I think it can be related.
3, I thought it might be relates because it's ghost nodes, but cannot reproduce, so let's forget it.

xavijam:
1. Look at the code, it is near impossible that the code is related with the cache stuff @javitonino (it only touches bubble legend).
2. I've found a bug trying to replicate the second problem. Fixing it.

xavijam:
I feel problems have been fixed :crossed_fingers: .

javitonino:
1. Yeah, I know. I tried to replicate again today and I cannot, although it happened to me 3 or 4 times yesterday. There probably was something wrong somewhere else.

Btw, in Firefox I'm getting the following errors when adding an analysis on top of a layer with bubble legends:
screenshot_20170301_094206
It does not seem to be a problem, though. I don't know why it says XML 😕

lib/assets/core/javascripts/cartodb3/data/legends/legend-bubble-definition-model.js
```diff
@@ -38,8 +39,9 @@ module.exports = LegendBaseDefModel.extend({
   },
 
   _onChangeStyle: function () {
-    if (!this.layerDefinitionModel.get('autoStyle')) {
-      this.set('fillColor', this._inheritStyleColor(this.layerDefinitionModel.styleModel));
+    var styleModel = this.layerDefinitionModel.styleModel;
+    if (!this.layerDefinitionModel.get('autoStyle') && !styleModel.hasNoneStyles()) {
+      this.save({ fillColor: this._inheritStyleColor(styleModel) });
```
xavijam: @nobuti, when there is a change in the styles, we choose another color for the bubbles, but we don't persist it. Sounds ok the change?. cc @javitonino

xavijam:
Weird @javitonino, but in any case, if you have the chance, try with the last changes please 💋 .

lib/assets/core/javascripts/cartodb3/data/legends/legend-bubble-definition-model.js
nobuti: If styles change, the legend manager should be aware of it, and in the last term, legend factory should save the legend. Why do we need to save it here?

xavijam: Because it was not being saved, checking those steps:

```
- Create points map
- Style size by value, and then color by value
- Refresh the map
- The bubble legend will be yellow (it should be the primary color of the color ramp).
```

xavijam: I'll check what's going on with legend-manager in this case.

javitonino:
Sure! Will do!

lib/assets/core/javascripts/cartodb3/data/legends/legend-bubble-definition-model.js
xavijam: Thanks @nobuti! 💋

nobuti: Yeah, it was a rhetoric question, what I mean is that we are changing the flow. I'm ok with the change if it fixes the issue, but it would be great not to have exceptions in the manner we handle the flow. Anyway, I'm in.

xavijam: No no, I wanted to ask you because I knew I was missing something. Checking that behaviour you mentioned. Stay tunned.

xavijam: Ok, I've found the thing following the steps:

- Create points map.
- Style size by value.
  · Legend manager checks that there was a change in the style, so `onChange` method is triggered. It ends up with a bubble legend created thanks to legend-factory (it was not created any before, so it is a new model).
- Style color by value.
  · Again legend manager checks that there was another change in the style, so `onChange` is called again of course. In this case, it doesn't create a new bubble legend, it basically modifies the current one, but the color is not present in the modified attrs (`LegendFactory.createLegend(layerDefModel, 'bubble', {title: size.attribute});`) so only modifies the title.

Do you think we should take the color and pass it like an argument [here](https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/deep-insights-integration/legend-manager.js#L60-L69)?

nobuti: Yep :)

xavijam: _ON MY WAY MAN!_ Thanks 💋

xavijam:
Please, @nobuti, take a look when you have a minute. In short:
- Removed bindings from `legend-bubble-definition-model`, because now `layer-manager` takes care of that 💃 .
- Changed tests.
- That's all.

NEWS.md
```diff
@@ -56,7 +56,8 @@ Development
 * Migrate to use GNIP v2 for twitter search connector (#10051, #11595)
 
 ### Bug fixes
-* Fixed missing metadata option in header when dataset is sync
+* Fixed problem with Bubbles legend when a new analysis is applied (#11666)
+* Fixed missing metadata option in header when dataset is sync (#11458)
```
nobuti: I guess this is a missing news from another issue, nothing to do with this one right?

xavijam: No no, it is below. I completed it with the issue number.

javitonino:
Ok, looking 🍏. Could not break it with several combinations of adding analysis, widgets and playing with legends. Rock solid!

I'll deploy after all tests pass.

xavijam:
Yay!, thanks both! 💋

Issue #11617, opened 2017-02-23
Create special slider for DO boundaries


xavijam:
The upcoming update for the DO analysis needs several new components, one of them is a special slider (check image), in order to select the boundaries. If you need more context, check the [main issue](https://github.com/CartoDB/cartodb/issues/9991).

![screen shot 2017-02-23 at 10 25 01](https://cloud.githubusercontent.com/assets/132146/23253259/246054ea-f9b4-11e6-9d46-b8e114e944d1.png)

I encourage to create a new special component for Backbone.Forms. Create and test it and leave it ready for the [integration](https://github.com/CartoDB/cartodb/issues/11619).

---
*Design: https://invis.io/5G9YKGISB.

saleiva:
This component will be used in more forms in the future

javierarce:
How does the component work? 

From the screenshot I imagine that each of the ticks are associated with a title (that is presented to the users) and a value (that's submitted with the form)… is that right? 

If that's true, I'll propose to make it very generic, supporting any type of value (numbers, strings, booleans, etc.). Plus, the component should allow to customize the grey bar through a CSS class.

xavijam:
Fine by me, but take into account we have the number input where the slider appears. But I agree to create a new generic one 👍 . @noguerol or @saleiva could you help @javierarce?

javierarce:
@xavijam yes, I think we could extend from that one or make it more generic, supporting both the simple case and the complex one.

xavijam:
Not sure about extend the number one, I'd go for the other way around. Create a generic one, and then extend then number from that one. But in any case, review the use cases and apply your magic, you know better than me the best solution 💋 .

nobuti:
This is already in master. Closed by #11668.

PR #11668, opened 2017-02-28
Add slider form component

javierarce:
Implements https://github.com/CartoDB/cartodb/issues/11617

This PR adds a new component required for the Data Observatory analysis called `slider`.
This PR also adds a custom class in the form fields (`.Editor-formInner--Slider`, `.Editor-formInner--Text`, etc.) , enabling more fine-grain customization of the elements.

slider

Since we'll want to use a slider in other forms than the DO one, the image of the component is customizable through a `direction` property. The property accepts 2 values, `asc` and `desc` and affects the direction of the background image. If the property is not defined, the default horizontal line of the slider will be used instead.

sliders

### How to use it

The input of the slider is an array of values that will be presented to the user depending on the position of the slider draggable UI. The output will be the position of said value or, if a list of values is submitted, the corresponding value from the array.

The slider also allows changing the value using the left and right keys.

```
slide: {
  type: 'Slider',
  title: 'Slider',
  direction: 'asc', // optional (possible values 'asc' and 'desc')
  values: ['a', 'b', 'c', 3, 4, 5, 6], // this is optional
  labels: ['Huckepack', 'Naseweis', 'Packe', 'Pick', 'Puck', 'Purzelbaum', 'Rumpelbold']
}
```
### How to do acceptance

Since the component is not used yet, in order to do acceptance you'll need to add it to an analysis. Here's a simple way of doing this (we'll we working with the sampling analysis):

1. Open `sampling-form-model.js` and add a couple of sliders inside of the `schema` hash of the  `_setSchema` method (use the code from above if you want).
2. Open the template `sampling-form.tpl`, search for the `data-fields` property and add the sliders you've just created to the list.

3. Add a "Subsample percent of rows" analysis and test the sliders.

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/components/form-components/editors/slider.js
+++ b/lib/assets/core/javascripts/cartodb3/components/form-components/editors/slider.js
@@ -11,6 +11,7 @@ Backbone.Form.editors.Slider = Backbone.Form.editors.Base.extend({
   className: 'CDB-OptionInput-container CDB-OptionInput-container--noMargin',
 
   options: {
+    direction: '',
     labels: [0, 5, 10]
   },
 
@@ -44,6 +45,7 @@ Backbone.Form.editors.Slider = Backbone.Form.editors.Base.extend({
   _initViews: function () {
     this.$el.html(
       template({
+        direction: this.options.direction,
         value: this.value
       })
     );
--- a/lib/assets/core/javascripts/cartodb3/components/form-components/editors/slider.js
+++ b/lib/assets/core/javascripts/cartodb3/components/form-components/editors/slider.js
@@ -1,0 +1,168 @@
+var $ = require('jquery');
+var _ = require('underscore');
+var Backbone = require('backbone');
+var template = require('./slider.tpl');
+require('jquery');
+require('jquery-ui');
+
+var TAB_KEY_CODE = 9;
+var LEFT_ARROW_KEY_CODE = 37;
+var RIGHT_ARROW_KEY_CODE = 39;
+var SPACE_KEY_CODE = 32;
+var ENTER_KEY_CODE = 13;
+
+Backbone.Form.editors.Slider = Backbone.Form.editors.Base.extend({
+
+  tagName: 'ul',
+  className: 'CDB-OptionInput-container CDB-OptionInput-container--noMargin',
+
+  // backbone-forms 0.14.1 has a 'number' validator:
+  // https://github.com/powmedia/backbone-forms/blob/v0.14.1/src/validators.js#L64
+  // We're on 0.14.0 so we have to use a 'regexp' validator
+  options: {
+    data: [0, 5, 10],
+    validators: [{
+      type: 'regexp',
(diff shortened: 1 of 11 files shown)


javierarce:
@urbanphes… maybe you could have a quick look to the variations of the component I created (see the images in the description above).

lib/assets/core/javascripts/cartodb3/components/form-components/editors/slider.js
```diff
@@ -0,0 +1,167 @@
...
+require('jquery-ui');
+
+Backbone.Form.editors.Slider = Backbone.Form.editors.Base.extend({
+
+  tagName: 'ul',
+  className: 'CDB-OptionInput-container CDB-OptionInput-container--noMargin',
+
+  options: {
+    direction: '',
+    labels: [0, 5, 10]
```
javierarce: True, we could remove that. I didn't required the `labels` until I finished the feature and then forget to delete that entry from the options hash.

lib/assets/core/javascripts/cartodb3/components/form-components/editors/slider.js
```diff
@@ -0,0 +1,167 @@
...
+    this._labels = this.options.labels;
+
+    this._initViews();
+  },
+
+  _initViews: function () {
+    this.$el.html(
+      template({
+        direction: this.options.direction,
+        value: this.value
```
javierarce: Nice catch.

lib/assets/core/javascripts/cartodb3/components/form-components/editors/slider.js
```diff
@@ -0,0 +1,167 @@
...
+    $(this.$('.js-tick').get(tickID)).addClass('is-highlighted');
+  },
+
+  _updateLabel: function (tickID) {
+    this.$('.js-label').text(this._labels[tickID]);
+  },
+
+  _getTicks: function (step) {
+    return [0].concat(_(this._labels.length - 1).times(function (n) {
+      return +(step * (n + 1)).toFixed(2);
```
javierarce: Nice! I like that extrapickyness, it makes total sense.

javierarce:
> A final comment. The widget is rendered always pointing to a tick, so visually it has always a value. Does the required validation make sense in this widget?

I think you are right, I'll modify the description to avoid the confusion.

lib/assets/core/javascripts/cartodb3/components/form-components/editors/slider.js
```diff
@@ -0,0 +1,167 @@
...
+
+    if (this.options.values && this.options.labels.length !== this.options.values.length) {
+      throw new Error('values and labels should have the same length');
+    }
+
+    this._values = this.options.values;
+    this._labels = this.options.labels;
+
+    this._initViews();
+  },
```
javierarce: I'm not sure about this, I think that since we take care of the initial value this is not really necessary (but I'm open to change it if you think is important).

javierarce:
After the first CR I've implemented the recommendations and also improved the specs.

ivanmalagon:
The slider is not used in any place so far. Merged to get it available for next branches.

Issue #11184, opened 2017-01-04
JOIN Analysis Fails Without Error Message


michellechandra:
### Context
A user is running into issues applying the JOIN analysis to his map. The query applies without error but no geometry data is shown on the map. I have a test map and datasets stored in my team account.

**User:** chandra
**Dataset:** counties
**Dataset:** im_jm_unemployment_new
**Map:** https://team.carto.com/u/chandra/builder/44af4324-d2d0-11e6-90bd-0e05a8b3e3d7/
 
I am also unable to update the counties table manually with a SQL query and suspect there is an issue with the counties dataset.

### Steps to Reproduce
1. Create a map from the counties dataset
2. Apply a JOIN analysis with the following settings:
![screen shot 2017-01-04 at 14 54 25](https://cloud.githubusercontent.com/assets/5423288/21662284/c0077f0c-d28d-11e6-82e7-cafd3c61cff8.png)

3. Analysis will apply but no geometry is shown on the map.
![screen shot 2017-01-04 at 14 54 34](https://cloud.githubusercontent.com/assets/5423288/21662296/cb90b9ba-d28d-11e6-841d-230c1dba6dc0.png)

### Current Result
Analysis applies but no geometry is shown on the map.

### Expected result
JOIN should be performed and geometry shown on map.

S/B: 10520054

cc @CartoDB/support

saleiva:
When working with I've found something similar.
If selecting the the_geom or the_geom_webmercator as any of the columns on the dropdown the analysis fails without giving error on the UI. 

I propose we don't let users to select the_geom or the_geom_webmercator in the column selector since we already have a combo dedicated for that just above.

javitonino:
In production

PR #11672, opened 2017-03-01
Don't provide the_geom or the_geom_webmercator as columns to keep in a join analysis

xavijam:
Related ticket: #11184.

Basically that, we should not provide any geometry column in the last step of the merge/join analysis.

screen shot 2017-03-01 at 11 18 24

- [ ] Check that in a join analysis in the last step (columns to keep) we don't provide geometry columns.


javitonino:
LGTM ✅ 

This is more or less a selective reversal of https://github.com/CartoDB/cartodb/pull/11367

PR #11675, opened 2017-03-01
Updating perfect-scrollbar dependency

xavijam:
That's all! cc @nobuti


xavijam:
Frontend tests passed, merging! 🔔

Issue #11687, opened 2017-03-02
"Edit this cell" box not opening when double-clicking an empty cell


arianaescobar:
### Context

In the table view, you can edit cells by double-clicking on them but when the cell is empty, it doesn't work. 

### Current Result

![7536dee4-fe7c-11e6-85aa-6af0b1ea8bbe](https://cloud.githubusercontent.com/assets/16247628/23503577/9790d06a-ff3c-11e6-96dc-20b0f776880b.gif)

### Expected result

The "Edit this cell" box should always appear on double-clicks.

xavijam:
Fixed thanks to https://github.com/CartoDB/cartodb/pull/11691.

Issue #11475, opened 2017-02-07
Start using same "read-only" logic from Builder in dataset page.


xavijam:
Finally we have fixed a problem related to how to check if a "node" is read only or not here -> https://github.com/CartoDB/cartodb/pull/11448. Basically we have passed that logic to the node definition model, and it has a special case when it is source type, because it could be `sync`, it could have a SQL applied, user could have or not permission for writting, etc.

So the JS architecture of how to check if a node is read-only is right now in Builder:

```js
Analysis-definition-node-source-model
├── Query-Schema-Model
├── Query-Geometry-Model
└── Table-Model
…
analysisDefNodeSource.isReadOnly();
analysisDefNodeSource.isCustomQueryApplied();
```

Bad thing is in dataset view, we don't use that "node" logic and we directly use a generated query-schema-model and query-geometry-model, and then we use the internal table-model in order to make those read-only checks:

```js
Dataset-view
├── Query-Schema-Model
├── Query-Geometry-Model
└── Table-Model
…
tableModel.isSync();
DatasetView.isCustomQueryApplied();
tableModel.hasWritePermission();
…
```

 But that's not ideal, we should follow the same we do in Builder. Using a node definition model as entry point and then keep using the same logic from that node.

xavijam:
Also, with this, we should review the table-view logic, because there are several methods checking if query is custom but it is 👎 .

xavijam:
Using it thanks to https://github.com/CartoDB/cartodb/pull/11691.

PR #11691, opened 2017-03-02
Applying query-rows-collection change over dataset view

xavijam:
Related https://github.com/CartoDB/cartodb/issues/11316.

After a big and necessary move using `query-rows-collection` inside an `analysis-definition-node-model`, we needed to address several tasks with the table view and of course with the dataset page (because it uses the table view). In short:

**TABLE VIEW:**
- Instead of taking `query-schema-model` and `query-geometry-model`, it uses `analysis-definition-node-source-model`, where it takes those `query-schema-model`, `query-geometry-model`, `table-model`,... and make use of the `isCustomQueryApplied` or `isReadOnly` methods created there. cc @javitonino @CartoDB/frontend 
- Fixed problem double clicking over empty cells (https://github.com/CartoDB/cartodb/issues/11687).
- There is only one event for clicking and double-clicking in `table-body-view`, improving performance.
- Fixed problem when table was initialized and `sort_order` arrow was pointing to desc instead of asc in the `cartodb_id` column.
- Make use of `listenTo` and required opts stuff.
- Clipboard updated and allowing Safari users to copy the cell value :dancer:.

**LAYER CONTENT VIEW (which contains table view in builder):**
- Fixed problem with context buttons, making use of a model in order to store the context (map or table).
- Initialize the table view in a proper way (no need to get table or permission models anymore, easier).

**DATASET VIEW:**
- Using same logic than in the Builder side for the table view. Initialize everything with an `analysis-definition-node-source-model` (https://github.com/CartoDB/cartodb/issues/11475).
- Make use of `listenTo` and required opts stuff.
- Don't try to save a new SQL applied over a table without write permissions (https://github.com/CartoDB/cartodb/issues/10126).

## THINGS TO TEST
**Dataset**
---
- [x] Test a dataset view is working for a standard dataset (your dataset).
- [x] Check that the sort order arrow appears in the cartodb_id column like ↑.
- [x] Sort by other column and check it works correctly.
- [x] Check cell options behaviour works correctly and you can copy the cell value with Safari browsers.
- [x] Test you can edit that dataset ^^^. Create column, create row, change name,...
- [x] Check double click works for cells with _null_ values.
- [ ] Apply a SQL and check the results are ok.
---
- [ ] Visit a dataset shared with you having READ permissions,  check that is rendered properly and you **CAN'T** edit anything.
- [x] Apply a custom query and check that it works and you can't edit anything.
---
- [x] Visit a dataset shared with you having WRITE permissions, check that is rendered properly and you **CAN** edit anything.
- [x] Apply a custom query and check that it works and you can't edit anything.
---

**Builder**
---
- [x] Create a map with several layers, points layer, lines layer, polygons layer. And also one synced layer. And finally import a layer shared with you.
- [x] Check the table view works for any layer.
- [x] Add some analyses and check that table view keeps working properly.
- [x] Being in the table view, apply a SQL and check that the "context menu buttons" (add geometries and context switcher) work properly.

---
cc @matallo


matallo:
woot, let's take a look!

matallo:
works like a charm, I just found some minor visual artifacts when accessing through safari, the rest is all golden, see the details below

---

>  Apply a SQL and check the results are ok.

in safari the pagination buttons look misplaced (working in Chrome, tho)

screen shot 2017-03-07 at 15 26 23
screen shot 2017-03-07 at 15 26 27

login organization doesn't seem to work in safari

screen shot 2017-03-07 at 16 40 48

share read and write permissions between organization members is not working when toggling the permission for all the organization instead of member by member (see attached)

c062cbab-0118-416d-88fc-8a040da40cc2

xavijam:
1) Only happening in Safari, could you open a ticket?
2) Just tested in production and working fine, probably a problem in that staging server?.
3) Already reported -> https://github.com/CartoDB/cartodb/issues/9933.

xavijam:
So,  merging! 🎉

Issue #11193, opened 2017-01-05
Ensure all analysis onboarding screens link to carto.learn guides


javierarce:
A link in the  "join from 2nd layer" analysis onboarding is pointing to an old map academy sql lesson instead of the new guide (https://carto.com/learn/guides/analysis/join-column-from-second-layer). 

We should change the link to point to right resource.

See the linked issue for a detailed discussion of the problem and reference screenshot.

javierarce:
cc: @saleiva

saleiva:
+1

mehak-sachdeva:
Looks great to me - thanks @xavijam !

nobuti:
Closed by #11706.

PR #11706, opened 2017-03-06
Using learn guides in all onboarding analyses

xavijam:
Related ticket: #11193.

So I've replaced/added those available guides in each proper analysis onboarding. Except cluster, where the onboarding is not available (should we create it?).

Could you take a look @namessanti or @andy-esch? Is it ok?

cc @javierarce

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -59,6 +59,7 @@ Development
 * Migrate to use GNIP v2 for twitter search connector (#10051, #11595)
 
 ### Bug fixes
+* Ensured all analysis onboarding screens link to carto.learn guides (#11193)
 * Fixed problem with Bubbles legend when a new analysis is applied (#11666)
 * Fixed missing metadata option in header when dataset is sync (#11458)
 * Fixed problem with dates when filtering time series widget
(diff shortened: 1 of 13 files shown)


andy-esch:
Woohoo! @mehak-sachdeva will review as she wrote most of the guides

mehak-sachdeva:
@xavijam - Thank you, this looks amazing! I reviewed all the links and everything looks good to me. Since the calculate clusters analysis is not very common, I feel it would be great to have an on-boarding for that for the users, as and when the schedule permits.
This is great! :)

nobuti:
I've found an issue:

screen shot 2017-03-07 at 11 11 21

not sure if we should tackle it in this issue or in a new one @xavijam

nobuti:
LGTM. Waiting green and 🚀

nobuti:
Vamos

namessanti:
Vamos! Nice push everyone. It looks great @xavijam 

@mehak-sachdeva we have clustering in a learn milestone already right?

mehak-sachdeva:
@namessanti : We have a guide on clustering already on the Learn platform - I guess what they meant was we don't have an on-boarding page for that yet after the analysis is done. The moment we do, we can link the guide straight it! :)

Issue #11686, opened 2017-03-02
The fill component gets closed when clicking in the icon modal window


javierarce:
Doing the acceptance [for this PR](https://github.com/CartoDB/cartodb/pull/11638) we found that if an user selects an icon from the asset modal window and clicks in the modal itself (that is, not on an icon or button but anywhere else), and then closes the modal window, the fill component gets closed.

### Steps to Reproduce

1. Open a map with some points.
2. Style the points with an icon using the modal window.
3. Click anywhere on the modal window.
4. Close it (by selecting an icon or pressing the esc key)

### Current Result

The fill component gets closed. 

### Expected result

The fill component should remain visible.

xavijam:
Found on that PR, but happening at that moment in production. The reason is, when a component (any dropdown) is opened, we enable a listener over the document in order to close that component if the user clicks out. 
So the solution should be disable that listener when the modal is opened, and when it is closed or the icon is chosen, enable it again.

xavijam:
Thanks @javierarce for the report 💋 .

xavijam:
Deployed!

PR #11718, opened 2017-03-07
Document click and ESC binding

nobuti:
This PR implements #11686.

Read the issue's description and try to reproduce. Try again playing with ESC key instead of document click.


matallo:
nevermind, I opened prod to check the current behaviour and this is totally right, my fault 😞

Issue #11709, opened 2017-03-06
Remove Builder activation notification for new users


noguerol:
### Context
Builder activation message currently shows on every new user, while it only should on users coming from Editor. 
 

### Steps to Reproduce

1. Create a new account

### Current Result
You get the "CARTO Builder has been activated on your account (...)" notification
 

### Expected result
No notification should appear, as Builder is now our only product.

saleiva:
PLEASE

xavijam:
Deployed!

PR #11720, opened 2017-03-08
Do not show builder activated notification for new users

javitonino:
Fix for #11709 

This hides the builder activated notification (black top bar) for users that never had the editor. We take the decision based on creation date, which should be pretty accurate.

**Acceptance**
- With an old user (`created_at < 2016-11-11`) with the notification active (you can delete the user from the `user_notifications` table).
  - [ ] Check that the notification appears at dashboard
  - [ ] Check that the notification appears at dataset view
  - [ ] Check that the notification appears at Builder
- With a brand new user (FREE signup)
  - [ ] Check that the notification doesn't appear at dashboard
  - [ ] Check that the notification doesn't appear at dataset view
  - [ ] Check that the notification doesn't appear at Builder


lib/assets/core/javascripts/cartodb3/dataset.js
```diff
@@ -190,7 +190,8 @@ if (tableModel.isOwner(userModel) && visModel.get('locked')) {
 }
 
 if (!configModel.get('cartodb_com_hosted')) {
-  if (userModel.get('actions').builder_enabled && _.isEmpty(dashboardNotifications)) {
+  if (userModel.get('actions').builder_enabled && Date.parse(userModel.get('created_at')) < new Date(2016, 11, 11) &&
```
nobuti: I'd use `moment.js` when working with dates. Date is not the most reliable thing in JS.

javitonino:
Back to dev then, let's do this right. Thanks @nobuti!

lib/assets/core/javascripts/cartodb3/dataset.js
```diff
@@ -190,7 +191,8 @@ if (tableModel.isOwner(userModel) && visModel.get('locked')) {
 }
 
 if (!configModel.get('cartodb_com_hosted')) {
-  if (userModel.get('actions').builder_enabled && _.isEmpty(dashboardNotifications)) {
+  if (userModel.get('actions').builder_enabled && moment(userModel.get('created_at')) < moment('2016-11-11') &&
```
xavijam: Why this date?

xavijam: And I'd move it as a "cons" var.

xavijam:
Although it could be an unnecessary worrying (because it is just a date), I'd not like to have that date in several places and without control. Why not passing it to the Frontend from Backend? Using it, for example, in the frontend-config method or just like another variable exposed in the HTML.

lib/assets/core/javascripts/cartodb3/dataset.js
javitonino: That is the date new signups started to get the Builder. Regarding constants, I see two options:
- Leave it as is. Not much reason to add constants, specially since we need at least two (cartodb and cartodb3)
- Use a different attribute from backend instead of creation date. A new attribute like `had_editor` or something like that, that is used just for this.

app/controllers/carto/api/user_presenter.rb
```diff
@@ -4,6 +4,7 @@ module Carto
   module Api
     class UserPresenter
       include AccountTypeHelper
+      BUILDER_ACTIVATION_DATE = Date.new(2016, 11, 11).freeze
```
xavijam: 👏

matallo:
checked changing the `created_at` with the `rails c` in staging which worked flawslessly, also checked the next urls:

https://{user}.{host}.com/profile
https://{user}.{host}.com/your_apps
https://{user}.{host}.com/account
https://{host}.com/account/{user}/plan

but in those cases the notification wasn't appearing

xavijam:
Not sure if that notification is present there @matallo.

matallo:
I don't think so, @noguerol ?

xavijam:
I feel we only displayed in dashboard, dataset and builder.

javitonino:
Here is the original ticket: https://github.com/CartoDB/design/issues/272

Right now, it is only displayed in dashboard, dataset and builder (i.e: this is not a regression). But maybe we want to add it to those extra urls.

xavijam:
Well, @noguerol decides, but I would not add that in every page

noguerol:
Why not? if its a global notification its weird to have them on some pages, even if they're not so visited.

xavijam:
Currently it is not added in every page, but ok. Back to development.

noguerol:
If there is not a reason for not having it on a certain page then we have no reason for not making it a global notification.

thanks buds <3

matallo:
I prepared the js assets for the notification, and it is already appearing in the profile pages above. The problem (as @javitonino advanced) is we need to expose the`@dashboard_notifications` object in the respective controllers.

Some backender available to help? 🙏 

--

also, I realized, the link in the dashboard (`cartodb`) and the map (`cartodb3`) templates are different:

one points to https://carto.com/learn/guides
the other https://carto.com/blog/categories/new-features

shouldn't they be the same?

xavijam:
About the link, yes, I feel we should link to guides. Because that one is updated.

matallo:
updated the link, waiting for some assistance with the controllers

matallo:
🚀 back to acceptance?

javitonino:
frontend CI is 🔴, is that expected?

matallo:
arg! damn tests, should be 🍏 soon

javitonino:
Backend CR 👍

app/views/admin/client_applications/api_key.html.erb
```diff
@@ -9,7 +9,7 @@
     var upgrade_url = "<%= current_user.upgrade_url(request.protocol) %>";
```
xavijam: Bad thing about adding it like another "entry" is that trackjs, for example, will be initializated again, for example. I'd try to add the necessary view/behaviour in the current page entry.

app/views/admin/users/account.html.erb
```diff
@@ -17,7 +17,7 @@
       var iframe_src = '<%= @google_plus_config.iframe_src %>';
```
xavijam: In fact we have the account entry shared in both...

xavijam:
I'll make it! 💋

xavijam:
Changes made, back to CR 💃 .

xavijam:
I guess all checks (described in description) are completed right?

xavijam:
Tested in local and it works like expected. Merging!

PR #11733, opened 2017-03-09
Fixed two problems found in Builder for IE11 browsers

xavijam:
1. .fill method for Array objects is not available in IE11.
2. `getAttribute` was not working properly in IE11.

@nobuti


nobuti:
Frontend test passing 🚀

Issue #7866, opened 2016-06-15
Clean up code that reads from non zero padded date keys


rafatower:
No rush for it, it shall not be done before 2016-07-11

See the PR #7761

rafatower:
PR here: https://github.com/CartoDB/cartodb/pull/11738

PR #11738, opened 2017-03-10
7866 clean up non zero padded reads

rafatower:
This fixes https://github.com/CartoDB/cartodb/issues/7866


rafatower:
### Acceptance
* Deploy to a staging

* Log in with a user into the staging

* Perform a few geocodings through the SQL API

* Check that you can get the quota used and it is consistent

* Check that no redis `ZSCORE` command reads non-zero padded keys

PR #11742, opened 2017-03-13
move public_common to core

matallo:
/cc @AleGoiko


AleGoiko:
replicated the structure (except for the client part which wasn't supposed to be public_common, but I got the point :P), I think we can close this. Vasito de agua para @matallo!!

matallo:
BOOM

Issue #11736, opened 2017-03-10
Problem with numeric column wants to be styled as cateogry


xavijam:
Problem found with TrackJS and FullStory.

## STR
1. Create a map of polygons.
2. Style it by value, using cartodb_id column.
3. Change quantification for category.
4. Click over the colors ramp.
5. 💥

xavijam:
Branch: fill-number-to-category

xavijam:
Closed I guess 😬

PR #11744, opened 2017-03-13
Fill number to category bug

xavijam:
Related ticket: https://github.com/CartoDB/cartodb/issues/11736.

After the big revision of the fill component done [here](https://github.com/CartoDB/cartodb/pull/11638), we have realized that the possibility to use a number column with categories is not working properly since that change. So this PR tries to fix that problem.

Apart from that, we have improved the states of the `input-color-categories` using a `viewStatus` model, storing the status (loading | idle | error) and the step of the `stackLayoutModel`.

### STTest

- Add a points | lines | polygons layer into a new map.
- Style it by value, first with a number column. It should show the ramps list.
- Choose one and edit any color of it, it should work properly.
- Now go back to columns list and choose a string one.
- Edit the colors, add any icon/image (only if it is a points layer, remember). It should work properly.
- Now let's choose another number column, fill component should select automatically the first ramp color (not the category colors previously selected).
- Change the quantification, selects `category` one.
- Loader first and then colors list should appear (and map should be refreshed of course).
- Change any color of the list.
- Get back again to column list and choose a number type.
- Ramp list should be selected by default, no category colors selected.
- Edit your ramp colors, everything should work properly.
- Go back (the last one) to columns list and select another number column, the previous edited ramp should be selected (not the default one).

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/components/form-components/editors/fill/input-color/input-categories/input-color-categories.js
+++ b/lib/assets/core/javascripts/cartodb3/components/form-components/editors/fill/input-color/input-categories/input-color-categories.js
@@ -37,6 +37,17 @@ module.exports = CoreView.extend({
     checkAndBuildOpts(opts, REQUIRED_OPTS, this);
     this._imageEnabled = opts.imageEnabled;
     this.listenTo(this.model, 'change:attribute', this._fetchColumns);
+    this._viewModel = new Backbone.Model({
+      step: 0,
+      status: 'idle'
+    });
+
+    this.listenTo(this._viewModel, 'change:status', this.render);
+
+    // If it comes from a quantification change
+    if (this.model.hasChanged('quantification') && this.model.get('quantification') === 'category') {
+      this._fetchColumns();
+    }
   },
 
   render: function () {
@@ -47,6 +58,7 @@ module.exports = CoreView.extend({
     var columnType = column && column.type;
 
     this.$el.append(template({
+      status: this._viewModel.get('status'),
       columnType: columnType,
       attribute: this.model.get('attribute'),
       quantification: this.model.get('quantification') || 'category'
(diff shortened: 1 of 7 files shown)


lib/assets/core/javascripts/cartodb3/components/form-components/editors/fill/input-color/input-color-value-content-view.js
```diff
@@ -41,11 +41,15 @@ module.exports = CoreView.extend({
   },
 
   _initBinds: function () {
-    this.model.bind('change:quantification', function () {
+    this.listenTo(this.model, 'change:quantification', function () {
+      if (this.model.previous('quantification') === 'category') {
+        delete this.model.attributes.range;
```
nobuti: Is there any reason to not to use `unset` with `silent` option?

lib/assets/core/javascripts/cartodb3/components/form-components/editors/fill/input-color/input-categories/input-color-categories.js
```diff
@@ -37,6 +37,17 @@ module.exports = CoreView.extend({
     checkAndBuildOpts(opts, REQUIRED_OPTS, this);
     this._imageEnabled = opts.imageEnabled;
     this.listenTo(this.model, 'change:attribute', this._fetchColumns);
+    this._viewModel = new Backbone.Model({
```
nobuti: ❤️

lib/assets/core/javascripts/cartodb3/components/form-components/editors/fill/input-color/input-color-value-content-view.js
xavijam: I don't know why I decided that, but yes, It is the same.

nobuti:
Everything works but I've found a thing, I don't know if it's an issue:

- Select a number column.
- Customise the ramp.
- Change the quantification to jenks (or quantiles if jenks it's selected). The custom ramp doesn't change.
- Change to category quantification, and then again to jenks or quantiles (the selected in the previous step). The ramp is the default one, not the custom one.

I feel this might be out of scope, but it's weird sometimes the custom ramp is recorded and other not.

Other than that works fine.

xavijam:
Well, once you go from jenks or quantiles or other quantification to category, that custom ramp is lost because category ramp is totally different, there is no a relation with it. At least that's the behaviour I understood.

nobuti:
Perfect then ✨

Issue #11745, opened 2017-03-13
Layer UI disappears after analysis selection when not having cdb_service_quota_info


rochoa:
### Context
*Please explain here below what you were doing when the issue happened*

 

### Steps to Reproduce && Current Result

 
![cdb_service_quota_info](https://cloud.githubusercontent.com/assets/45346/23855436/e2395a3a-07f5-11e7-8863-9e7daf8acef4.gif)

### Expected result
Don't fail when `cdb_service_quota_info` function does not exist if the analysis doesn't require it.

### Browser and version
Google Chrome Version 56.0.2924.87 (64-bit)

xavijam:
In production!

PR #11748, opened 2017-03-13
When quota API is not available, it should not required the quota infobox

xavijam:
Related ticket: https://github.com/CartoDB/cartodb/issues/11745.

It should only add the quota info if the form is valid and the analysis (to be added) requires the quota. It is likely I'm wrong, what do you think @nobuti?

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-controls-view.js
+++ b/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-controls-view.js
@@ -65,21 +65,26 @@ module.exports = CoreView.extend({
     this.clearSubViews();
     var type = this._formModel.get('type');
     var fetchingState = this._viewModel.get('userFetchModelState');
-    var requiresQuota;
+    // If no dataservice, requiresQuota returns false because the model doesn't exist and it's converted to boolean
+    var requiresQuota = AnalysesQuotaOptions.requiresQuota(type, this._quotaInfo);
     var html;
 
     if (this._isAnalysisDone() && !this._hasChanges()) {
       html = this._html();
     } else {
-      if (fetchingState === 'fetching' || fetchingState === 'error') {
-        html = this._createQuotaView();
-      } else if (fetchingState === 'fetched') {
-        // If no dataservice, requiresQuota returns false because the model doesn't exist and it's converted to boolean
-        requiresQuota = AnalysesQuotaOptions.requiresQuota(type, this._quotaInfo);
-        if (this._canSave() && requiresQuota) {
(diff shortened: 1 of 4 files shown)


lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-controls-view.js
```diff
@@ -77,7 +76,7 @@ module.exports = CoreView.extend({
       } else if (fetchingState === 'fetched') {
         // If no dataservice, requiresQuota returns false because the model doesn't exist and it's converted to boolean
         requiresQuota = AnalysesQuotaOptions.requiresQuota(type, this._quotaInfo);
-        if (this._canSave() && (requiresQuota || !apiReady)) {
+        if (this._canSave() && requiresQuota) {
```
xavijam: If the API is not ready, it will go into `this._createQuotaView`, and the error message will appear. But, if it doesn't require quota, it will go into the `_html` function, just displaying the "Apply" button.

xavijam: @nobuti ^^^

nobuti: Makes sense

javitonino:
Taking this acceptance. We should test in a server without dataservices installed, right?

xavijam:
Yep!

javitonino:
I am getting the following message:
screenshot_20170313_163551
- After adding an analysis without quota (subsample). This popups for a second, then hides and the analysis works. Still, ugly.
- I cannot edit the previous subsample analysis (change % of rows). This message popups and removes the `apply` button.

I also tested with the lon/lat geocoder, but it worked ok on creation except for the message (same as above) 🙂

javitonino:
Wait, it appears the wrong branch got deployed! Retesting! Sorry 🙏

xavijam:
😱

javitonino:
Nevermind, I deployed the correct branch but `git` is crazy sometimes. The commit id of the deployed version is correct `f7b3dd71a0355c63c6ba5ea906bcc88d0bb8b8bd`, but it doesn't correctly display the branch name for some reason.

Anyway, the bugs I reported above are real. Time to 🔧 them!

xavijam:
Where did you deploy it?

javitonino:
ded07

xavijam:
Fixed problem with last change, adding some tests.

xavijam:
Here we go @nobuti

javitonino:
I tested in a server with quotas and a server without them and it seems this works ok in both cases.
 :+1:

xavijam:
💃

Issue #11698, opened 2017-03-06
Research: improve JS tests development


ivanmalagon:
### Context

As we continue to increase the number of JS specs, the task of developing them has become unsustainable. We suffer several inconveniences:

* Build time takes long: the time since `watch` notices something has changed to the moment specs have been regenerated  lasts several seconds. In the meantime, if we change some other content, the build process doesn't catch it and we need to be very on top of the build process to be sure that everything has been built.

* Browser debugging: we generate one file for all the specs. This file and its sourcemap are huge. Developer tools have a hard time just loading it and the process of debugging is painful.

* Specs divided in CLI and Browser: our development process is focused on passing the specs on the browser, but we have a few hundred of them set to pass only via Command Line. We use to forget about them. We need to unify them, moving CLI one to run to the browser in order to have only one way of developing specs.

### Task

The aim of this ticket is to do some research about how to improve DX (Developer Experience) in the testing process.

Desired outcomes:

- Decrease tests build time.
- Faster debugging.
- No test must be left behind. Avoid noticing that a test is failing in the CI process.
- No test division between CLI and browser.

### First approach

As a first POC the research will focus only on Builder specs, the ones under `spec/cartodb3`.

What we will try is to generate only the specs related to the current feature branch, avoiding to generate and pass specs that are not going to be affected at all because of the branch changes. If we're successful with this approach, the desired outcomes can be reached releiving developers of being aware of the specs subset to generate. That will be formed automatically.

The ongoing process will be documented in this ticket.

PR #11750, opened 2017-03-13
Tests refactor

ivanmalagon:
Implements #11698


ivanmalagon:
### STATE

Next, I'll describe the current state of this task.

TL;DR We're in the good direction! 💃 

#### Achievements · · · · · · · · · · · · · · · · · ·

**New Grunt task**

There is a new grunt task, called `grunt affected_dev` that build only the affected specs by changes in the current branch. This is made through different steps:

1. A custom node script called `branchFiles` output the changed files in this branch compared to master. The already committed and the unstaged ones. This could be useful for other tasks too.

1. The output feeds another custom script the traverse all Builder specs to build their dependency tree. The outcome is a list of specs affected by the changed files in this branch (the one that comes from the previous step).

1.  This list of affected specs feed a browserify grunt task that builds only those specs.

1. We watch changes on Builder code and specs.

The whole loop once it's being in watch mode takes around 10 seconds on my computer.


**Development separated from testing**

We dropped all specs generation from `grunt dev` task. This way, the watch process is way faster, generating only relevant code, not specs while we're doing regular code.

`grunt dev` and `grunt affected_dev` can run in paralell in case you want to check changes in both code and specs.

**Bundle size** ⚠️ 

Specs bundle size has not decreased as expected. As an example, generating only 170 specs out of a total of 2300 only decreased the bundle size from 31 MB to 22MB. That's it, generating only 7% of total specs only reduces the bundle size to 70%. It's obvious that common dependencies take a big chunk of the total size.

The main problem is that sourcemap is still quite big. We don't run so many specs, which is great and a big difference, but in the moment where you need to debug, we still suffer from a slow process.

As said before, now specs generation is separated from regular code base generation. As we don't affect regular development, next step is using webpack for bundling the specs, separating vendor code from our specs code and generate sourcemaps only for us.

I'll keep you informed.

ivanmalagon:
## Webpacking

TL;DR Builder tests has been  successfully bundled with Webpack but, so far, it's not an ideal solution.

### Main problem: build time

The big issue with webpack and our build needs is that it needs to be triggered as part of a grunt workflow. The workflow is:

1. Copy files from our core / client folders to the dist folder.

2. Bundle the tests: as something has changed, we must compile the tests to reflect those changes. Using Webpack / Browserify / you name it.

3. Build Jasmine file: HTML file using the generated bundle.

4. Get a local server up to serve that page.

5. Watch changes from folders in point 1 in order to copy the modified ones and trigger again this workflow.

Our problem is in point 2. We have to run the bundling process after the modified files (whether regular code or tests) have been copied. And then, we must trigger point 3 after bundling has finished. The problem is that Webpack doesn't have a cache mechanism between two different runs. It uses its own cache **if it's run with its own** `watch`. But we already have one `watch` running (the one for pluggable frontends). Moreover, `Grunt` doesn't behave well with tasks that are always running. If a task doesn't finish, the next one in the queue doesn't trigger. `Grunt` only works well with its own `watch` to have an always running process.

To sum it up, we need to trigger Webpack from scratch every time that the process it's triggered and, since it doesn't reuse any cache, it takes a looong time (30 seconds aprox.) to bundle specs. 

### Good thing: multiple outputs

Webpack allows us to divide easily the whole bundle in two parts: vendor code and our own code. This can be done via two mechanisms.

- Common chunks: this plugin packs everything every time and, as it traverses the dependencies, it bundles the code if different chunks. The most common pattern is to separate modules within `node_modules` in a vendor chunk and the rest of the code in the main chunk

- DLL: this plugin allows us to pack once a set of modules and then use this bundle as reference for our regular code. This is useful for common dependencies that won't change often, as vendor code. Then, when bundling our code, instead of bundling these common dependencies, it only uses the reference. This enables faster builds as Webpack doesn't have to bundle any of those modules again.

**Benefits**

1. Sourcemaps size: as we divide the whole bundle, the only sourcemap that interest us is our code's. The whole bundle for our code only weights 4 GB (agains the former 32 GB). 

2. Webpack quirks: being able to bundle the whole test suite has not been straightforward. In the process, we've discovered several issues that pave the way when we use Webpack for the whole app. Some of this hurdles:
   - TPL loader: we've created a fork of a existing TPL loader because all of them uses lodash in generated code instead of underscore. It behaves slightlt different, enough to break our existing templates.
   - CommonJS shim: several modules in cartodb.js need CommonJS shimming as they're not correct modules. In Browserify we used `browserify-shim` and we need to use a similar strategy here.
    - Global objects: jQuery, for instance, it's supposed to be globally accessible in some parts of our code. 

### Possible strategies

Webpack will be very useful for us in our workflow if we overcome the build time problem. These are two possible ways to tackle it:

1. Use a cache mechanism between runs: not straightforward. There's not a set way to do it. There are plugins for this but they don't work with our code. I've tried [HardSource plugin](https://github.com/mzgoddard/hard-source-webpack-plugin) but it can't serialize our cache. It says clearly that you can run into problems since Webpack cache is quite complex. I've tried other approached too with no success. I'll give it another try for a day and if they still don't work..

2. Rewrite our workflow to use `Webpack watch`: another approach is to rewrite our whole process to run outside Grunt and using `webpack watch`. For me, this is the last step to take.

Any questions? Ping me then 😄

matallo:
can't manage to make it work (yet), when merging with another branch i.e. https://github.com/CartoDB/cartodb/pull/11873

```
Running "browserify:editor" (browserify) task
>> TypeError: iterator must be a function while parsing file: cartodb/lib/assets/javascripts/cartodb/common/dialogs/create/listing/datasets/datasets_paginator_view.js
Warning: Error running grunt-browserify. Use --force to continue.

Aborted due to warnings.
```

also the specs tab throw a lot of red 😛 

screen shot 2017-03-28 at 16 23 42

matallo:
nevermind, running `grunt && grunt dev` just again makes it work 🤔

ivanmalagon:
1. I agree. It's weird that the connect page is the same for both cases. It should show old Editor ones if running `grunt dev`and the new Builder ones in `grunt affected_specs`. We should have two different HTMLs for each case.

2. That's expected, like said at the new CONTRIBUTING.md

```
You will see in the first terminal that a lot of changes build the bundle again. That's normal. The first step of the point 3 is to copy all needed files, so the `watch` of `grunt dev` triggers. Don't worry about it.
```

The first thing to do in `grunt affected_specs` is to copy all the files to have them available for bundling. It's needed in case you don't run it simultaneously with `grunt dev`. That triggers the `grunt dev` watch, but only once.

3. Any console trace?

4. :confused: That's weird. I've `rm -rf node_modules` several times while working with this and I've never found a problem with lodash. I'll investigate.

Thanks for the feedback, Mata.

lib/build/tasks/webpack/webpack.config.js
```diff
@@ -0,0 +1,79 @@
...
+          filename: '[file].map',
+          exclude: /vendor/
+        })
+      ],
+      target: 'web',
+      node: {
+        fs: 'empty'
+      },
+      stats: 'normal'
+      // resolve: {
```
nobuti: Do we need this?

matallo:
@ivanmalagon I think the only problem is lodash not being in the `package.json` so after removing the `node_modules` it is still missing thus the errors

+1 to have different htmls, maybe with a description to each other and a reminder of the available commands

ok to the explanation in the CONTRIBUTING 👍 

thank you for the awesome work

nobuti:
Tested in a couple of medium sizes issues and only the `lodash` flaw 🙌

ivanmalagon:
After testing this branch with one developed by Buti `fine-tune-tests` I realized that locale files were out of the picture.

I've added them to the copy and watch tasks.

xavijam:
Great @ivanmalagon. Let me merge it with master 💋 .

matallo:
just spotted when there's a syntax error, the error handling was better in the former specs (where it simply won't bundle and throw the warning), now it's easy for it to get lost in the red

screen shot 2017-04-04 at 14 09 55

finally I also noticed changes in`lib/assets/core/locale/en.json` won't retrigger affected specs, but as we don't transform the i18n strings I don't think there's a problem right know, just wanted to mention it

anyway, if you introduce a syntax error, and then modify another js file it would retrigger and show the syntax error in the `en.json` file so maybe it would be worth it

ivanmalagon:
Thanks @matallo 

Locale files problems was spotted yesterday (as stated in a comment above yours) and added. Please pull the last changes.

You're right about syntax errors. I'm going to take a look to do it in the best way possible. At least we need that the developer is conscious about something bad happened.

xavijam:
Tested in staging the popups and looks 👌 .

javitonino:
Don't worry about backend tests failing. I killed them after the process got stuck; they are ✅

matallo:
just found an edge case, when adding a new suite, i.e. `text.spec.js` the `affected_specs` won't take it into account as the related component `text.js` won't change but there's indeed this new suite to include into the affected specs, what do you guys think?

xavijam:
And edge case, could you open a ticket for that maptallo?

matallo:
=> https://github.com/CartoDB/cartodb/issues/11919

Issue #11778, opened 2017-03-16
Context menu view not being correctly disposed


ivanmalagon:
### Context
While passing tests, we ran into a weird random problem that left orphan DOM nodes. After doing some research and thanks to the magic touch of @xavijam we discovered the source of the problems.

`context-menu-view` is used as a subview of `data-layer-view`. When hiding the context menu, we call `remove` instead of `clean`. This leaves context menu events attached to document.

Current function
```
  _hideContextMenu: function () {
    this._menuView.remove();
    this.removeView(this._menuView);
    delete this._menuView;
  }, 
```

Tasks:
1. Change that `.remove` call to `clean`

1. Look in other views that hides the context menu in the same way and refactor them also.

Note:
Changing `.remove` to `.clean` fixed the problem, but it's fixed in another research branch. That's why I've opened this ticket, to fix it as soon as possible in the regular codebase.

xavijam:
Merged and deployed!

PR #11781, opened 2017-03-16
Using clean method instead of remove for context-menu-view

xavijam:
Thanks to https://github.com/CartoDB/cartodb/issues/11778.

## STTest
- Check context menus work properly in table view (in dataset and builder pages).
- Check data-layer menu works as expected (layer options in layers list).
- Check map options work as expected (where delete, export, duplicate map [etc] are placed).
- Check context menu when you want to remove a custom basemap works as expected.


nobuti:
Works as expected 💯

xavijam:
You can merge it 💃

Issue #11793, opened 2017-03-21
Remove new_geocoder_quota feature flag


rafatower:
Remove all usages of `new_geocoder_quota` feature flag that was useful
when migrating quotas to redis but should be considered true in all
cases nowadays.

rafatower:
PR here: https://github.com/CartoDB/cartodb/pull/11794

rafatower:
@ethervoid or @jgoizueta could you review this, please? I need it as prior work to other quota issues.

Mind that some of the decisions I made along the way are reasoned in the *commit messages*. I did not delete any code at my own leisure :smile:

rafatower:
### Acceptance
Deploy to a staging server and check the following
#### Editor
- [ ] set up a user in central: to use the editor, have here as its geocoding provider and just a few credits to geocode.
- [ ] check that geocoding a table with a few rows consume quota.
- [ ] check that it cannot geocode passed its quota.
#### Builder
- [ ] set up the user to use Builder instead. Give it a bit more quota.
- [ ] check that geocoding a table with a few rows consume quota.
- [ ] for completeness (not part of these changes but in ds api + ui) check that it cannot geocode passed its quota.

PR #11794, opened 2017-03-21
Remove new_geocoder_quota feature flag

rafatower:
Remove all usages of `new_geocoder_quota` feature flag that was useful
when migrating quotas to redis but should be considered true in all
cases nowadays.

Related ticket: https://github.com/CartoDB/cartodb/issues/11793


services/dataservices-metrics/spec/unit/geocoder_usage_metrics_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      expect {
+        @geocoder_usage_metrics.send(
+          :check_valid_data,
+          :geocoder_internal,
+          :total_requests,
+          -42)
+      }.to raise_exception(ArgumentError, 'Invalid geocoder metric amount')
+    end
+
+    it 'validates that the amount passed cannot be zero' do
```
jgoizueta: Why is zero not a valid value?

services/dataservices-metrics/spec/unit/service_usage_metrics_spec.rb
```diff
@@ -27,4 +27,35 @@
...
+      }.to raise_exception(ArgumentError, 'Invalid metric amount')
+    end
+
+    it 'validates that the amount passed cannot be negative' do
+      expect {
+        @usage_metrics.send(:assert_valid_amount, -42)
+      }.to raise_exception(ArgumentError, 'Invalid metric amount')
+    end
+
+    it 'validates that the amount passed cannot be zero' do
```
jgoizueta: ditto: why cannot be zero?

lib/tasks/user_overquota.rake
```diff
@@ -34,11 +34,7 @@ namespace :cartodb do
         number_cycles.times do |cycle|
           date_from = billing_date << cycle
           date_to = billing_date << (cycle-1)
-          if args[:redis]
-            geocodings = user.get_new_system_geocoding_calls(from: date_from, to: date_to)
-          else
-            geocodings = user.get_db_system_geocoding_calls(date_from, date_to)
-          end
+          geocodings = user.get_new_system_geocoding_calls(from: date_from, to: date_to)
```
jgoizueta: I guess you're not removing the :redis argument from the rake task to maintain compatibility with any scripts using this task.

rafatower:
Arguably passing a zero to increment a counter is a programming error on the client. Other than that it would produce an undesired request to the redis storage. But as I cannot guarantee 100% that the client code is guarded against that I'll relax that condition on the service class. Thanks for the review!

PR #11809, opened 2017-03-21
Optimize quota usage calculation

rafatower:
### Contents
- A refactor of `User.last_billing_cycle`, extracted to a mixin.
- Added a method `ServiceUsageMetrics.get_sum_by_date_range` that minimizes the requests to redis, pretty much tailored to what quota checks do.
- Added a method `ServiceUsageMetrics.get_date_range` to get a hash of `{date => metric_value}`, an interface aimed at fitting in the `UsageMetricsRetriever`
- New methods integrated into clients of `ServiceUsageMetrics` derived classes.

### Acceptance
- [x] On a staging, run the rake task `cartodb:user_overquota:calculate` and check the results got against master (through the console, with `OverquotaUsersService.get_stored_overquota_users` )
- [x] Provision a user with here geocoder provider and a few credits.
- [x] Make sure it can geocode below its quota and that the remaining quota is displayed correctly
- [x]  Make sure that it cannot run passed its quota.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -117,6 +117,7 @@ group :development, :test do
   gem 'byebug'
   gem 'rack'
   gem 'zeus'
+  gem 'timecop', require: false
 
   # Server
   gem 'thin',                           require: false
--- a/Gemfile
+++ b/Gemfile
@@ -117,7 +117,6 @@ group :development, :test do
   gem 'byebug'
   gem 'rack'
   gem 'zeus'
-  gem 'timecop', require: false
 
   # Server
   gem 'thin',                           require: false
--- a/lib/carto/metrics/mapviews_usage_metrics.rb
+++ b/lib/carto/metrics/mapviews_usage_metrics.rb
@@ -32,7 +32,7 @@ module Carto::Metrics
 
     def get_date_range(_service, _metric, date_from, date_to)
       stats = CartoDB::Stats::APICalls.new
-      ret = {}
+      map_views = {}
       (@organization ? @organization.users.map(&:username) : [@username]).each do |username|
         MAPVIEWS_REDIS_KEYS.each do |redis_key|
           user_map_views = stats.get_api_calls_from_redis_source(
@@ -41,10 +41,16 @@ module Carto::Metrics
             from: date_from,
             to: date_to
           )
-          ret.merge!(user_map_views) { |date, accum, value| accum + value }
+          map_views.merge!(user_map_views) { |_date, accum, value| accum + value }
         end
       end
-      ret.reduce({}) { |h, (key, val)| h[Date.parse(key)] = val; h}
+
+      # Return a hash of {Date => Number} pairs instead of { String => Number }
+      # in order to abide to the interface.
+      map_views.reduce({}) do |new_hash, (date_str_key, value)|
+        new_hash[Date.parse(date_str_key)] = value
+        new_hash
+      end
     end
   end
 end
--- a/lib/carto/metrics/mapviews_usage_metrics.rb
+++ b/lib/carto/metrics/mapviews_usage_metrics.rb
@@ -27,5 +27,22 @@ module Carto::Metrics
         end
       end
     end
+
+    def get_date_range(_service, _metric, date_from, date_to)
+      stats = CartoDB::Stats::APICalls.new
+      ret = {}
+      (@organization ? @organization.users.map(&:username) : [@username]).each do |username|
+        MAPVIEWS_REDIS_KEYS.each do |redis_key|
+          user_map_views = stats.get_api_calls_from_redis_source(
+            username,
+            redis_key,
+            from: date_from,
+            to: date_to
+          )
+          ret.merge!(user_map_views) { |date, accum, value| accum + value }
+        end
+      end
+      ret.reduce({}) { |h, (key, val)| h[Date.parse(key)] = val; h}
+    end
   end
 end
--- a/lib/carto/metrics/mapviews_usage_metrics.rb
+++ b/lib/carto/metrics/mapviews_usage_metrics.rb
@@ -1,5 +1,7 @@
+require_relative 'usage_metrics_interface'
+
 module Carto::Metrics
-  class MapviewsUsageMetrics
+  class MapviewsUsageMetrics < UsageMetricsInterface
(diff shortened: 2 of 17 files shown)


rafatower:
@javitonino can you please take a look? especially the part concerning the `UsageMetricsRetriever` thanks!

Gemfile
```diff
@@ -117,6 +117,7 @@ group :development, :test do
   gem 'byebug'
   gem 'rack'
   gem 'zeus'
+  gem 'timecop', require: false
```
javitonino: We got [`Delorean`](https://github.com/bebanjo/delorean) already installed for this, it seems to serve the same purpose.

services/dataservices-metrics/lib/service_usage_metrics.rb
```diff
@@ -39,6 +39,40 @@ def get(service, metric, date = DateTime.current)
       total
     end
 
+    def get_sum_by_date_range(service, metric, date_from, date_to)
```
javitonino: Couldn't this function just do something like `get_date_range(service, metric, date_from, date_to).values.sum`? It seems like a lot of code is duplicated between these two.

lib/carto/metrics/usage_metrics_retriever.rb
```diff
@@ -15,11 +15,15 @@ def metrics
     def get_range(user, org, service, metric, date_from, date_to)
       usage_metrics = @cls.new(user.try(:username), org.try(:name))
 
-      result = {}
-      date_from.upto(date_to).each do |date|
-        result[date] = usage_metrics.get(service, metric, date)
+      if usage_metrics.respond_to? :get_date_range
```
javitonino: Nice to get this refactor applied to this code as well. This should make Marlango queries much better. Thanks! ❤️ 

About this, I see an alternative way to avoid the `respond_to?`. Since there are only two possible things that can go into `usage_metrics` (either a subclass of `ServiceUsageMetrics` or `Carto::Metrics::MapviewsUsageMetrics`). What do you think about always calling `get_date_range` here, and implementing `get_date_range` inside `Carto::Metrics::MapviewsUsageMetrics` (basically, just moving the code in the else there)?

Not opposed to this approach, but I don't trust Ruby's metaprogramming too much (although this class is an example of it 😆). Anyway, just a suggestion. not sure if it's better.

Gemfile
rafatower: arg, I did not notice, thanks for pointing out...

services/dataservices-metrics/lib/service_usage_metrics.rb
rafatower: sure, gonna try and do right that (I implemented them in the reverse order)

lib/carto/metrics/usage_metrics_retriever.rb
rafatower: Fully agree with your comment.

However, let me consider it, depending on the effort needed to implement that in `MapviewsUsageMetrics` **and** in `Carto::Metrics::TwitterImportsRetriever` (if I'm not mistaken), as it's a bit beyond the refactor.

services/dataservices-metrics/lib/service_usage_metrics.rb
rafatower: done

lib/carto/metrics/usage_metrics_retriever.rb
javitonino: A quick summary of how the metrics retriever stuff work, for context (I realize the names are not the best):

From `MetricsControllerHelper`, I call "metrics retrievers" to get the metrics. The "metrics retrievers" implement the interface `get_range`. There are two of them `UsageMetricsRetriever` and `TwitterImportsRetriever`.

At the same time, `UsageMetricsRetriever` is a generic implementation that works with "UsageMetrics" objects. These objects implement the interface `get(service, metric, date)`. All of them are subclasses of `ServiceUsageMetrics` except for `MapviewsUsageMetrics`. (You can see which are used in `MetricsControllerHelper::USAGE_METRICS_CLASSES`.

So, adding `get_date_range` to the mapviews retriever should be enough.

That said, I agree that this might go beyond the scope of the PR, that's why I only suggested the change. Up to you whether to implement this here or not, I am not so sure myself.

rafatower:
@javitonino one last look, please? I wrote a `UsageMetricsInterface` and made `ServiceUsageMetrics` and `MapviewsUsageMetrics` implement it.

lib/carto/metrics/mapviews_usage_metrics.rb
```diff
@@ -37,7 +37,7 @@ def get_date_range(_service, _metric, date_from, date_to)
 
       # Return a hash of {Date => Number} pairs instead of { String => Number }
       # in order to abide to the interface.
-      map_views.reduce({}) do |new_hash, (date_str_key, value)|
+      map_views.inject({}) do |new_hash, (date_str_key, value)|
```
javitonino: I prefer `reduce` over `inject`. If you agree, we can revert this and edit the rule in the `rubocop.yml` file.

rafatower: I do prefer reduce and map but don't know how to rewrite that rule. Actually trying to solve some warnings in my local rubocop I came to think that we should add it to the Gemfile to run it along the rest of the bundle, shouldn't we?
```
bundle exec rubocop my_amazing_file.rb
```

javitonino: I got rubocop installed locally in my machine, but you are right that it might make sense to add to the development section of the Gemfile.

About the rule, it is supposed to be configured, but it seems there is some Hound default (that is not in rubocop) that interferes. We had the same issue with detect/find. Lets add `reduce:` in rubocop.yml and see if it helps (we did something similar with find and it worked):
```
Style/CollectionMethods:
  Description: Preferred collection methods.
  StyleGuide: https://github.com/bbatsov/ruby-style-guide#map-find-select-reduce-size
  Enabled: true
  PreferredMethods:
    collect: map
    collect!: map!
    find_all: select
    find:
    detect: find
    reduce:
    inject: reduce
```

If it doesn't work, I am in favour of ignoring the damn dog and using reduce. Much more readable IMHO

Issue #11803, opened 2017-03-21
Removing a layer within the proper layer is throwing a JS error


xavijam:
## STR
- Create a map of points.
- Once created, go into the layer.
- Try to remove the layer from the layer options.
- 💥 .

Smoke tests: #SMOKE_MG04

xavijam:
Deployed, thanks @matallo 💋

PR #11810, opened 2017-03-21
Don't go back to layers-view when model is destroyed because it is not removed from collection yet

xavijam:
Related ticket: #11803

Thing is, when model is destroyed, remove event is not thrown yet, so it is already available in the collection, and it makes the application fail when it tries to render all available layers. So better to go back to `layers-view` when it is removed.

## STT
- Create a map of at least two layers.
- Go into one of them.
- Delete the layer from the options menu.
- Check that there is no JS errors.


lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-view.js
```diff
@@ -254,7 +254,7 @@ module.exports = CoreView.extend({
   _initBinds: function () {
     this.listenTo(this.model, 'change:context', this._renderContextButtons);
 
-    this.listenTo(this._layerDefinitionModel, 'destroy', this._onClickBack);
+    this.listenTo(this._layerDefinitionModel, 'remove', this._onClickBack);
```
nobuti: That should not happen if the destroy request has the `wail` param, right?

xavijam:
I tested that and it worked, but if I wanted to go back manually while the layer was destroying, the layers-view tried to render it and the JS error appeared.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-view.js
xavijam: I tested that and it worked, but if I wanted to go back manually while the layer was destroying, the layers-view tried to render it and the JS error appeared.

PR #11813, opened 2017-03-21
Respect geometry_types when data from table visualization is parsed

xavijam:
Related ticket: #11799 

Basically, when table model is fetched, we take data from `table_visualization` attribute but we got rid of `geometry_types` and it made failed this part of code where we decide which styles should be included for the new added layer.

## STT
- Create a map with any layer.
- **Import** a layer of polygons/lines.
- Check that the style tab is reflecting the same colors than in the map.

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/data/table-model.js
+++ b/lib/assets/core/javascripts/cartodb3/data/table-model.js
@@ -38,7 +38,13 @@ module.exports = Backbone.Model.extend({
     // attribute, because it returns the canonical visualization data. So we have to take
     // the data from there.
     if (r.table_visualization && !_.isEmpty(r.table_visualization)) {
-      r = r.table_visualization;
+      r = _.extend(
+        {},
+        r.table_visualization,
+        {
+          geometry_types: r.geometry_types
+        }
+      );
     }
 
     if (r.synchronization) {
--- a/NEWS.md
+++ b/NEWS.md
@@ -135,6 +135,7 @@ Development
 * Fixed overflow on loaders.
 * JOIN Analysis Fails Without Error Message (#11184)
 * Fix problem with perfect-scrollbar in Edge browsers (CartoDB/perfect-scrollbar/#2)
+* Correctly autostyle layers based on geometry when adding layers from modal (#11813)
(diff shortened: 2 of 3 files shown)


lib/assets/core/javascripts/cartodb3/data/table-model.js
```diff
@@ -38,7 +38,13 @@ module.exports = Backbone.Model.extend({
     // attribute, because it returns the canonical visualization data. So we have to take
     // the data from there.
     if (r.table_visualization && !_.isEmpty(r.table_visualization)) {
-      r = r.table_visualization;
+      r = _.extend(
+        {},
```
javitonino: Is this `{}` needed? Probably there is some reason to do this in JS world... 😛

lib/assets/core/javascripts/cartodb3/data/table-model.js
```diff
@@ -38,7 +38,13 @@ module.exports = Backbone.Model.extend({
...
+        r.table_visualization,
+        {
+          geometry_types: r.geometry_types
```
javitonino: Only this one?

This is the complete list of attributes at top-level, take a look and see if there is something we use. A lot of these are duplicated at visualization level, so it "works".
```
id:                           
name:                         
privacy:                      
schema:                       
updated_at:                   
rows_counted:                 
table_size:                   
map_id:                       
description:                  
geometry_types:               
table_visualization:          
dependent_visualizations:     
non_dependent_visualizations: 
synchronization:
```
I think this is mostly ok. Maybe `rows_counted` is also needed, but I think the rest are duplicated in visualization. One example of uglyness here, is `id`. You are going to be sending sometimes the UUID of the table, and other times the UUID of the visualization. The controllers in backend side have some magic to try to guess which ID is getting, so it should work anyway.

javitonino:
@xavijam should we move this to acceptance?

xavijam:
Yep, although we need to add the News check.

javitonino:
Tested:
- Adding layers from dashboard
- Adding layers from Builder (existing dataset)
- Adding layers from Builder (importing dataset)

The correct style gets applied both in CartoCSS and the styles properties. 👍

PR #11815, opened 2017-03-21
Shrinkwrap updated pointing to last Deep-insights changes

xavijam:
cc @piensaenpixel and @nobuti in case you see something weird in the generation.


xavijam:
There are changes in cartodb.js#v4, those changes are ok right @donflopez @alonsogarciapablo ?

xavijam:

Issue #11819, opened 2017-03-22
After filtering a widget, style pane is not working


xavijam:
## STR
- Create a map of points.
- Add an analysis (subsample, for instance)
- Add a category widget, for example, to the last node just created.
- Go to the style panel within the layer.
- Being there, select a category.
- Check that style panel is stuck in loading state.

xavijam:
cc @saleiva

nobuti:
In production.

PR #11822, opened 2017-03-22
If there is any query change over an analysis node, style | legend | infowindows panel should react to that change

xavijam:
Related ticket: https://github.com/CartoDB/cartodb/issues/11819.

So, I've just discovered thanks to @rochoa @dgaubert that when a widget is filtered (category or histogram), the query of the referenced node changes. That means, in our application, we should fetch the data (query, geometry and rows) and render again the panel where the user is.

So, we have taken into account that possibility and those three views will listen to query changes and render again when the fetches have finished.

## STR
- Follow the steps described in https://github.com/CartoDB/cartodb/issues/11819 for each type of panel (legends | infowindows | style).


ivanmalagon:
Acceptance passed.

I selected / unselected categories of a widget and checked that style / popup / legend panels didn't freeze.

One minor issue spotted while doing this acceptance, but it's not been created by this PR. More info at https://github.com/CartoDB/cartodb/issues/11825

PR #11848, opened 2017-03-27
Add free users back to overquota calculation

rafatower:
After implementing some optimizations for (over)quota calculation, add back FREE users to overquota calculation.

@jgoizueta do you mind doing this small review? thanks!


rafatower:
In production, excluding free users it went from 11 minutes to less than 2 minutes. The expected gain would be 11 / 2 = 5.5 times faster. With all users it peaked at 500 minutes so the expected exec time after merging this is 90 minutes.

rafatower:
For the record: tested in staging. I'm merging and deploying it and will check tomorrow the duration of the job and hopefully will close the original issue as well.

Issue #11872, opened 2017-03-28
Error tracker doesn't take starting errors and it makes the app provide other falsy errors


xavijam:
After checking a pick of TrackJS errors, I've noticed that if there is a JS error that happens at the beginning, TrackJS doesn't take it because it is not loaded, so it is not tracked there.
Let's try to move the script to the top of the scripts (http://docs.trackjs.com/tracker/installation) in order to fix this problem.

xavijam:
cc @CartoDB/frontend and @nobuti because we saw this in the past.

xavijam:
I'm thinking how to test this and I'm wondering if we could "enable" (basically add the trackjs credentials) in any staging server and test it there. Would it be possible @CartoDB/builder-backend?

javitonino:
We can do it for all staging servers (via chef), or we can do it for one server temporarily (disabling chef, and adding credentials manually). Shouldn't be a problem either way, we know how to cook.

xavijam:
It would be great if we can add them for a while, at least for this acceptance...

javitonino:
Done in ded07 for a few days. Remember to tell us to undo it when done :)

nobuti:
@javitonino already tested, you can undo your balck magic.

javitonino:
Magic undone. Welcome to the regular boring non-magical world. Hope you are happy.

nobuti:
Shipped.

PR #11873, opened 2017-03-28
Move TrackJS in order to obtain better results when any error is thrown at the beginning

xavijam:
Related ticket: https://github.com/CartoDB/cartodb/issues/11872.

Basically that. Where TrackJS is required, we are moving it to the top of the `<script>` files in order to follow their suggestions (http://docs.trackjs.com/tracker/installation).

Due to the fact that we don't load TrackJS in staging, we should review carefully we haven't broken anything important.

cc @CartoDB/frontend

### DIFF
--- a/app/views/admin/visualizations/embed_map_error.html.erb
+++ b/app/views/admin/visualizations/embed_map_error.html.erb
@@ -30,7 +30,6 @@
       </div>
     </div>
 
-    <%= insert_trackjs('embeds') %>
     <%= insert_google_analytics('embeds', true) %>
     <%= insert_hubspot() %>
   </body>
--- a/app/views/admin/visualizations/embed_map.html.erb
+++ b/app/views/admin/visualizations/embed_map.html.erb
@@ -49,7 +49,7 @@
 <% end %>
 
 <%= content_for(:js) do %>
-  <%= render :partial => 'admin/visualizations/public/embed_map_inline_js' %>
   <%= insert_trackjs('embeds') %>
+  <%= render :partial => 'admin/visualizations/public/embed_map_inline_js' %>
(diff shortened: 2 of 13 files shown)


matallo:
NEWS!

matallo:
isn't there a track.js snippet in the public dataset view? https://github.com/CartoDB/cartodb/blob/master/app/views/admin/visualizations/public_dataset.html.erb

app/views/admin/visualizations/embed_map_error.html.erb
```diff
@@ -30,7 +30,6 @@
       </div>
     </div>
 
-    <%= insert_trackjs('embeds') %>
```
matallo: we don't have any map here, but maybe google analytics, hubspot or any other vendor scripts can throw some error?

xavijam: in the embed map error we don't provide any custom js. Take into account it is a html, just that. And if any of those fail, that's their problem, right?

xavijam:
Further to public_dataset, is it possible that we already have it in the layout?

matallo:
indeed, just removed my comment ><

but I couldn't find it in the dataset view i.e. "https://team.carto.com/u/matallo/dataset/untitled_table_38"

matallo:
updated the NEWS myself, just made the comment about the track.js script missing in the dataset view, the rest looks legit 👍

xavijam:
Let me add it in the dataset view 💋 .

xavijam:
Thanks for NEWs update by the way.

xavijam:
Added!

nobuti:
Works like a charm ✨

Issue #11857, opened 2017-03-27
No geometry messages are displayed after a new geometry is drawn


alonsogarciapablo:
### Context
I was testing something else.
 

### Steps to Reproduce

1. Create a new empty map.
2. Click on the new layer.
3. Draw any geometry.

### Current Result
Unless I refresh, I'm seeing the following messages/things:

<img width="929" alt="screen shot 2017-03-27 at 17 15 43" src="https://cloud.githubusercontent.com/assets/390398/24363845/51a212c4-1311-11e7-9557-eb8178d59d0d.png">
<img width="819" alt="screen shot 2017-03-27 at 17 13 20" src="https://cloud.githubusercontent.com/assets/390398/24363846/51a550a6-1311-11e7-86cc-f96f4a758377.png">
<img width="641" alt="screen shot 2017-03-27 at 17 14 02" src="https://cloud.githubusercontent.com/assets/390398/24363847/51b10108-1311-11e7-9488-13413d372406.png">
 
### Expected result
Since I just drew a new geometry, I shouldn't see any of those messages and I should only be able to draw geometries of the same type I just drew.

xavijam:
In production?

alonsogarciapablo:
Yep.

saleiva:
Can you guys get involved @josecruz in this one too? Even if it looks apparently a simple fix, I really want designers to be doing more acceptance so new stuff can appear!

nobuti:
Of course, @josecruz count me in for whatever you need, I'll ping you
tomorrow to deploy it to staging.

josecruz:
Ok so I see this working! I've found another user but is not related with this on (https://github.com/CartoDB/cartodb/issues/11898)

So from my side, :bandera:!

nobuti:
Closed by #11875.

PR #11875, opened 2017-03-28
Feature edition buttons and style panel fixed when adding geometry to an empy layer

ivanmalagon:
Implements #11857 

**How to tackle acceptance**

1. Add an empty layer.

1. Add some geometry via `feature edition`.

1. That layer panels should behave properly according to the geometry added. That is:
    - Style tab has to show the default style for that kind of geometry.
    - Pop-up tab should show something instead of '
There's no geometry in your data to add a popup.'
    - Legend tab should show content instead of '
There's no geometry in your data to add a legend.'
    - Context buttons (the blue ones to add geometry) should only show the kind of geometry added.
    - The layer should not have any warning regarding the layer not having geometry (icon in layer lists).

Now, the other way round.

1. Picking a layer with custom geometry, delete it in order to obtain an empty layer.

1. Check that:
   - Context buttons allows the user to add any kind of geometry.
   - Layer panels show a warning regarding the layer being empty (style, popup and legend).
   - An icon should show in layer list warning that the layer has no geometry.

These steps should be replicated by all kind of geometries: point, line and polygon.

Check also that an action that converts the kind of geometry from one kind to another behaves the same as before. For instance, pick a points layer and apply an area of influence analysis to convert it to polygon.


lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-view.js
```diff
@@ -276,10 +276,13 @@ module.exports = CoreView.extend({
 
     // Fetch geometry type if there isn't one yet
     var queryGeometryModel = this._getQueryGeometryModel();
-    if (!queryGeometryModel.hasValue() && queryGeometryModel.shouldFetch()) {
+    if (queryGeometryModel.shouldFetch()) {
       queryGeometryModel.once('change:simple_geom', function (mdl, simpleGeom) {
         // Reset styles form in order to prevent a sync problem with styles
-        this._layerDefinitionModel.styleModel.setDefaultPropertiesByType('simple', simpleGeom, true);
+        if (mdl.hasValue()) {
+          this._layerDefinitionModel.styleModel.setDefaultPropertiesByType('simple', simpleGeom, true);
```
ivanmalagon: This code was already called before this issue but I wonder: is it possible for a layer with existing geometry in backend to enter this point with `simple_geom` empty? If so, the code will fetch the `queryGeometryModel` and then, when receiving the right type of geometry we would reset the styles, overwriting the existing ones.

I'd say it's no possible to run into this scenario in a normal situation since it's not a bug that we have but I think it's possible.

nobuti:
To ship it after green.

nobuti:
Vamonos.

Issue #11856, opened 2017-03-27
Replace NPM use by Yarn


xavijam:
Basically, after a quick test done by @rochoa, we checked that installing our dependencies took 30 seconds with yarn instead of ~120 using NPM.
Why?
[Yarn](https://yarnpkg.com/en/) is faster because it uses parallelization and caches every package. 
Also it uses its own "shrinkwrap", called `yarn.lock`.

- [ ] Replace use of NPM and start using Yarn in:
  - [x] CI tests.
  - [ ] Development environment: basically checks our grunt tasks.
  - [ ] Staging environment: check deploys are correct, basically the grunt build task.
- [x] Change all documentation, like CONTRIBUTING.
- [x] Compare results from old (npm) to new (yarn).
- [ ] Migrate duplicated dependencies checker to new Yarn.lock*.

*About last bullet: we have a "checker" made by @rochoa and @javisantana (if I'm not wrong), where we check if the same dependency is used in other place with a different version. So, we should keep this checker but using the new `yarn.lock` file instead of the `npm-shrinkwrap.json`.

xavijam:
Some work done in this [PR](https://github.com/CartoDB/cartodb/pull/11882).

rochoa:
What is the status of this one? Are you actively working on it?

xavijam:
I work on it in my free time :(. It would be great if you have anything to add please 🙏 .

jorgesancha:
Hi guys,

after discussing with @rochoa and @CartoDB/systems and to avoid delaying LDS ON Premise, we are going to continue using NPM for the time being. We'll aim to use YARN for the next LDS On Premise release.

Removing from TODO column

xavijam:
So, are we ready to start using it @jorgesancha then?

jorgesancha:
The only reason to delay it (if I remember correctly) was because we didn't want to include it in this onprem release yet. 

@rporres @jvillarf do you see any problem to get this going and to keep it in mind for the next onprem release?

rporres:
The problem with `yarn` for onpremise is the lack of support for `--build-from-source` when compiling windshaft

jorgesancha:
So is the answer "No way" then? Or is it something along the lines of "Not sure how to do it, we have to investigate"?

If you could give us a bit more info on what the options are (if any) to use this onprem so that we can decide whether it makes sense to go ahead with this or not, that'd be great. 

Thanks!

rporres:
[...]

PR #11882, opened 2017-03-28
Migrating to yarn

xavijam:
Related ticket: https://github.com/CartoDB/cartodb/issues/11856.

That's all.

## How to test it?
- [ ] Change things and test that the CI is working properly.
- [ ] Same as previous but checking that staging deploy works correctly.
- [ ] Compare times and check that CI tests and deploys should be slightly faster.


.travis.yml
```diff
@@ -1,15 +1,15 @@
 language: node_js
 node_js:
-  - "0.10"
+  - 6
 
 before_install:
   - git submodule update --init --recursive
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 0.18.1
```
rochoa: Recent versions of Travis have yarn installed when you use Node.js >=4.

Gruntfile.js
```diff
@@ -27,44 +27,44 @@
...
-
-    var duplicatedModules = duplicatedDependencies(require('./npm-shrinkwrap.json'), SHRINKWRAP_MODULES_TO_VALIDATE);
-    if (duplicatedModules.length > 0) {
-      grunt.log.fail("############### /!\\ CAUTION /!\\ #################");
-      grunt.log.fail("Duplicated dependencies found in npm-shrinkwrap.json file.");
-      grunt.log.fail(JSON.stringify(duplicatedModules, null, 4));
-      grunt.log.fail("#################################################");
-      process.exit(1);
-    }
+    // function preFlight(done) {
```
rochoa: We need to adapt this to the version we want to use.

package.json
```diff
@@ -159,6 +159,6 @@
     "test-debug": "node-debug --no-preload ./node_modules/.bin/jasmine --",
     "test-watch": "watch 'npm test' $npm_package_config_npmJsDir",
     "preupdate-internal-deps": "echo 'DEPRECATED TASK, use `npm run-script update-internal-deps` instead'",
-    "update-internal-deps": "rm -rf node_modules && npm install --no-shrinkwrap && npm shrinkwrap"
+    "update-internal-deps": "rm -rf node_modules && yarn install"
```
rochoa: We should use `yarn upgrade`.

Gruntfile.js
xavijam: yep, already added in the ticket as a task to do.

.travis.yml
xavijam: I needed to install this specific version because after that it introduced something it made fail the build 😱  (stack-overflow ™️ ).

package.json
xavijam: :+1:

Issue #11802, opened 2017-03-21
Orchestrate all requests involved for a DO measurement


nobuti:
After https://github.com/CartoDB/cartodb/issues/11617 and https://github.com/CartoDB/cartodb/issues/11618, we need to handle some more request in order to feed a measurement form group. Take this [gist](https://gist.github.com/talos/554c7759bafe53faafd23a9711305220) as an example of api's usage.

nobuti:
Blocking this until we get a clear idea about the new column field and how camshaft deals with the data from the UI.

xavijam:
Blocking again due to new camshaft analysis is not supporting yet `boundaries`, `normalize` and `timestamp` attributes. For context (https://github.com/CartoDB/camshaft/compare/data-observatory-multiple-measures).

nobuti:
Normalisation is integrated already, waiting for the final node in camshaft.

talos:
@nobuti is there some specific code you would like me to review?

talos:
Found the PR, working on a review.

xavijam:
Blocked until new issues are fixed.

xavijam:
So, ladies and gentlemen, say hello to the new DO UI 😄  ❤️  💋

Issue #11621, opened 2017-02-23
Create a way to add several "items" in the same (Backbone)Form


xavijam:
The upcoming update for the DO analysis will let the user add as many measurements as he/she wants, so we should implement the way to do that.

- With one item:
![screen shot 2017-02-23 at 10 25 18](https://cloud.githubusercontent.com/assets/132146/23257216/2964f504-f9c3-11e6-9865-a353cf2a9d1f.png)

- With two items (showing possibility to remove the second one):
![screen shot 2017-02-23 at 10 25 49](https://cloud.githubusercontent.com/assets/132146/23257209/21d22c8a-f9c3-11e6-907b-00a467d7d34a.png)

Of course it should check how to implement it with Backbone.Forms, already checked and it was a bit painful. Create and test it and leave it ready for the [integration](https://github.com/CartoDB/cartodb/issues/11619).

---
*Design: https://invis.io/5G9YKGISB.

xavijam:
Main branch for starting adding the functionality: `data-observatory-analysis-ui`.

xavijam:
Closing because it is already implemented.

Issue #11619, opened 2017-02-23
Integrate all new components in order to update the DO analysis UI


xavijam:
Basically this should be the final ticket before closing [Data Observatory Analysis UI milestone](https://github.com/CartoDB/cartodb/milestone/135). It should integrate:

- [x] [Slider for DO boundaries](https://github.com/CartoDB/cartodb/issues/11617).
- [x] [Dropdown for DO measurament](https://github.com/CartoDB/cartodb/issues/11618).
- [ ] [Orchestrate all requests for a DO measurement](https://github.com/CartoDB/cartodb/issues/11802)
- [ ] [Add several "items" in the same (Backbone)Form](https://github.com/CartoDB/cartodb/issues/11621).
- [ ] [Questions about implementation](https://github.com/CartoDB/design/issues/620).

Main branch where pull request should apply : `data-observatory-analysis-ui`

And of course, it should work as it is described here: https://github.com/CartoDB/cartodb/issues/9991.

---
* Design: https://invis.io/5G9YKGISB

xavijam:
It is gonna be blocked until all points described are done.

nobuti:
Also, [here](https://github.com/CartoDB/cartodb/issues/11663) there is relevant information that is needed for this issue.

jorgesancha:
Dependent on https://github.com/CartoDB/observatory-extension/issues/263

noguerol:
The mockup has been updated: https://invis.io/5G9YKGISB with no "Bounding box" nor "global" options in the region dropdown. The first one has been discarded due to its cost, the second because we still don't have global measurements in DO.

rafatower:
Posting here as I think this is the "master issue". This was blocked time ago but the original blockers are removed AFAIK.

Things are falling into place. But there's still a "known unknown" non-functional issue that I believe has to do with size of geometries and/or complexity validity. Fortunately we're able to reproduce it: https://github.com/CartoDB/cartodb-platform/issues/3081#issuecomment-302357552

xavijam:
Closing because it is already being tackled in the main issue.

PR #11883, opened 2017-03-28
[WIP] DO UI analysis

nobuti:
This PR implements #11802, #11621 and #11619.

do-ui

Things to do:

- [ ] More tests
- [ ] Column name for every measurement
- [ ] Get rid of move cursor for measurement item
- [ ] Test behaviour deeply

In addition to above, we need to fine tune the data. For instance, every region is returned enclosed in `"`, this happens as well for other fields. It should be nice to avoid data formatting.

Also, I think maybe camshaft should change to reflect the new payload, something like this:
```
{
   type: 'data-observatory-measure',
   source: 'a0',
   region: 'wadus',
   measurements: [
       {
           segment_name: 'wadus',
           column_name: 'wadus'
       },
       ....
   ]
}
```


nobuti:
This PR should go against the base branch, not master. Closing and creating a new one.

Issue #11663, opened 2017-02-28
Data observatory frontend integration


nobuti:
In order to revamp the Data Observatory analysis form (design here https://invis.io/5G9YKGISB), the current SQL call is not enough, and we need to use all the helpers already available in the API. 

@talos we need some guidance, let me do some questions about the Data Observatory:

- Could you point us to the proper documentation? I know we have all signatures listed here https://github.com/CartoDB/dataservices-api/blob/development/client/renderer/interface.yaml but it should be great to know what method to use when.
- How do we need to manage the geometry param? I can see some methods accept a `bounds` param of type geometry, but not sure how to deal with it. Can we pass a country name? Do you have an example of this param using the current bounding box?
- Related to above, how do we get the regions available? If you see the design, there is a number next to the region's name, I guess is the number of measurements available in that region. Does the API provide this number?
- We can't see the new column name to augment the dataset, how does it work?

A measurement has a structure like this, according to the API response:
```
{
    numer_id: "xxx",
    numer_name: "...",
    numer_description: null,
    numer_weight: "3",
    numer_license: null,
    numer_source: null,
    numer_type: "Numeric",
    numer_aggregate: "sum",
    numer_extra: null,
    numer_tags: {
         section/tags.ca: "Canada",
         unit/tags.people: "People",
         subsection/tags.language: "Language",
         license/ca.statcan.license.statcan-license: "Statistics Canada Open Licence Agreement",
         source/ca.statcan.license.statcan-census-2011: "Statistics Canada Census of Population 2011"
    },
    valid_denom: false,
    valid_geom: false,
    valid_timespan: false
},
```
- In order to implement the filters, should we use tags?
- I see the description is null in almost every result, how we get the description of a measurement? and for filters?
- In order to run the analysis, what parameters are mandatory? region and measurement? all but normalize?

@rafatower do you see any troubles in order to get the quota info? Take in mind we can have multiple measurements embedded in the same form.

urbanphes:
@nobuti I'll try to answer some questions:

- _In order to implement the filters, should we use tags?_
This was thought and designed far time ago but I can't remember why didn`t we make it. Perhaps @saleiva or @talos can explain you better. But anyway, I update the [**InVision project**](https://invis.io/5G9YKGISB) and  I added another screen for the flow where it can see when you have for instance 2 filters applied. This mention appears on entry point for filters. You can see the sample:

[...]

PR #11885, opened 2017-03-29
DO UI analysis

nobuti:
This PR implements #11802, #11621 and #11619.

do-ui

Things to do:

- [x] More tests
- [x] Column name for every measurement
- [x] Get rid of move cursor for measurement item
- [ ] Test behaviour deeply

### How to test

- [ ] We can add/remove multiple measurements.
- [ ] Every measurement is working independently.
- [ ] Check the new columns are generated properly after an analysis is executed. Also, the new column name is a combination of measurement, denominator and timespan selected values.
- [ ] Check the quota limits is working properly for multiple measurements. The rule is each analysis will consume `n measurements * y rows`.
- [ ] Check that we can edit the analysis once saved.
- [ ] Check the form's validation. Only `measurement` field is required.
- [ ] Check the save button is working properly:
    - A user can't save an invalid form
    - Once saved, the button is disabled until the form changes.
- [ ] This PR introduces a new component for this analysis. The first dropdown allows filtering on the frontend side showing up to 100 items to select. We should check the performance is good.
- [ ] Check that tracking css classes are generated properly.


lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/layer-content-analyses-view.js
```diff
@@ -190,17 +190,23 @@ module.exports = CoreView.extend({
...
   },
 
-  _selectedNodeId: function () {
-    if (arguments.length === 0) {
+  _selectedNodeId: function (id, options) {
+    if (!this._viewModel) {
+      return;
+    }
+
+    if (id === undefined) {
```
xavijam: Q: `if (id === undefined && options === undefined)` ?

nobuti: well, this is a refactor trying to avoid `arguments` in favour of named params. If `id` is undefined, `options` parameter is irrelevant, because we are returning the cached value.

xavijam: :+1:

nobuti:
I've updated the on boarding template:

screen shot 2017-05-11 at 16 09 41

nobuti:
The tests fail because the version of camshaft reference, it's 🍏  local.

app/assets/stylesheets/editor-3/forms.css.scss
```diff
@@ -142,3 +156,8 @@
 .RampItem-bar {
   width: 20px;
 }
+
+// TODO: fix this in CartoAssets
```
xavijam: Don't we have it there?

app/assets/stylesheets/editor-3/legend-list-form.css.scss
```diff
@@ -24,6 +29,18 @@
   }
 }
 
+.ListItem-deteleButton {
+  position: absolute;
+  top: 6px;
+  right: -14px;
+  opacity: 0;
+
+  .ListItem:not(:first-child):hover & {
```
xavijam: first child?

lib/assets/core/javascripts/cartodb3/components/form-components/editors/data-observatory/measurement-item.js
```diff
@@ -0,0 +1,47 @@
...
+    this.nestedForm = new NestedForm({
+      className: 'Editor-formInner--nested',
+      model: modelInstance,
+      idPrefix: this.cid + '_',
+      fieldTemplate: 'nestedField',
+      template: _.template('<form data-fields="*"></form>'),
+      trackingClass: this.options.trackingClass
+    });
+
+    this._observeFormEvents();
```
xavijam: observeFormEvents?

lib/assets/core/javascripts/cartodb3/components/form-components/editors/data-observatory/measurement-item.js
```diff
@@ -0,0 +1,47 @@
...
+
+    this.listenTo(this.nestedForm, 'change', this._onChangeForm, this);
+    // Render form
+    this.$el.html(this.nestedForm.render().el);
+
+    if (this.hasFocus) this.trigger('blur', this);
```
xavijam: I'd try to use it in another line.

lib/assets/core/javascripts/cartodb3/components/form-components/editors/list/list.js
```diff
@@ -39,9 +47,11 @@ Backbone.Form.editors.List = Backbone.Form.editors.Base.extend({
       return editors[type];
     })();
 
-    this._userModel = this.options.userModel;
-    this._configModel = this.options.configModel;
-    this._modals = this.options.modals;
+    _.each(OPTIONS, function (item) {
```
xavijam: Why don't we use `checkAndBuildOpts` helper?

lib/assets/core/javascripts/cartodb3/components/form-components/form.js
```diff
@@ -4,7 +4,7 @@ Backbone.Form = Backbone.Form.extend({
 
   initialize: function (options) {
     this.options = options;
-    this.constructor.__super__.initialize.apply(this, arguments);
+    Backbone.Form.Original.prototype.initialize.call(this, options);
```
xavijam: Thanks!

lib/assets/core/javascripts/cartodb3/components/form-components/nested-form-custom.js
```diff
@@ -0,0 +1,98 @@
+var _ = require('underscore');
```
xavijam: Nested custom form 😱  💥  🏋️

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-quota/analyses-quota-options.js
```diff
@@ -42,13 +50,15 @@ var ANALYSES_WITH_QUOTA_MAP = {
...
-    name: _t('editor.layers.analysis-form.quota.analysis-type.data-observatory-measure'),
-    api: 'observatory',
-    user: 'obs_general',
-    needsQuota: true,
-    transform: unity
-  },
+  'data-observatory-measure': dataObservatoryQuota,
+  'data-observatory-multiple-measures': _.extend(dataObservatoryQuota, {
+    transform: function (input, formModel) {
+      var measurements = formModel.get('measurements');
```
xavijam: 👏

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/data-observatory-multiple-measure-list.tpl
```diff
@@ -0,0 +1,10 @@
+<div class="u-grow">
+  <div data-items class="u-bSpace--xl">
+
+  </div>
+  <div class="u-flex u-justifyEnd" style="width: 100%;"> <!-- We need a fucking u-full helper -->
```
xavijam: u-full? fucking? 😅

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/legend/legends-view.js
```diff
@@ -86,7 +86,7 @@ module.exports = CoreView.extend({
       overlayModel: this._overlayModel
     });
     this.addView(view);
-    this.$('.js-content').append(view.render().el);
+    this.$el.append(view.render().el);
```
xavijam: ?

app/assets/stylesheets/editor-3/forms.css.scss
nobuti: AFAIK no

app/assets/stylesheets/editor-3/legend-list-form.css.scss
nobuti: the first child should not show the trash icon to delete the item.

lib/assets/core/javascripts/cartodb3/components/form-components/editors/data-observatory/measurement-item.js
nobuti: It inherits from Object editor in Backbone.Form: https://github.com/powmedia/backbone-forms/blob/master/src/editors/object.js#L86-L97

lib/assets/core/javascripts/cartodb3/components/form-components/editors/data-observatory/measurement-item.js
```diff
@@ -0,0 +1,47 @@
...
+  },
+
+  render: function () {
+    var data = this.value || {};
+    var NestedModel = this.schema.model;
+
+    // Wrap the data in a model if it isn't already a model instance
+    var modelInstance = (data.constructor === NestedModel) ? data : new NestedModel(data, this.options);
+
+    this.nestedForm = new NestedForm({
```
nobuti: Object editor already does it https://github.com/powmedia/backbone-forms/blob/master/src/editors/object.js#L72-L76

lib/assets/core/javascripts/cartodb3/components/form-components/index.js
```diff
@@ -12,6 +12,8 @@ _.extend(
   }
 );
 
+Backbone.Form.Original = Backbone.Form;
```
nobuti: nested-form-custom.js inherits from Backbone.Form. Due we have already an object that inherits from the same object, nested-form-custom enters in a loop.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/legend/legends-view.js
nobuti: This fixes the overlay in the legends panel view.

xavijam: 💃

lib/assets/core/javascripts/cartodb3/components/form-components/editors/list/list.js
nobuti: mmm it doesn't fit in the required-options module. Not all options are required 🤔

xavijam:
Should we cut off the text of the licenses? Or, should we display the whole information?

screen shot 2017-05-23 at 17 18 10

nobuti:
license

xavijam:
UI things we saw the other day with @saleiva:

- [ ] Boundaries slider "drag ball" is not correctly centered (checked in Safari).
screen shot 2017-05-29 at 11 03 53
- [ ] Text of the boundaries slider is displaced when "timespan" is selected (check image above, checked in Safari).
- [ ] If boundaries component is disabled, it should not have hover animation.
- [ ] You can remove the measurement although there is only one item (icon doesn't appear but you can click it).
 - [ ] Not sure about this (open for discussion with @saleiva), but if we select a region, does it make sense to preserve all the measurements you already set? I'd clean them, just offering the first one and totally cleared.

nobuti:
Almost all UI fixes done, two left waiting comments.

xavijam:
Added some more found in the last round @nobuti @urbanphes.

urbanphes:
@xavijam Removing Licence label it's ok for me. Now the text begins on the left side and helps to read it better.

nobuti:
New slider up and running.

xavijam:
F****** master, could you deploy those changes to Rui server?

On Wed, May 31, 2017 at 18:50, Buti <notifications@github.com> wrote:
New slider up and running.

nobuti:
They are in ruis.

xavijam:
Working like a charm dude! Congrats \o/.

saleiva:
ok, can we add @urbanphes to the acceptance?

nobuti:
It should be nice that someone with a deeper DO knowledge could take a look, what do you think @talos?

lib/assets/core/javascripts/cartodb3/components/form-components/editors/data-observatory-measurements/measurements-dialog-view.js
```diff
@@ -0,0 +1,67 @@
...
+
+  isVisible: function () {
+    return this.model.get('visible');
+  },
+
+  _toggleVisibility: function () {
+    this.$el.toggleClass('is-visible', !!this.isVisible());
+  },
+
+  clean: function () {
```
xavijam: We don't need it right now, do we?

lib/assets/core/javascripts/cartodb3/components/form-components/editors/list/list.js
xavijam: :+1:

lib/assets/core/javascripts/cartodb3/components/form-components/editors/select/select-view.js
```diff
@@ -127,8 +127,9 @@ Backbone.Form.editors.Select = Backbone.Form.editors.Base.extend({
     var item = this.collection.getSelectedItem();
     if (item) {
       return item.getValue();
+    } else {
```
xavijam: Why this change?

lib/assets/core/javascripts/cartodb3/components/form-components/editors/slider/slider.js
```diff
@@ -3,16 +3,17 @@ var _ = require('underscore');
 var Backbone = require('backbone');
```
xavijam: Hello new slider

lib/assets/core/javascripts/cartodb3/components/form-components/editors/slider/slider.js
```diff
@@ -36,78 +37,78 @@ Backbone.Form.editors.Slider = Backbone.Form.editors.Base.extend({
...
+      fillClass: 'rangesliderFill',
+      handleClass: 'rangesliderHandle',
+      onSlide: this._onSlideChange
     });
 
     this._addTicks(step);
-    this._updateUI(tickID);
+    this._updateUI();
+
+    this.$('.js-slider').on('change', this._onValueChange);
```
xavijam: Could we move this to the Backbone events place?:

```js
events: {
  'change .js-lider': '_onValueChange'
}
```

lib/assets/core/javascripts/cartodb3/components/form-components/index.js
```diff
@@ -2,6 +2,7 @@ var $ = require('jquery');
 var _ = require('underscore');
 var Backbone = require('backbone');
 require('backbone-forms');
+require('rangeslider.js');
```
xavijam: Why do we nee it here?

lib/assets/core/javascripts/cartodb3/components/form-components/editors/data-observatory-measurements/measurements-dialog-view.js
nobuti: Nope, this comes from my template. Fixing it..

lib/assets/core/javascripts/cartodb3/components/form-components/editors/select/select-view.js
nobuti: In a form redraw, it could happen that the component would have a value but the options would not be there yet because the data collection was the result of a fetch operation.

lib/assets/core/javascripts/cartodb3/components/form-components/index.js
nobuti: It's a jquery plugin, and we need to plug in. Not sure about the best place, but I assumed that the slider will be used as part of a form control.

lib/assets/core/javascripts/cartodb3/data/data-observatory/boundaries-collection.js
```diff
@@ -0,0 +1,45 @@
+var _ = require('underscore');
+var BaseCollection = require('./data-observatory-base-collection');
+var BaseModel = require('../../components/custom-list/custom-list-item-model');
+
+var BOUNDARIES_QUERY_PLAIN = 'SELECT * FROM OBS_GetAvailableGeometries((SELECT ST_SetSRID(ST_Extent(the_geom), 4326) FROM (<%- query %>) q),  NULL, {{{ measurement }}}, NULL, NULL) denoms WHERE valid_numer IS TRUE ORDER BY numgeoms DESC';
+
+var BOUNDARIES_QUERY_NORMALIZE = 'SELECT * FROM OBS_GetAvailableGeometries((SELECT ST_SetSRID(ST_Extent(the_geom), 4326) FROM (<%- query %>) q), NULL, {{{ measurement }}}, {{{ normalize }}}, NULL) denoms WHERE valid_numer IS TRUE AND (valid_denom IS TRUE OR {{{ normalize }}} IS NULL) ORDER BY numgeoms DESC';
```
talos: I think you're using the variable name `normalize` here where `denom_id` would be more appropriate and match the Observatory API docs better, where `normalization` has another (different meaning).

* `denom_id`: A column ID for a denominator, like `us.census.acs.B01003001`
* `normalization`: How the returned value will be normalized, for example `areaNormalized` or `denominated`.

lib/assets/core/javascripts/cartodb3/data/data-observatory/boundaries-collection.js
```diff
@@ -0,0 +1,45 @@
...
+
+var BOUNDARIES_QUERY_TIMESPAN = 'SELECT * FROM OBS_GetAvailableGeometries((SELECT ST_SetSRID(ST_Extent(the_geom), 4326) FROM (<%- query %>) q), NULL, {{{ measurement }}}, NULL, {{{ timespan }}}) denoms WHERE valid_numer IS TRUE AND valid_timespan IS TRUE ORDER BY numgeoms DESC';
```
talos: Ordering by `numgeoms` makes sense to populate the slider in order; I'm not seeing code elsewhere to default the initial selection to the geometry with the highest `score`, though.   We should make sure to do so.

lib/assets/core/javascripts/cartodb3/data/data-observatory/column-name.js
```diff
@@ -0,0 +1,75 @@
...
+var cdb = require('cartodb.js');
+var errorParse = require('../../helpers/error-parser');
+var checkAndBuildOpts = require('../../helpers/required-opts');
+
+var REQUIRED_OPTS = [
+  'configModel',
+  'nodeDefModel'
+];
+
+var OBS_META_QUERY = 'SELECT OBS_GetMeta((SELECT ST_SetSRID(ST_Extent(the_geom), 4326) FROM (<%- query %>) q), {{{ metadata }}})';
```
talos: It looks like we make one of these requests for each column.  Considering they are added one at a time, that makes sense, but I just want to note that it's totally kosher to pass in many columns to `OBS_GetMeta` at once.

lib/assets/core/javascripts/cartodb3/data/data-observatory/regions-collection.js
```diff
@@ -0,0 +1,24 @@
+var _ = require('underscore');
+var BaseCollection = require('./data-observatory-base-collection');
+var BaseModel = require('../../components/custom-list/custom-list-item-model');
+
+var REGIONS_QUERY = "SELECT count(*) num_measurements, tag.key region_id, tag.value region_name FROM OBS_GetAvailableNumerators((SELECT ST_SetSRID(ST_Extent(the_geom), 4326) FROM (<%- query %>) q)) numers, Jsonb_Each(numers.numer_tags) tag WHERE tag.key like 'section%' GROUP BY tag.key, tag.value";
```
talos: 👍

lib/assets/core/locale/en.json
```diff
@@ -193,6 +193,11 @@
       "custom-column": "This column contains the data requested from the Data Observatory",
       "description": "The Data Observatory enrichment is the result of the geometry location and the measure requested."
     },
+    "data-observatory-multiple-measures": {
+      "title": "Enrich from the Data Observatory",
+      "custom-columns": "These columns contain the data requested from the Data Observatory",
+      "description": "The Data Observatory enrichment is the result of the geometry location and the measure requested."
```
talos: "...the *measures* requested."

talos:
Looking really good!  A few requested changes above.  I think we could avoid some future confusion if we use the `denom_id` terminology instead of `normalize`, which means something different in the API.

lib/assets/core/javascripts/cartodb3/data/data-observatory/boundaries-collection.js
jorgesancha: cc @rafatower

jorgesancha:
Hey @saleiva it would be useful for you to also do acceptance of this one if you have a mo

jorgesancha:
Tiny issue: autocomplete should be case insensitive, no?

doui

kevin-reilly:
Folks, I'm seeing a bunch of bugs here:

## Analysis Only Returns Nulls

I ran an analysis that looked at Zillow Rental Value Index Single Family (2016-8 timespan) and Non-US Citizen Population (2011-2015 timespan) on the Wal-Mart Dataset.  Only NULL values where returned.

screen shot 2017-06-08 at 3 34 39 pm

I tried this for a couple of different maps, datasets, and analysis and was never able to get proper values to appear in the new column.

## Measurement Selector Showing Error But Not Letting Me Correct

* Selected New York as the Region
* Selected a Measurement
* Scrolled down to Timespan
* Measurement disappeared
* Selected Timespan
* Went back up to measurement to re-add it but couldn't get the pull down again.  No response when I clicked on the field.

do-test-4

This same behavior happened in a few others cases all when New York was selected as the region.

## Region Selector Not Correctly Filtering Datasets on First Load

* Navigating to the DO Analysis screen for the first time in the session
* Select "United States" in Region field
* Click on Measurement field and see "Population Density per Kilometre" which appears to be a Canadian dataset
* Select "Population Density per Kilometre" 
* On some refresh event, this dataset is disappeared from the selection
* Went back to the Measurement selector and saw a different list datasets, these might be US only

do-test-region

## Sort Order in Timespan Pull Down is Random

The sort order for the Timespans appears random.  These should be most recent first, then descending.

screen shot 2017-06-08 at 3 18 36 pm

## Display of Measurement Name

Many of the datasets in our DO have long names.  Often they have common prefixes (i.e. "Zillow Home Value Index {something}."  The current field is too short and is truncating the name.  Once you've selected it, you cannot click back in to see what your selection was.  It is impossible to check to make sure you got the right Measurement.

We should fix it by either having the text wrap (ideally) or at least showing the full value on-click / on-hover.

screen shot 2017-06-08 at 3 22 41 pm

## Choosing the Same Measurement

Nothing prevents a user from choosing the same Measurement with the same Timespan.  However, when you run the analysis, you get a cryptic error message that says the column names are ambiguous.  I assume this is because it is trying to create two columns with the same name.  The easiest thing to do here is to make the error more user friendly.  Better, would be to warn the user when they make the selection that this won't work.

screen shot 2017-06-08 at 3 43 01 pm

nobuti:
About the `null` values, @ethervoid checked the dump we have in staging is too old. We think there is no data for those parameters, once we load a fresh dump, we can make sure about this. But there is a problem related with this, there is no disk space to load the full range of data: https://github.com/CartoDB/cartodb-platform/issues/3027

nobuti:
About the measurement name being cut, I'm afraid it's a problem we have across the whole interface. @urbanphes @saleiva do you see any good solution?

xavijam:
Hey @kevin-reilly, as @nobuti has pointed that text-cut "issue" is something we have across the whole builder. But if you make over the item for a few seconds the whole text will be displayed, like in any other web application. 
Further to the last problem, @nobuti has added a validation but it is something we need to check with the design team.

saleiva:
Hi guys, I had tons of things that are already reported by @kevin-reilly (I can reproduce it all), so I'm adding only the other ones here:
- [ ] We are missing measurements descriptions (this might help with the naming issue that @kevin-reilly is talking about) as they were designed in the mockups.
- [ ] Spacing between different measurement blocks is wrong so it's hard to see the separation between them, it should be 16px.
- [ ] Some measurements show a tons of different boundaries and the component becomes almost unusable. I'm not sure if we shouldn't hide some since they look very specific. For example `13.5.34.1.6. Workers employed in firms in construction`

kevin-reilly:
@xavijam @nobuti Yes, I've noticed the "clipped text" issue through out the site.  But given the number of different measurements available here, it seems like it could be more of an issue.

The least we should do now is display the full name of the selected Measurement on-hover or top result when the input box is clicked.  This doesn't seem to be happening right now.  

To be honest, I don't remember which Measurement I selected in the GIF below and I can't figure it out given the way the feature works now.

do-test-5

nobuti:
> Some measurements show a tons of different boundaries and the component becomes almost unusable. I'm not sure if we shouldn't hide some since they look very specific.

Do we know the max number of boundaries to show?

nobuti:
About the second issue of the list above, I've been investigating all day. In the first place, I thought we were dealing with a race condition, but none of the fixes applied worked, so I took a look at the data returned by the API. Taking in mind [this comment](https://github.com/CartoDB/cartodb/issues/11663#issuecomment-285115216) to establish if a measurement is valid, it turns out that the API it seems inconsistent when the query includes the region param. For instance, for the NY region, none of the measurements includes the `subsection` tag (or `license`):

screen shot 2017-06-09 at 18 53 42

So I'm not sure how to deal with it, should we show all the measurements no matter if they have or not a `subsection` tag?

kevin-reilly:
Love the quick fixes @nobuti. Thanks!  The Timespans are now sorted correctly.  The full description of the Measurement is showing up both on-hover and on the top of the list when you click the input field.  It appears that Measurements are correctly filtering based on the selected Region.  And the Measurements include details on the dataset which helps discovery.  All good.

I'm not sure I understand the question above about sub-sections.  We will need to talk about that one.

Some additional comments after testing today:

## Loop Behavior

It looks like I made the system go into a bit of a loop.  Again, I was using New York as the region, so my guess is that this is related to @nobuti's comment about sub-sections above.  But I selected a measurement and it loaded and on-loaded in a loop until I deleted the analysis.  It also told me that I didn't have enough credits even though I do (but I didn't get a screenshot).  The GIF below is about a 10 second loop for reference.

do-test-6

## Insufficient Credit Error

I setup an analysis and it told me I had "Insufficient Credits" but said it cost 0 credits and I have plenty of credits.  When I clicked Confirm, the analysis ran without an issue.

screen shot 2017-06-09 at 2 46 05 pm

## Choosing Same Measurement

It looks like a user is no longer able to select the same Measurement and Timespan (great).  However, if a user does do this, the "Apply" button is greyed out with no indication of why.  We should show an error that says "You cannot run an analysis selecting the same Measurement and Timespan multiple times.  Please select a new Measurement, a different Timespan, or remove one of the duplicates."  (Edits to copy welcome.)

## Normalize and Boundaries

Sometimes Normalize and Boundaries are not selectable after I have chosen a Measurement.  This may be correct behavior, but it is not obvious why.  Can someone explain this too me?

## Sort Order for Measurements

The sort order for Measurements seems to vary by Map and even layer.  Is there a reason for this?
  It seems like the order should be consistent.

rafatower:
When reporting issues it is quite important to specify what input data you use, at least if it is points or polygons (even better: provide a .carto file), because the DO ability to return meaningful data.

From the "Data Observatory API Architecture Guide":
image

Unfortunately we may have some trouble reflecting that in the UI.

ethervoid:
Taking a look to all of this, I think we should filter the regions we get in order to only show regions with data that has subsections (as @talos said we should only show elements that has subsection tag).

I've modified the query to get the region adding filtering for subsection:
```sql
SELECT count(*) num_measurements, tag.key region_id, tag.value region_name
FROM (
    SELECT * FROM OBS_GetAvailableNumerators((SELECT ST_SetSRID(ST_Extent(the_geom), 4326) 
    FROM (<%- query %>) q)) 
    WHERE jsonb_pretty(numer_tags) LIKE '%subsection/%'
) numers,  Jsonb_Each(numers.numer_tags) tag WHERE tag.key like 'section%'
GROUP BY tag.key, tag.value;
```

I've checked other options but this one is pretty lightweight.

I have a question for @stuartlynn or @michellemho. We should filter all the data that doesn't have `subsection`? are there any corner cases? Thx.

michellemho:
@ethervoid, I agree that we should only show data that has a `subsection` tag. The New York City region data is unique from other regions because the data is mostly categorical text describing building and property lot types. The exception is US Spielman-Singleton population segments, which are categorical. I also think the NYC region data was intended for a special NYC real estate data package and not inclusion in the DO, and therefore do not have a `subsection` tag. The current 23 `subsection` tags mostly correspond to usual census topics like "age and gender", "families", "housing", "language", etc.

ethervoid:
@michellemho So just to be sure, we don't need to show the user data without the subsection tag because it's going to be used "internally" or is not intended to be used in the DO?

ethervoid:
@nobuti has found an issue with geom and the timespans, we have data that when we filter by numer and denom we get:

```
cartodb_staging_user_3c8e3afc-8117-4edb-bc99-eb065d907e76_db=> SELECT timespan_id, timespan_name FROM OBS_GetAvailableTimespans((SELECT ST_SetSRID(ST_Extent(the_geom), 4326) FROM (select * from lightrailstations) q), NULL, 'us.census.acs.B08201002', 'us.census.acs.B11001001') denoms WHERE valid_numer IS TRUE AND (valid_denom IS TRUE OR 'us.census.acs.B11001001' IS NULL) ORDER BY timespan_name DESC
;
-[ RECORD 1 ]-+------------
timespan_id   | 2015 - 2015
timespan_name | 2015 - 2015
-[ RECORD 2 ]-+------------
timespan_id   | 2014 - 2014
timespan_name | 2014 - 2014
-[ RECORD 3 ]-+------------
timespan_id   | 2011 - 2015
timespan_name | 2011 - 2015
-[ RECORD 4 ]-+------------
timespan_id   | 2010 - 2014
timespan_name | 2010 - 2014
-[ RECORD 5 ]-+------------
timespan_id   | 2010 - 2010
timespan_name | 2010 - 2010
-[ RECORD 6 ]-+------------
timespan_id   | 2006 - 2010
timespan_name | 2006 - 2010
```

but if we use that timespan id to filter the geom, we get an empty result because the timespan is stored differently. The numer:
```
obs_2017_05_04_f20ee1131c=# select numer_id, timespans from observatory.obs_meta_numer where numer_id = 'us.census.acs.B08201002';
-[ RECORD 1 ]------------------------------------------------------------------------------------
numer_id  | us.census.acs.B08201002
timespans | {"2006 - 2010","2010 - 2010","2010 - 2014","2011 - 2015","2014 - 2014","2015 - 2015"}
````
the denom
```
-[ RECORD 1 ]------------------------------------------------------------------------------------
denom_id  | us.census.acs.B11001001
timespans | {"2006 - 2010","2010 - 2010","2010 - 2014","2011 - 2015","2014 - 2014","2015 - 2015"}
```
and the geom for that numer and denom:
```
obs_2017_05_04_f20ee1131c=# select timespans FROm observatory.obs_meta_geom where 'us.census.acs.B08201002' = ANY(numers) and 'us.census.acs.B11001001' = ANY(denoms) group by timespans;
-[ RECORD 1 ]-----
timespans | {2015}
-[ RECORD 2 ]-----
timespans | {2014}
```

Looks like the geoms are stored by year so we're not able to filter by geoms using the timespan_id. But we have other cases:
```
-[ RECORD 1 ]----------------------------------------------------------------------------------------------------------------
numer_id  | us.bls.avg_wkly_wage_10
timespans | {2012Q1,2012Q2,2012Q3,2012Q4,2013Q1,2013Q2,2013Q3,2013Q4,2014Q1,2014Q2,2014Q3,2014Q4,2015Q1,2015Q2,2015Q3,2015Q4}
denoms    | {NULL}
geoms     | {us.census.tiger.county,us.census.tiger.county_clipped}
```
here you can see timespan id like `2012Q1` which can't be used to filter geoms because these are the available timespans for geom data:
```
obs_2017_05_04_f20ee1131c=# select timespans from observatory.obs_meta_geom group by timespans;
-[ RECORD 1 ]---------
timespans | {2011}
-[ RECORD 2 ]---------
timespans | {2010}
-[ RECORD 3 ]---------
timespans | {2013}
-[ RECORD 4 ]---------
timespans | {20150101}
-[ RECORD 5 ]---------
timespans | {2016}
-[ RECORD 6 ]---------
timespans | {2014}
-[ RECORD 7 ]---------
timespans | {2015}
-[ RECORD 8 ]---------
timespans | {2016v2}
```

What do you think @michellemho @stuartlynn did we misunderstand anything?

**Comment by csobier** [comment]
What is the expected behavior when there is an existing Data Observatory analysis? For example:
- I opened an older map with the "Enrich from Data Observatory" applied in staging, and the ANALYSIS tab still shows the old options. (The manually added `NEW COL NAME` result still appears in the Data View as well). 
- If I click + to add a new analysis to the chain and select "Enrich from Data Observatory", the new options appear.

Is this the intended design? Should we suggest that users with an existing Data Observatory analysis add a new analysis that includes more granular options to enrich their data? Or will there be some UI text?

**Comment by xavijam** [comment]
In which staging @csobier? Because the new UI is only deployed in **RUI** servers...
We will not provide the option to use the old Data Observatory analysis UI, but it will be compatible with the new version, so the previous attributes will appear in the new analysis UI.

**Comment by ethervoid** [comment]
I've updated my comment about timespans including the query to the API

**Comment by xavijam** [comment]
So, the status. Several UI issues have been fixed, and DO problems are being evaluated. In short:

### Pending UI issues
- Insufficient Credit Error: @nobuti is checking it, and we will fix it.
- Choosing same measurement: this is a difficult one because we haven't implemented a way to "warn" the user when you can't apply an analysis due to a validation with different fields. @nobuti has added a validation and if that case happens, the analysis will not be available. In our opinion, it could add us more complexity (take into account the size of this PR), I feel we should tackle it after merging this one.

### Pending Data Observatory (data) issues
- [Issue with geom and the timespans](https://github.com/CartoDB/cartodb/pull/11885#issuecomment-307841579): discovered yesterday by @nobuti and @ethervoid. We will create a ticket in the DO repository and we will discuss there if we have to face it now or later (reference ticket when it is created).
- [Issue with data returning nulls](https://github.com/CartoDB/cartodb/pull/11885#issuecomment-307733476): we need to check with design how to improve the workflow to avoid that problem.

**Comment by saleiva** [comment]
@xavijam can you elaborate on what is going on regarding the nulls?

**Comment by xavijam** [comment]
> @xavijam can you elaborate on what is going on regarding the nulls?

I feel @rafatower will do better than me.

**Comment by jorgesancha** [comment]
@ethervoid is going to be helping DO/Backend team with DO issues in the short term, so I suggest he picks it up from here.

@ethervoid Can you create the appropriate tickets on the observatory-extension repo so that we can track UI vs DO discussions separately?

**Comment by ethervoid** [comment]
Sure :)

**Comment by csobier** [comment]
@xavijam , I'm using mamut-st.carto-staging.com for testing.... Let me know if I should be using something else

**Comment by ethervoid** [comment]
Created the issues and update [the comment](https://github.com/CartoDB/cartodb/pull/11885#issuecomment-308060872) of @xavijam changing the links of the observatory issues.

**Comment by nobuti** [comment]
Insufficient credit error estimation should be fixed now.

**Comment by xavijam** [comment]
> Insufficient credit error estimation should be fixed now.

👏  👏  👏

**Comment by kevin-reilly** [comment]
I was just testing this tonight.  It was very sluggish and I couldn't get any analysis to run properly.  

I''m using [this](https://mamut-st.carto-staging.com/builder/4adf7297-9bd3-48c0-907c-cedbad6f4e89) map.

I was trying to add the Hispanic population measurement to the lightrailstations dataset.  I kept getting this error:

screen shot 2017-06-13 at 9 30 23 pm

The associated .carto file is attached.

[KR Test (on 2017-06-14 at 01.30.44).carto.zip](https://github.com/CartoDB/cartodb/files/1073292/KR.Test.on.2017-06-14.at.01.30.44.carto.zip)

**Comment by nobuti** [comment]
@kevin-reilly ready for another round, those issues you found yesterday should be fixed.

**Comment by kevin-reilly** [comment]
I'm still using [this](https://mamut-st.carto-staging.com/builder/4adf7297-9bd3-48c0-907c-cedbad6f4e89) map.  I attached the .carto file.

I was successfully able to run and analysis with non-null values that included:

1. Hispanic population, population normalized, 2015 added fine
2. Hispanic population, area normalized, 2015 added fine
3. African American, population, population normalized, 2015 added fine.
4. Zillow - got the error below the first time.  Did not finish running the second time.

I added the four measurements both one-at-a-time (add one, run, add the second, run, etc.) and then in all at once.

One thing that I have noticed, is that after I run an Analysis, the formatting changes.  For example, my points were `red` before the Analysis and turned to `yellow` immediately after I ran the Analysis.  Is this on purpose?

## Credits

The credits seem to be calculating correctly.  

When a user adds a new Measurement to an existing Analysis chain, it does seem like we're overcharging.  For example, in step one above, we charge 54 credits (which is correct).  Then I added the second measurement, it charged me 108 credits.  Arguably, this should only be another 54 since the original Measurement was already added to my data set.  

Is this by design?  I don't think this is a blocker though.  Unless I learn that credits are really expensive.

## Post-Delete Error

Sometimes after I delete an analysis, I receive the error below saying one of the columns in the analysis doesn't exist.  This does not occur all the time.

screen shot 2017-06-14 at 4 10 30 pm

## Changing Date Range

I tried to change the date range on an analysis and I got the error below.  I think this is occurring  because I was using this column as the value for the auto-formatting.

screen shot 2017-06-14 at 9 19 38 pm

Once I cleared the error and attempted to run it again, it wouldn't finish.

**Comment by nobuti** [comment]
@kevin-reilly about the styles being reset, there is already an issue to fix it: https://github.com/CartoDB/cartodb/issues/10700. 

About the credits, it's an important question and I feel you are right. @rafatower do we have a better solution for this? The obvious workaround is to split every measurement in its own analysis.

**Comment by xavijam** [comment]
- Credits: as @nobuti has pointed out, I feel it is not possible to solve right now. And, yes, the option would be to split the analysis again, but that's something new for this analysis.
- Post delete error: I haven't suffered yet, let me keep doing more tests in case I find it.
- Changing date range: yeah, if you have that column to be used in the styles, and you want to change it to a date column, the turbo-style won't work ([issue already opened](https://github.com/CartoDB/cartodb/issues/11059) and engine team is already aware).
- Null values: I'm a bit worried about the results because I've tried an easy dataset (SF neighborhoods), applying a standard measurement and the results are always null ([map](https://mamut-st.carto-staging.com/builder/8e7c4d97-bc9c-4e93-8f07-1d81221e372d) and [.carto](https://www.dropbox.com/s/7i5hxj6rnoy6mgc/DO%20UI%20%2B%20Jam%20%28on%202017-06-15%20at%2008.54.37%29.carto?dl=0)).

**Comment by xavijam** [comment]
Apart from that ^^^, @nobuti and I have thought about "releasing" this new analysis under feature flag (1), the sooner we have it in production the better. And then, if we need to change anything, we could create new issues and manage them in any Kanban.

Sounds good?

1. That means, user with that feature flag will have available the new DO analysis when wants to create a new one. If the user doesn't have that feature flag, old DO analysis will appear.

**Comment by rafatower** [comment]
Re: credits and applying some analysis multiple times

The analysis applies to the whole dataset and has no notion of state (done/not done for a particular row/measurement). Adding such state would over-complicate things and sounds to me like a mid-sized feature in itself.

The same applies to other analyses (eg: geocoding, then inserting a few more rows and geocoding again), but it is particularly complex in the case of DO with multiple measurements.

If we were to improve it, I'd try to think in broader terms: that is, think about the requirements for all analyses involving credits consumption.

**Comment by saleiva** [comment]
we shouldn't "release" this until fixing the critical stuff (ie: null values)

**Comment by xavijam** [comment]
> we shouldn't "release" this until fixing the critical stuff (ie: null values)

Yeah, absolutely. We propose to release it under feature flag and not enable it or open to everybody until we feel confident with it.

**Comment by ethervoid** [comment]
I'm working on the [NULL issue](https://github.com/CartoDB/observatory-extension/issues/290) right now

File: app/assets/stylesheets/editor-3/entry.scss [hunk 24]
```diff
@@ -4,6 +4,7 @@
 @import '../cartoassets/entry';
 @import '../colorpicker/bootstrap-colorpicker/bootstrap-colorpicker';
 
+@import 'vars';
```
xavijam: :+1:

saleiva:
The thing is that then we will lose access to the analysis that works in favour of an analysis that doesn't work. To be it sounds too early, but @kevin-reilly to decide

xavijam:
Agree @saleiva, we should wait until we have the NULLs issue fixed 👍 . Go go go go!

ethervoid:
I've created [a PR](https://github.com/CartoDB/camshaft/pull/312) that fixes the reported NULL problem (returning NULL could be driven by multiple causes. In this case was that we're creating the "boundaries" of the dataset and we're including the NULL values and that seems to cause a misbehavior that leads the DO to think that it's dealing with polygons.

In any case, any use of DO with polygon datasets is could not be possible, we need to know if [we're able to interpolate the measure](https://github.com/CartoDB/observatory-extension/issues/290#issuecomment-308085190) with the polygons. If you want to know more about this, please ask me.

saleiva:
About the polygons part... @ethervoid , can you elaborate? Not sure I follow

ethervoid:
Sure, DO could interpolate the measure value using polygons but we have some especial cases where we can't (you can't see it [here](https://github.com/CartoDB/observatory-extension/issues/290#issuecomment-308085190)). 

So in some cases (I don't fully understand how DO data works yet), we're not able to return a value so we return NULL. Talking with @xavijam both of us agree in giving the user some kind of feedback in this cases or filter that measures directly (we have to previously analyze the dataset). I've added a note in the roadmap to take it into consideration.

Hope I've explained 🤞

kevin-reilly:
Did the staging version revert to the production version somehow?  Am I missing something?

I'm seeing this somehow:

kr test carto 2017-06-15 17-52-06

xavijam:
Sorry @kevin-reilly, something happened with the new feature flag and it was not enabled. It is enabled again. Apologies again :(.

xavijam:
There is something I've suffered several times (and reported by @kevin-reilly too), but not sure how to manage it. Steps:

- Create a new DO analysis, with a measurement A, timestamp B.
- Analysis will generate a column called *A_B* (for example).
- After finishing it successfully, the user styles the layer by that column (obtaining a choropleth map 💃 ).
- Turns out, he/she wants to change that DO analysis, and selects another timestamp, in this case, *C*.
- So, the analysis will generate another column, in this case, will be called *A_C*.
- When the user applies the new analysis it will fail because the Map instantiation will fail too, style will be referencing to a non-existant column A_B.

What should we do in this case @saleiva @kevin-reilly? Not sure if we should reset the styles of the analysis applied.

ethervoid:
Fix for the NULL issue reported by @kevin-reilly deployed

kevin-reilly:
Regarding @xavijam comment about formatting, ideally if A_B is used for formatting that formatting should be applied to A_C when the Analysis is changed.

If that is not possible, if A_B no longer exists because of the change to A_C, than the formatting should return to the default.

How is this handled outside of Analysis?  For example, if I have a column called `Sales` that I use for formatting, what happens if I drop that column using SQL?

kevin-reilly:
Couple more things I'm seeing after testing today.  All tests using [this map](https://mamut-st.carto-staging.com/builder/4adf7297-9bd3-48c0-907c-cedbad6f4e89) and the lightrailstations dataset unless otherwise noted.

## Normalizing
Using the Hispanic Population 2015 dataset:

If I **do not** select Normalize, I cannot select a boundary; the selector defaults to "No values."  The analysis completes with different values for most points but with some null values.

If I **do** select Normalize > Population, the selector allows me to choose a boundary, but the Analysis returns a null for each point.  I tried this with both US Census Blocks and Shoreline clipped Elementary School Districts.

If I **do** select Normalize > Per sq km, I cannot select a boundary; the selector defaults to "No values"  and the Analysis returns a null for each point.

## Boundaries
* These should be called "Geographic Units"
* If there is both a Shoreline clipped and a non-shoreline clipped geography, we should only present the option for Shoreline clipped and remove the "Shoreline clipped" prefix from the description.  This is confusing
* How are these sorted?  It should be largest to smallest but I'm not clear it works that way. 

## Timespans

The Timespan options are a bit strange for many datasets.  Looking at Hispanic Population again, I see:

* 2015-2015
* 2014-2014
* 2011-2015
* 2010-2013
* 2010-2010
* 2006-2010

The single year timespans should just show "2015" and not 2015-2015, for example.

## Region

Using the lightrailstation dataset which is only in the US, the Region selector only presents that as an option when I click the pulldown.  This is correct, but if there is only one selection, it should populate by default and not make me select it.

Using the Irish Counties dataset (which is only in Ireland...), when I select the Region menu, it only offers me "United Kingdom."  I believe this should also offer me EU datasets as well.

I created a dataset called Dummy which only includes two points, one in Texas and one in Ireland.  When I try to create an Analysis for this data set, the Region menu offers me United States, Canada, Spain, and European Union but not United Kingdom.  This seems wrong.

jorgesancha:
Hey @kevin-reilly 

I'm not seeing the same results that you are seeing regards normalization. I was able to select boundaries with that dataset and I always got results back, except with the first case where there were sometimes null values.

Not normalized
image

Normalized
image

Normalized > Per Square Km
image

Regarding everything else, while I agree with pretty much all of your feedback, I don't think we should use this ticket to solve what are fundamentally bad decisions on the design of the DO architecture or inconsistencies in the data. Bear in mind that what we show in the UI ("Shoreline clipped", etc... or the actual Timespans) is what we get from the DO API, and if we want to change something, we should change it there. Making adjustments on the client to cater for these things is not the right way to go.

Regarding "Boundaries" vs "Geographic Units", I would think twice before changing it; or if we do, we should probably change it everywhere else (in the DO, I mean). So far, everyone has referred to this as boundaries, so is quite a significant change.

If there are no more client side issues to fix, I would suggest we put this in production under feature flag and open it to team, and use this interface to:

- Better explore the DO data and potential inconsistencies
- Come up with 
  - a) a list of issues we want to prioritise on the DO (whether changes to the DO or the actual Data) before we announce it 
  - b) any potential changes to the UI to make things clear or better based on the feedback we get

What do you think? Should we get together to discuss briefly this afternoon?

cc/ @juanignaciosl @saleiva

xavijam:
> If that is not possible, if A_B no longer exists because of the change to A_C, than the formatting should return to the default.

Right now we don't have any logic when an analysis has changed and a "conditional" style was set. But we have it when an analysis is removed, trying to keep the previous style.

> How is this handled outside of Analysis? For example, if I have a column called Sales that I use for formatting, what happens if I drop that column using SQL?

If the user styles the layer by a column called "example":
- and drops that column, the map instantiation will fail (error will appear), and will need to change the style in order to visualize it again.
- and change the SQL only selecting, the `cartodb_id`, `the_geom`, `the_geom_webmercator`, it will "reset" the style and apply it the default one.

saleiva:
Hi guys, I might be arriving late to the party (:tada:) but remember we ideally should show the default time-span as selected on the input. If not, it looks like a needed step.

juanignaciosl:
I agree with jsancha: +1000 to feature-flag this and open issues with fixes and improvements.

ethervoid:
@saleiva right now, AFAIK, we don't have a way from DO that defines a time span as the default for a numerator measurement. We could add a new column for it in the metadata part of the database that points the last timespan as the default one but the closest we're is to order them and pick the first of it. 

The problem with this approach is that we have only strings for the time spans so it's not 100% reliable. I've added a [new issue](https://github.com/CartoDB/bigmetadata/issues/178) that would make us possible to filter/sort the time spans.

kevin-reilly:
I'm ok with opening this up to more _internal_ users via feature flag.  I know there are people looking forward to testing this so that will help expose any other issues.

I am a bit concerned that @jorgesancha was not able to reproduce my normalization issue.  I checked this exact thing a number of times on Friday to make sure I wasn't doing something wrong.  What would cause us to see different results?

I agree absolutely that Builder should present all datasets made available by the API.  The API is the correct place to address my comment about the 'shoreline' clipped datasets.  We should put that in the backlog.

ethervoid:
@kevin-reilly we're currently working on the clipped part [here](https://github.com/CartoDB/observatory-extension/issues/293)

Regarding the normalization issue, it's weird and concerning so we should take a look to it :\

nobuti:
Guys, this PR has grown too much, it's unmanageable, so we are creating single issues for every fix we need to make in order to improve the DO analysis. So, I'm merging this into the `data-observatory-analysis-ui` and opening PR from there.

Issue #11775, opened 2017-03-15
Enable Google Maps export for "Export as an image"


javierarce:
Currently the "Export as an image" feature doesn't export Google Maps basemaps. We should enable this so we always export geometries and the underneath basemap.

See [this issue](https://github.com/CartoDB/cartodb/issues/10822) for some hints on how to implement this feature using Google's Static Map API.

- [x] Export defaults / non-styled Google Maps basemaps.
- [x] Export styled Google Maps basemaps.
- [x] Allow panning the basemap when the image export widget is active.
- [x] Apply Google Maps limits in the export image form and warn the users.
  - [x] For regular users.
  - [x] For Google Maps Premium users.
- [x] Add attribution to the exported image.
- [x] Add CARTO logo to the exported image.
- [x] Remove menu in the export image form header. 

PR: https://github.com/CartoDB/cartodb/pull/11891

xavijam:
So, we have discussed it offline, but we should not be worried about Google's Premium Plan members because there are not that type of customers.
But in any case, we could use the current "fields" in order to add new params and make use of them wherever we want.

noguerol:
Talked offline, just updating the ticket:

· Only users with a "pay-as-you-go" license should use a digital signature
· Digital signatures could be appended if needed to the API key as a single text string, so changes are needed in the admin panel

Besides, there are image size limits: https://developers.google.com/maps/documentation/static-maps/usage-limits

· Free and pay-as-you-go: 640x640px
· Premium annual plan: 2048x2048px

javierarce:
Capture with our logo and the attribution. What do you think? 

![untitled map 1](https://cloud.githubusercontent.com/assets/4933/24395591/74424bf2-13a0-11e7-82c1-c42cdc8ed666.png)

xavijam:
I love it :dancer:!

javierarce:
In certain cases, when the license of the Google Maps is very long, we could end up having this issue:

![untitled map 1 17](https://cloud.githubusercontent.com/assets/4933/24397530/c3e0ecda-13a6-11e7-9718-23a468180097.png)

… so a solution is to check for the width of the attribution in the html and apply a tiny margin:

![untitled map 1 21](https://cloud.githubusercontent.com/assets/4933/24398258/26005476-13a9-11e7-9b1a-5862e80704f9.png)
[...]

PR #11891, opened 2017-03-29
Allows exporting static maps with Google Basemaps

javierarce:
This PR extends our current static map export and allows exporting google maps basemaps.

untitled map 1 44

### Work done in this PR: 

- [x] Export defaults / non-styled Google Maps basemaps.
- [x] Export styled Google Maps basemaps.
- [x] Allow panning the basemap when the image export widget is active.
- [x] Apply Google Maps limits in the export image form and warn the users.
  - [x] For regular users.
  - [x] For Google Maps Premium users.
- [x] Add attribution to the exported image (for regular basemaps and GMaps)
- [x] Add CARTO logo to the exported image.
- [x] Remove menu in the export image form header. 

### How to test this

- Enable the `static-maps` feature flag and then simply create a map (with or without geometries) and export it. 

- Enable the Google Maps API in your user.

- Change the basemap and export it again. Use GMaps or default ones. Check that the attribution and our logo appears in the image. 

- Check that with Google Basemaps the max dimensions of the image are these: 

  - **Default**: 640x640px
  - **Premium plan**: 2048x2048px (those that use a `&signature=` in the `google_maps_key`)


javierarce:
There's a couple of minor backend changes (to expose the google maps api key to the frontend) that probably should be reviewed by a backender.

javierarce:
No sure what's going on with the build, checking…

xavijam:
Can any @CartoDB/backend er check this one please?

lib/assets/core/javascripts/cartodb3/editor/export-image-pane/export-image-form-model.js
```diff
@@ -67,12 +77,22 @@ module.exports = Backbone.Model.extend({
   _validateDimension: function (value, formValues) {
```
xavijam: I miss some tests about this validation control.

lib/assets/core/javascripts/cartodb3/editor/export-image-pane/export-image-pane.js
```diff
@@ -95,45 +108,245 @@ module.exports = CoreView.extend({
...
-      zoom: vis.map.getZoom(), // eslint-disable-line
+  _getStaticMapURL: function () {
+    return this._vis.getStaticImageURL({
+      zoom: this._vis.map.getZoom(),
       width: this._exportImageFormModel.get('width'),
       height: this._exportImageFormModel.get('height'),
       lat: this._exportImageFormModel.get('lat'),
       lng: this._exportImageFormModel.get('lng'),
-      format: format
+      format: 'png' // we always use 'png' to allow merging layers
```
xavijam: so there is no option to choose png or jpg?

lib/assets/core/javascripts/cartodb3/editor/export-image-pane/export-image-pane.js
```diff
@@ -95,45 +108,245 @@ module.exports = CoreView.extend({
...
+      if (style) {
+        result.push('style=' + style);
+      }
+    });
+
+    return result.join('&');
   },
 
-  _getDataUri: function (url, format, callback) {
+  _exportImage: function () {
```
xavijam: Great function!

juanignaciosl:
@xavijam it looks ok now, isn't it?

xavijam:
One question @javierarce. Do we limit the resize of the "canvas" to the user limits or we only warn the user?

lib/assets/core/javascripts/cartodb3/editor/export-image-pane/export-image-pane.js
javierarce: Yes, but since we need to merge a data layer with a google basemap, we need to set PNG here (so we have transparency). Also, the final format is decided later in the code, when we merge the layer and the logo.

javierarce:
> One question @javierarce. Do we limit the resize of the "canvas" to the user limits or we only warn the user?

We just warn users. Resizing the canvas to the biggest dimension allowed is something we could do, but I prefer that they do the action.

lib/assets/core/javascripts/cartodb3/editor/export-image-pane/export-image-pane.js
javierarce: Thanks!

lib/assets/core/javascripts/cartodb3/editor/export-image-pane/export-image-form-model.js
javierarce: True, let me add a couple of tests.

xavijam:
> We just warn users. Resizing the canvas to the biggest dimension allowed is something we could do, but I prefer that they do the action.

👍

lib/assets/core/javascripts/cartodb3/editor/export-image-pane/export-image-pane.js
xavijam: Ahhhh 🐦  (pájaro).

xavijam:
@javierarce, could you deploy your last changes in `ded07`? Thanks cc @saleiva in case you want to test it.
* Remember, `static-maps` feature flag should be enabled in the user you are testing.

javierarce:
This is ready on ded07

xavijam:
Backend tests failing 😱 .

javitonino:
It's a random Jenkins connection failure. I relaunched them.

noguerol:
So, checking again, I found a new the following:

· We have some image size limit only in jpg. Trying to export a 3500x1000 png will work, but not in jpg. If this is a limitation of the format export itself then that should be reflected on the size text input fields limitation warning.

· When dragging the crop field there is a normally small but sometimes very large mismatch between the shadow box and the crop marks (being this ones the real export field), which may me confuse the user on which area is actually exported

captura de pantalla 2017-04-03 a las 11 48 41

captura de pantalla 2017-04-03 a las 11 51 39


· Copywriting for the export pane needs a review:

captura de pantalla 2017-04-03 a las 11 53 05

"You can also edit in width and height" sounds a but condescending and should be "Manually enter the image size"
"Change column names and type in table mode" makes no sense so we can put something like "Choose a lossless (png) or lossy (jpg) format"

javierarce:
> We have some image size limit only in jpg. Trying to export a 3500x1000 png will work, but not in jpg. If this is a limitation of the format export itself then that should be reflected on the size text input fields limitation warning.

Was that tested with a Google Maps basemap? The limit for regular baseamps are 8192x8192px in both cases.

javierarce:
> · When dragging the crop field there is a normally small but sometimes very large mismatch between the shadow box and the crop marks (being this ones the real export field), which may me confuse the user on which area is actually exported

I can't reproduce this :/ When does this happen?

javierarce:
> "You can also edit in width and height" sounds a but condescending and should be "Manually enter the image size"
> "Change column names and type in table mode" makes no sense so we can put something like "Choose a lossless (png) or lossy (jpg) format"

I like these changes. I'll update the texts.

noguerol:
This is happening also on our basemaps.

xavijam:
That's not related with the basemaps @noguerol, it is related with the UI. Let's find the pattern, if it is a consistent problem, let's fix it. If not, let's open a ticket in order to investigate it, because it is true we could improve the way we manage that "focus" area when the user changes its size.

javierarce:
> · When dragging the crop field there is a normally small but sometimes very large mismatch between the shadow box and the crop marks (being this ones the real export field), which may me confuse the user on which area is actually exported

Copy and that bug fixed too. I'll deploy again. 

PS: The problem with exporting a 3000x1000 image appeared on production, not in staging or dev.

noguerol:
Yes, i've not been able to replicate the size problem in staging :-/

xavijam:
That's good man! So, could we move this to deploy then?

javierarce:
> Yes, i've not been able to replicate the size problem in staging :-/

It's likely that we'll find that issue in production.

> That's good man! So, could we move this to deploy then?

Fine by me. What about you, @noguerol?

javierarce:
Update regarding the export size: I was able to export a 8000x1000 px image in production.

noguerol:
"Export button" is enabled even with out of the limit size, exporting an image only to the max size permitted and opening some inconsistencies in https://github.com/CartoDB/cartodb/issues/11775

noguerol:
Another thing: when exporting with a google basemap now works, but switching back to any other basemap and exporting it returns in an image with no basemap, only geometries cc @alonsogarciapablo

xavijam:
> Another thing: when exporting with a google basemap now works, but switching back to any other basemap and exporting it returns in an image with no basemap, only geometries

Noooo 😱

ivanmalagon:
Just noticed the same issues as Noguerol.

1. Exporting an image with our basemaps only paints geometry, but no basemaps.

1. When exporting a Google basemap image out of bounds, the portion of Google Maps respect the size limit but it doesn't paint well what falls outside it.

untitled map 2 1

javierarce:
> When exporting a Google basemap image out of bounds, the portion of Google Maps respect the size limit but it doesn't paint well what falls outside it.

Yes, that's a limitation of their service. Once I fix the disable button bug, it won't be possible to export that kind of images.

xavijam:
@noguerol found this problem -> https://github.com/CartoDB/cartodb/issues/11910, and we will take care of it in that issue. So, ready to be merged right guys? 💃 .

javierarce:
Yes, let's do it!

Issue #11894, opened 2017-03-30
Problem with bounding box in maps with google basemaps and widgets


xavijam:
If the user creates a map with basemap of Google, adds a widget, and then reloads the map, a "state" stored with wrong bounding box will be saved :(.

## STR
- Create a map.
- Change basemap to a Google basemaps (of course, you have to have it enabled 😬 ).
- Add a widget.
- Move the map wherever you want.
- Reload the map.
- Everything should work.
- Reload again the map.
- Center is moved to [0,0][0,0].

More info: https://github.com/CartoDB/support/issues/590#issuecomment-290451205
Support ticket for context: https://github.com/CartoDB/support/issues/590.

xavijam:
@noguerol @jorgesancha, this is important for migrations.
cc @alonsogarciapablo

jorgesancha:
Please check it out first and let's get a sense of effort. I assume this
affects everyone with widgets and google basemaps, right?

xavijam:
I've just seen this in CARTO.js https://github.com/CartoDB/cartodb.js/blob/v4/src/geo/gmaps/gmaps-map-view.js#L146

```js
  getBounds: function () {
    if (this._isReady) {
      var b = this._gmapsMap.getBounds();
      var sw = b.getSouthWest();
      var ne = b.getNorthEast();
      return [
        [sw.lat(), sw.lng()],
        [ne.lat(), ne.lng()]
      ];
    }
    return [ [0, 0], [0, 0] ];
  },
```

saleiva:
Are we fixing the other issue together with this? the one for saving zoom and x,y instead of bbox?

xavijam:
Blocking it until we have something we agree in DI, where the source of the problem is placed.

xavijam:
PR about DI -> https://github.com/CartoDB/deep-insights.js/pull/533

xavijam:
So, there are two PR:
- Builder: https://github.com/CartoDB/cartodb/pull/11901#discussion_r109371772
- DI: https://github.com/CartoDB/deep-insights.js/pull/533

If we check everything is 👍 , we should:
- Merge deep-insights PR into master.
- Change package to point to master and update dependencies in the Builder PR.
- Push it and wait for a 🍏 .

xavijam:
Deployed

PR #11901, opened 2017-03-31
Change way to listen when map or widgets have changed in deep-insights

xavijam:
Related ticket: https://github.com/CartoDB/cartodb/issues/11894

After a talk with @alonsogarciapablo, we have reviewed that there is no need to check if the map or the widgets collection have changed. With only listen to state changes is enough, and it avoids the current initialization problems we have.

## STT (Review the ticket too)
- Create a map of points.
- Add a category widget, for example.
- Move the map.
- Reload and check that the map is placed where it has to be.
- Change the basemap to GMaps.
- Move the map.
- Reload and check everything is ok.
- Reload again and check, no joke.
- Filter the widget by a category.
- Move the map.
- Reload and check that the widget is filtered and map is centered where it has to be*.

*Be careful with this [bug](https://github.com/CartoDB/cartodb.js/issues/1610) in Google Maps basemaps.


xavijam:
cc @alonsogarciapablo

lib/assets/core/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -109,10 +109,8 @@ var F = function (opts) {
   this._vis().on('reload', this._visReload, this);
   this._vis().on('change:error', this._visErrorChange, this);
 
-  var saveStateDefinition = this._saveStateDefinition.bind(this);
-  this._diDashboard._dashboard.widgets._widgetsCollection.bind('change', _.debounce(saveStateDefinition, 500), this);
-  this.visMap().on('change', _.debounce(saveStateDefinition, 500), this);
-
+  var saveStateDefinition = _.debounce(this._saveStateDefinition.bind(this), 500);
+  this._diDashboard.onStateChanged(saveStateDefinition, true);
```
nobuti: Not sure about the `true` parameter here. I think this could lead us to a double binding on changes in some scenarios. When the dashboard is created, we rely on `opts.share_urls` [here](https://github.com/CartoDB/deep-insights.js/blob/master/src/api/create-dashboard.js#L144-L146) to bind map and widgets changes. But we are calling it again with true by default.

nobuti: In any case, a test to cover it would be nice

alonsogarciapablo: 👎  to boolean parameters. They are really hard to understand from the perspective of the client. Alternatives:

- Create a new method (which might use `onStateChanged` internally) and give it a proper name (eg: `onStateChangedWhenSomethingIsTrue(...)`.
- Give that param a name by passing an options object (eg: `onStateChanged(saveStateDefinition, { somethingMeaningful: true })`).

Anyway, 👍  to removing the 🚂  of calls (`this._diDashboard._dashboard.widgets._widgetsCollection`).

xavijam: Well, take into account that, if the parameter is not true, callback will not be triggered.
https://github.com/CartoDB/deep-insights.js/blob/master/src/api/dashboard.js#L59-L77

xavijam: We could do something more explicit here like:

```js
var triggerCallbackWhenChange = true;
```

xavijam: Which test do you mean it is worth it @nobuti? I've already added some, but I'm open to add more if they make sense 🙏 .

nobuti: Nevermind, I was confused, the thing is we need multiple bindings for several callbacks.

xavijam: > Nevermind, I was confused, the thing is we need multiple bindings for several callbacks.

Yep

alonsogarciapablo: From what I can see, that method is doing at least two things:

```
  onStateChanged: function (callback, shareURLs) {
    this._dashboard.vis.once('dataviewsFetched', function () {
      this._dashboard.widgets._widgetsCollection.initialState();
      this._dashboard.widgets._widgetsCollection.each(function (m) {
        m.applyInitialState();
      }, this);
      shareURLs === true && this._bindChange(callback);
    }, this);
  },
```

There're several smells here (IMO):

1. Method is doing two things:

- Adding a bind to the `dataviewsFetched` event.
- (Maybe) binding the given callback to "state" (widgets + map) changes.

The fact that we have two different responsibilities is probably the reason why we're using a boolean parameter in the first place. For this use case, we probably don't want to bind `dataviewsFetched`. We'd only want to listen to state changes, right?

We could split the method in two and, wherever we're using this already (https://github.com/CartoDB/deep-insights.js/blob/7f106bf59c969999fd635789b100b2a311a26d2b/src/api/create-dashboard.js#L144-L146), do something like:

```
dashboard.onWidgetsFetched(...);
if (sharedURLs) {
  dashboard.onStateChange(...);
}
```

We can then just use `dashboard.onStateChange(...);` in the new code... 👍  or 👎 ?

2. Both `onStateChanged` and `_bindChange` are only using stuff from `this._dashboard`,  so this code probably belongs to that class (whatever it is). We could probably need to keep methods in this class (cause it's part of the API), but they can just act as proxies to the methods in `this._dashboard`.

xavijam: Hum... It is true we should not "trigger" any state change until dataviews have been fetched. In Builder, we could bind that state event and be launched before it is "everything initialized" and happened the same problem we are trying to fix. So, IMHO, we should not get rid of it.

xavijam: But it doesn't mean (^^^) I like how it is implemented. Are you proposing this thing @alonsogarciapablo? :
```js
var saveStateDefinition = _.debounce(this. _saveStateDefinition.bind(this), 500);
var onStateChange = dashboard.onStateChange(saveStateDefinition);
dashboard.onWidgetsFetched(onStateChange);
```

alonsogarciapablo: I didn't explain myself correctly, sorry. Yes, we might want to wait until dataviews have been fetched, but what about:

```
      this._dashboard.widgets._widgetsCollection.initialState();
      this._dashboard.widgets._widgetsCollection.each(function (m) {
        m.applyInitialState();
      }, this);
```

Do we want to call those in this situation? What I'm saying is that those should probably be extracted to a separate method.

xavijam: Yeah... let me elaborate a PR in deep-insights. I'll ping you both there 💋 .

alonsogarciapablo: Basically, this is what I'm suggesting:

```
// BEFORE

function doAAndMaybeB(doB) {
  // doA
  if (doB) {
    // doB
  }
};

// client
doAAndMaybeB(doB);

// AFTER

function doA {
  // doA
}

function doAandB {
  doA();
  // doB
}

// client
if (doB) {
  doAandB();
} else {
  doA();
}
```

xavijam: Not sure if Builder should know a lot about if it has to request one thing or other... not sure.

lib/assets/core/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -109,10 +109,8 @@ var F = function (opts) {
   this._vis().on('reload', this._visReload, this);
   this._vis().on('change:error', this._visErrorChange, this);
 
-  var saveStateDefinition = this._saveStateDefinition.bind(this);
-  this._diDashboard._dashboard.widgets._widgetsCollection.bind('change', _.debounce(saveStateDefinition, 500), this);
-  this.visMap().on('change', _.debounce(saveStateDefinition, 500), this);
-
+  var saveStateDefinition = _.debounce(this._saveStateDefinition.bind(this), 500);
+  this._diDashboard.onStateChanged(saveStateDefinition);
```
alonsogarciapablo: 💖

PR #11922, opened 2017-04-05
11476 default zoom is 3

javitonino:
Fix for #11476 

Refactors `Carto::BoundingBoxHelper`. It also changes the behaviour when the bounds cannot be calculated (empty dataset) so it returns `nil` instead of the default bounds. This allows the map model to be able to set default center and zoom (instead of setting the center and zoom calculated from the default bounds).

About the refactor:
- Reunites `BoundingBoxHelper` and `Carto::BoundingBoxHelper`. Added an `execute_in_user_database` function to user db services, that offers the same interface but runs the query with the corresponding ORM.
- Splits `BoundingBoxHelper` into two (as a bonus, it removes it from `app/helpers`):
  - `BoundingBoxUtils`: static functions related to bounding box calculations, such as generating the SQL for a bounding box query.
  - `BoundingBoxService`: functions tied to an user database, to calculate the bounding box of a table. It uses the `Utils` package

**Acceptance**
- [x] Create an empty map. Zoom should be 3
- [x] Create a map from a dataset. Zoom should be something reasonable, close to the data bounding box


app/models/carto/map.rb
```diff
@@ -30,10 +30,10 @@ class Carto::Map < ActiveRecord::Base
   # So for now, we are just treating them as strings (see the .to_s in the constant below), but this could be improved
   DEFAULT_OPTIONS = {
     zoom:            3,
-    bounding_box_sw: [Carto::BoundingBoxHelper::DEFAULT_BOUNDS[:minlat],
-                      Carto::BoundingBoxHelper::DEFAULT_BOUNDS[:minlon]].to_s,
-    bounding_box_ne: [Carto::BoundingBoxHelper::DEFAULT_BOUNDS[:maxlat],
-                      Carto::BoundingBoxHelper::DEFAULT_BOUNDS[:maxlon]].to_s,
+    bounding_box_sw: [Carto::BoundingBoxHelper::DEFAULT_BOUNDS[:miny],
```
juanignaciosl: Currently, every access to `(Carto::)BoundingBoxHelper::DEFAULT_BOUNDS` is done with min/max - lat/lon. How come this is changed to min/max x/y?

app/models/carto/map.rb
```diff
@@ -173,8 +173,11 @@ def process_privacy_in(layer)
   end
 
   def recalculate_bounds!
-    set_boundaries(get_map_bounds)
-    save
+    bounds = get_map_bounds
```
juanignaciosl: Previously boundaries were always reset if this method is called. Now it's only if you get the bounds.

app/controllers/api/json/tables_controller.rb
```diff
@@ -97,7 +96,7 @@ def update
           @stats_aggregator.timing('georeference') do
             @table.georeference_from!(:latitude_column => latitude_column, :longitude_column => longitude_column)
           end
-          BoundingBoxHelper.update_visualizations_bbox(@table)
+          @table.update_bounding_box
```
gfiorav: Have you considered this to be a hook in the update of the model? Is it performance you're worried about?

app/models/carto/map.rb
```diff
@@ -30,10 +30,10 @@ class Carto::Map < ActiveRecord::Base
   # So for now, we are just treating them as strings (see the .to_s in the constant below), but this could be improved
   DEFAULT_OPTIONS = {
     zoom:            3,
-    bounding_box_sw: [Carto::BoundingBoxHelper::DEFAULT_BOUNDS[:minlat],
-                      Carto::BoundingBoxHelper::DEFAULT_BOUNDS[:minlon]].to_s,
-    bounding_box_ne: [Carto::BoundingBoxHelper::DEFAULT_BOUNDS[:maxlat],
-                      Carto::BoundingBoxHelper::DEFAULT_BOUNDS[:maxlon]].to_s,
+    bounding_box_sw: [Carto::BoundingBoxUtils::DEFAULT_BOUNDS[:miny],
```
gfiorav: Why did you go for x,y instead of lat,lon?

app/models/carto/user_db_service.rb
```diff
@@ -27,5 +27,17 @@ def self.build_search_path(user_schema, quote_user_schema = true)
     def public_user_roles
       @user.organization_user? ? [CartoDB::PUBLIC_DB_USER, @user.database_public_username] : [CartoDB::PUBLIC_DB_USER]
     end
+
+    # Execute a query in the user database
```
gfiorav: Is this doc something we're doing now? Why isn't it on every other method you've added? (curious)

app/models/carto/user_db_service.rb
```diff
@@ -27,5 +27,17 @@ def self.build_search_path(user_schema, quote_user_schema = true)
...
+    # @param [String] query, using $1, $2 ... for placeholders
+    # @param ... values for the placeholders
+    # @return [Array<Hash<String, Value>>] Something ({ActiveRecord::Result} in this case) that behaves like an Array
+    #                                      of Hashes that map column name (as string) to value
+    # @raise [PG::Error] if the query fails
+    def execute_in_user_database(query, *binds)
+      @user.in_database.exec_query(query, 'ExecuteUserDb', binds.map { |v| [nil, v] })
```
gfiorav: What's this `ExecuteUserDb` thing for?

app/models/carto/user_db_service.rb
```diff
@@ -27,5 +27,17 @@ def self.build_search_path(user_schema, quote_user_schema = true)
...
+    rescue ActiveRecord::StatementInvalid >= exception
+      raise exception.cause
```
gfiorav: Is this always populated? I know that in some cases AR exceptions just leave `.record` blank, that's why I'm asking. Also, why `.cause` and not just raise the exception itself or rails `.message`?

app/models/carto/user_db_service.rb
```diff
@@ -27,5 +27,17 @@ def self.build_search_path(user_schema, quote_user_schema = true)
...
+
+    # Execute a query in the user database
+    # @param [String] query, using $1, $2 ... for placeholders
+    # @param ... values for the placeholders
+    # @return [Array<Hash<String, Value>>] Something ({ActiveRecord::Result} in this case) that behaves like an Array
+    #                                      of Hashes that map column name (as string) to value
+    # @raise [PG::Error] if the query fails
+    def execute_in_user_database(query, *binds)
+      @user.in_database.exec_query(query, 'ExecuteUserDb', binds.map { |v| [nil, v] })
+    rescue ActiveRecord::StatementInvalid >= exception
```
gfiorav: I think you meant 

```ruby
rescue ActiveRecord::StatementInvalid => exception
```

Or maybe `>=` is something I don't know about 🙈

app/models/common/map_common.rb
```diff
@@ -1,16 +1,29 @@
...
     bounds = get_map_bounds
-    set_boundaries(bounds)
-    recenter_using_bounds(bounds)
-    recalculate_zoom(bounds)
+    if bounds
+      set_boundaries(bounds)
+      recenter_using_bounds(bounds)
+      recalculate_zoom(bounds)
+    else
+      set_boundaries(Carto::BoundingBoxUtils::DEFAULT_BOUNDS)
```
gfiorav: proposal! define `set_boundaries` as:

```ruby
def set_boundaries(bounds: Carto::BoundingBoxUtils::DEFAULT_BOUNDS)
  ...
end
```

Or

```ruby
def set_boundaries(bounds = Carto::BoundingBoxUtils::DEFAULT_BOUNDS)
  ...
end
```

This way you don't have to define the defaults out of the scope of interest, and you can simply write:

```ruby
if bounds
  set_boundaries(bounds: bounds)
  ...
else
  set_boundaries
  ...
end
```

or 

```ruby
if bounds
  set_boundaries(bounds)
  ...
else
  set_boundaries
  ...
end
```
respectively.


Also, you can do the same for `recenter_using_bounds` and `recalculate_zoom`.

Moreover, since the three methods use the same param, you could maybe create a single method that does these three things.

app/models/table.rb
```diff
@@ -1246,6 +1246,15 @@ def table_visualization
     @user_table.table_visualization
   end
 
+  def update_bounding_box
+    update_table_geom_pg_stats
+    bounds = Carto::BoundingBoxService.new(owner).get_table_bounds(name) || Carto::BoundingBoxUtils::DEFAULT_BOUNDS
```
gfiorav: Proposal:

It feels weird to pass a name to a bounding box thingy. How about passing a table? Also, maybe you can have the thingy decide internally whether it will return default (it also seems weird to access that from here). All in all:

```ruby
bounds = Carto::BoundingBoxService.new(owner).bounds(table)
```

or even better

```ruby
bounds = Carto::BoundinBoxService.new(table).bounds
```

Since 1 bounding box amounts to 1 table.

app/queries/carto/visualization_query_builder.rb
```diff
@@ -312,7 +312,7 @@ def build
     end
 
     if @bounding_box
-      bbox_sql = BoundingBoxHelper.to_polygon(@bounding_box[:minx], @bounding_box[:miny], @bounding_box[:maxx], @bounding_box[:maxy])
+      bbox_sql = Carto::BoundingBoxUtils.to_polygon(@bounding_box[:minx], @bounding_box[:miny], @bounding_box[:maxx], @bounding_box[:maxy])
```
gfiorav: ⏰

lib/carto/bounding_box_service.rb
```diff
@@ -0,0 +1,75 @@
+# encoding: utf-8
+require_relative '../../lib/carto/table_utils'
+require_dependency 'carto/bounding_box_utils'
+
+class Carto::BoundingBoxService
+  include Carto::TableUtils
+
+  def initialize(user)
+    @user = user
+  end
```
gfiorav: As I commented elsewhere, but maybe here it's more relevant:

1 bounding box <-> 1 table.

It feels more natural to me to have this be:

```ruby
def initialize(table)
  @table = table
end

def bounds
  ...
end
```

The interface seems more reasonable:

```ruby
Carto::BoundingBoxService.new(table).bounds
```

Also, you don't have to pass table around that way. It overall feels more natural to me. Let me know your thoughts.

lib/carto/bounding_box_utils.rb
```diff
@@ -0,0 +1,45 @@
...
+    maxy: 90
+  }.freeze
+
+  def self.bound_for(value, minimum, maximum)
+    [[value, DEFAULT_BOUNDS.fetch(minimum)].max, DEFAULT_BOUNDS.fetch(maximum)].min
+  end
+
+  def self.to_polygon(minx, miny, maxx, maxy)
+    return nil unless check_bounds_for(minx, miny) && check_bounds_for(maxx, maxy)
+    %{ST_Transform(ST_Envelope('SRID=4326;POLYGON((#{minx} #{miny}, #{minx} #{maxy}, #{maxx} #{maxy}, #{maxx} #{miny}, #{minx} #{miny}))'::geometry), 3857)}
```
gfiorav: ⏰

lib/carto/bounding_box_utils.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def self.to_point(x, y)
+    return nil unless check_bounds_for(x, y)
+    %[ST_GeomFromText('POINT(#{x} #{y})',3857)]
+  end
+
+  def self.parse_bbox_parameters(bounding_box)
+    bbox_coords = bounding_box.split(',').map { |coord| Float(coord) rescue nil }.compact
```
gfiorav: I agree!

app/controllers/api/json/tables_controller.rb
javitonino: Meh, I am not sure what that part of the code does. According to Mario, this (`visualizations.bbox` field) should not even be in use currently. But I prefer not to change the calls in this case, as I don't think it is called in all updates to the model, and it might be important. TL;DR: 🐔

app/models/carto/map.rb
javitonino: I standardized a mixture of `maxx` and `maxlon` usages, and I did go for `x, y` since it is used in a lot more places, including user-facing apis (in parameters and presenters) which we probably cannot change. So this way, we use the same thing everywhere, even if I would have preferred to go for lat/lon.

app/models/carto/user_db_service.rb
javitonino: We are doing this only in the Gears API. But not for internal code. I just felt like it, because there are some non-trivial details in the usage (how binds are used) that I wanted to document. So I just followed the same style of docs as we use elsewhere.

app/models/carto/user_db_service.rb
javitonino: Just a logging tag, it shows up in AR console logs. e.g: it would replace the `Carto::User Load` here:
```
Carto::User Load (54.4ms)  SELECT "users".* FROM "users" LIMIT 1
^^^^^^^^^^^^^^^^
```

app/models/carto/user_db_service.rb
javitonino: Oops! 👍

app/models/common/map_common.rb
javitonino: There is a bit of a trick in here with `recalculate_zoom`, as it's default is different of what the result of calculating the zoom on the default bounds would be.

I could do something like:
```
def recalculate_zoom(bounds: nil)
  if bounds
    ...
  else
    self.zoom = 3
  end
end
```

but I think that this is actually worse.

With that in mind, I think having a function that groups all three loses a bit of meaning, since it would only be used once. OTOH, it could help readability (something like `set_default_map_bounds` is better than the current approach).

I'll give it a 🤔, open to more suggestions!

app/models/carto/user_db_service.rb
javitonino: It should always be populated.

In this case, the exception is `ActiveRecord::StatementInvalid` and the cause (nested exception) is `PG::Error`. For the Sequel case, the exception is `Sequel::DatabaseException`, but the cause is the same as before (`PG::Error`).

Basically, AR and Sequel are both wrapping the underlying database exception in their own wrappers, in order for the caller code to avoid working with DB specific errors. In our case, we'd rather do that to be able to catch the same exception in caller code, independent of the model. 

This is useful for making the same code compatible with both models, even in exception handling. See for example: https://github.com/CartoDB/cartodb/pull/11922/files#diff-d0b2c8542eb925ab903c3fdd9226d568R58

lib/carto/bounding_box_service.rb
javitonino: Yes, it makes some sense now that this only works for tables... 🤔 

I am not sure we should pass the entire model, as we are just using the name. Also, I think we still need to pass the user anyway (to make sure it is the right class for ORM + user table refactor stuff). Uhm, I'll try it and run the tests and see if it works.

javitonino: Tested, it doesn't work. Due to how the order of callbacks works, we don't have the persisted table model when calling this (this is called as part of the `create` hooks of `UserTable`). Still, I will move `table_name` to an instance variable of the service, it should help reduce passing it around internally.

javitonino:
I think I covered all your requests, @gfiorav. Care to take a second look?

lib/carto/bounding_box_service.rb
```diff
@@ -0,0 +1,76 @@
...
+
+class Carto::BoundingBoxService
+  include Carto::TableUtils
+
+  def initialize(user, table_name)
+    @user = user
+    @table_name = table_name
+  end
+
+  def get_bounds
```
gfiorav: Minor, but why the `get_`? I mean, it's not like we do `user.get_username`. Don't fight rubiness or some person in the future will have to do the same thing we did for `is_available?`-> `available?` :D

lib/carto/bounding_box_service.rb
```diff
@@ -0,0 +1,76 @@
...
+  # Postgis stats-based calculation of bounds. Much faster but not always present, so needs a fallback
+  def current_bbox_using_stats
+    # Less ugly (for tests at least) than letting an exception generate not having any table name
+    return nil unless @table_name.present?
+
+    # (lon,lat) as comes from postgis
+    JSON.parse(@user.db_service.execute_in_user_database(%{
+      SELECT _postgis_stats ('#{@table_name}', 'the_geom');
+    }).first['_postgis_stats'])['extent'].symbolize_keys
+  rescue => e
```
gfiorav: Can we check for a specific exception? this catch-all's are discouraged and I can see why!

Maybe there's a direct way to check for stats? I like the exception driven approach, but not when we have to catch all and parse messages :(

app/models/common/map_common.rb
```diff
@@ -1,18 +1,39 @@
...
   def set_default_boundaries!
     bounds = get_map_bounds
-    set_boundaries(bounds)
-    recenter_using_bounds(bounds)
-    recalculate_zoom(bounds)
+    if bounds
+      set_viewport_from_bounds(bounds)
+    else
+      set_default_viewport
+    end
```
gfiorav: Just to be obnoxious ( :trollface: ), how about:

```ruby
set_viewport(bounds)
```

or the likes 😁 

You can just asume that null is the default inside the method right?


Also, really nice chance to:

```ruby
bounds ? set_viewport_form_bounds(bounds) : set_default_viewport
```

Minor, of course!

lib/carto/bounding_box_service.rb
javitonino: Sure, that's why I made the user service function always raise the same type of exception 👍

javierarce:
Acceptance is 👍 , just a comment regarding this: 

> Zoom should be something reasonable, close to the data bounding box

screen shot 2017-04-10 at 10 55 14
screen shot 2017-04-10 at 10 55 01

I think we could probably add another zoom level in this case.

javitonino:
The hard thing here is that the zoom level is calculated from backend, where we have no idea of the browser window size, so we just guesstimate it.

But yeah, maybe an extra zoom level would be cool as monitor sizes increase 🤔

PR #11936, opened 2017-04-06
Map samples squash

vishalya:


vishalya:
It was sent by accident. Not really meant for the upstream.

Issue #11812, opened 2017-03-21
Wrong value shown when georeferencing on georeference analysis


saleiva:
### Steps to Reproduce

1. Add a layer
2. Georeference it by street adress
3. Apply

### Current Result
![screen shot 2017-03-21 at 17 44 39](https://cloud.githubusercontent.com/assets/132149/24158976/2ab363d2-0e5e-11e7-8581-c28085eddb22.png)

TYPE Dropdown is showing a wrong value (and growing more than supposed to do).

 
After going to the layer list, and entering back to the analysis, it looks correct

xavijam:
Have we fixed this one in the new street geocoding PR @matallo ?

matallo:
yes!

xavijam:
Hey, it should be fixed thanks to the thing mentioned before. Closing it and if we find it again, let's open a new one 💋 .

Issue #11662, opened 2017-02-28
Georeference by street address is lacking custom fields addition


saleiva:
### Context
Previously, users could add extra custom fields for specifying the rules of georeferencation by street address. It's not longer there!

### Steps to Reproduce

1. Create a layer with this file [Relacion Camaras.csv.zip](https://github.com/CartoDB/cartodb/files/806596/Relacion.Camaras.csv.zip)
2. Georeference by street address.

### Current Result
It's impossible since number and street are in different fields.

### Expected result
As in editor, you should be able to add fields to the form right after street one

xavijam:
Please, add design link when it is ready 💋 .

urbanphes:
These are the visuals when the option "Street Addresses" is selected

![analysis - georeference 04](https://cloud.githubusercontent.com/assets/14546701/23413664/9ea2e776-fdd9-11e6-82fb-40eba175646d.png)
![analysis - georeference 05](https://cloud.githubusercontent.com/assets/14546701/23413668/a2f5d4f0-fdd9-11e6-955e-6a2a610d1419.png)
![analysis - georeference 06](https://cloud.githubusercontent.com/assets/14546701/23413672/a75858d8-fdd9-11e6-85fa-760556028477.png)

ivanmalagon:
Is this supported / implemented in Camshaft?

According to the last [camshaft reference](https://github.com/CartoDB/camshaft/blob/master/reference/versions/0.47.0/reference.json#L543) there are no additional columns available for this analysis. They are the same as shown in the UI.

I don't know if we can send additional info to `street_address_column` encoded somehow in the string or not.

Summoning @rochoa to answer this.

ivanmalagon:
In editor we were doing this by calling to `geocodings` API filling `formatter` field with comma-separated values.

```
endpoint: POST /api/v1/geocodings/

payload:
{
  "kind":"high-resolution",
  "formatter":"{calle}, {numero_de_finca}",
  "table_name":"relacion_camaras",
  "state":""
}

```

javisantana:
Hey designers, one question. The typical usage for this is "{street_adress_column}, 28010 (Madrid)" rather than a column concatenation.

Don't you think we should have a more open way to do it using templates? cc @noguerol

saleiva:
Although I love the idea (and I'm tempted to add it anyway) I think is risky because:
[...]

PR #11939, opened 2017-04-07
advance georefence street address

matallo:
re: #11662 https://github.com/CartoDB/cartodb/issues/11812

many changes here:

- added Toggle editor (based in Radius)
- added Placeholder add-on to Codemirror
- added `autocompleteChars`, `autocompleteSuffix`, `autocompletePrefix `, `autocompleteTriggers` (these two come usually together) options to Codemirror
- added mustache mode to Codemirror
- extend custom Textarea editor from custom Text editor
- added `editor` option to custom Textarea editor, which will turn it into a Codemirror editor passing the option `editor: true` this won't have any effect for current Textarea editors
- added placeholder option to Textarea custom editor
- made `querySchemaModel`, and `layerDefinitionModel` optional for hints
- added and extended the necessary tests for the changes above

**acceptance**

- current codemirror hints should keep working as usual
- new textarea codemirror should appear when selecting advance in the toggle of georeference analysis
- new textarea codemirror hints should appear
- normal georeference analysis should work (d'oh!)
- advance georeference analysis should work (d'oh!)

how to test the codemirror hints:

after writing "{{", or first two characters of columns, it should suggest column names, and autocomplete "}}" suffix, as well as "{{" prefix if it has not been typed

codemirror hints should work on demand with "Ctrl + Space" shortcut, with the same behaviour

they should work normally in previous codemirror hints such sql in data tab

**STAGING**: rui servers


xavijam:
cc @noguerol ^^^

urbanphes:
Some things;

- Less left and right margin interior on the darker input. We can test with 4px to each instead of 14px.
- Change the info by: “Use columns or free text to compose your address schema”
- Change the text color to difference the text inside {{}} and the text outside. Maybe we can use the white for the text outside like a “{“
- The autocomplete should appear when you start typing columns even if you don’t put keys.
- When there is an error, the input doesn't mark as red.

matallo:
- [x] Less left and right margin interior on the darker input. We can test with 4px to each instead of 14px.
- [x] Change the info by: “Use columns or free text to compose your address schema”
- [ ] Change the text color to difference the text inside {{}} and the text outside. Maybe we can use the white for the text outside like a “{“
- [x] The autocomplete should appear when you start typing columns even if you don’t put keys.
- [x] When there is an error, the input doesn't mark as red.

Changes made, the only pending point is the syntax color which I will experiment a bit, as we need to make our custom syntax coloring, this is deployed again in the ruis, and tests are passing 🤞

saleiva:
@matallo @urbanphes @noguerol should we do another pass then? The point that is missing could be a nice-to-have at some point...

matallo:
it's ready to test again 👍

I'm doing some quick experiment now to see if I can get the syntax coloring working, in any case

lib/assets/core/javascripts/cartodb3/components/code-mirror/code-mirror-view.js
```diff
@@ -27,16 +29,22 @@ module.exports = CoreView.extend({
   initialize: function (opts) {
     if (!opts) throw new Error('options for codemirror are required.');
     if (!opts.model) throw new Error('Model for codemirror is required.');
-    if (opts.model.get('content') === undefined) throw new Error('Content property for codemirror is required.');
-    if (!opts.tip) throw new Error('tip message is required');
+    if (opts.model.get('content') === void 0 &&
```
xavijam: I'd use the nice and great `var checkAndBuildOpts = require('../../helpers/required-opts');` here. What do you think?

lib/assets/core/javascripts/cartodb3/components/code-mirror/code-mirror-view.js
```diff
@@ -95,44 +115,92 @@ module.exports = CoreView.extend({
...
+  _completeIfAfterOpeningCurly: function (cm) {
+    return this._completeAfter(cm, function () {
+      var cur = cm.getCursor();
+
+      return cm.getRange(CodeMirror.Pos(cur.line, cur.ch - 1), cur) === '{';
+    });
+  },
+
+  _completeIfAfterCtrlSpace: function (cm) {
+    var self = this;
```
xavijam: Why do we need the `self` here?

lib/assets/core/javascripts/cartodb3/components/code-mirror/code-mirror-view.js
```diff
@@ -27,16 +29,22 @@ module.exports = CoreView.extend({
   initialize: function (opts) {
     if (!opts) throw new Error('options for codemirror are required.');
     if (!opts.model) throw new Error('Model for codemirror is required.');
-    if (opts.model.get('content') === undefined) throw new Error('Content property for codemirror is required.');
-    if (!opts.tip) throw new Error('tip message is required');
+    if (opts.model.get('content') === void 0 &&
+        opts.placeholder === void 0) throw new Error('Content property or placeholder for codemirror is required.');
+    if (opts.tip === void 0) throw new Error('tip message is required');
 
+    this._autocompleteChars = opts.autocompleteChars || 3;
```
xavijam: This 3 number should be a "static" var at the top. But in any case, if this is a view, you could create the `options` variable inside the Class and specify whatever you want (same for mode).

lib/assets/core/javascripts/cartodb3/components/code-mirror/code-mirror-view.js
```diff
@@ -95,44 +115,92 @@ module.exports = CoreView.extend({
...
           }
-        }
+        }, opts);
       }, 150);
     }
   },
 
-  _showAutocomplete: function (cm) {
+  _completeAfter: function (cm, pred, opts) {
+    var self = this;
```
xavijam: Same as previously said with `self`.

lib/assets/core/javascripts/cartodb3/components/form-components/editors/radio.tpl
```diff
@@ -3,7 +3,7 @@
     <input type="radio" class="CDB-Radio u-iBlock<% if (item.className) { %> <%- item.className %> <% } %>"
     name="<%- item.name %>" value="<%- item.value %>" id="<%- item.id %>" 
         <% if (item.selected) { %>
-          selected="selected"
+          checked="checked"
```
xavijam: Won't it make fail or work weird for the rest of the Radio components?

lib/assets/core/javascripts/cartodb3/components/form-components/editors/textarea.js
```diff
@@ -1,15 +1,38 @@
...
+
+Backbone.Form.editors.TextArea = Backbone.Form.editors.Text.extend({
+  tagName: 'textarea',
 
-Backbone.Form.editors.TextArea = Backbone.Form.editors.TextArea.extend({
   className: 'CDB-Textarea',
 
   initialize: function (opts) {
     Backbone.Form.editors.Base.prototype._setOptions.call(this, opts); // Options
-    this.constructor.__super__.initialize.apply(this, arguments);
```
xavijam: Hey, I feel we should not remove this line, should we?

lib/assets/core/javascripts/cartodb3/components/form-components/editors/textarea.js
```diff
@@ -1,15 +1,38 @@
...
+
+    this.value = this.model.get(this.options.keyAttr);
   },
 
   render: function () {
     this.setValue(this.value);
+
+    if (this.options.editor) {
```
xavijam: Hum... 🤔  about if we should add this behaviour to the default textarea in the Backbone.Forms...

saleiva:
taking it then

xavijam:
Hey @saleiva, not sure if it is deployed in any staging with last changes, @matallo?

matallo:
it is!

saleiva:
Works nice but I found a few things:
- [ ] Syntax errors are kept. Try with {{calle}, When applying it removes some text and do the analysis. We should validate on the back and give an error, or keep the text as it is.
- [ ] After applying the analysis, if you go to the analysis and modify only the text on the dark box, the button doesn't get activated back. 
- [ ] If changing the text on advanced, clicking back on normal mode, and going back to advanced, the template is not being saved.
- [ ] If being in the Street Address form, and clicking on the dropdown for selecting Admin regions, the form doesn't get updated.

lib/assets/core/javascripts/cartodb3/components/code-mirror/code-mirror-view.js
matallo: this is a tricky one as we have some required opts indeed as the `model` but some other cases such either `content` or `placeholder` is required, and finally some optional opts like prefix/suffix, ...

I'm not really sure, thoughts @nobuti ?

nobuti: Agree, we are mixing options with attributes, I don't think we can use the helper for all.

xavijam: _flojos_

matallo: haha, actually in this PR I replaced the `required_opts` for `optional_opts` (as redundant as it might seem) https://github.com/CartoDB/cartodb/pull/11939/files#diff-c262f4f35ff0590e20b9390bbc8316e3R6

so maybe it's the time to set a proper name for this where we can have both the required and the optional _whatever_ with a different loop for each one

lib/assets/core/javascripts/cartodb3/components/code-mirror/code-mirror-view.js
matallo: 👍 also moved the default attributes for the model in that `options:`

lib/assets/core/javascripts/cartodb3/components/code-mirror/code-mirror-view.js
matallo: 👍

lib/assets/core/javascripts/cartodb3/components/code-mirror/code-mirror-view.js
matallo: 👍

lib/assets/core/javascripts/cartodb3/components/form-components/editors/radio.tpl
matallo: from MDN https://developer.mozilla.org/en/docs/Web/HTML/Element/input

> Use the checked attribute to indicate whether this item is selected by default

checked (pun intended) in several other places and it doesn't seem to have further consequences

lib/assets/core/javascripts/cartodb3/components/form-components/editors/textarea.js
matallo: now we're extending from custom `Text` editor instead of Backbone forms `Textarea` editor, where explicitly prevents this https://github.com/powmedia/backbone-forms/blob/master/src/editors/textarea.js#L9

lib/assets/core/javascripts/cartodb3/components/form-components/editors/textarea.js
matallo: how to better proceed? Having a new editor extending from `Textarea`? How that new editor will be called?

lib/assets/core/javascripts/cartodb3/components/form-components/editors/radio.tpl
xavijam: 😱

lib/assets/core/javascripts/cartodb3/components/form-components/editors/textarea.js
xavijam: I'm not sure, it depends the number of uses. Will we need more features for this textarea? or more types? I feel we could leave it as it is for the moment, but if it grows, create it in a separate it editor.

lib/assets/core/javascripts/cartodb3/components/form-components/editors/textarea.js
xavijam: Well, not sure about that, but you have tested it so 👍 .

matallo:
- [ ]  Syntax errors are kept. Try with {{calle}, When applying it removes some text and do the analysis. We should validate on the back and give an error, or keep the text as it is.

I'm not sure how to proceed with the syntax error :/ anyway, there was a bug and now the string is being sent as is, not removed 👍 

- [x] After applying the analysis, if you go to the analysis and modify only the text on the dark box, the button doesn't get activated back.

fixed some stuff that prevented changes in the codemirror to be populated to the form

- [x] If being in the Street Address form, and clicking on the dropdown for selecting Admin regions, the form doesn't get updated.

🙈 well spotted, fixed now

- [ ] If changing the text on advanced, clicking back on normal mode, and going back to advanced, the template is not being saved.

this one is not easy, but I can further look into it, I can share which is the process now, though: when changing context I'm copying the value in the `street_address_column` (select) to `street_address_template` (textarea) adding the wrap `{{*}}`, if I change back as the values won't coincide the select has not any option selected, and if I change back again, as the value is copied, the value copied is empty

as said, I'll take a look if I can workaround this somehow

matallo:
after discussing with buti, we cannot see a straight-forward solution for keeping the `street_address_template` value, also none of the forms work this way now, i.e.:

1. select the columns for longlat
1. change type
1. go back to longlat

the columns are not preserved

saleiva:
@matallo ok, I'd keep that as an enhancement for the time being
cc/ @noguerol

saleiva:
About the first point, can we validate the syntax of the expression? I talked to @rochoa about that

matallo:
Yeah, we'd probably have to think about it globally, for every form.

rochoa:
@saleiva, @matallo 

> About the first point, can we validate the syntax of the expression? I talked to @rochoa about that

The backend will keep the string as it is when the template is something like `{{calle} {{number}}, {{city}}`, replacing `{{number}}` and `{{city}}`, but keeping `{{calle}` like it is.

app/assets/stylesheets/editor-3/codemirror.css.scss
```diff
@@ -465,6 +465,19 @@ span.CodeMirror-selectedtext { background: none; }
   color: white !important;
 }
 
+// mustache mode
+.cm-mustache,
+.cm-mustache-text {
+  color: #e9eded;
+}
+
+.cm-mustache-error {
```
nobuti: I think we are not using this anymore.

matallo: we are! we should get rid of it if we go ahead with https://github.com/CartoDB/cartodb/pull/11965 though

lib/assets/core/javascripts/cartodb3/components/code-mirror/code-mirror-view.js
```diff
@@ -95,44 +120,93 @@ module.exports = CoreView.extend({
...
+
+      return cm.getRange(CodeMirror.Pos(cur.line, cur.ch - 1), cur) === '{';
+    });
+  },
+
+  _completeIfAfterCtrlSpace: function (cm) {
+    var opts = {};
+    var cur = cm.getCursor();
+
+    var openingMustache = cm.getRange(CodeMirror.Pos(cur.line, cur.ch - 2), cur) !== '{{';
```
nobuti: I feel this should not be here, it should be an option. The base view should not know anything about mustache or any other mode the user is using.

lib/assets/core/javascripts/cartodb3/components/form-components/editors/textarea.js
nobuti: Same concerns, this is an exception, and as it is, I think it should be a new editor. We are adding a lot of code that the majority of instances don't need.

lib/assets/core/javascripts/cartodb3/components/form-components/editors/toggle.js
```diff
@@ -0,0 +1,36 @@
+var _ = require('underscore');
+var Backbone = require('backbone');
+var template = require('./toggle.tpl');
+
+Backbone.Form.editors.Toggle = Backbone.Form.editors.Radio.extend({
```
nobuti: Sweet.

lib/assets/core/javascripts/cartodb3/components/form-components/editors/textarea.js
matallo: same concerns than @xavijam then, happy to move it out to its own editor 😄

xavijam: 🎉

matallo:
back to development again, for someone to take those changes, thanks for the review mates!

matallo:
what about now @xavijam @nobuti ?

xavijam:
At a first sight looks what we said so 👍 , but in any case, when you have some minutes @nobuti, could you take a look please? 💋

nobuti:
👍

saleiva:
After another pass I found the following issues:
- [ ] After applying the analysis, the name of the type is weird on the selector (looks like it's not internacionalized) and it only appears the option to switch to lat/lon and not all.
screen shot 2017-04-18 at 15 36 11
screen shot 2017-04-18 at 15 36 49

- [ ] After applying the analysis I cannot apply it again. The button is always disabled.

I tested with [Relacion.Camaras.csv.zip](https://github.com/CartoDB/cartodb/files/929820/Relacion.Camaras.csv.zip)

matallo:
>  After applying the analysis, the name of the type is weird on the selector (looks like it's not internacionalized) and it only appears the option to switch to lat/lon and not all.

This is already happening in production :( investigating how to fix it, and pushing it to the PR

matallo:
thank you @nobuti very much, for maanging the `dsReady` issue, I just updated the syntax color for latin characters, anything else?

(still in the ruis)

xavijam:
@nobuti has found something weird with the behaviour of the "apply" button but it doesn't have to do with this PR, so :+1:. To acceptance again?

matallo:
promoted!

matallo:
and back to development, with the last changes there seems to be some condition by which applying a georeference analysis > subsample analysis > edit georeference analysis the subsample analysis gets stuck `enqueued` 

this is not happening in production but in staging

xavijam:
How come? I mean, how is that possible? JS error? take into account node states come from CARTO.js.

matallo:
no errors are shown in the console, I'm investigating further

matallo:
dismiss my last comment this seems to be an issue with staging environments rather than georeference

PR #11951, opened 2017-04-10
Added jasmine build missing task

ivanmalagon:
When removing test building from `dev` task, I wiped out also the task of building the jasmine spec runners. It's needed for the old editor specs.

It has come out with a build from scratch.


xavijam:
Testing it!

xavijam:
Like a charm!

Issue #11947, opened 2017-04-10
Remove feature-flag code/checks for new org notifications functionality


xavijam:
That means, we should:

- [x] Remove the **feature flag checker** in order to show/hide the notifications panel.
- [x] Remove the **feature flag** (don't forget to remove it from production and staging).

xavijam:
@noguerol and @saleiva, when you decide.

saleiva:
IMO, we should open this right now.

juanignaciosl:
Ey, you can open it to everybody just unchecking the "restricted" flag at superadmin. The tasks at this issue are just for code cleanup.

xavijam:
It is already opened yep, we want to remove the code, yes.

xavijam:
Changed title, sorry for the mixed up.

javitonino:
Although this is ready, I am holding because I found a couple of UI bugs during acceptance that I think should be fixed before opening this to the public: https://github.com/CartoDB/cartodb/issues/11957

xavijam:
Added in the TODO column due to the fact that it belongs to a current milestone.

PR #11955, opened 2017-04-10
Remove notifications feature flag

javitonino:
Part of #11947 

**Acceptance**
- [ ] Ensure that notifications work without the feature flag active:
  - [ ] Send a notification from org panel
  - [ ] Check that it is received by a user
  - [ ] Check that it can be marked as read (open/close the bell notification list)

**Deploy**
- Notify in https://github.com/CartoDB/docs/pull/1196, so the docs can be merged


javitonino:
I did not find any instance of the feature flag in frontend. Maybe @matallo can tell us if he added such a check. Next time, let's use a less generic name than "notifications" for feature flags 🙏, it makes it easier to make sure you did not forget anything when removing it.

matallo:
here's one occurence https://github.com/CartoDB/cartodb/blob/master/app/views/admin/shared/_org_subheader.html.erb#L26

xavijam:
I feel there is no other place, so let's go! 💃

javitonino:
Unblocking since the FF is unrestricted anyway (my previous reasoning was bogus).

Issue #11543, opened 2017-02-15
Disable geometry edition button in read-only layers


ivanmalagon:
After a small discussion in #10905, @saleiva suggested to show always the geometry edition buttons but disabling them in case of read-only layers. Comment link: https://github.com/CartoDB/cartodb/issues/10905#issuecomment-278362912

![image](https://cloud.githubusercontent.com/assets/1078228/22969763/f790e186-f36f-11e6-9e1a-55f13047b421.png)

Current result is that we hide that button.

![image](https://cloud.githubusercontent.com/assets/1078228/22969802/16295cfe-f370-11e6-9038-53952232e5d7.png)

A read only layer can be obtained doing an analysis on that layer.

Not sure if we have a defined style for disabling those kind of button. Summoning here @CartoDB/design

saleiva:
I don't think so, @urbanphes ?

noguerol:
i would make it dissapear. let's go for the cleaner option.

noguerol:
But, to make this clearer, in the layer onboarding i'll draw a fake add geometry button that appears in that step, so we can also give a hint to those users that unfortunately trigger the onboarding with a readonly layer

urbanphes:
@ivanmalagon the method for disabled buttons is low the opacity to 24%.

![5 buttons](https://cloud.githubusercontent.com/assets/14546701/23994037/5d822574-0a43-11e7-9cba-48255681af36.png)

... but in this case can be a bit of weird because on the background we have the basemap.


I agree with @noguerol about disappear the button instead of disabled the button. IMHO is better to offer to the user only that they can do.

saleiva:
I would not remove the buttons so we embrace visibility and discovery of the UI.

noguerol:
IMO hiding a layer is an active action and a need for certain use cases, in which case I don't find a need for either encouraging that user to make it visible nor discover the rest of the UI, while I agree with @juanma that offering what is not available is normally frustrating.

saleiva:
It's not an active action, this also happens when a layer is sync (for example)

noguerol:
Hiding is still an action. If a layer is synced it makes no sense either to show the buttons. But hey, let's not block this: @saleiva your call

saleiva:
The thing is that showing this kind of actions gives us a possibility to explain the steps needed to enable those options. In this case a) sync tables can be edited if being changing options to non-sync and b) read-only tables can be edited if duplicated in the users account.

[...]

PR #11959, opened 2017-04-11
11543 - Disabled edit geometry buttons in read-only and not visible layers

nereski:
This pull request solves the following issue CartoDB/cartodb#11543.

The main goal was to show edit geometry buttons always but disabled for read-only and not visible layers, as you can see in the following GIF.

edit-geometry-disabled

Also, I've added a tooltip to the buttons so the user understands why this functionality can't be used in the current layer.

Every edit geometry buttons are disabled in this case, so instead of disabling each one individually, I've disabled all at once by disabling their container. This way I've been able to change the opacity of the buttons' background and reduce the instances where we evaluate if the current layer is read-only or not visible.

In the future, if each button can be disabled individually they should be individual buttons, each one with their background and space around. In that case, each button should evaluate the conditions to be or not be disabled, but for now, the current solutions works for its purpose.

I've added the corresponding tests to check that buttons are disabled and the tooltip is visible only when the layer is read-only or not visible.

### DIFF
--- a/app/assets/stylesheets/editor-3/context-switcher.css.scss
+++ b/app/assets/stylesheets/editor-3/context-switcher.css.scss
@@ -15,6 +15,9 @@
   overflow: hidden;
   z-index: 4;
 }
+.Editor-contextSwitcher.is-disabled {
+  background: rgba($cBlue, 0.5);
+}
 .Editor-contextSwitcher.is-moved {
   right: 392px;
 }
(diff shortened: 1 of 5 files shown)


Cartofante:

Cartofante:

xavijam:
Hey @nereski, thanks for all 💋 , closing!

Issue #11957, opened 2017-04-11
Issues with org notifications UI


javitonino:
I think these issues block removing the FF and opening up to the public! cc @xavijam 

1. Firefox and Chrome. Pressing cancel in delete a notification dialog, removes it anyway
![out-a](https://cloud.githubusercontent.com/assets/7569673/24899540/f98171c0-1ea0-11e7-872a-7600a7719640.gif)

2. Only Firefox. You have to press "Send" twice for the notification to be sent
![out-b](https://cloud.githubusercontent.com/assets/7569673/24899556/03d118e2-1ea1-11e7-969e-ab7e2b9d0fc8.gif)

javitonino:
Hey @saleiva! Not only Firefox. The deletion dialog thing happens in Chrome as well. Also, I know the issue title sucks, and I should have opened two of them, but seemed like a waste for such minor issues).

xavijam:
Deployed!

PR #11961, opened 2017-04-11
fix organization notifications issues

matallo:
re: https://github.com/CartoDB/cartodb/issues/11957

**acceptance** (test mostly in firefox, but also this should still work in the rest)

- cancel button should work as ESC and close icon
- form submit should work in chrome as well as in Firefox, with Ctrl + Enter, and button click


lib/assets/core/javascripts/cartodb/organization/organization_notification/organization_notification_view.js
```diff
@@ -67,5 +71,11 @@ module.exports = cdb.core.View.extend({
     });
     this.remove_notification_dialog.appendToBody();
     this.addView(this.remove_notification_dialog);
+  },
+
+  clean: function () {
+    this.sendButton.unbind('submitForm');
```
xavijam: This is not necessary if `sendButton` is a view because all bindings pointing to it are removed when it is cleaned.

javitonino:
Acceptance results are in: the delete dialog is working perfectly now.

However, the send notification form has still the same issues. You have to press the "Send" button twice, and control+enter does not work (in Firefox).

xavijam:
Development!

matallo:
😞 sorry for the duplicated work testing, I'l definitely take a look but I was sure this was solved when pushing the changes

sorry again

javitonino:
Haha, don't worry. Now it's when you discover this was due to the change requested in CR, and shift the blame to @xavijam (now that he can't hear us :P)

matallo:
just found out when you modify a form and then try to submit via javascript firefox prevents behaviour, the solution is cloning the form and appending to the body :/

xavijam:
Tested in staging, works like a charm in firefox. If you wanna try with your Linux ( 😛 ) @javitonino, it is deployed in ded07. 💋

javitonino:
Giving my personal seal of approval for Firefox/Linux compatibility.

Yay! 🙌

PR #11965, opened 2017-04-11
codemirror: mustache template

matallo:
made changes in the codemirror parser to comply with this request in https://github.com/CartoDB/cartodb/pull/11939#issuecomment-292559200

> Change the text color to difference the text inside {{}} and the text outside. Maybe we can use the white for the text outside like a “{“

it's much less sophisticated now, so I'm not sure about it, what's your opinion @nobuti ?

screen shot 2017-04-11 at 18 04 41

/cc @urbanphes

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/components/code-mirror/mode/mustache.js
+++ b/lib/assets/core/javascripts/cartodb3/components/code-mirror/mode/mustache.js
@@ -3,56 +3,22 @@ module.exports = function (CodeMirror) {
     return {
       token: function (stream, state) {
         var ch;
-
         if (stream.match('{{')) {
-          ch = stream.peek();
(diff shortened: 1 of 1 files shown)


nobuti:
the simpler, the better. I think to show parsing errors is a nice thing though.

matallo:
yeah, actually there's an ongoing discussion about syntax validation https://github.com/CartoDB/cartodb/pull/11939#issuecomment-293290750

I need to talk to @rochoa about that when he has a minute

xavijam:
Hey guys, what about this PR? can we close it? Should we keep pushing?

matallo:
it's done, pending feedback and/or acceptance as it changes a bit how moustache strings are validated. They slightly change the color for better represent the template (see attached)

screen shot 2017-07-06 at 15 05 37

this is prod ^^^

Issue #11855, opened 2017-03-27
Geometries cannot be edited from the dataset view


alonsogarciapablo:
### Context

I was just doing some testing.
 
### Steps to Reproduce

1. Create a new map with an empty layer.
2. Click on the layer.
3. Draw any geometry.
4. Refresh (because of https://github.com/CartoDB/cartodb/issues/11857)
5. Click on the layer again.
6. Open the dataset view.
7. Double click on any of the editable cells for the geometry you just created (eg: name)
8. Type something and hit Enter.

### Current Result

This:

<img width="358" alt="screen shot 2017-03-27 at 17 05 41" src="https://cloud.githubusercontent.com/assets/390398/24363275/a989b8cc-130f-11e7-9cb4-5ac3963b2027.png">

Backend is returning the following error:

```
{"errors":["PG::Error: ERROR:  parse error - invalid geometry"]}
```

This is the payload that is being sent to the server:

```
{
  "cartodb_id": 1,
  "the_geom": "{\"type\":\"Point\",\"coordinates\":[4.921875,9.10209674]}",
  "the_geom_webmercator": "{\"type\":\"Point\",\"coordinates\":[547900.61874814,1017529.72067584]}",
  "name": null,
  "description": "b"
}
```

### Expected result

Geometry is updated and no error message appears.

xavijam:
Could it be a backend issue @juanignaciosl @javitonino? Could you take a look whenver you can?

alonsogarciapablo:
Just to clarify, the problem persists after page is reloaded.

javitonino:
It seems the problem is that the frontend is sending the webmercator column, which is not sent from anywhere else (for example, `/dataset` or geometry edition). 

I think the easiest solution would be for the frontend to stop sending `the_geom_webmercator` in that view if we do not care about supporting edition of `the_geom_webmercator`. If not, we can take a deeper look, it seems it could be a problem with projections.

xavijam:
Let me take a look tomorrow @javitonino :kiss:.

xavijam:
Fixed!

alonsogarciapablo:
👏👏👏

PR #11981, opened 2017-04-12
Dataset issues

xavijam:
- Bug reported here -> https://github.com/CartoDB/cartodb/issues/11855.
- Pagination position in Safari browser.
- Renaming a table breaks edition until reloaded (https://github.com/CartoDB/cartodb/issues/11969).

We should test that all those things are tested, just follow the issues.


lib/assets/core/javascripts/cartodb3/components/table/table-view.js
```diff
@@ -157,6 +157,10 @@ module.exports = CoreView.extend({
     });
   },
 
+  getColumnsCollection: function () {
```
matallo: being ultra picky here, shouldn't we have a little spec for this? even if we pass the `columnsCollection` as an option 🙏

xavijam: _Bueeeeno_

xavijam:
Moving it to acceptance, ok @matallo ?

matallo:
acceptance straight-away

javierarce:
@xavijam what does "Pagination position in Safari browser." mean?

All the other points in the description were successfully fixed and are accepted.

xavijam:
There was a problem reported in twitter by a user where the pagination was placed over the create map button in the dataset view.

javierarce:
I see, thanks. That seems to work alright too.

xavijam:
\o/

xavijam:
Frontend tests passed with Travis, merging after solved a conflict with master \o/.

Issue #11874, opened 2017-03-28
Problem when user filters a widget and it was displaying the widget form view


xavijam:
We have a problem with the widget form and related when the user filter something in the category/histogram widget (@saleiva @noguerol @jorgesancha). Due to it, we have had a big impact of JS errors yesterday. Fortunately, we have researched a bit and found the problem.

## STR
- Create a map of points.
- Add an analysis (subsample is enough).
- Add a category widget pointing to that analysis node.
- Go to edit this widget.
- Filter by the first result of the widget.
- Check that the carousel of the widget form is not providing several options, only one.
- Try to change the color of the auto-style.
- 💥 .

This is a problem simiilar to https://github.com/CartoDB/cartodb/pull/11822.

xavijam:
Hey @javierarce, I feel we should talk about this, do you want to have a quick recap?

xavijam:
Also, I started some work in `11874-issue` branch.

xavijam:
Recapping what I have talked with some frontenders. @saleiva found a problem where panels were not reacting to changes coming from widgets (https://github.com/CartoDB/cartodb/issues/11819). That opened a box we didn't know, so we decided to just listen to query changes in several views and make the necessary requests in order to "update" the query objects (`query-schema-model`, `query-geometry-model` and `query-rows-collection`). That doesn't scale, so we have to find another way to update those models/collections in an effective way.

A possible solution would be to update those models when an "updated" node is ready in the builder side (info coming from CARTO.js), and make those requests.

But there are several concerns about it:

- We should only request those models from the top node of the layer. It is not necessary to request all nodes data because we only "visualize" the data referenced to the last/top node.
- We must not request the models data at the beginning. We don't need it (although geometry info, but it is requested in other way), and it could provoke 3x the analyses the map has. 5 analyses x 3 objects (QSM, QGM, QRQ) = 15 requests :|

javierarce:
What does 💥  mean exactly? Errors in the console, lack of updates, both?

Also: I've been unable to get the step #5 of your list (`Check that the carousel of the widget form is not providing several options, only one.`). I'm always presented all the options for widgets.

xavijam:
💥  -> JS error in the console.

If you have doubts, I'll show you live 😬 .

javierarce:
Yes, please, I'm probably doing something wrong, because I can't reproduce it on my machine or in production.

xavijam:
Sorry, my bad. The description was not that accurate.

[...]

PR #11989, opened 2017-04-12
fetch analysis node query data when it is ready, and it is node on top

matallo:
from https://github.com/CartoDB/cartodb/issues/11874#issuecomment-292944611

- We should only request those models from the top node of the layer. It is not necessary to request all nodes data because we only "visualize" the data referenced to the last/top node.
- We must not request the models data at the beginning. We don't need it (although geometry info, but it is requested in other way), and it could provoke 3x the analyses the map has. 5 analyses x 3 objects (QSM, QGM, QRQ) = 15 requests :|

I made this a separate PR to properly track changes, I think the first point is achieved with no unintended consequences (as far as I tested). The second point seems to be tackled in the original PR.

Added the essential tests, and removed some failing that were not needed anymore.

Happy to discuss this further @xavijam


matallo:
leaving the fetch as it was, moving the responsibility of is node on top to the collection, now just noticing the collectiong that the model is ready

/cc @xavijam @ivanmalagon

javierarce:
👋  It seems that thew NEWS.md is not updated, @matallo.

app/views/admin/organization_users/_form.html.erb
```diff
@@ -184,7 +184,7 @@
 
     <div class="FormAccount-row">
       <div class="FormAccount-rowLabel">
-        <label class="CDB-Text CDB-Size-medium is-semibold u-mainTextColor">Extra Here Isolines</label>
+        <label class="CDB-Text CDB-Size-medium is-semibold u-mainTextColor">Extra Routing API calls</label>
```
javierarce: Does this change belong to this issue?

lib/assets/core/javascripts/cartodb3/data/analysis-definition-nodes-collection.js
```diff
@@ -66,6 +66,28 @@ module.exports = Backbone.Collection.extend({
         table_name: d.tableName
       }
     });
-  }
+  },
+
+  nodeReady: function (model) {
```
javierarce: I find the name of this method a bit confusing. I would never imagine what it does just looking at it. I would suggest to add a verb (suggestion: `fetchIfOnTop`, `fetchDataIfOnTop`?)

matallo:
actually the original (and non-existent-yet) PR would be from the branch `11874-issue`, thus I didn't make any further change. Shall I point this to master now @xavijam?

xavijam:
Wait, let me add several comments too.

javierarce:
> actually the original (and non-existent-yet) PR would be from the branch 11874-issue, thus I didn't make any further change. Shall I point this to master now @xavijam?

Oh, sorry, I didn't notice that it wasn't pointing to master.

app/views/admin/organization_users/_form.html.erb
xavijam: ?

lib/assets/core/javascripts/cartodb3/data/analysis-definition-node-model.js
```diff
@@ -63,6 +63,10 @@ module.exports = Backbone.Model.extend({
     }
   },
 
+  nodeReady: function () {
```
xavijam: I would not invoke a collection method with a node method too. Why don't you like an event?

xavijam: And let's not use ready for this purpose, that's already used to know if an analysis has finished...

lib/assets/core/javascripts/cartodb3/data/analysis-definition-nodes-collection.js
xavijam: Already dixit, node ready is already used and it is confusing.

app/views/admin/organization_users/_form.html.erb
matallo: seems derived from not having `11874-issue` updated from `master` now they are

lib/assets/core/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -368,7 +367,7 @@ F.prototype._tryToSetupDefinitionNodeSync = function (m) {
 
       if (status === 'ready') {
         if (!m.__initialization) {
-          m.fetchQueryObjects();
+          m.nodeReady();
```
xavijam: Same as above.

lib/assets/core/javascripts/cartodb3/data/analysis-definition-nodes-collection.js
xavijam: Agree with @javierarce

matallo:
made the changes, and moved the method (now renamed) to an event

matallo:
updated master, and now PR points to `master` as requested

xavijam:
Could we change the title and the description of this PR please. Because now it has everything related to the ticket.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/data/data-content-view.js
```diff
@@ -7,6 +7,7 @@ var dataErrorTemplate = require('./data-content-error.tpl');
 var actionErrorTemplate = require('../../sql-error-action.tpl');
 var QuerySanity = require('../../query-sanity-check');
 var OverlayView = require('../../../components/overlay/overlay-view');
+var checkAndBuildOpts = require('../../../../helpers/required-opts');
 
 var renderLoading = true;
```
javierarce: I know this was here before, but does it make sense to use a global var? Wouldn't it be better to use `this._renderLoading` or as a field inside of a model?

matallo: actually I think it makes sense to have it in the `modelView` I'll proceed if you're ok with this

javierarce: +1

lib/assets/core/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -110,7 +110,7 @@ var F = function (opts) {
   this._vis().on('change:error', this._visErrorChange, this);
 
   var saveStateDefinition = _.debounce(this._saveStateDefinition.bind(this), 500);
-  this._diDashboard.onStateChanged(saveStateDefinition);
+  this._diDashboard.onStateChanged(saveStateDefinition, true);
```
xavijam: This `true` is not necessary. I feel I added in my branch but it is not necessary anymore.

javierarce:
I think this is good to go… what do you think, @xavijam?

xavijam:
I feel It can go to acceptance!

xavijam:
Deploying to ded07.

app/assets/stylesheets/editor-3/form-placeholder.css.scss
```diff
@@ -12,6 +12,9 @@ $br: 2px;
   position: relative;
   margin-bottom: 30px;
 }
+.FormPlaceholder-paragraphLoader {
```
xavijam: Uhm? ^^^

matallo: https://github.com/CartoDB/cartodb/blame/8fc88fee00742c6c48c1cbc260d1cec34e93d4b5/app/assets/stylesheets/editor-3/form-placeholder.css.scss#L15

🤷

xavijam:
Something weird is happening with style, data and rest of layer panels. They don't get loaded when widgets are filtered. I'll check at length tomorrow.

app/assets/stylesheets/editor-3/form-placeholder.css.scss
xavijam: :joy: sorry.

Issue #11927, opened 2017-04-05
On layer list the collapse state is not save


urbanphes:
### Context
Being in the layer list and collapse any of the layers I have if I go to one for instance to change the color and I return to the layer list, the layer that was collapsed appear open.

### Steps to Reproduce
1. On the layer list tries to collapse any layer.
2. Go to another or the same layer for instance to change the color.
3. Return to the layer list.

### Current Result
The layer was collapse appear open.

### Expected result
The layer was collapse should appear collapse.

### Browser and version
Safari Versión 10.1 (12603.1.30.0.34)

PR #11992, opened 2017-04-12
Save collapse state in layer definition model.

nobuti:
This PR implements #11927.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -101,6 +101,7 @@ Development
 * Multiple file upload through "upload file" tab (#11952)
 
 ### Bug fixes
+* Save collapse state for layer list (#11927)
 * Disable export image button if not validated (#11949)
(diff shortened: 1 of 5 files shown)


nobuti:
The test for the collapse functionality was already there. What additional
test do you suggest?

matallo:
following @ivanmalagon approach I'd add a new spec for every new method, in this case `toggleCollapse` even if we need to rewrite/rearrange the existing specs to match

happy to discuss and approve, though, I don't have a very strong opinion on this specially if we had this already covered, maybe is just a matter of styleguide

nobuti:
💋

javierarce:
Accepted: 💚

matallo:
merged, bumping assets in `master` for deploy

matallo:
aaand deployed 🚀

PR #12004, opened 2017-04-17
Fix a problem with conditionals geometries in tangram

donflopez:
https://github.com/CartoDB/tangram-carto/issues/37

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -101,6 +101,7 @@ Development
 * New dashboard notifications added (#11807).
 * Multiple file upload through "upload file" tab (#11952)
 * Change setView by flyto.
+* Update tangram to fix layer geometry conditionals.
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -101,6 +101,7 @@ Development
 * New dashboard notifications added (#11807).
 * Multiple file upload through "upload file" tab (#11952)
 * Change setView by flyto.
+* Update tangram to fix layer geometry conditionals.
```
alonsogarciapablo: I don't think we need to include all changes to all the dependencies here... Otherwise we would end up having a huge NEWS.md file (changes to BUILDER + changes to all deps).

donflopez: But it fails otherwise

Issue #11838, opened 2017-03-23
Auto-style not working on Google enabled Builder embeds


noguerol:
### Context
Autostyle is not working on embeds coming from Google enabled Builder accounts.
 

### Steps to Reproduce
*Please break down here below all the needed steps to reproduce the issue*

1. Make a map with both a google basemap and some widgets
2. Click on the autostyle button in the embed view

### Current Result
Geometries in the map doesn't change
 

### Expected result
Geometries get auto styled

This comes from: https://github.com/CartoDB/support/issues/577

ivanmalagon:
Deployed.

PR #12005, opened 2017-04-17
Updated dependencies.

nobuti:
This PR implements #11838.

It updates cartodb.js to bring https://github.com/CartoDB/cartodb.js/pull/1623

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -104,6 +104,7 @@ Development
 * Update tangram to fix layer geometry conditionals.
 
 ### Bug fixes
+* Autostyling for google basemaps (#11838)
 * Styling falsy categories (#11421)
 * Fixed bug editing geometries from dataset view (#11855)
 * Fixed pagination position in Safari browser
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "cartodb-ui",
-  "version": "4.6.73",
+  "version": "4.6.73-ded03-stg",
   "dependencies": {
     "backbone": {
       "version": "1.2.3",
@@ -86,7 +86,7 @@
                             },
                             "readable-stream": {
                               "version": "2.0.6",
-                              "from": "readable-stream@>=2.0.5 <2.1.0",
+                              "from": "readable-stream@>=2.0.0 <2.1.0",
                               "resolved": "https://registry.npmjs.org/readable-stream/-/readable-stream-2.0.6.tgz",
                               "dependencies": {
                                 "core-util-is": {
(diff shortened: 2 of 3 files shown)


matallo:
not really sure if this is related but the widget won't activate until you interact with the map, i.e. click zoom controls

once the map works normally, if you enable autostyle from editor view, when disabling in the embed it would still keep the autostyle style, you have to check and uncheck again

autostyle

if you reload in the layers list with autostyle enabled, there seems to be an error

pageshot of untitled map 2 carto 2017-04-02-1134 57
 
I made sure I did a clean `npm install`

nobuti:
About the first point, there is an open issue already to fix it.
I think the layer list issue is not linked with the changes deployed here, they only related with the named_map template, but I'll check it out.

matallo:
I'm still experiencing the same issues with "matatagoog02" user in staging :(

even worse, having the autostyling ON on the first load the tiles loaded are not styled, this works once you interact with the map though

xavijam:
Take into account also this issue in CartoDB.js https://github.com/CartoDB/cartodb.js/issues/1610.

matallo:
oh, if both the initialization and check/uncheck are known and reported issues, then it'd be good to go

xavijam:
I feel it is related with the issue described above (almost 100% sure).

nobuti:
I'm looking into the check/uncheck thing.

nobuti:
As far I can see, when the dataview fails requesting data, deep-insights fails generating the styles for the autostyle, because the dataview has no data:

```
marker-fill: ramp([type_store], (), ());  
```
screen shot 2017-04-18 at 15 38 57

But even the request to the maps api fails, the style is set into the `initialState` attribute, so if the user disable widget's autostyle, then it tries to set this attribute again, without success. 

Not sure if we could check the styles validity before use them @xavijam.

xavijam:
Yeah, I understand. This is a different problem than the thing we want to solve. I'd open a ticket in deep-insights with this problem.

xavijam:
Already talked with @nobuti, it is a problem we have already @matallo. If happens in the embed and in builder. If you move the map to the place where there is data, everything will work as expected.

nobuti:
Both issues should be fixed in DI. Heading ded03.

nobuti:
User `nobuti-ded03` just in case.

ivanmalagon:
Acceptance passed.

- The check / uncheck issue in embed view doesn't happen anymore.

- The error while reloading the Builder view doesn't happen anymore.

- Still having issues with widgets not being initialised until we interact with the map. So a layer with autostyle doesn't show the proper colors until we move / zoom the map. Then, the interaction behaves as expected. This is going to be tackled at [https://github.com/CartoDB/cartodb.js/issues/1610](https://github.com/CartoDB/cartodb.js/issues/1610)

ivanmalagon:
Deployed to production.

PR #12028, opened 2017-04-20
Removed ff checks

ivanmalagon:
Implements #11847 

**Acceptance**

Be sure that neither the user nor the whole system has the feature flag `static-maps`.

The user should be able to export an image of the map both in editor pane and settings pane.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,7 +17,7 @@ Development
 * Adding Google-Maps basemaps (#11447)
 * Improve dialog forms to render them floated. (#7786)
 * Adds slider component to the forms (#11617)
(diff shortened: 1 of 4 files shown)


javitonino:
Acceptance 👍 

Only found a little bug (unrelated to the FF), but I think we can deploy this before waiting for that fix:
https://github.com/CartoDB/cartodb/issues/12029

@xavijam: Ok to deploy?

xavijam:
deploy it! If only affects Firefox, let's go on and fix it later.

javitonino:
Going out like there is no tomorrow. I'll delete the FF after the deploy is complete.

Issue #12018, opened 2017-04-18
Deprecated SQL Function analysis UI


rochoa:
To support https://github.com/CartoDB/camshaft/pull/280 at builder level, we need to pull a new UI for it.

cc @xavijam

rochoa:
Our initial proposal would be something like:

![screen shot 2017-04-18 at 18 36 24](https://cloud.githubusercontent.com/assets/45346/25142245/8b253434-2466-11e7-996d-285798f4d0e6.png)

saleiva:
Like this :dab:

![analysis - deprecated sql node](https://cloud.githubusercontent.com/assets/132149/25142387/fbe3f6ec-2466-11e7-9086-221a49fff1f6.png)

I don't think we need more designs since this will be automatically generated.

saleiva:
cc/ @noguerol @xavijam

xavijam:
Let's not forget the onboarding text and the animation.

saleiva:
Not sure we can have an onboarding text since it will depend on the function... About the animation. Ok, we can prepare something quick

noguerol:
Update: we have just decided we will have no animation nor onboarding.

xavijam:
Before doing this task, [this ticket](https://github.com/CartoDB/cartodb/issues/12025) is absolutely necessary.

xavijam:
Basically this analysis will let the user use a function "created" in the DB with parameters. The use case is how @saleiva and @rochoa have explained it.

- We need to make a query in order to take all the available functions under the user (query will be done by @rochoa 💋 ).
- That query will give us three things:
   - the name of the function.
   - the name of the arguments accepted.
   - the type of the arguments accepted.
- With that, the second part of the form will be created, giving the user the possibility to fill up all those arguments and launch the analysis.

That's all!

xavijam:
Yep, looks like it is ok! One question, the order of the `params_name` is the one the analysis API is waiting to receive?

rochoa:
Yes.

ivanmalagon:
Even not having animation nor onboarding I think we need some graphic to show in the analyses view.

![image](https://cloud.githubusercontent.com/assets/1078228/25333736/c101cac0-28eb-11e7-9d49-cad645b38a5a.png)

If not, it feels unfinished. Perhaps only a graphic like `y = f(x)` in the same colors as the rest (but no animation).

What do you think @saleiva @noguerol ?

xavijam:
[...]

PR #12036, opened 2017-04-24
12018 sql function

ivanmalagon:
Implements #12018 

This adds to the analyses suite one to run custom SQL functions. Those SQL functions must follow the guidelines described at [deprecated-sql-function-analysis](https://github.com/CartoDB/camshaft/blob/ca43718cc25563d491f27616a3c7891439171b51/docs/deprecated-sql-function.md#deprecated-sql-function-analysis)

Since this is not a trivial task, several functions are provided in the DED01 environment to do this acceptance.

Acceptance
-----------

### Credentials

This version is deployed in DED01. Ping me for credentials.

### Analysis acceptance

- Create a map with a couple of datasets.
- Select one layer, go to `analysis` tab and click on `add analysis`
- `SQL Function`should be there. Click on it.
- There are three main example functions:
    - DEP_EXT_buffer: this only has an integer parameter. Check that it works. This analysis is interesting because it has no secondary node. If it works, it means that the validations of the analysis params are working properly.
    - DEP_EXT_interpolation: this has text and number parameters. I won't say a thing about it in order for you to discover edge cases playing with the input. This one has a mandatory secondary node.
    - DEP_EXT_fixed_buffer: this has a boolean parameter. Just to test that the form renders boolean params.

After applying the analysis with correct input, the typical overlay page informing the user that the analysis has finished should show up. If not, there's something wrong.

This analysis is intended for 'pro' audience and it's generated automatically, so there's no much information in case something goes wrong. Play with it and write down your thoughts about what can be improved (if you think that).

Feel free to add your own functions following the guideline pointed above. @javisantana and @AbelVM did it already with mixed results.

### Feature flag acceptance

This analysis is under a feature flag. Go to central and enable / disable the feature flag on the user profile and check that the analysis option shows / hides. 

### Data services enabled analyses

I had to do a mini-refactor to the function that tells if an analysis can be executed or not. Please check that the next analysis simply work running an analysis:
- georeference-city
- georeference-street-address
- routing (sequential or single point)
- Enrich data from data observatory

Ping me if you need further clarifications or help.


ivanmalagon:
@AbelVM 

Regarding your problem with your DEP_EXT_curvature function.

When I developed this task I talked with @rochoa and he told me that all function arguments should be mandatory. In your function, there's one that should go empty according to your dataset. That's the first thing.

Locally, I've bypassed the required validation and then I get the error
```
column reference "the_geom_webmercator" is ambiguous
```

that suggest that either your function or the code that runs the function is creating the column `the_geom_webmercator` in at least two subtables and the `select` operation doesn't know which one is the result referring to.

We need to argue on the required validation. We should agree on either we leave all fields optionals or not. We can't select which ones because this is a dynamic analysis and so far we only have the type description of the params.

AbelVM:
Roger that. WIll go back on it tomorrow!!

javisantana:
hey, one thing: the_geom_webmercator is something that can NOT be used in analysis. That column is meant for rendering and is dropped when a table enters in the analysis workflow

@rochoa I think i'd make sense to add it to the doc -> https://github.com/CartoDB/camshaft/pull/291

AbelVM:
I create the_geom_webmercator within my function because I need to make some geometry calcs that make no sense in spheric coordinates. The test function I/O has only the_geom

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/deprecated-sql-function-form-model.js
```diff
@@ -0,0 +1,312 @@
...
+    this.set('fetchStatus', 'unfetched');
+    this._functionsInfo = [];
+
+    this._setSchema();
+    this._initBinds();
+    this._fetchAvailableFunctions();
+  },
+
+  _initBinds: function () {
+    this.listenTo(this, 'change:fetchStatus', this._onFetchStatusChanged.bind(this));
```
xavijam: Is It necessary the `.bind(this)`?

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/deprecated-sql-function-form-model.js
```diff
@@ -0,0 +1,312 @@
...
+            {val: 'false', label: 'false'}
+          ]
+        };
+        initialValue = true;
+        break;
+      }
+    }
+    if (item) {
+      schema[key] = item;
+      if (this.get(key) === void 0) {
```
xavijam: `this.has(key)` ?

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/deprecated-sql-function-form-model.js
```diff
@@ -0,0 +1,312 @@
...
+    if (selectedFunction) {
+      _.each(_.keys(selectedFunction.params), function (key) {
+        values.push(this.get(key));
+      }, this);
+
+      this.set('function_args', values);
+    }
+  },
+
+  // TODO: delete it when merge with 'Find nearest' is complete
```
xavijam: ^^^

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/deprecated-sql-function-form-model.js
```diff
@@ -0,0 +1,312 @@
...
+  _isPrimarySource: function (sourceAttrName) {
+    return sourceAttrName === this.get('source');
+  },
+
+  // TODO: delete it when merge with 'Find nearest' is complete
+  _isFetchingOptions: function () {
```
xavijam: Same as above

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/deprecated-sql-function-form-model.js
```diff
@@ -0,0 +1,312 @@
...
+    return this._analysisSourceOptionsModel.get('fetching');
+  },
+
+  // TODO: delete it when merge with 'Find nearest' is complete
```
xavijam: Same as above

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/deprecated-sql-function-form-model.js
```diff
@@ -0,0 +1,312 @@
...
+  _isSourceDisabled: function (sourceAttrName) {
+    return this._isPrimarySource(sourceAttrName) || this._isFetchingOptions();
+  },
+
+  // TODO: delete it when merge with 'Find nearest' is complete
```
xavijam: Yiiija!

lib/assets/core/test/spec/cartodb3/data/analyses.spec.js
```diff
@@ -7,7 +7,7 @@ var camshaftReference = require('camshaft-reference').getVersion('latest');
 var UnknownTypeFormModel = require('../../../../javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/unknown-type-form-model');
 var FallbackFormModel = require('../../../../javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/fallback-form-model');
 
-describe('cartodb3/data/analyses', function () {
+describe('cartodb3/data/analysesjs', function () {
```
xavijam: analyses.js?

package.json
```diff
@@ -21,11 +21,11 @@
     "bootstrap-colorpicker": "2.5.0",
     "browserify": "13.0.0",
     "browserify-shim": "3.8.12",
-    "camshaft-reference": "<2.0",
+    "camshaft-reference": "0.32.0",
```
xavijam: :+1: cc @rochoa

lib/assets/core/javascripts/cartodb3/components/form-components/editors/select/select.tpl
```diff
@@ -8,8 +8,8 @@
       <svg class="CDB-LoaderIcon-spinner" viewBox="0 0 50 50">
         <circle class="CDB-LoaderIcon-path" cx="25" cy="25" r="20" fill="none"></circle>
       </svg>
+      <span class="u-lSpace u-secondaryTextColor"><%- _t('components.backbone-forms.select.loading') %></span>
```
xavijam: I'm curious about this change, what is it for?

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/available-functions.tpl
```diff
@@ -0,0 +1,41 @@
+WITH dep_fns AS (
```
xavijam: Querizaca

ivanmalagon:
After some tweaking made by @AbelVM on his function, it works flawlessly. This is the result.

image

The "problem" here is that this function needed an optional parameter and the UI doesn't allow that. This is on purpose following the specs of the issue. We ran the function in my machine bypassing this requirement.

So take into account that every parameter is mandatory if you develop SQL functions. If future functions will need to have optional parameters we should address it as soon as possible. It affects the acceptance.

lib/assets/core/javascripts/cartodb3/components/form-components/editors/select/select.tpl
ivanmalagon: The spinner appeared in another line. Quite ugly.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/available-functions.tpl
ivanmalagon: Quericaza

xavijam:
Don't forget to remove the unnecessary code already moved to another file thanks to the nearest analysis.

rochoa:
> The "problem" here is that this function needed an optional parameter and the UI doesn't allow that. This is on purpose following the specs of the issue. We ran the function in my machine bypassing this requirement.

Thanks a lot for debugging that case. I just updated the documentation to state all _arbitrary` arguments should be mandatory: https://github.com/CartoDB/camshaft/blob/34800212d32f7dc01208f56f2f0f50f51aabadaa/docs/deprecated-sql-function.md.

> All arbitrary arguments are mandatory, if you need to support optional arguments, it's better to have two functions with two signatures.

juanignaciosl:
Acceptance notes:

- With that user, I get the infamous `:
"PL/Proxy function cdb_dataservices_client._cdb_service_quota_info(2): [cartodb_staging_user_63e60c6d-e907-4a8a-a9ff-251c623231e1_db] PQconnectPoll: FATAL:  no pg_hba.conf entry for host "10.129.221.115", user "geocoder_api", database "cartodb_staging_user_63e60c6d-e907-4a8a-a9ff-251c623231e1_db", SSL on↵FATAL:  no pg_hba.conf entry for host "10.129.221.115", user "geocoder_api", database "cartodb_staging_user_63e60c6d-e907-4a8a-a9ff-251c623231e1_db", SSL off↵"` error when I get into Builder. Probably not relevant.
- At the issue @urbanphes said that "Maybe we can make it as a typewriter. As if the text is written." but box is empty.
- DEP_EXT_interpolation (which is actually DEP_EXT_spatialinterpolation):
  - using any existing column (such as `name`) as `val_column` displays this error: "Analysis A1 failed: column "name" specified more than once". It's probably fine, but not straightforward. Using a nonexisting column displays "Analysis A1 failed: column _source.name2 does not exist". Other that that, I haven't been able to make it crash, but I didn't come up with any result. I don't know what it does.


@ivanmalagon do you want me to create additional custom functions or test DEP_EXT_spatialinterpolation more properly? :-)

ivanmalagon:
- The typewriter stuff won't be made. I talked with @xavijam and decided to not include any animation for this analysis.

- Happy to hear that you didn't make it crash! The other errors are expected. Since this runs custom functions we show to the user the raw error coming from the API. The errors will be different depending on the function. If you want to be successful in that particular function, apply it to `lineas_metro` insert this arguments:

```
target node: B0
val_column: place
method: 1
number_of_neig: 1
decay_order: 1
```

lib/assets/core/javascripts/cartodb3/editor/layers/analysis-views/composite-layer-analysis-view.js
```diff
@@ -55,22 +55,24 @@ module.exports = CoreView.extend({
...
+    if (nodeDefModel) {
+      var view = nodeDefModel.get('type') === 'source'
+        ? new SourceLayerAnalysisView({
+          model: nodeDefModel,
+          analysisNode: this._analysisDefinitionNodesCollection.get(nodeDefModel.id),
+          layerDefinitionModel: this._layerDefinitionModel,
+          showId: !!(layerDefModel && this._isOwnedByOtherLayer(layerDefModel))
+        })
+        : new DefaultLayerAnalysisView({
+          tagName: 'div',
```
ivanmalagon: I didn't came up with this. I just move the code into the conditional. Do you want me to research this?

xavijam: No, not now.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/deprecated-sql-function-form-model.js
ivanmalagon: You're right. It's not needed.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/deprecated-sql-function-form-model.js
ivanmalagon: Changed.

juanignaciosl:
@ivanmalagon let me know if you want me to do more testing!

lib/assets/core/test/spec/cartodb3/data/analyses.spec.js
ivanmalagon: I did this to filter better this particular spec. If not, since `cartodb3/data/analyses` is also the prefix of another specs files I got more tests than desired.

lib/assets/core/javascripts/cartodb3/editor/layers/analysis-views/composite-layer-analysis-view.js
ivanmalagon: I didn't write this particular line. Only move it to include inside a conditional.

xavijam: Ok ok!

lib/assets/core/javascripts/cartodb3/data/analysis-definition-node-model.js
```diff
@@ -111,18 +111,23 @@ module.exports = Backbone.Model.extend({
     for (var name in paramsForType) {
       var param = paramsForType[name];
       var val = this.get(name);
+      var includeParam = true;
```
nobuti: I find this a bit harder to understand, I tend to get lost between positive and negative values (always keep in mind my slow mind). `param.optional` seems to affect only to `includeParam`, why not use it without any proxy variable?
```
if (param.type === 'node') {
    var node = this.collection.get(val);
    if (!node && !param.optional) {
      throw new Error('no node found for param "' + name + '" with id "' + val + '" in node ' + JSON.stringify(this.attributes));
    } else {
      val = node.toJSON(options);
    }
  }

  if (!param.optional) {
    paramsforJSON[name] = val;
    analysisParams.push(name);
  }
```

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/base-analysis-form-model.js
```diff
@@ -36,13 +36,16 @@ module.exports = Backbone.Model.extend({
...
   _getSourceOption: function () {
     var sourceNodeDefModel = this._layerDefinitionModel.findAnalysisDefinitionNodeModel(this.get('source'));
-    return [{
-      val: sourceNodeDefModel.id,
-      type: 'node',
-      nodeTitle: this._analyses.title(sourceNodeDefModel),
-      layerName: this._layerDefinitionModel.getName(),
-      color: layerColors.getColorForLetter(sourceNodeDefModel.letter())
-    }];
+    if (sourceNodeDefModel) {
```
nobuti: maybe `return sourceNodeDefModel && ...` and we can skip one return to implicitely return `undefined`.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/deprecated-sql-function-form-model.js
```diff
@@ -0,0 +1,283 @@
...
+        this.set('fetchStatus', 'error');
+        if (errors && errors.responseJSON) {
+          track(errors.responseJSON);
+        }
+      }.bind(this)
+    });
+  },
+
+  _parseFunctionsMetadata: function (metadata) {
+    return _.map(metadata.rows, function (data) {
```
nobuti: what if metadata has no rows? working on DO, I found scenarios where the returned data is null.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/deprecated-sql-function-form-model.js
```diff
@@ -0,0 +1,283 @@
...
+        info.params[data.params_names[i]] = data.params_types[i];
+      }
+      return info;
+    }, this);
+  },
+
+  _capitalizeFunctionName: function (name) {
+    if (name) {
+      var depExtRegex = /^dep_ext/i;
+      return name.replace(depExtRegex, 'DEP_EXT');
```
nobuti: the same comment as above, in this case, the second return is an explicit `return undefined` right?

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/deprecated-sql-function-form-model.js
```diff
@@ -0,0 +1,283 @@
...
+
+  _getSelectFunctionAttrs: function () {
+    var fetchStatus = this.get('fetchStatus');
+    var status = {
+      disabled: true
+    };
+
+    switch (fetchStatus) {
+      case 'unfetched':
+        status = {
```
nobuti: only `break`makes the deal

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/deprecated-sql-function-form-model.js
```diff
@@ -0,0 +1,283 @@
...
+        label: info.functionName
+      };
+    }, this);
+  },
+
+  _serializeFunctionArgs: function () {
+    var values = [];
+    var selectedFunction = this._getSelectedFunction();
+    if (selectedFunction) {
+      _.each(_.keys(selectedFunction.params), function (key) {
```
nobuti: reduce better?

lib/assets/core/javascripts/cartodb3/vis-notifications.js
```diff
@@ -30,16 +30,28 @@ var VisNotifications = {
     }
   },
 
-  _getNotificationAttrs: function () {
+  _getNotificationAttrs: function (message) {
+    var body = _t('notifications.vis.failed.body');
+    if (message) {
+      body = this._period(message) + ' ' + body;
```
nobuti: we could use a local template here instead string concatenation.

lib/assets/core/test/spec/cartodb3/data/analyses.spec.js
```diff
@@ -168,4 +170,78 @@ describe('cartodb3/data/analyses', function () {
       expect(analyses.MAP.closest.onboardingTemplate).toEqual(jasmine.any(Function));
     });
   });
+
+  describe('checkIfxxx several checks', function () {
+    var checkIfxxxTest = function (testTypes, dataServicesFlag) {
```
nobuti: sweet!

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-form-models/deprecated-sql-function-form-model.js
xavijam: 👍

ivanmalagon:
PR last changes deployed at ded01 (pointing to the latest `cartodb.js` version). We have to wait to the changes in camshaft regarding `node_id`and then pass a quick round of acceptance.

AbelVM:
Once deployed to staging, is there any checklist of things to test? Or may I just try to break it, as usual?

image

ivanmalagon:
The first comment of this PR has a series of steps to do the acceptance. But in any case, if you want to enter a bar to order beers for your birthday, you're free to go ahead.

Wait until the last version of camshaft is deployed, by the way. I'm on it.

AbelVM:
By the way, if you want to test another function within this brand new node: **DEP_EXT_curvature**. 

This function calculates the center and radius of curvature of each point in a point dataset that represents tracks (typically a GPX file source) with sequential locations.

Params:
* **cat_column**: track ID column, if you have more than one track (`0` if only one track)
* **temp_column**: time column  (`0` defaults to cartodb_id)

Collinear points (infinite curvature radius) will result in NULL geom and radius.

The results keep the cartodb_id of the original point, so you can join it later.

PR #12063, opened 2017-04-25
Disabling geometry buttons for read-only layers

rjimenezda:
This PR makes new geometry buttons look disabled (instead of hiding them) for read-only layers.

I've moved the is-disabled to the parent view, which requires a small change in styles & a slight change in the click handler as well.

This is how it looks like (like isVisible false)

image

I've updated the tests blindly because the cartodb3 tests don't run locally for me. It looks like the previous ones are broken (it expected .length 3 in both cases, which makes sense now, not before), so I don't know if my changes are correct or not.

We could add more coverage by testing the behaviour with isVisible as well, which produces the same result.

### DIFF
--- a/app/assets/stylesheets/editor-3/context-switcher.css.scss
+++ b/app/assets/stylesheets/editor-3/context-switcher.css.scss
@@ -48,11 +48,13 @@
     border-right: 0;
   }
 }
-.Editor-contextSwitcherItem.is-disabled {
-  opacity: 0.24;
+.Editor-contextSwitcher.is-disabled {
+  .Editor-contextSwitcherItem {
+    opacity: 0.24;
 
-  .Editor-contextSwitcherButton {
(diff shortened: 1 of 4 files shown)


Cartofante:

Cartofante:

rjimenezda:
Yeah so I actually had a problem on the spec, thanks travis :heart_eyes:

rjimenezda:
Mmmm ok tests were still broken, just pushed another fix. this one should be fine I think.

I am setting the visible on each test, which should trigger a _renderContextButtons.

The tests were originally 'wrong' (always expecting 3, instead of 3 and 0) because it never re-renders, so it's always 3. Which I guess was done on purpose to keep the tests, but not break them.

Not being able to run tests locally sure is a handpicap.

xavijam:
Closing!

Issue #11746, opened 2017-03-13
Big mismatch in actual vs estimated row count


noguerol:
### Context
On the analysis quota warning, consumption estimation shows a huge mismatch between rows and credits.

Related to https://github.com/CartoDB/camshaft/issues/238

### Steps to Reproduce
*Please break down here below all the needed steps to reproduce the issue*

1. Open a two point layer
2. Add AOI / Georeference analysis and select settings

### Current Result
Quota notification shows a 510 credit consumption estimation
 
![captura de pantalla 2017-03-13 a las 14 15 30](https://cloud.githubusercontent.com/assets/70603/23855993/237762f6-07f8-11e7-83cc-3a496a76e00b.png)

### Expected result
2 credit consumption estimation (or at least something close)

### .csv file

[two_point_table.zip](https://github.com/CartoDB/cartodb/files/838270/two_point_table.zip)

 

### Additional info
*Please add any information of interest here below*

nobuti:
I can't reproduce it in production, under my team account, could you give me more details?

noguerol:
try with this .carto: 
[quota test map (on 2017-03-13 at 18.14.41).zip](https://github.com/CartoDB/cartodb/files/839228/quota.test.map.on.2017-03-13.at.18.14.41.zip)

nobuti:
This seems to happen only in @noguerol's account. The `explain` query returns:

```
{"rows":[{"QUERY PLAN":"Seq Scan on untitled_table_4  (cost=0.00..11.53 rows=510 width=132)"}],"time":0.002,"fields":{"QUERY PLAN":{"type":"string"}},"total_rows":1}
```

with an isoline with only 1 tract.

any clue @rafatower?

rafatower:
@noguerol what's the username of the account you used?

rafatower:
There's a difference between importing a table and populating an empty table with a couple of rows:
* When importing,`ANALYZE` is run as part of the process.
* When populating an empty table, `ANALYZE` is not run *after* inserting the rows. That's why you cannot reproduce the issue by importing a `.carto` file.

E.g: if you run this:

```sql
CREATE TABLE test_row_estimate_stats (my_attr text);
ANALYZE test_row_estimate_stats;
INSERT INTO test_row_estimate_stats (my_attr) VALUES ('foo'), ('bar');
EXPLAIN SELECT * FROM test_row_estimate_stats;
```

most likely you're going to get an unexpected number of rows:
```

[...]

PR #12064, opened 2017-04-26
11746 mismatch estimation

nobuti:
This PR implements #11746.

In order to test it, check that estimation match the right value for analysis with quota. Try running an analysis importing a new layer as well.


lib/build/affectedFiles/affectedFiles.js
```diff
@@ -148,11 +148,11 @@ var main = function (testsFolder, replacements, modifiedFiles, filesRegex) {
         logTriggers(specsInfo.triggers);
```
xavijam: Not yellow anymore?

Issue #11962, opened 2017-04-11
Make changes in the assets build process


xavijam:
#What does it mean? Basically improve the way we prepare the assets for development or production. 
The main purposes are:

- Improve development speed: watch, bundle, initialization,...
- Improve deployment speed: don't uglify already minified vendor files, for example, don't run unnecessary development tasks here, etc.

It has a lot of implications, but the main tasks should be:

## Main tasks
- [x] Migrate from Browserify to **Webpack**: already done in the test part thanks to @ivanmalagon.
- [x] Taking into account the previous bullet, we should not migrate the old code (editor) to this new "workflow". That means only, for the moment, **builder**, **dataset** and **public** embed views will be "migrated".
- [x] Start using **lib-sass** for compiling the CSS instead of **~compass~** for the whole project. We should leave this last one only for the old assets.
- [x] We should check if new tests task needs to be revisited after those new changes.

## Caveats
- [x] We need to keep using the **core <> client** behaviour for assets touched "outside" of our usual workflow (mainly for solutions engineers).
- [ ] ~Vendor files (like tangram) don't need to be uglified, we should avoid this in order to gain time in the deploys~ _Decided to leave it as it is for now (follow thread)_.
- [x] When a new file is added, it should be taken by the watch process.

## Improvements
- [x] Research how to improve javascript source-map support for staging and production.

## Issues:
- [x] Some modals are not properly rendered.
- [x] CSS sourcemaps are not working.
- [x] Revamp cartodb.js and deep-insights to include cartoassets in bundle time. Right now we have this dependency twice in those styles -> https://github.com/CartoDB/cartodb/pull/12203.
- [x] As a continuation from above, use partials `_whatever` to include code not compilable directly -> https://github.com/CartoDB/cartodb/pull/12203.

## Nice to have
- [ ] Hot module replacement
- [ ] Smart build: only build touched code. If the edited/new code belongs to Builder, only build the Builder. If the code belongs to Editor, only build the Editor. If there is code for both sides, build Builder and Editor.

@CartoDB/frontend

matallo:
considering editor js code won't be migrated, does this imply we can split now the bundling tasks for them? i.e.: now to start developing you need to do `grunt && grunt dev` which will create the sprites from css, ...

could we have something similar *just* for builder? in any case, the build task would take editor into account when releasing new versions

xavijam:
IMO:

- For editor we should use `grunt editor dev`.
[...]

PR #12065, opened 2017-04-27
move from browserify to webpack assets bundling, and split compass in builder assets

matallo:
re: https://github.com/CartoDB/cartodb/issues/11962

this PR improves the assets bundling, moving from a `browserify` based one (via `grunt` task) to a `webpack` based one. For the moment this task will also be triggered through `grunt`, to keep the current front-end workflow, but it is already prepared to be executed independently as an `npm` task (or `yarn`!!!).

In this same PR we've split the `compass` task, and created a new `sass` one to compile Sass to CSS using `node-sass`.

These changes mostly affect Builder assets, which we are actively developing for and was the main pain point. As Editor assets will eventually be dropped, no further improvements have been done in this area.

- while the deploy time remained in the threshold of 17-18 minutes, the time to start developing `grunt dev` plunged **from ~3:20s to ~1:40s**

- the builds have been targeted for Editor/Builder and `grunt dev:builder` can be launched, which makes this figure even smaller to merely **~35s**, being compass one of the big chunks in assets generation

- rebuild times between changes in development are almost immediate, with **less than 1s**

- In order to run tests, the task `grunt test` task has to be launched normally, which replaces Builder specs bundled with `browserify` with the new _affected_specs_ workflow in `browserify` (kudos @ivanmalagon!)

Editor, and Builder specs can be launched independently, via `grunt editor_specs`, and `grunt affected_specs` respectively.

---

Secondary actions have been taken:

1. Simplified `copy` task targets:

No more `js_core_cartodb`, `js_client_cartodb`, ... Now they both are part of the same target, using files array format [1]

2. Removed `js_test_jasmine_cartodb3` target, jasmine specs were moved to browser specs. 

3. No watch for editor stylesheets.

This for sure will be a pain (as already happened in the past, now being more aggressive with watched targets) but will push to migrate old Compass stylesheets. _Just a reminder that only few selected Editor js folders are watched._

4. Removed `/common` folder from `importPath` options in `compass` (and not added to `sass`). From documentation:

> Makes files under the specified folder findable by Sass's `@import` directive.

In a quick check no `@import` from `/common` is done, just `/cartoassets`, and if any we should add `/old_common`, that is explicitly imported now (and to be honest, I prefer it this way).

[1] https://gruntjs.com/configuring-tasks#files-array-format

---

During this process I noticed several points to discuss (aka nice to haves, work not done).

1. I'm not sure `make update_cdb` is the optimal task to include Cartodb.js dependency (maybe I'm losing some context here), I'd rather install and include this dependency through `npm`. We should improve Cartodb.js build if necessary.

2. While Core/Client workflow remains no further optimization steps have been done. Ideally including client frontends should be parametrized, so tasks such copy client folder are optional scraping some ms to the final build time.

3. This is the same case than copy specs tasks. When deploying to production tests are not being checked, however specs files are copied. As those tasks are not finally part of the deploy workflow this is not a very intensive task.

If we really wanted to improve frontend builds to the full extent, neither client nor specs copy tasks should be run, unless some `--with-client --with-specs` parameter is passed.

4. ~~Editor tests are still bundled through Browserify in every build, with no further improvements. This could be done similarly to current Builder specs, but is a penalty I wouldn't invest removing, as its days are numbered.~~

5. ~~Despite an initial approach to split development and production builds for editor and Builder to work standalone (as in you could only build one and work normally without the other) [1] this was quickly dismissed because of the interdependencies in both js and css (i.e. add layer dialog in map view). We should start migrating to Builder process extensively.~~

6. Hot module replacement

[1] https://github.com/CartoDB/cartodb/issues/11962#issuecomment-293303420

/cc @CartoDB/frontend


xavijam:
Hey! Nice improvement! But IMHO, I don't like so much the `grunt && grunt dev` task. Don't we like the thing I proposed in the issue? For example, in development: `grunt editor dev` or `grunt builder dev` ?

matallo:
ok, will try to rearrange tasks in that fashion, in any case take into account my comment

> Despite an initial approach to split development and production builds for editor and Builder to work standalone (as in you could only build one and work normally without the other) [1] this was quickly dismissed because of the interdependencies in both js and css (i.e. add layer dialog in map view). We should start migrating to Builder process extensively.

so, despite we can have a `grunt editor dev` and `grunt builder dev` (not sure about the final syntax) we'll need to run both to have a fully functioning Builder experience

xavijam:
Yep, we already talked that yesterday. I feel we could create an "alias" to run the "editor" assets task and then run the builder task with the watch.

matallo:
as requested, now you can launch `grunt dev:builder` to just build js/css for the builder part, of course Dashboard won't be functional, but it's way faster (~30s)

to build Builder && Editor as normal, `grunt dev` will be enough

---

`cartodb3-specs` has been removed from browserify, and we're already pulling from webpack

I'd love @ivanmalagon to take a look at my changes, specially https://github.com/CartoDB/cartodb/pull/12065/commits/bfd9fa6edac461d9e8bad89b038a773e1094b75a#diff-84673bdcf887decd1e1ddd95d2ef9a1aR28

As a final note (this would need to be added to the `NEWS.md`), as well as you can launch `grunt affected_specs` for builder, editor specs can be launched with `grunt editor_specs`

all tests are passed with `grunt test` normally (for travis)

lib/build/tasks/browserify.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
           }]
         ],
 
-        // enables watchify when grunt is run with a watch task, e.g. `grunt dev`
+        // enables watchify when grunt is run with a watch task
         // must be evaluated lazily (using template var) to allow override by setConfig task.
         watch: '<%= env.browserify_watch %>',
```
nobuti: Do we need watchify? I'm not familiarised with all the uses cases

matallo: actually no, `browserify` is only being used for Editor specs, which are not watched anymore

package.json
```diff
@@ -85,10 +85,10 @@
     "shim-loader": "0.1.0",
     "source-map-support": "0.4.0",
     "stringify": "5.1.0",
-    "tpl-loader": "git://github.com/ivanmalagon/tpl-loader.git#master",
+    "tpl-loader": "git://github.com/ivanmalagon/tpl-loader.git#5cc2b785d10d873f2530fa885d3cb69894569a2f",
```
xavijam: Hey, could we move that "fork" to CartoDB organization @ivanmalagon ?

matallo:
from https://github.com/CartoDB/cartodb/issues/11962#issuecomment-300148528 (continuing the discussion here)

- [x] Some modals are not properly rendered.

this has been fixed after **always** running compass task, one of the drawbacks of having editor assets mixed with builder (onboardings, dialogs, ...) probably the pain/slowness of compass will make us move to sass faster 😛 

- [x] CSS sourcemaps are not working.

they work now, however there is an artifact: they only work properly when the file has been generated through sass. I'll explain, if the source stylesheet has been generated with compass or just copied to be concatenated after, the sourcemaps appear garbled. If you make this concat in sass entry files the sourcemaps look good. I moved cartoassets to this entry point instead of concat the produced cartoassets. However they don't seem to be right 😞 I'll investigate further.

Also, I moved the sass-generated editor3 files from concat to an editor3 entry point, and at the same time prefixed them with `_`

- [x] The dashboard is not built anymore using the `builder`. We knew that, but we need to discuss the need to generate the files initially but not to watch them later.

Same compass issue than above, solved.

- [ ] ~~Get rid of invalidation grunt task~~

I don't think this was needed.

- [ ] Revamp cartodb.js and deep-insights to include cartoassets in bundle time. Right now we have this dependency twice in those styles.

I think this might be out of scope

- [x] As a continuation from above, use partials `_whatever` to include code not compilable directly.

BOOM. Done.

I have this branch working in ded04 for testing.

xavijam:
Niiiice push man! 💪

matallo:
closer and closer! let's see if I can find a workaround to the sourcemaps stuff, but we should move to sass entry points the sooner the better 😛

matallo:
**and sourcemaps working for builder**, I had to add a new `common_editor3` entry to concat the common files generated with compass and split the builder entry, but I think we can live with that

in any case, this is an actual improvement as we don't have css sourcemaps in builder currently

matallo:
some final thoughts but I'd leave this ready for merge:

css goes from scss components in `/app/assets/stylesheets/editor-3/entry.scss` to sass generated css in `/tmp/sass/editor-3/entry.css` to concat file in `/public/assets/VERSION/editor3.css`

while sass generates the sourcemaps too, it does with the same name than the source file, thus we und up with a bunch of `entry.css` and the corresponding `entry.css.map`

however when we concat those with some other files (cartoassets, deep-insights, cartodb.js, ...) the resulting file has references to a relative sourcemap which has a different name, if we stick to our original example, we end up with a `editor3.css` which has a sourcemap pointing to a relative `entry.css.map`

you can check the whole list where those files are in https://github.com/CartoDB/cartodb/blob/11962-assets-build-webpack/lib/build/files/css_files.js

this can be worked out generating directly the `editor3.css` (and `editor3.css.map`) from sass, but tried different combination and had no luck making them work in firefox, also when trying in staging the map files aren't uploaded so they won't work at all

in any case, in production the sourcemaps and other comments in the css are stripped so they won't work either

js sourcemaps work as expected

tl;dr: I'd leave the css sourcemaps inlined as they make no harm

ivanmalagon:
Found two issues:

- All resources related to `bootstrap-colorpicker` return 403 and it doesn't get well rendered.

- In a histogram time-series widget, if you select a range, the widget tab gets faded as it's reloading forever.

Since I'm not sure these problems happen because of the new pipeline, I'm investigating a bit further.

ivanmalagon:
Image URLs issue fixed and CSS watch added.

The problem with colorpicker images was that they were using a URL helper from Compass. We're not using Compass anymore so it failed silently, simply setting a non existing URL.

We've modified the URLs to let libsass to form them well.

While fixing this, we realized that we lost CSS files watch. The solution is far from perfect because it launches `npm start` and `grunt watch:css` at the same time. That makes the webpack output invisible. If we end up finding this annoying we should run the two tasks in different terminals.

Regarding the another issue, I couldn't reproduce it. It's not happening now.

ivanmalagon:
Deployed again, checked colorpicker and everything works great. Rock solid! 🗿 

Kudos to @matallo and @nobuti !!

xavijam:
What about merging this one on monday? @CartoDB/frontend

nobuti:
I want to clarify, kudos to @matallo, he did the 99% of the work 👏  👏  👏

xavijam:
There is a problem with the last merge with master and I have to take a look :(. But in any case, after fixing it, I'll merge it mates.

xavijam:
Ok, so we are suffering [this problem](https://github.com/npm/npm/issues/14042) with npm.

xavijam:
Ouch, I've found a problem with the API keys page:

With new changes:
screen shot 2017-06-13 at 12 55 46


With old changes:
screen shot 2017-06-13 at 12 56 54

cc @CartoDB/frontend

xavijam:
Seems like we have fixed those problems 💃 . MERGING @CartoDB/frontend .

alonsogarciapablo:
VAMOS! 👏 👏 👏 👏 👏

rjimenezda:
👏 👏 👏 👏 👏 

Hyperspace!

Issue #12045, opened 2017-04-25
UI Notifications - Markdown italics not rendering


asimcox:
### Context
*Please explain here below what you were doing when the issue happened*

From Admin Org UI -> Create a notification to all users with markdown text as follows
Try Markdown

After I hit send, the notifications show up with this text as plain text instead of being in Italics. But when I see this notification in my Admin Org UI sent notifications list, it shows in Italics - see image in referenced ticket below.

noguerol:
cc @xavijam

nobuti:
Up and running

PR #12068, opened 2017-04-27
Fixed styles within Notifications dropdown

xavijam:
Related ticket: https://github.com/CartoDB/cartodb/issues/12045

screen shot 2017-04-28 at 14 05 06
screen shot 2017-04-28 at 14 09 08

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -110,7 +110,7 @@ Development
 
 ### Bug fixes
 * Autostyling for google basemaps (#11838)
-
+* Fixed problem with markdown in organization notifications (#12045)
 * Save collapse state for layer list (#11927)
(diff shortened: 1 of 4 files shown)


xavijam:
I feel it is not needed to change the NEWs.

javitonino:
Italic works fine, but bold text does not display correctly in the notifications bubble. Also, we do not support code (fixed-width) text, which maybe we should.

screenshot_20170428_130447

I think this is back to dev, although this is not explicitly stated in the ticket, I feel it should be addressed here. 🔴

xavijam:
Agree! Not sure about the code one but, yeah, let's check it.

nobuti:
Works great ✨

PR #12070, opened 2017-04-28
Get rid of Blob.

nobuti:
This PR implements #12066.

Blob is a pita in Safari. Basically the solution was to put the mime type(data:image/svg+xml) in the
variable containing the svg, this way you don't need to use the blob anymore and then to set the image src with the svg.


nobuti:
@javierarce could you take a look from a user point of view? I'm not familiarised on how it was working in Safari before this change. Right now, in Safari a new window opens, forcing the user to save the image manually.

javierarce:
Yes, that's how it has been working (in Editor, for example) and that was the idea for the behaviour in Builder. As far as I know, it's the only option for Safari… but did you find a workaround for that? Does this PR changes the way you save the image in Safari?

nobuti:
No, it's working the same way, the only change was to get rid of Blob to get the logo. But I was not sure about how was working before it because I didn't test it Safari before.

xavijam:
❤️

javierarce:
Oh, ok! That's the way it should work then. It's a limitation on Safari's part.

xavijam:
`code`

matallo:
I didn't manage to get it working under Safari with "matata1ded09" user :/

xavijam:
Change the package and deploy again, just in case.

nobuti:
@matallo it's working for me, with that user and Safari 10.1, what version of Safari are your trying?

matallo:
screen shot 2017-04-28 at 17 38 55

matallo:
Also not working in IE11. In Edge it opens a blank page, but won't work either.

matallo:
shall we move back to development?

nobuti:
Yes, it has to be something related with the mime type, all major browsers support data URIs: http://caniuse.com/#feat=datauri

nobuti:
Both IE issues are production right now 😞

nobuti:
Some status. I've been working on this, and IE's issues still remain due to Security Error. IE seems to be tainting all images filled with SVG for security reasons: http://stackoverflow.com/questions/33364396/svg-tained-canvas-ie-security-error-todataurl. I made some tests using https://github.com/canvg/canvg but I could n't get it to work. We need to test uploading the generated image when this error raises and load the image from there.

xavijam:
Working in:

- Firefox.
- Opera.
- Safari (tested 10.1, but it was not working in my machine before).
- Chrome.
- IE11.
- IE Edge.

nobuti:
Did you check in Vivaldi?

xavijam:
😂

lib/assets/core/javascripts/cartodb3/editor/export-image-pane/export-image-pane.js
```diff
@@ -24,7 +25,7 @@ var ATTRIBUTION_WIDTH = 15;
 var ATTRIBUTION_HEIGHT = 16;
 
 var NOTIFICATION_ID = 'exportImageNotification';
-
+var LOGO_PATH = '/unversioned/images/carto.png';
```
nobuti: Thinking about pluggable frontend, should this have an agnostic name? Like export-logo.png or something. Not sure if we have this scenario where a company wants to upload their logo.

nobuti: Well, they could overwrite this file, so no need.

xavijam: yep, they could create another logo file and then change the "filename". Or basically, as you said, overwrite the file.

xavijam:
Deploying it to ded05.

PR #12072, opened 2017-04-28
Fixing problem about saving metadata

xavijam:
Related ticket: https://github.com/CartoDB/support/issues/656

We didn't set the "internal" properties for the TextArea editor component, and in the commit function it was required:

```js
…
  commit: function(options) {
    var error = this.validate();
    if (error) return error;

    this.listenTo(this.model, 'invalid', function(model, e) {
      error = e;
    });
    this.model.set(this.key, this.getValue(), options);

    if (error) return error;
  },
…
```

That was the reason why description was not updated after editing it in the metadata modal 💃 .

## STTest
- Check **map** description is stored after being edited in metadata dialog.
- Check **dataset** description is stored after being edited in metadata dialog.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -109,6 +109,7 @@ Development
 * Update tangram to fix layer geometry conditionals.
 
 ### Bug fixes
+* Fixed problem with the textarea editor (cartodb/support#656)
 * Autostyling for google basemaps (#11838)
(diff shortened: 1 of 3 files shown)


nobuti:
Looking good 👍

Issue #12044, opened 2017-04-24
UI Notifications distorted pop up


asimcox:
### Context
*Please explain here below what you were doing when the issue happened*

Showing a weird cut popup on IE browser. See referenced ticket below for more detail and screenshot.

related to #11676

AleGoiko:
looks like this: 
![](https://cloud.githubusercontent.com/assets/25391188/25313690/8ca86e0e-2801-11e7-9ffa-e707dcb48b7b.png)

xavijam:
Added in Builder TODO.

nobuti:
In production.

PR #12073, opened 2017-04-28
ie11 fix for dropdowns with scrollview

matallo:
re: https://github.com/CartoDB/cartodb/issues/12044

fixes ie11 issue when using a scrollview inside a dropdown made it collapse, if the dropdown content is <300px tall the content remains as is, if it is taller a scrollview is appended to the dropdown containing the content and a new class with the fix added

otherwise we'd make all dropdowns to be 300px height

photo5771840639258962207

**acceptance**

- <300px and >300px height dropdowns should work normally in chrome, and ie11


nobuti:
Looks great 🍝

Issue #12025, opened 2017-04-19
Possibility to enable/disable an analysis by feature flag


xavijam:
Right now we take the decision about offering an analysis (or not) if it is enabled by config in the case of it needs any service. Example, for data-observatory we have ([analyses.js file](https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/data/analyses.js#L80)):

```js
…
'data-observatory-measure': {
   …
    checkIfEnabled: function (configModel) {
      if (!configModel) throw new Error('configModel is required');
      return configModel.dataServiceEnabled('data_observatory');
    },
    …
  },
…
```

So, as you can see, we only check the availability with the configModel. One solution would be to add another parameter with the `userModel` for the `checkIfEnabled` function.

javierarce:
Should we extend this new check to all the analysis or just the ones that are currently being checked by `dataServiceEnabled`?

xavijam:
Well, if we pass it as an argument, and it is placed in the second position, we could "make it visible" unless it is need within the checker function.

```js
…
checkIfEnabled: function (configModel) {
   return configModel.get('whatever');
},

…
checkIfEnabled: function (configModel, userModel) {
   return userModel.hasFeatureFlagEnabled('whatever');
},

…
```

javierarce:
For the feature flags I was thinking in using the same key we are using for the `dataServiceEnabled` check plus a prefix and a suffix:`analysis-` + key + `-enabled`.

So we could have:

- `analysis-data_observatory-enabled`
- `analysis-geocoder_internal-enabled`
- etc.

xavijam:
Sorry but I don't get you. Why do you need to specify it right now? I mean, it should be added when it is developed right? not automatically...

javierarce:
I thought that the original request was adding another check with the userModel, something like: 

```
…
checkIfEnabled: function (configModel, userModel) {
  return checkIfAnalysisIsEnabled('geocoder_internal', configModel, userModel);
},
…
```

```
var checkIfAnalysisIsEnabled = function (key, configModel, userModel) {
  if (!configModel) {
    throw new Error('configModel is required');
  }

  if (!userModel) {
    throw new Error('userModel is required');
  }

  return !!configModel.dataServiceEnabled(key) && userModel.featureEnabled(key);
};
```

xavijam:
[...]

PR #12082, opened 2017-05-03
Adding userModel parameter for analysis check-if-enabled

xavijam:
Related ticket: #12025 

Basically that ^^^. We will have the chance to check if an analysis is enabled by any userModel attribute from now on. Example, we can use it in `analyses.js` like:

```js
…
  'routing-to-single-point': {
    title: _t('analyses.routing.short-title'),
    FormModel: FallbackFormModel,
    checkIfEnabled: function (configModel, userModel) {
      return userModel.featureEnabled('routing-analysis');
    }
  },
…
```


xavijam:
cc @ivanmalagon

ivanmalagon:
Closing this PR. The code has been already merged in #12036

PR #12083, opened 2017-05-03
12048 admin user actions

### LINKED ISSUE #12048: Admin user actions

javitonino:
Closes #12048 

**Acceptance**
You need to enable the `organization-admins` FF for all users with access to this. I suggest making it unrestricted in staging.
I prefer not to give specific instructions for this one, as it is a big one with many moving parts. I'll give some directions instead:
- [ ] Tests that users cannot access anything
- [ ] Test that organization owners can do everything
- [ ] Test that organization admins can: manage users, groups and notifications, but not change organization/auth settings
  - [ ] Admins should not be able to edit other admins/owner, create new ones, etc. Only owner can do this.

List of functionality to test
- [ ] Org auth settings
- [ ] Org settings
- [ ] Org user management
  - [ ] Including user invitations (test a signup from an invitation sent by an admin!)
- [ ] EUMAPI
- [ ] Notifications
- [ ] Groups
- [ ] Dropdown menu has a link to "Your organization" (only in private part, the public part does not have this link in current master)

---

Progress tracking so I don't forget anything.

**Backend**
- [x] Allow admins to create/edit/remove other users
  - [x] But not admin/owner users
  - [x] Cannot create (or convert plain users into) admin users
  - [x] Tests
  - [x] EUMAPI
- [x] Invite other users
- [x] Manage groups
- [x] Notifications
- [x] Tests for negative case: admins cannot change org settings

**Frontend**
- [x] Show [ADMIN] tag in user list
screenshot_20170504_145119

- [x] Disable link to edit other admin users
screenshot_20170504_145135

- [x] Remove redirection when entering org admin pages and not being the owner
- [x] Show organization settings in side panel for admins
- [x] Show organization settings in avatar popup for admins (private parts, it's not present in public views menu)
- [x] Remove inaccessible tabs from org settings menu for admins
- [x] Hide `admin` toggler when creating/editing users as an admin

**Some interesting notes**
- Fixed a bug where you could change your own quota https://github.com/CartoDB/cartodb/pull/12083/commits/94c58138944d0bbe398358cd0e76382801833590
- Cleaned up the UserPresenter, there was a confusion between `current_user` and `current_viewer` https://github.com/CartoDB/cartodb/pull/12083/commits/e9559828973b1b8dedee6bbbe3279b6a99605f64. Changed options hash to keyword argument and added tests to cover this behaviour.


app/controllers/carto/api/user_presenter.rb
```diff
@@ -58,14 +60,8 @@ def to_public_poro
           viewer:           @user.viewer?
         }
 
-        if @options[:fetch_groups] == true
-          poro.merge!(groups: @user.groups ? @user.groups.map { |g| Carto::Api::GroupPresenter.new(g).to_poro } : [])
-        end
-
-        if @options[:current_user] && @options[:current_user].organization_owner? && @user.belongs_to_organization?(@options[:current_user].organization)
```
juanignaciosl: 🤔  No longer needed?

app/models/carto/invitation.rb
```diff
@@ -23,7 +23,7 @@ class Invitation < ActiveRecord::Base
     private_class_method :new
 
     def self.create_new(inviter_user, users_emails, welcome_text, viewer)
-      raise CartoDB::InvalidUser.new("Only owners can create invitations") unless inviter_user.organization_owner?
+      raise CartoDB::InvalidUser.new(" ") unless inviter_user.organization_admin?
```
juanignaciosl: How come `" "`?

app/views/admin/organization_users/_form.html.erb
```diff
@@ -86,7 +86,7 @@
 
     <div class="FormAccount-rowData">
       <div class="CDB-Text CDB-Size-medium u-rSpace--xl">
-        <input class="CDB-Radio js-userViewerOption" type="radio" name="user[viewer]" value="true" <%= @user.viewer? ? 'checked' : '' %> <%= (current_user.organization.remaining_viewer_seats > 0 || @user.viewer?) && !@user.organization_owner? ? '' : 'disabled' %> required>
+        <input class="CDB-Radio js-userViewerOption" type="radio" name="user[viewer]" value="true" <%= @user.viewer? ? 'checked' : '' %> <%= (current_user.organization.remaining_viewer_seats > 0 || @user.viewer?) && !@user.organization_owner? && !(@user.organization_admin? && !current_user.organization_owner?) ? '' : 'disabled' %> required>
```
juanignaciosl: Consider extracting this (long!) check to a method, maybe at a helper if it doesn't fit into the model.

app/controllers/carto/api/user_presenter.rb
javitonino: No. This was a hack added because `current_user` was being used instead of `current_user`.

For presenting a user from `carto/api/organizations_controller.rb`, before this PR:
1. `current_user` was passed
2. As there is no `current_viewer`, `to_poro` calls `to_public_poro` instead.
3. In `to_public_poro`, this lines were added as a hack to ensure the needed data is presented

After this PR:
1. The controller passes `current_viewer`
2. `to_poro` always renders the three fields listed here

The commit https://github.com/CartoDB/cartodb/pull/12083/commits/e9559828973b1b8dedee6bbbe3279b6a99605f64 has links to the PRs that introduced this.

app/models/carto/invitation.rb
javitonino: Fixing it, U_U

app/controllers/admin/organizations_controller.rb
```diff
@@ -145,7 +146,8 @@ def auth_update
 
   def load_organization_and_members
     @organization = current_user.organization
-    raise RecordNotFound unless @organization.present? && current_user.organization_owner?
+    raise RecordNotFound unless @organization.present? && current_user.belongs_to_organization?(@organization) &&
```
javitonino: Uhm, you are right that the check is redundant, as the organization already comes from the `current_user`. It looks like copy-paste from other parts of the code where the `@organization` comes from the `params`.

I'll simplify it, always looking at an excuse to remove multi-line statements.

app/controllers/admin/organization_users_controller.rb
```diff
@@ -264,13 +274,15 @@ def initialize_google_plus_config
 
   def check_permissions
     raise RecordNotFound unless current_user.organization.present?
-    raise RecordNotFound unless current_user.organization_owner? || ['edit', 'update'].include?(params[:action])
+    raise RecordNotFound unless current_user.organization_admin?
```
javitonino: Only kind of. If there is a non-org user with the `org_admin` flag set, this could be a problem. I propose the following:
- Remove the previous check as you suggested.
- Add a validation so `org_admin` cannot be set for non-org users.
- Maybe explicitly check for the organization in `organization_admin?`

This way we can remove some of these redundant checks.

app/controllers/admin/organization_users_controller.rb
```diff
@@ -75,6 +76,10 @@ def create
 
     # Validate password first, so nicer errors are displayed
     model_validation_ok = @user.valid_password?(:password, @user.password, @user.password_confirmation) && @user.valid?
+    if @user.organization_admin? && !current_user.organization_owner?
```
javitonino: The flag is also used for password validation, so maybe we cannot remove it completely. I'll try moving this to the model, even if still needs to be called explicitly from the controller, but at least we remove some of the duplication.

app/views/admin/organization_users/_form.html.erb
xavijam: Agree

lib/assets/core/javascripts/cartodb/organization/organization_users/organization_user.jst.ejs
```diff
@@ -8,6 +12,8 @@
           <%- username %>
           <% if (isOwner) { %>
             <span class="OrganizationList-userInfoLabel">OWNER</span>
+          <% } else if (isAdmin) { %>
+            <span class="OrganizationList-userInfoLabel grey">ADMIN</span>
```
xavijam: IMHO, it should be `OrganizationList-userInfoLabel--admin`.

lib/assets/core/javascripts/cartodb3/data/user-model.js
```diff
@@ -86,18 +86,25 @@ var UserModel = Backbone.Model.extend({
...
+  isOrgOwner: function () {
     if (this._organizationModel) {
       return this._organizationModel.getOwnerId() === this.get('id');
     }
     return false;
   },
 
+  isOrgAdmin: function () {
+    if (this._organizationModel) {
+      return this.get('org_admin') === true;
```
xavijam: We need to specify that a user is/isn't an admin in an org (not an owner). Do you think there is another place/name better for this purpose @nobuti @ivanmalagon?

nobuti: Could you give us more context? What are you worried about?

xavijam: I don't like `org_admin` name 😬 . Also, I'm not sure if we should expose that in the attributes root. I feel that info, although it would complicate it more, should be inside the organization model.

nobuti: Agree, something like `orgModel.isOrgAdmin(userModel)` should work.

xavijam: Yep... :)

xavijam:
Reviewed a little bit the organization users list:

screen shot 2017-05-08 at 12 20 53
screen shot 2017-05-08 at 13 04 30

javitonino:
@xavijam There is something weird going on when the email address is longer than the username + tags, it aligns the tags to the right of the email:

screenshot_20170508_134753

Can you take a look? 🙏

xavijam:
> @xavijam There is something weird going on when the email address is longer than the username + tags, it aligns the tags to the right of the email Can you take a look? 🙏 

Sure!

app/views/admin/organization_users/_form.html.erb
```diff
@@ -84,9 +84,10 @@
       <p class="FormAccount-rowInfoText">Build and read</p>
     </div>
 
+    <% admin_editing_admin = @user.organization_admin? && !current_user.organization_owner? %>
     <div class="FormAccount-rowData">
       <div class="CDB-Text CDB-Size-medium u-rSpace--xl">
-        <input class="CDB-Radio js-userViewerOption" type="radio" name="user[viewer]" value="true" <%= @user.viewer? ? 'checked' : '' %> <%= (current_user.organization.remaining_viewer_seats > 0 || @user.viewer?) && !@user.organization_owner? ? '' : 'disabled' %> required>
+        <input class="CDB-Radio js-userViewerOption" type="radio" name="user[viewer]" value="true" <%= @user.viewer? ? 'checked' : '' %> <%= (current_user.organization.remaining_viewer_seats == 0 && !@user.viewer?) || @user.organization_owner? || admin_editing_admin ? 'disabled' : '' %> required>
```
javitonino: I tried to simplify this a bit, extracting the `admin_editing_admin` variable, which also should make this more readable. Not 100% sold on extracting this to a helper, since the same conditions are repeated below, in order to display an appropriate message, but it's hard to evaluate.

Anyway, take a look as an external observer, and if you still think this should be extracted, I will. My preference would be a helper, since this check is very specific for this view.

javitonino:
Ready for second round of reviews.

app/controllers/admin/organization_users_controller.rb
```diff
@@ -263,14 +266,15 @@ def initialize_google_plus_config
   end
 
   def check_permissions
-    raise RecordNotFound unless current_user.organization.present?
-    raise RecordNotFound unless current_user.organization_owner? || ['edit', 'update'].include?(params[:action])
+    raise RecordNotFound unless current_user.organization_admin?
```
juanignaciosl: How come `['edit', 'update'].include?(params[:action])` was enough, before? Won't this change block some kind of individual edition?

app/controllers/carto/api/groups_controller.rb
```diff
@@ -134,12 +134,14 @@ def validate_organization_or_user_loaded
...
         unless @organization.id == current_user.organization_id
-          render json: { errors: ["Not organization owner"] }, status: 400
+          render json: { errors: ["Not organization user"] }, status: 400
         end
       end
 
-      def org_owner_only
-        render json: { errors: ["Not org. owner"] }, status: 400 unless @organization.owner_id == current_user.id
+      def org_admin_only
+        unless current_user.belongs_to_organization?(@organization) && current_user.organization_admin?
```
juanignaciosl: Minor

`current_user.belongs_to_organization?(@organization) && current_user.organization_admin?` happens often. Maybe this would be useful and less error-prone: `@organization.admin?(current_user)`.

app/controllers/carto/api/user_presenter.rb
juanignaciosl: PS: " because current_user was being used instead of current_user" -->  "because current_user was being used instead of current_viewer"?

app/models/carto/user.rb
```diff
@@ -32,7 +32,8 @@ class Carto::User < ActiveRecord::Base
                    "users.api_key, users.database_schema, users.database_name, users.name, users.location," \
                    "users.disqus_shortname, users.account_type, users.twitter_username, users.google_maps_key, " \
                    "users.viewer, users.quota_in_bytes, users.database_host, users.crypted_password, " \
-                   "users.builder_enabled, users.private_tables_enabled, users.private_maps_enabled".freeze
+                   "users.builder_enabled, users.private_tables_enabled, users.private_maps_enabled, " \
+                   "users.org_admin".freeze
```
juanignaciosl: 🙄

app/views/admin/organization_users/_form.html.erb
juanignaciosl: IMHO, if something is _actually_ duplication it should be extracted, but I know I'm quite obsessive about that. If you saw the extraction and disliked it, let it be.

juanignaciosl:
PS: great full stack work :D

app/controllers/admin/organization_users_controller.rb
javitonino: Yes, this blocks that individual edition. It was allowing plain organization users to edit their own quota, as long as they knew the URL and disabled JavaScript (there was a JS redirection to get out of this page if you accidentally got to it).

app/controllers/carto/api/groups_controller.rb
javitonino: Taking a look. I saw this during development and I thought about `@user.organization_admin?(@organization)` but it felt off. Putting it in the organization model looks better to me, to the point I'm actually going to change it right now.

Thanks for the suggestion! ❤️

app/controllers/admin/organization_users_controller.rb
juanignaciosl: Oh, that's the one you told me about! I didn't realize. 👍

lib/assets/core/javascripts/cartodb3/data/user-model.js
```diff
@@ -86,18 +86,25 @@ var UserModel = Backbone.Model.extend({
...
 
-  isOrgAdmin: function () {
+  isOrgOwner: function () {
     if (this._organizationModel) {
       return this._organizationModel.getOwnerId() === this.get('id');
     }
     return false;
   },
 
+  isOrgAdmin: function () {
```
xavijam: :kiss:

lib/assets/core/javascripts/cartodb3/data/organization-model.js
```diff
@@ -30,5 +30,11 @@ module.exports = Backbone.Model.extend({
 
   getOwnerEmail: function () {
     return this._ownerModel && this._ownerModel.get('email');
+  },
+
+  isOrgAdmin: function (user) {
```
xavijam: :kiss:

javitonino:
Backend tests need a quick review after the extraction of `@organization.admin?` (it seems we need some `try`. On it.

ivanmalagon:
Problems found:

- Trying to create a user with the email `spliffstar-07@gmail.com` triggers a validation error but the address is correct based on the requirements.

image

... Acceptance still on its way ...

javitonino:
Some nice finds by @ivanmalagon up there. The quotas one was tricky for sure. I pushed some comments merging master, addressing his comments and a further one I found (soft limits togglers not correctly deactivated in some corner case which should not be possible unless you have a dirty dirty DB).

Anyway, let's wait for 🍏 and go again.

app/views/admin/organization_users/_form.html.erb
```diff
@@ -169,11 +178,11 @@
       </div>
       <div class="FormAccount-rowData">
         <div class="Toggler">
-          <%= f.check_box :soft_geocoding_limit, id: "soft_geocoding_limit", disabled: !current_user.soft_geocoding_limit %>
+          <%= f.check_box :soft_geocoding_limit, id: "soft_geocoding_limit", disabled: !current_user.organization.owner.soft_geocoding_limit %>
```
juanignaciosl: Minor: why not stopping this madness of limits propagation? Suggestion (for this and other cases, not only for limits but also `current_user.organization` everywhere):

```ruby
@organization.soft_geocoding_limit

# which needs...

@organization = current_user.organization

# which needs...

def soft_geocoding_limit
  owner.try(:soft_geocoding_limit)
end
````

javitonino: The only reason I had not to do it was to minimize changes as this is already pretty big. But if the CRer is ok with it... refactor time!

javitonino: I changed this as discussed, only in this part of the code (another interesting change would be in the soft limits validation, but the check is done more cleanly there, and it's not such a clear refactor).

Some notes:
- Data observatory soft limits are missing from the organization user form
- It's easy for the owner to disable soft limits for himself, which then makes it impossible to edit other users soft limits :/ But it has been like this forever, so...

javitonino:
Implemented the suggested refactor.

javitonino:
No frontend changes since last CR, moving to acceptance.

ivanmalagon:
Did acceptance again.

This time the button appears, I'm able to modify quotas being an admin and I've played with the `extra` buttons. Everything works as expected.

Bandera 🇯🇲

juanignaciosl:
Then, only @noguerol Seal of Approval™ is pending. This should cover everything about multiple admin but the feature of organization deletion.

juanignaciosl:
Merged master to remove conflict cc @xavijam @javitonino

ivanmalagon:
@noguerol came to us with a hammer but even with this violent attitude he approved this. So another bandera. 🇸🇪

xavijam:
:dancer: 💃

Issue #12085, opened 2017-05-03
Reset styles per node is not working anymore after adding a new analysis


xavijam:
After checking @AbelVM 's summit demo, we realized that the "hacky" thing we introduced to reset the styles when a new node was added is not working properly. 

Seems like when we added this CartoDB/cartodb/pull/11989, we broke the mentioned one.

## STR
1. Create a map of points, for example.
2. Style the layer, for example, make a choropleth map.
3. Add a 'filter by column value' analysis (you can try with any other).
4. After finishing it, go to styles tab.
5. Or none style is selected or it says there is no styles, also legend is still there.

cc @ivanmalagon because we mentioned after the talk/demo and @nobuti

PR #12086, opened 2017-05-03
It is not necessary to trigger fetchs for reset-style-per-node when node itself launches one when it is ready/finished

xavijam:
Related ticket: https://github.com/CartoDB/cartodb/issues/12085

When we introduced https://github.com/CartoDB/cartodb/pull/11989, we didn't realize we f***** up reset-styles thing. Because it was resetting the status of the query objects from the deep-insights-integration node changes (https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/deep-insights-integrations.js#L347-L375).

In the future, we will remove this and we will apply the specific style for each type of analysis (cc @saleiva @makella).

In order to test this, I'd check that when a new analysis is added, reset styles works properly. Visit styles panel and check that the simple one is selected when the analysis is finished.

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/helpers/reset-style-per-node.js
+++ b/lib/assets/core/javascripts/cartodb3/helpers/reset-style-per-node.js
@@ -83,8 +83,13 @@ module.exports = function (nodeDefModel, layerDefModel, forceStyleUpdate, resetQ
 
     saveDefaultStylesIfStillRelevant();
 
-    queryGeometryModel.bind(CHANGE_STATUS, saveDefaultStylesIfStillRelevant);
-    querySchemaModel.bind(CHANGE_STATUS, saveDefaultStylesIfStillRelevant);
+    if (!queryGeometryModel.isFetched()) {
+      queryGeometryModel.bind(CHANGE_STATUS, saveDefaultStylesIfStillRelevant);
+    }
+
+    if (!querySchemaModel.isFetched()) {
+      querySchemaModel.bind(CHANGE_STATUS, saveDefaultStylesIfStillRelevant);
+    }
   };
 
   if (resetQueryReady) {
--- a/lib/assets/core/javascripts/cartodb3/helpers/reset-style-per-node.js
+++ b/lib/assets/core/javascripts/cartodb3/helpers/reset-style-per-node.js
@@ -83,20 +83,8 @@ module.exports = function (nodeDefModel, layerDefModel, forceStyleUpdate, resetQ
 
     saveDefaultStylesIfStillRelevant();
 
-    if (!queryGeometryModel.isFetched()) {
-      queryGeometryModel.bind(CHANGE_STATUS, saveDefaultStylesIfStillRelevant);
-    }
-
-    if (!querySchemaModel.isFetched()) {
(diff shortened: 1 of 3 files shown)


nobuti:
NEWS?

xavijam:
Added!

lib/assets/core/javascripts/cartodb3/helpers/reset-style-per-node.js
```diff
@@ -90,13 +90,6 @@ module.exports = function (nodeDefModel, layerDefModel, forceStyleUpdate, resetQ
     if (!querySchemaModel.isFetched()) {
       querySchemaModel.bind(CHANGE_STATUS, saveDefaultStylesIfStillRelevant);
     }
-
-    if (queryGeometryModel.shouldFetch()) {
```
nobuti: How did you find this solution? Could this cause the same issue to something else in somewhere else? I mean, we are fetching those models in a few places as well.

xavijam: That's a good question. We are fetching those objects in several places, but it is true in the code we added when analysis has changed and ready (https://github.com/CartoDB/cartodb/pull/11989), we *force* (not check if we should or not) to fetch them and it makes some other places fail. Like this example.

And it is true (too), that we could make that change in that place, just checking if the fetch should be triggered or not.

nobuti: >  it makes some other places fail.

Should we fix those as well as part of this PR? Do we get them identified?

xavijam: I feel there isn't any other. But, as I've said, maybe it is worth it to change the way we fetch all query objects when analysis has changed and is ready... what do you think?

nobuti: I think it's worth if it supposed fewer side effects.

xavijam: Arg, I'm remembering that there is another thing. Due to the fact that we have in deep-insights-integration the "analysis" attributes update (https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/deep-insights-integrations.js#L347-L375), when something happens, the status of the query objects changes to 'unfetched', and that's the thing that is not working properly for us (https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/helpers/reset-style-per-node.js#L53-L54).

javitonino:
Hey, I am testing this and I cannot reproduce the issue in production with the steps from the ticket. Is this something that happens always? Only some of the time?

xavijam:
Oh! 😱 , let me check it again.

xavijam:
Well, it happens. I've tried this time with 'Filter by column value' and I have added more info about the problem in the issue description.

screen shot 2017-05-05 at 12 21 21

javitonino:
I have been able to reproduce with a bigger dataset and a more complex analysis. Continuing..., thanks!

javitonino:
Looking good. I tested with multiple analysis, including those which change geometry type. Also with different initial geometries (points, lines, polygons). It resetted styles every time. ✨ Great job, guys :)

PR #12092, opened 2017-05-04
Recreate table overviews on synchronizations

jgoizueta:
When a sync table had overviews created on the initial import, these persisted unmodified for ensuing synchronizations,
while the newly updated base table was lacking the _feature_count column. So we had two problems:
* Base table without _feature_count is inconsistence with the existence of overview tables
* Overviews tables where out of date with base table


NEWS.md
```diff
@@ -4,6 +4,7 @@ Development
 * An attacker could execute commands in the server running the queues by importing a file with a carefully crafted filename. Fixed in #11782
 
 ### Features
+* Overiviews are synchronized now (#12092)
```
ethervoid: Typo s/Overiviews/Overviews

services/importer/lib/importer/overviews.rb
```diff
@@ -79,6 +88,10 @@ def create_overviews!
           @service.create_overviews! @table_name
         end
 
+        def delete_overviews!
```
ethervoid: We could add here an `update_overviews` method that act as recreate

app/models/synchronization/adapter.rb
```diff
@@ -48,6 +50,28 @@ def run(&tracker)
         raise exception
       end
 
+      def recreate_overviews(result)
```
ethervoid: We have two methods to deal with overviews, one in the importer and this one. 

Maybe we should move this logic to [the overviews class](https://github.com/CartoDB/cartodb/blob/master/services/importer/lib/importer/overviews.rb). There we could do exactly what we're doing here or in [here](https://github.com/CartoDB/cartodb/blob/master/app/connectors/importer.rb#L116-L128)

app/models/synchronization/adapter.rb
```diff
@@ -36,6 +37,7 @@ def run(&tracker)
           overwrite(table_name, result)
           setup_table(table_name, geo_type)
           run_index_statements(index_statements)
+          recreate_overviews(result)
```
ethervoid: Here we could do: `@overviews_creator.dataset(result.name).update_overviews()` for example

NEWS.md
jgoizueta: Fixing it, thanks!

app/models/synchronization/adapter.rb
jgoizueta: After having taken a look at this I also think we should move the overview-recreation logic out of here. But I think I'll keep this method here handling exceptions.

jgoizueta: I'm no longer sure if I prefer the current situation ir moving logic out of here! 😅 

Let's see:

The `Overviews` class takes care of  overviews maintenance logic, exposed through Overviews::Dataset.
(and using the low level overview operations encapsulated in `OverviewsService`)

Consumers of overviews are:
* Imports (`Connector::Importer` app/connectors/importer.rb)
* Syncs (`Synchroniztion::Adapter` app/models/synchronizations/adapter.rb)

And they should just request overview maintenance and handle exceptions.

### A. (current situation)

* `should_create_overviews?` information is independent of overview creation.

This allows adding logic in the consumers, such as additional logging, or aborting sync creation if `interval` is too short for overviews

### B. (proposed)

`Overviews` knows if overviews should be created so it handles this internally:

```ruby
def create_overviews
  @service.create_overviews! if should_create_overviews? 
end

def update_overviews
  delete_overviews!
  create_overviews
end
```

* Then we cannot easily do additional logging at the consumer.
* To allow sync limits we should continue exposing `should_create_overviews?` (probably with a better name: `will_have_overviews?`)

I'll hold off this decision, in hope that tomorrow there will be not so many bands doing soundchecks almost inside CARTO :guitar:. Feedback is welcome in the meantime.

spec/models/synchronization/member_overviews_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+require_relative '../../../services/data-repository/repository'
+require_relative '../../../app/models/synchronization/member'
+require 'helpers/unique_names_helper'
+require 'helpers/file_server_helper'
+require_relative '../../helpers/feature_flag_helper'
+
+include UniqueNamesHelper
+include CartoDB
+
+describe Synchronization::Member do
```
jgoizueta: Yep. Sure. Lots of lines.

jgoizueta:
I've added some tests, fixed a problem and refactored a little bit the overviews-handling logic. @ethervoid: a little review?

rochoa:
What is the status of this PR? I see changes were requested in the CR.

ethervoid:
@rochoa My fault, he asked me for a review. I'll do it in a minute and we'll move this forward

ethervoid:
@jgoizueta I've checked the changes and looks good to me, good catch with the bug

rochoa:
🎉

PR #12108, opened 2017-05-10
Fix analysis quota view leak

nobuti:
This PR fixes a leak related to the quota analysis view. In some scenarios, we found the analysis-control-view is not properly removed when the parent view was removed. This could cause an inconsistent state of this view when the analysis finishes or fails. Basically, it gets stuck in the fetching state.

### Context

In analyses-view.js we have [a bind](https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-view.js#L129) for when the analysis form collection is reset or an analysis is removed. When an analysis is done, [the source changes](https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/layer-content-analyses-view.js#L123), so we [reset the analysis form collection](https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/layer-content-analyses-view.js#L203) with the new data. This follows [a complete render of the parent view](https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-view.js#L132-L153). But the thread in the actual view is not finished, so when it get back to [this point](https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/layer-content-analyses-view.js#L204) the former view is rendered again, this time in memory because it was removed from the DOM when the parent view was re-rendered, but still listening and rendering subviews.

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/components/tab-pane/tab-pane-view.js
+++ b/lib/assets/core/javascripts/cartodb3/components/tab-pane/tab-pane-view.js
@@ -23,6 +23,9 @@ module.exports = CoreView.extend({
   },
 
   render: function () {
+    this.clearSubViews();
+    this.$el.empty();
+
     this.$el.html(this.template(this.options));
 
     this._renderCollection();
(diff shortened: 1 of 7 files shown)


nobuti:
As far as I know, no news needed.

xavijam:
Tested in staging and working like a charm

xavijam:
When green, I'll merge it 💃 . Good catch man!

xavijam:
Frontend passing, merging 💃 .

PR #12119, opened 2017-05-12
Adding z-index to file-input button

xavijam:
Related ticket: https://github.com/CartoDB/support/issues/690.

Silly change, and certainly, I have no idea why it was not already there.


xavijam:
Tested in Chrome, Firefox, Safari and Opera, working as expected, merging.

Issue #12029, opened 2017-04-21
Export image button is not disabled while generating


javitonino:
STR
1. Got to export image
2. Click the export button quickly a bung of times (even while the button is disabled "generating")

You will get a bunch of popups to download the image. You should only get one.

![out](https://cloud.githubusercontent.com/assets/7569673/25270249/36aec5b6-2680-11e7-8701-aed7a300e408.gif)

Chrome is a bit smarter than Firefox, but the download is still triggered multiple times. Instead of downloading a bunch of times, the following is shown:
![screenshot_20170421_105113](https://cloud.githubusercontent.com/assets/7569673/25270332/8ac67a4a-2680-11e7-98ce-767add8b08aa.png)

xavijam:
deployed

PR #12120, opened 2017-05-12
12029 disable export

nobuti:
This PR implements #12029.


xavijam:
Tested in local (not need to check it in staging). All good 👍 .

PR #12125, opened 2017-05-12
Link map and table view when adding and removing features.

### LINKED ISSUE #11759: map is not being refreshed after row deletion in layer view
### Context
when removing a row with geometry neither the panel nor the map is being refreshed

if the url is reloaded the map works as expected 

### Steps to Reproduce

1. create a map
2. enter an empty layer
3. add a feature (point)
4. go to table view (bottom right action button)
5. remove row with geometry
6. go to map view (bottom right action button)
 

### Current Result
panel has point forms, and point is still represented in the map

### Expected result
map should refresh and style panel should warn the layer has no geometry, no errors in the console

### Additional info

5a403d75-1f2b-46fd-aa51-01b234afe44d
b271e3f2-7590-4fe2-8c98-fac80d7ec8f7

/cc @xavijam this is happening in production, indeed

### LINKED ISSUE #11978: Data view doesn't update with custom geometries
### Context
When adding custom geometries, data view is not updated nor provided in the case it was originally empty.

### Steps to Reproduce

1. Create a new map with an empty layer
2. Add some points via geometry edition
3. Switch to table view

### Current Result
No data

### Expected result
The resulting point table should be shown

nobuti:
This PR fixes #11978 and fixes #11759.


lib/assets/core/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -319,6 +328,8 @@ F.prototype._onAnalysisDefinitionNodeIdChanged = function (m, changedAttributes)
 };
 
 F.prototype._onAnalysisDefinitionNodeRemoved = function (m) {
+  m.queryRowsCollection.on('remove', this._invalidateMap, this);
```
xavijam: `off` instead of `on`?

lib/assets/core/javascripts/cartodb3/editor/layers/edit-feature-content-view.js
```diff
@@ -272,8 +272,8 @@ module.exports = CoreView.extend({
     });
   },
 
-  _onSaveFeatureSuccess: function () {
-    this._onUpdateFeature('save');
+  _onSaveFeatureSuccess: function (op) {
```
xavijam: If it means operation, let's use `operation` 🍇 .

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindows-view.js
```diff
@@ -101,11 +102,11 @@ module.exports = CoreView.extend({
   },
 
   _onStatusChanged: function () {
-    if (this._querySchemaModel.isFetched() &&
-        this._queryGeometryModel.isFetched() &&
-        this._queryRowsCollection.isFetched()) {
-      this.render();
-    }
+    fetchAllQueryObjectsIfNecessary({
```
xavijam: This is not the same, it will fetch the query objects is possible, and then it will render it. Are we expecting that?

lib/assets/core/javascripts/cartodb3/helpers/fetch-all-query-objects.js
```diff
@@ -17,11 +23,15 @@ module.exports = function (params, callback) {
...
       params.queryRowsCollection.fetch(opts);
     } else {
-      callback && callback();
+      // we need to check if all are fetched because shouldFetch include the fetching state
+      allFetched() && callback && callback();
     }
   };
 
-  if (params.queryGeometryModel.shouldFetch()) {
+  var rows = params.queryRowsCollection.size();
```
xavijam: Take into account that at that moment, queryRowsCollection has not been fetched...

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindows-view.js
nobuti: Well, take this case in mind. We have a map with one point, we are in the style panel (or infowindow) and we change the map view to the table view. Then we remove the row, we should render the content panel in order to show the `No geometry` thing.

lib/assets/core/javascripts/cartodb3/deep-insights-integrations.js
nobuti: Yes, my bad.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/infowindow/infowindows-view.js
xavijam: Ok, gotcha.

lib/assets/core/javascripts/cartodb3/helpers/fetch-all-query-objects.js
nobuti: Even in that case, rows will be 0, I think it will work as same as today.

lib/assets/core/javascripts/cartodb3/deep-insights-integrations.js
xavijam: I'll do the change.

xavijam:
Working pretty good man. I've tested it in local and in staging.

Issue #8468, opened 2016-07-04
Auto-styling on widgets shoulnd't be available when having a point dataset aggregated by hexabins, squares or admin regions.


saleiva:

saleiva:
Looks like this could be happening again
https://www.fullstory.com/ui/2A7A7/session/5402923928387584:(5676081478631424!5629499534213120(5707702298738688(763015)))

saleiva:
Yep I can replicate it. After aggregating in the style panel, if you go to the data panel of the layer, and select one of the columns, when adding the widget, the auto-style button will still be there.

saleiva:
Would love to prioritize this, but leaving this to you guys :)

xavijam:
Deploying to ded04 for testing purposes.

PR #12127, opened 2017-05-15
Autostyle aggregation

nobuti:
This PR implements #8468. It relays on https://github.com/CartoDB/deep-insights.js/pull/541, so merge it before deploy it. Also don't forget to point package to the right branch.


lib/assets/core/javascripts/cartodb3/editor.js
```diff
@@ -284,8 +284,7 @@ var mapCreation = function () {
     show_empty_infowindow_fields: true,
     state: stateJSON,
     interactiveFeatures: true,
-    layerSelectorEnabled: false,
-    autoStyle: true
```
xavijam: We should do the same for the public view. I'll do it.

xavijam:
Merging!

Issue #12144, opened 2017-05-17
User should not be able to change the name of a shared dataset


arredond:
### Context
*Please explain here below what you were doing when the issue happened*

When editing the name of a table that has been shared by another organization user with read/write permission, the browser's URL changes from:

`https://carto.lan/user/player1/dataset/**carto**.f5xm8i41fotsozyu85kqjq_points`

to:

`https://carto.lan/user/player1/new_name`

Since it loses the owner's name in the URL, the dataset becomes uneditable (although visible). If you refresh the page, you get a 404. When going back to the DATASETS view and clicking on the dataset with the new name (in X SHARED WITH YOU), the correct URL is loaded and the map can be edited:

`https://carto.lan/user/player1/dataset/**carto**.new_name`

### Steps to Reproduce
*Please break down here below all the needed steps to reproduce the issue*

1. Create two accounts in an on-prem instance and share a dataset from one to the other with write permission
2. Log into the **other** and change the dataset's name
3. Try editing a cell (will fail). Try refreshing

### Current Result
*Please describe here below the current result you got*

Browser refreshes to a bad link when editing a different user's dataset name

### Expected result
*Please describe here below what should be the expected behaviour*

Browser should refresh to the correct link, including the other user's name

### Browser and version
*What internet browser (Chrome, Firefox, etc) and version was you using and version*

Chrome, 56.0.2924.87

xavijam:
The problem is that a user *CAN'T* change the name of a dataset, no matter if the user has read&write permissions.

We have to disable the edition after double click.


<img width="449" alt="screen shot 2017-05-17 at 15 25 48" src="https://cloud.githubusercontent.com/assets/132146/26156056/5b2841e8-3b15-11e7-88b0-d36f91c0bd85.png">
<img width="426" alt="screen shot 2017-05-17 at 15 25 52" src="https://cloud.githubusercontent.com/assets/132146/26156055/5b253afc-3b15-11e7-9799-373abb19cec4.png">

nobuti:
Deployed.

PR #12146, opened 2017-05-17
Don't provide a way to rename a dataset shared with you

xavijam:
Related ticket: #12144 

If a user A shares a dataset with read&write permissions with a user B, dataset view should NOT provide a way to rename the dataset. It is true user B can edit the data, but not the name or the sync options.


nobuti:
It works great ✨

nobuti:
To be deployed first thing in the morning.

PR #12147, opened 2017-05-17
Disable geometry buttons instead of hiding

### LINKED ISSUE #11543: Disable geometry edition button in read-only layers

rubenmoya:
Fixes #11543

test

### DIFF
--- a/lib/assets/core/test/spec/cartodb3/editor/layers/layer-content-view.spec.js
+++ b/lib/assets/core/test/spec/cartodb3/editor/layers/layer-content-view.spec.js
@@ -144,6 +144,7 @@ describe('editor/layers/layer-content-view/lcv', function () {
   describe('._renderContextButtons', function () {
     var analysisDefNodeModel;
     var layerDefinitionModel;
+    var queryGeometryModel;
     function hasTipsy (subviews) {
       var newGeomItems = _.filter(subviews, function (elem) {
         return elem.el.classList.contains("js-newGeometryItem");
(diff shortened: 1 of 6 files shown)


lib/assets/core/test/spec/cartodb3/editor/layers/layer-content-view.spec.js
```diff
@@ -147,8 +143,23 @@ describe('editor/layers/layer-content-view/lcv', function () {
 
   describe('._renderContextButtons', function () {
     var analysisDefNodeModel;
+    var layerDefinitionModel;
+    var queryGeometryModel;
+    function hasTipsy (subviews) {
```
nobuti: nice!

rubenmoya:
It makes sense, to move it to the container, we thought about it but decided to put it in the items just in case we wanted to change the text later, but you're right.

nobuti:
Works like a charm ✨

nobuti:
Congrats @rubenmoya @rjimenezda, your first PR is about to be merged 🎉

xavijam:
VAMOS!

Issue #12057, opened 2017-04-25
"Find nearest" analysis lack of support for being applied to source as target


AbelVM:
### Context
Let's say you want to apply "Find nearest" to get the 5 closest neighbors of your own kind to make some aggregations. Like "give me the average KPI of my area".

You can't. There is no way to select the source node as target in the UI

### Steps to Reproduce
Just try to find nearest features in the same node you're using as source

cc @ivanmalagon @saleiva @noguerol

saleiva:
hi hi,

I am not sure this is the correct analysis for those use cases. BUT, you can still run it agains the dataset (not the layer tho). When you "give me the avg KPI for my area" you mean with a widget, right?

Give us more context! :)

AbelVM:
Yep, maybe. Just think on getting the highest value of the "neighborhood" of each feature, being that the current feature plus its 5 closest neighbors. You can even use this as a way of "smoothen" your distribution or finding missing values (your feature may have a null value for that field)

saleiva:
Uhm... for that I'd think on something like "summarize nearby"...

AbelVM:
2nd use case: a unique dataset of fast-food restaurants, categorized per brand: I want to know my closest competitor per brand.

noguerol:
IMO those use cases sound pretty uncommon, as clients will almost always get their own data from a different table than competitors, and of there is an exception they can always reconnect the table. That said, duplicating the table is not optimal and from a long term perspective enabling source as target is not a bad idea as may open some future use cases, but not a priority right now I think.

AbelVM:
:+1:  That's it. It looks quite simple to enable the source as target, but no rush.

arredond:
Hey guys, just wanted to let you know that not only Abel and I have been wanting to use this analysis with matching source and target layers. [Here's a SB ticket](https://cartodb.supportbee.com/tickets/12328381) from GSTV (if I'm correct, the guys who sponsored this analysis) where they state that _"For each GSTV station, I want to find the closest other GSTV station"_

I've used this analysis four or five times since it was released and, in all cases except one, I've wanted to do a "Self Nearest" as well. It really isn't that uncommon, and the workaround (which is what GSTV is using) of adding the layer again, then filtering out the results with distance = 0, is really ugly.

noguerol:
Understood and prioritized.

nobuti:
Deployed.

saleiva:
[...]

PR #12150, opened 2017-05-18
Possibility to include the source node in the node/dataset options list

xavijam:
Related ticket: https://github.com/CartoDB/cartodb/issues/12057

In the closest analysis we need to provide the source node as an option in the target list, that's all. By default, we will not provide it (in any other analysis), but it could be added by the `includeSourceNode` parameter.

And about how to test it, we should check we provide that option in that analysis and also check if in the rest of them we don't provide it in the target editor.


nobuti:
Accepted!

Issue #12133, opened 2017-05-16
Once you open an operators component, you can't edit it anymore


xavijam:
Spotted in this [PR](https://github.com/CartoDB/cartodb/pull/12106), but it is happening right now in production.

## STR
- Create a map of points (with just a few is enough).
- Go to style tab.
- Apply a "squares" style.
- Change the aggregation options and select `operation` by `sum` `cartodb_id`.
- Done that, try to change the aggregation option again.
- Dropdown doesn't appear again.

xavijam:
![screen shot 2017-05-16 at 13 28 56](https://cloud.githubusercontent.com/assets/132146/26103855/a0c97b70-3a3b-11e7-8d9c-7467859ec32f.png)

xavijam:
Deployed!

AleGoiko:
@asimcox i dont think so, it should be on carto staging though

PR #12152, opened 2017-05-18
Hide operator list view instead destroy it.

nobuti:
This PR implements #12133.

In order to test it:
- [x] Check edition is working in style panel.
- [x] Check analyses that use the component:
    - [x] intersect layer
    - [x] centroid
    - [x] group points into polygons


xavijam:
Tested in local, not need to check it in staging.

PR #12153, opened 2017-05-18
Engine enabled should be inherited from organization by default

javitonino:
https://github.com/CartoDB/cartodb-management/issues/4941

Organization users created from the cloud were getting engine enabled by default, ignoring organization settings.

**CR**
Don't forget to CR https://github.com/CartoDB/cartodb-central/pull/1772 as well

**Acceptance**
Deploy this and https://github.com/CartoDB/cartodb-central/pull/1772
- [x] Org user creation (as an owner). Check `engine_enabled` in cloud and central (should be `nil`)
- [x] Org user signup. Check `engine_enabled` in cloud and central (should be `nil`)
- [x] Normal signup. Check `engine_enabled` in cloud and central (should be `false` as in FREE price plan)

**Deploy**
Deploy cartodb and central.
Do not move this task to done until we fix the affected users (see the management ticket)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -251,6 +251,7 @@ Development
 * Fixed problem resetting styles per node after adding a new analysis (#12085)
 * Docs, fixed some minor spelling and grammar errors in the content.
 * Docs, updated "More Info" url from street addresses georeference options to new, related guide.
+* Organizations users now get engine_enabled from the organization by default (#12153)
 * Color picker disappears in CartoCSS editor after clicking (#12097).
 
 ### NOTICE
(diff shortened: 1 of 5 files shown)


mayoral:
I've created a new organization in central with the engine flag disabled. After that, I created some users and in fact, they get the "Discover Engine" module, but when I see the info about that users in central they still show the engine flag enabled.

Checking in the database, the box shows the correct flags (user flag is nil and organization flag is false), but when checking this same data in central, the user flag is true and the organization flag is false. So, for some reason, the propagation to central isn't working OK.

Moving back to development.

mayoral:
Acceptances test are all OK!

👍 

Moving to Deploy.

Issue #12132, opened 2017-05-16
Loading button


piensaenpixel:
We need to implement the new loader button
- Related: https://github.com/CartoDB/design/issues/658

All the code is merged in CartoAssets
https://github.com/CartoDB/CartoAssets/pull/161

If you have any questions ping me :)

xavijam:
Could you enumerate the places where we should start using it?

piensaenpixel:
sure they are here https://github.com/CartoDB/design/issues/658
but this is the list
- [ ] Run analysis.
- [ ] Update the map.
- [ ] Export as an image.
- [ ] Apply CSS.
- [ ] Apply SQL.

xavijam:
Hey @nobuti, I feel we will need a small explanation about how to update cartoassets package and npm-shrinkwrap thing.

rubenmoya:
- [ ] Run analysis.
- [ ] Update the map.
- [x] Export as an image.
- [ ] Apply CSS.
- [ ] Apply SQL.
- [ ] Publish

rubenmoya:
Since we're showing the loading icon, we won't show the "loading", "publishing", "exporting" texts anymore, right?

xavijam:
@piensaenpixel @urbanphes  ^^^ ?

piensaenpixel:
yeah that's right @rubenmoya

rubenmoya:
@piensaenpixel For the run analysis case, should it show the loading indicator as long as the analysis has not finished?

nobuti:
Deployed.

xavijam:
NICE!

On Tue, May 23, 2017 at 15:30, Buti <notifications@github.com> wrote:
Closed #12132 [https://github.com/CartoDB/cartodb/issues/12132] .

piensaenpixel:
\o/

PR #12156, opened 2017-05-18
Use new loading buttons styles

rubenmoya:
Uses the new loading buttons in Builder as requested in #12132.

Example:
loading_button

Places where it has to be updated:
- [x] Run analysis.
- [x] Update the map.
- [x] Publish
- [x] Apply CSS.
- [x] Export as an image.
- [x] Apply SQL.

### How to test
Check the list above. In addition to this, check if we should update the analysis quota view as well.

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/components/modals/publish/publish-button.tpl
+++ b/lib/assets/core/javascripts/cartodb3/components/modals/publish/publish-button.tpl
@@ -1,22 +1,17 @@
-<button class="Publish-modalButton CDB-Button CDB-Button--primary CDB-Button--small u-rSpace js-button <% if (isDisabled) {%>is-disabled<% } %>">
+<button class="Publish-modalButton CDB-Button CDB-Button--loading CDB-Button--primary CDB-Button--small u-rSpace js-button
+  <% if (isDisabled) {%> is-disabled<% } %>
+  <% if (isLoading) {%> is-loading<% } %>
+">
   <div class="CDB-Button-Text CDB-Text is-semibold CDB-Size-small u-upperCase u-flex">
-    <% if (isLoading) { %>
-      <div class="CDB-LoaderIcon CDB-LoaderIcon--small u-iBlock">
-        <svg class="CDB-LoaderIcon-spinner" viewBox="0 0 50 50">
-          <circle class="CDB-LoaderIcon-path" cx="25" cy="25" r="20" fill="none"></circle>
-        </svg>
-      </div>
-      <% if (isPublished) { %>
-      <%- _t('components.modals.publish.updating-btn') %>
-      <% } else { %>
-      <%- _t('components.modals.publish.publishing-btn') %>
-      <% } %>
+    <% if (isPublished) { %>
+    <%- _t('components.modals.publish.update-btn') %>
     <% } else { %>
-      <% if (isPublished) { %>
-      <%- _t('components.modals.publish.update-btn') %>
-      <% } else { %>
-      <%- _t('components.modals.publish.publish-btn') %>
-      <% } %>
+    <%- _t('components.modals.publish.publish-btn') %>
     <% } %>
+  </div>
+  <div class="CDB-Button-loader CDB-LoaderIcon is-white">
+    <svg class="CDB-LoaderIcon-spinner" viewbox="0 0 50 50">
+      <circle class="CDB-LoaderIcon-path" cx="25" cy="25" r="20" fill="none"/>
+    </svg>
   </div>
 </button> <%- publishedOn %>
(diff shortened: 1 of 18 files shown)


xavijam:
Only 17 files? 😱 :joy:

lib/assets/core/javascripts/cartodb3/components/modals/publish/publish-button.tpl
```diff
@@ -1,22 +1,17 @@
...
-      <%- _t('components.modals.publish.update-btn') %>
-      <% } else { %>
-      <%- _t('components.modals.publish.publish-btn') %>
-      <% } %>
+    <%- _t('components.modals.publish.publish-btn') %>
     <% } %>
   </div>
-</button> <%- publishedOn %>
\ No newline at end of file
+  <div class="CDB-Button-loader CDB-LoaderIcon CDB-LoaderIcon--small is-white">
```
nobuti: are we rendering always the loader? Not sure this is a good idea in terms of performance.

lib/assets/core/javascripts/cartodb3/components/undo-redo/undo-redo-view.js
```diff
@@ -21,6 +21,7 @@ module.exports = CoreView.extend({
     this._trackModel = opts.trackModel;
     this._editorModel = opts.editorModel;
     this._clearModel = opts.clearModel;
+    this._applyStatusModel = opts.applyStatusModel;
```
nobuti: in what scenario can this model be optional?

lib/assets/core/javascripts/cartodb3/components/undo-redo/undo-redo-view.js
```diff
@@ -54,6 +56,10 @@ module.exports = CoreView.extend({
       this._clearModel.bind('change', this.render, this);
       this.add_related_model(this._clearModel);
     }
+
+    if (this._applyStatusModel) {
+      this._applyStatusModel.bind('change:loading', this._applyStatusChanged, this);
```
nobuti: `this.add_related_model(this._applyStatusModel);` missing if we use `bind` or `on`.

lib/assets/core/javascripts/cartodb3/dataset/dataset-options/dataset-actions-edition-view.js
```diff
@@ -25,7 +27,14 @@ module.exports = CoreView.extend({
   initialize: function (opts) {
     checkAndBuildOpts(opts, REQUIRED_OPTS, this);
 
+    this._applyButtonStatusModel = new Backbone.Model({
+      loading: false
```
nobuti: should we set initially to this._querySchemaModel.isFetching()?

saleiva:
good catch with that button @nobuti. This button should be on every place that we disable the CTA while loading. cc/ @urbanphes

lib/assets/core/javascripts/cartodb3/components/undo-redo/undo-redo-view.js
rjimenezda: 🤔  I guess if we ever have an undo / redo pattern that doesn't trigger an async event

lib/assets/core/javascripts/cartodb3/components/modals/publish/publish-button.tpl
rubenmoya: If we remove it and rerender it may be more expensive than letting the css handle it (it uses opacity)

https://github.com/CartoDB/CartoAssets/blob/master/src/scss/cdb-components/buttons.scss#L186

rjimenezda:
@nobuti @saleiva so after spending ~5h installing the stuff to get the analysis quota view button working, turns out that that button disappears when the analysis starts...so I guess no need for a state change in that button, right?

But hey we got the dataservices-api working, which rocks 🤘

lib/assets/core/javascripts/cartodb3/components/modals/publish/publish-button.tpl
nobuti: Changing opacity does not trigger any geometry changes, which is good. But since it is a visual property, it will cause painting to occur. Painting is typically a super expensive operation, so we should be cautious. At least we should check if `visibility` it's more appropriate in terms of performance.

nobuti:
Sorry guys, I was not sure, that's why I suggested to check it in acceptance 😞

saleiva:
but while it's being check, the button displays a loading indicator, right?

rjimenezda:
While the number of credits that it will cost is being checked, there's no button, just a panel with a loading indicator. Same thing when you hit 'confirm', the button is replace with a 'Running Analysis' loading indicator in a panel.

nobuti:
Agree, in order to apply the button, we need a revamp of that view.

nobuti:
Testing in ded02.

nobuti:
It works pretty well 🚀

Issue #12155, opened 2017-05-18
The layerDefinitionModel model is not correct when the basemap is google maps


xavijam:
Basically when we build that `layerDefinitionModel` model in the dataset view, we are not taking into account that a basemap could be a GMaps one.

[Context](https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/dataset.js#L60-L63):

```js
var layerDataModel = layersCollection.find(function (mdl) {
  var kind = mdl.get('kind');
  return kind !== 'tiled' && kind !== 'plain';
});
```

TrackJS error: https://my.trackjs.com/messages/filtered?hash=f484e536ab6b4d80

nobuti:
Deployed!

PR #12157, opened 2017-05-18
Using layer-types-and-kinds helper instead of guessing-comparision

xavijam:
Related ticket: #12155

We were not taking into account google basemaps and we were building incorrect layer-definition models when user had google basemaps enabled.

## STR
- Take a user with google maps enabled.
- Import a new dataset.
- Go to the dataset view and open the sql panel.
- Open the codemirror "hints" dialog with CTRL + Space.
- If it works 🎉 , if not, :boom:.
- Then, take another user (or the same one) without google key.
- Import a new dataset.
- Check the same as above.

💃


nobuti:
It just works 💅

Issue #11546, opened 2017-02-15
Boolean fields are not showing in dropdown menu from Filter by col. value analysis


ramiroaznar:
### Steps to Reproduce

1. Create map with `chicago_crimes` dataset.
2. Add Filter by column value analysis.
3. "Select" `arrest` (a boolean column) from the dropdown menu.

### Current Result

* Boolean columns are not showed on the dropdown menu.

![bug](https://cloud.githubusercontent.com/assets/5215798/22973618/c192006a-f37e-11e6-81be-82c8919f679d.png)

### Expected result

* Boolean columns should appear on the dropdown menu.

### Browser and version

* Google Chrome v54.

### .carto file

* Because the map size is too big, you can download a sample data from [here](https://team.carto.com/u/ramirocartodb/tables/builder-demo.chicago_crimes_sampled/public/map?redirected=true).

saleiva:
moving to next since this is part of the filter analysis (which is quite common)

nobuti:
@jorgesancha @noguerol @xavijam PR opened [here](https://github.com/CartoDB/cartodb/pull/12158) but the thing is, it won't work until https://github.com/CartoDB/cartodb/issues/10000 it's fixed in engine side. Basically it's the same issue, the tiler returns the whole set of features.

ethervoid:
We have the same problem in different projects and is how we treat boolean values. 

Issue #10000 is a turbo-carto issue but the one that is blocking Butis is related to camshaft. I'm going to ope an issue there

ethervoid:
The analysis is filtering the values properly. But we have, at least in local, some weird behavior painting the values. Funny thing is that the analysis resulting dataset is filtered correctly.

ethervoid:
I've been checking and the filtering is working fine. Is there any step to reproduce the case where the analysis is returning the whole set of features @nobuti ? thx

nobuti:
I tried with only two features and it seems to work, but still, there are weirdos with the boolean fields in the form.
- [ ] Validators don't play nice with those fields. @matallo is doing his magic here.
- [ ] If you try to select `true` and `hide` combination, it replaces automagically to `true` and `show`.
Anyway, it seems a full frontend issue now.

xavijam:
Deployed

Issue #10000, opened 2016-10-04
Styling by value does not work with boolean fields


ramiroaznar:
Styling by value does not work with boolean fields (Turbo-Carto\* is not applied).
### STR
- Create a map with a dataset with a boolean field.
- Click on `STYLE`.
- Click on color `FILL`.
- Select `BY VALUE` and choose the boolean field. The expected result would be a categoric map, instead what is showing is a simple map (screenshot). 

![boolean](https://cloud.githubusercontent.com/assets/5215798/19070322/f1ccb242-8a2a-11e6-986c-86d0223bb5bd.png)

*CartoCSS:

``` css
#layer {
  marker-width: 7;
  marker-fill: ramp([arrest], (#5B3F95, #1D6996), ("false", "true"));
  marker-fill-opacity: 1;
  marker-allow-overlap: true;
  marker-line-width: 1;
  marker-line-color: #FFF;
  marker-line-opacity: 1;
}
```

AbelVM:
Tested editing the CartoCSS by hand, and it doesn't either work with:
* true / false (without quotes)
* 0 / 1

saleiva:
Is this still happening?

ramiroaznar:
Yes, it is still happening.

![boolean](https://cloud.githubusercontent.com/assets/5215798/20593891/4a5a8b88-b234-11e6-8598-5934e7f9a615.png)

michellechandra:
S/B: 10679281

alapiere:
reproduced here.

ztephm:
Happening for S/B 12242644

Not using quotes in the conditional CartoCSS values worked in [this test](https://team.carto.com/u/stephaniemongon/builder/90f25382-2e85-11e7-bac7-0e3ebc282e83/embed):

![screen shot 2017-05-01 at 12 02 02 pm](https://cloud.githubusercontent.com/assets/1779444/25584949/2bd27032-2e66-11e7-9214-d859b71ca2cb.png)

saleiva:
we might need to prioritize this as a bug @jorgesancha

jorgesancha:
Prioritizing

ethervoid:
PR: cartodb/turbo-carto#67

ethervoid:
## Acceptance

Deploy the tiler (windshaft-cartodb) branch `fix_booleans_turbocarto` and follow the STR that Ramiro write down [here](https://github.com/CartoDB/cartodb/issues/10000#issue-180853250)

[...]

PR #12158, opened 2017-05-18
Filter by column value working with booleans.

nobuti:
This PR implements #11546. It:
- Applied different query for booleans when fetching.
- Special treatment for booleans when formatting attributes.

Still work in progress, that analysis lacks tests, it should be nice to add some.

But, **one important thing**, even the part of Builder is working, the tiler doesn't manage booleans properly, it returns the whole set of features. This is the same issue as https://github.com/CartoDB/cartodb/issues/10000. So in order to fix this completely, some work on the engine side must be done.


rubenmoya:
Working in staging! 👍

PR #12164, opened 2017-05-19
Updated copies for export image & download map

### LINKED ISSUE #12114: Improve copy of "Export map" and "Export as image"
There's a lot of confusion in our users from the similarity between both names and actions. We propose the following changes to avoid this confusion: 

Change **Export map** to **Download map**
Change **Export as Image** to **Export image**

rjimenezda:
Fix #12114 

screen shot 2017-05-19 at 12 24 47


nobuti:
🚀

PR #12166, opened 2017-05-20
Prevent Icon from saving when the upload fails

### LINKED ISSUE #11980: Icon is saved even when the upload fails
### Context
If you are trying to upload an icon, and it fails (because of the size validation for instance), the icon is saved in `Your uploads`. This happens for single or multiple uploads:

icon-upload-error

### Steps to Reproduce

1. Open the icon dialog. 
2. Select a source that doesn't fit into the expected file. For instance, the max size is 1024x1024. Choose a file bigger than that.
3. Once it fails, go to `Your uploads` tab.

### Current Result
The icon is saved but the image source is obviously broken.

### Expected result
The icon is not saved.

kutyel:
Fixes #11980 

carto

### DIFF
--- a/lib/assets/core/test/spec/cartodb3/components/form-components/editors/fill/input-color/asset-picker/assets-view.spec.js
+++ b/lib/assets/core/test/spec/cartodb3/components/form-components/editors/fill/input-color/asset-picker/assets-view.spec.js
@@ -28,7 +28,7 @@ describe('components/form-components/editors/fill/input-color/assets-view', func
     this.view = new AssetsView({
       model: this.model,
       modalModel: {
-        destroy: function () {}
+        destroy: function () { }
       },
       configModel: this._configModel,
       userModel: this._userModel
@@ -115,7 +115,7 @@ describe('components/form-components/editors/fill/input-color/assets-view', func
       this.view = new AssetsView({
         model: this.model,
         modalModel: {
-          destroy: function () {}
+          destroy: function () { }
         },
         configModel: this._configModel,
         userModel: this._userModel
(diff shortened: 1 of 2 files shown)


Cartofante:

Cartofante:

nobuti:
Nice. Could you include a test?

kutyel:
Sure thing! Here you go 😉 👍

lib/assets/core/test/spec/cartodb3/components/form-components/editors/fill/input-color/asset-picker/assets-view.spec.js
```diff
@@ -235,6 +235,14 @@ describe('components/form-components/editors/fill/input-color/assets-view', func
       expect(this.view._stateModel.get('error_message')).toBe('hi, I am an error');
       expect(this.view._stateModel.get('status')).toBe('error');
     });
+
+    it('should preserve the number of assets when upload fails', function () {
+      var previousAssetsCount = this.view._userAssetCollection.length;
+
+      mockCreate.call(this, 'error');
```
nobuti: Correct me if I'm wrong but I think `mockCreate` doesn't do anything on its own, you need to "upload" an image to force a call to the create method of the collection, then the error callback.

kutyel: You are right, I played along with the tests and couldn't find an actual way to test this behaviour, for example I tried spying on the `this.view._userAssetCollection.remove' method and it didn't work either. Would really appreciate some hints on how to approach this unit test.

nobuti: I think the approach is good, you only have to find the way to simulate an upload. If you connect the dots, from the moment the user click on the upload button, you'll find the answer.

lib/assets/core/test/spec/cartodb3/components/form-components/editors/fill/input-color/asset-picker/assets-view.spec.js
```diff
@@ -235,6 +235,16 @@ describe('components/form-components/editors/fill/input-color/assets-view', func
...
+
+      this.view.$('.js-fileInput').trigger('change');
```
kutyel: @nobuti what about this? 😅

nobuti: Nice!

xavijam:
Hey @kutyel, thanks for this 🤗 , closing it.

PR #12174, opened 2017-05-23
C1774 signup changes

javitonino:
CartoDB/cartodb-central#1731

**Multi-part PR**
We are tracking this in the central PR: CartoDB/cartodb-central/pull/1774


ivanmalagon:
Added `do not merge yet` because we need to open a new ticket only for the migration.

app/controllers/admin/visualizations_controller.rb
```diff
@@ -257,7 +257,7 @@ def public_map
       response.headers['Cache-Control'] = "no-cache,max-age=86400,must-revalidate, public"
     end
 
-    @name = @visualization.user.name.present? ? @visualization.user.name : @visualization.user.username.truncate(20)
+    @name = @visualization.user.name_or_username
```
juanignaciosl: Shouldn't `truncate` behavior be taken into account, maybe at frontend?

javitonino: I removed it because it didn't make much sense to me. The fact is, that `Name` is a free-form text without limits, and that is not truncated. The fact that username was truncated, but the full name didn't looked very strange to me and I was feeling a little :psycho: at the time.

Also, it is already truncated at frontend where the space is tight (in the breadcrumb) and not truncated where there is more space (near the bottom):
screenshot_20170525_151354

While testing, I found a place where maybe it _should_ be truncated, but it is not too bad (it just displays a wider popup):
screenshot_20170525_151519

In summary, I think it's okay as it is.

juanignaciosl: 👍

xavijam:
Both fields should take the same space than website field, for example.

screen shot 2017-05-26 at 12 09 10

javitonino:
Did the backend change.

xavijam:
Acceptance accomplished! Found just those 3 things. Taking care of the frontend ones.

app/assets/stylesheets/common/account_forms.css.scss
```diff
@@ -280,6 +280,9 @@ $sLabel-width: 140px;
 .FormAccount-input--med {
   width: 274px !important;
 }
+.FormAccount-input--small {
+  width: 131px !important;
```
javitonino: :trollface:

xavijam:
The problem about the name in the dropdown comes from the `user_decorator`, it only includes the first name, not both, could you take care of it @javitonino ?

app/assets/stylesheets/common/account_forms.css.scss
xavijam: Following the same rules already described ;)

PR #12181, opened 2017-05-23
Improved error handling when max layer limit is reached

### LINKED ISSUE #12167: Don't allow to add new layer when limit is reached
Everything is explained [here](https://github.com/CartoDB/design/issues/634#issuecomment-284705838), but in any case, let me add an images for context.

f8cf6dc4-0337-11e7-96f3-9a51d08f711f

rjimenezda:
Fix #12167 

screen shot 2017-05-23 at 16 27 41

This PR adds an infobox when users have reached their max layer limit. The error message & the subsequent 'upgrade' action are different, so it should be tested during acceptance.

Also, copies for Organization Users (both admin & non-admins) should be reviewed by someone.


lib/assets/core/javascripts/cartodb3/editor/editor-pane.js
```diff
@@ -194,6 +214,65 @@ module.exports = CoreView.extend({
     this.listenTo(this._layerDefinitionsCollection, 'add', function () {
       this._visDefinitionModel.fetch();
     });
+    this.listenTo(this._layerDefinitionsCollection, 'add remove', this._onLayerCountChange);
+  },
+
+  _getMaxLayerInfoBoxForCurrentUser: function () {
+    var self = this;
```
nobuti: why self here? It seems to be in the same context.

lib/assets/core/javascripts/cartodb3/editor/editor-pane.js
```diff
@@ -204,6 +283,33 @@ module.exports = CoreView.extend({
...
+
+  _getTranslatedLayersLabel: function (count, max) {
+    return _t('editor.tab-pane.layers.title-label', {
+      count: count,
+      maxCount: max
+    });
+  },
+
+  _enableAddButton: function () {
+    this.$('.js-add').removeClass('is-disabled');
```
nobuti: I tend to not trust in CSS classes to disable or enable a button. I think it's better to use the disabled attribute, the styles for the disable button should be implicit IMO.

lib/assets/core/javascripts/cartodb3/editor/editor-pane.js
```diff
@@ -220,21 +326,32 @@ module.exports = CoreView.extend({
...
         break;
       case 'elements':
         // TODO: trigger element creation
         break;
     }
   },
 
-  _addItem: function () {
+  _addItem: function (e) {
+    if (this.$('.js-add').hasClass('is-disabled')) return false;
```
nobuti: See the comment above, if we disable the button, we can get rid of this.

lib/assets/core/javascripts/cartodb3/editor/editor-pane.js
```diff
@@ -194,6 +214,65 @@ module.exports = CoreView.extend({
...
+          infoboxOpts.body = _t('editor.layers.max-layers-infowindow.org-admin.body', { maxLayers: self._getMaxCount() });
+          infoboxOpts.mainAction = { label: _t('editor.layers.max-layers-infowindow.org-admin.upgrade') };
+        } else {
+          infoboxOpts.body = _t('editor.layers.max-layers-infowindow.org.body', { maxLayers: self._getMaxCount() });
+          infoboxOpts.mainAction = { label: _t('editor.layers.max-layers-infowindow.org.upgrade') };
+        }
+
+        baseState.mainAction = function () { window.open('mailto:' + self._userModel.upgradeContactEmail()); };
+      } else {
+        baseState.mainAction = function () { window.open('https://carto.com/pricing/'); };
```
nobuti: should we add this to the locale file?

lib/assets/core/javascripts/cartodb3/editor/editor-pane.js
rjimenezda: We actually discussed adding pointer-events: none to is-disabled, but agree on not trusting CSS. I don't know if disabled attribute will have the same styling effect

lib/assets/core/javascripts/cartodb3/editor/editor-pane.js
rjimenezda: Good point, we might have different pricing URLs based on locale

lib/assets/core/javascripts/cartodb3/editor/editor-pane.js
rjimenezda: So, setting disabled obviously removes the pointer events, but it doesn't apply disabled styles. 

We can either set disabled to true, and the is-disabled class, or @piensaenpixel can add the same styles for .CDB-Button[disabled]

🤷‍♂️

lib/assets/core/javascripts/cartodb3/editor/editor-pane.js
rjimenezda: I actually don't know, maybe we had a different impementation at some point which required it, but it doesn't make sense, so we can blame my own ✨  incompetence ✨

rjimenezda:
@brittanymicek Can you please verify some new copies we have?

Depending on the type of user we are showing slightly different messages when the max layer limit is reached.

All users get the title `"You have reached the %{maxLayers} layer limit"`

Regular users get the message: `"Upgrade your account to add more than %{maxLayers} layers to your map"` and a button that reads `"Upgrade your account"`

Organization users that are **not** admins get the message: `"Contact your organization admin to upgrade your account"` and a button that reads `"Contact admin"`

Organization owners / admins get the message: `"Contact support to upgrade your account"`  and a button that reads `"Contact support"`

brittanymicek:
@rjimenezda Sure thing! Taking a look at it now.

brittanymicek:
Revised Copy:
**For all users (title):** "You have reached the maximum layer limit for this account."

**Regular users (message):** "Upgrade your account to add more than %{maxLayers} layers to your map."

**Org users that are NOT admin (message):** "Increase your layer limit by contacting the account's administrator."

**Org owners/admins (message):** "Increase the layer limit for your organization by contacting CARTO support."

Question(s): 
 - Will upgrading require a full upgraded plan (i.e. from free to personal)? 
- Or, can settings be adjusted for plans (i.e.? A user wants to increase the layer limit but keep the rest of a personal plan's functionality. Sort of a la carte).
- I have removed the word "upgrade" from the copy because at this stage a user's primary issue is not "upgrading" but rather *increasing* their layer limit. Of course, in order to increase they will need to upgrade but that is not the primary need. I have only left "upgrade" in the copy for regular users because they are solely responsible for their account settings.

Let me know if this helps/works!

Brit

rjimenezda:
@brittanymicek Thanks a lot!

As far as I know, non-enterprise users are bound to the upgrade path we offer. A la carte upgrades are more a thing for enterprise users.

matallo:
when doing CR I like to checkout to development branch locally (call it acceptance) and realized once you hit the max layers AND reload the button is disabled and no notice appears

pageshot of untitled map carto 2017-05-04-1135 20

back to development

lib/assets/core/javascripts/cartodb3/editor/editor-pane.js
matallo: my approach in these cases is using the disable attribute AND toggling the class depending on the value for that attribute

xavijam:
Enough for moving it to development again guys. It is important.

urbanphes:
For to give more context but without having a permanent message we can add a tooltip when the user pass over the button “Add” saying something like the bottom infobox: “You have reached the maximum layer limit (8) for this account”. 

What do you think?

rjimenezda:
@urbanphes I'm ok with the tooltip. I think that's better than **always** showing the layer limit box, which is very very intrusive visually, and we shouldn't punish users that are using exactly the amount of layers they need, which might be the max we allow.

Just to clarify, you're seeing an empty description because you are running a custom / open source version of CARTO, so there's no point to a CTA. I can add an extra explanation, but I think it's kind of redundant.

On a SaaS you'd see the following warnings (please ignore the ugly cross, will fix asap):

Regular users
screen shot 2017-05-25 at 17 44 09

Regular user that belong to an organization:
screen shot 2017-05-25 at 17 46 54

Organization admin:
screen shot 2017-05-25 at 17 48 46

saleiva:
We can do several things.
- Remove the buttons and add the links directly on the text. For example in the last option, "contacting CARTO support" could link directly to contact support.
- Cut the titles as @urbanphes says. "You have reached your max layer limit."

rjimenezda:
I really don't like the links on text, mainly because we have to write html on localization strings.

I fixed the cross issue:

screen shot 2017-05-26 at 11 40 56

By the way, the Contact admin / contact support are a **mailto:** link, towards the appropriate email (organization admin or enterprise support)

urbanphes:
We can:
- Leave the title: "You have reached your max layer limit."
- And keep the text, where is explained the max number layers and keep the CTA link.

rjimenezda:
Ok, so I'll cut the title so it doesn't split in two lines (in English): "You have reached your maximum layer limit."

So if we want to be consistent across all types of accounts, I've done this for open source / custom CARTO installations:

screen shot 2017-05-29 at 10 09 42

The CTA takes you to the contact form

lib/assets/core/locale/en.json
```diff
@@ -1882,8 +1882,13 @@
         "visualize": "Layer doesn't have geometry"
       },
       "max-layers-infowindow": {
-        "title": "You have reached the maximum layer limit for this account.",
+        "title": "You have reached your maximum layer limit.",
```
nobuti: I'd change `your` for `the`.

xavijam:
Acceptance?

Issue #11783, opened 2017-03-16
[SMOKES] Making a loading decision, raster or vector?


donflopez:
### Context
We are loading vector tiles under a flag, with the beta release we have to set VT by default, but we are a step behind raster in some scenarios.

1. [tangram-cartocss not supported features](https://docs.google.com/spreadsheets/d/1zvFU4F07u_RLW605AB7u8qbjZNMKbnRaF3QfmFXNGHo/edit#gid=1244395638)

2. cartocss dynamic features not suported in tangram.

3. Huge amount of shapes to draw where tangram rendering can take a lot of time.

4. Drastic visual differences between raster and vector.

Based on those reasons, we need a dynamic decision maker, allowing us to bring the best user experience in those different scenarios.

### Options:

All of them can be combined.

|Action|Side| Type & Fix|Benefit|Cost|
|------|-----|-----------|-------|-----|
| Send the total amount of data to make a decision | server / client | Preventive - Fix 3 | Have a good performance | Medium |
| Run tangram-cartocss to detect not supported features | server / client | Preventive - Fix 1 & 2 | Have a good UX | Medium |
| Raster fallback if the map fails | client | Palliative - Fix 1 & 2 | Not seem broken as last resource | Medium |
| Let the user choose in the meantime | server & client | Preventive - fix everything (new maps or edited ones) | To have freedom | Medium |
| Run tangram and raster and compare screenshots to see huge differencies automatically | server | Preventive - Fix 4 (old maps) | Have a good UX | High |

NOTE: add here other solutions ^^

cc/ @javisantana & @rochoa

javisantana:
Some comments: 

about 4, shouldn't it be based on cartocss? is there any other thing that adds differences between raster and vector?

donflopez:
Yeah @javisantana, they're mainly cartocss problems, we have 2 factors from cartocss but the 3rd is out of it.

1. Unsupported features (the best).
2. Dynamic features supported as static in tangram.
3. Differences at rendering level, ex: textures, border point (now fixed).

donflopez:
After looking this in detail we have some conclusions for the first step:

1.- Take the rendering decision on the client side (we have more information there).
2.- Send from the server features statistics of each layer, at the start the number of vertex per layer.
3.- This info will be sent next to the instance of the map.
4.- We'll check the number of features and  cartocss compatibility.

Some new ideas that have appeared are:
- Send the precision with the tile request.
- Get feature density from PSQL statistics.
[...]

PR #12182, opened 2017-05-23
Vector VS Raster

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb/issues/11783.


alonsogarciapablo:
@donflopez we need to add `vector:false` to [these options](https://github.com/CartoDB/cartodb/blob/5b582c1d6f61788de84a7148757e2f9919a3bc5b/lib/assets/core/javascripts/cartodb3/editor.js#L279-L288) to ensure vector rendering is disabled in BUILDER. I saw you removed the `vector` flag from vizjson (:+1:), but that will cause CartoDB.js to choose between raster and vector both in BUILDER and embed.

donflopez:
[It is](https://github.com/CartoDB/cartodb/blob/vector-vs-raster-1624/lib/assets/core/javascripts/cartodb3/editor.js#L288), isn't it?

alonsogarciapablo:
> It is, isn't it?

alt text

👍 👍 👍

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cartodb-ui",
-  "version": "4.7.37",
+  "version": "4.7.214",
```
javitonino: This is a big version bump :P

alonsogarciapablo: Yes! We'll fix this before merging! :+1:

alonsogarciapablo:
👏 👏 👏 👏 👏

PR #12186, opened 2017-05-24
New style for analysis button

### LINKED ISSUE #11902: New style for the add analysis button
We've realized the **+** button for adding more analyses isn't working as good as we expected so we're tweaking the style a bit to see if the performance improves. We're also adding tooltips to each analysis node. 
 
I'm assigning this to @piensaenpixel since he wanted to apply this changes himself. 

632ee19e-1577-11e7-893c-2303e250da74

rjimenezda:
Fixes #11902 

click

Additionally I've added an IE11 fix for the horizontal scroll. It's temporary, since it uses a fixed height, but I've tried to make it as least crappy as possible. I'll give it a go as soon as I can use a decent IE11 workstation

I've added a line in ruby config to serve assets zipped on development, @javitonino check it out please.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -72,6 +72,9 @@ CartoDB::Application.configure do
   # Generate digests for assets URLs
   config.assets.digest = true
 
+  # Serve gzipped files
+  config.middleware.insert_before ActionDispatch::Static, Rack::Deflater
+
   config.assets.initialize_on_precompile = true
 
   config.action_controller.asset_host = Proc.new do
--- a/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-item-view.js
+++ b/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-item-view.js
@@ -1,8 +1,11 @@
 var CoreView = require('backbone/core-view');
+var $ = require('jQuery');
 var template = require('./analyses-workflow-item.tpl');
 var nodeIds = require('../../../../value-objects/analysis-node-ids');
 var layerColors = require('../../../../data/layer-colors');
 var AnalysisTooltip = require('../../analysis-views/analyses-tooltip-error');
+var TipsyTooltipView = require('../../../../components/tipsy-tooltip-view');
+var Analyses = require('../../../../data/analyses');
 
 module.exports = CoreView.extend({
 
@@ -47,6 +50,7 @@ module.exports = CoreView.extend({
         bgColor: bgColor,
         isSelected: isSelected,
         nodeId: this._analysisNode.id,
+        name: Analyses.title(this._analysisNode.get('type')),
         hasError: status === 'failed'
       })
     );
@@ -57,6 +61,14 @@ module.exports = CoreView.extend({
     });
     this.$el.toggleClass('is-selected', isSelected);
     this.$el.toggleClass('has-error', status === 'failed');
+
+    var tooltip = new TipsyTooltipView({
+      el: this.$el.find('.js-tooltip'),
+      title: function (e) {
+        return $(this).data('tooltip');
+      }
+    });
+    this.addView(tooltip);
 
     return this;
   },
--- a/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-item.tpl
+++ b/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-item.tpl
@@ -1,5 +1,5 @@
 <% if (isDone) { %>
-  <div class="HorizontalBlockList-item-actionBlock CDB-Text CDB-Size-small u-upperCase">
+  <div class="HorizontalBlockList-item-actionBlock CDB-Text CDB-Size-small u-upperCase js-tooltip" data-tooltip="<%- name %>">
     <span class="HorizontalBlockList-item-text">
       <%- nodeId %>
     </span>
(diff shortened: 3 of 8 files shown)


config/environments/development.rb
```diff
@@ -72,6 +72,9 @@
   # Generate digests for assets URLs
   config.assets.digest = true
 
+  # Serve gzipped files
+  config.middleware.insert_before ActionDispatch::Static, Rack::Deflater
```
javitonino: Also pinging @juanignaciosl for awareness. Rationale behind this is to have the built-in server gzip the assets before sending, which is something very useful to test in mobile devices connected to the local instance and not have killer loading times.

It should not impact anything as this is the development environment, so it's a 👍 from me.

rjimenezda: For instance, the vendor.js goes from 12MB to 2.4MB, which is pretty neat.

There are some options in the same file that mention compression, but I guess they don't influence thin

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-item-view.js
```diff
@@ -47,6 +50,7 @@ module.exports = CoreView.extend({
         bgColor: bgColor,
         isSelected: isSelected,
         nodeId: this._analysisNode.id,
+        name: Analyses.title(this._analysisNode.get('type')),
```
ivanmalagon: Just being a bit paranoid here.

If `this._analysisNode.get('type')`returns `undefined` it provokes an error and nothing gets rendered.
We can pass the empty string if it returns `undefined` and `Analyses.title` will return the string `unknown` and nothing breaks.

It's true that both `analysisNode` or `formModel` seem to always have the `type` attribute. Hence the `paranoid` tag to this comment.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-item-view.js
```diff
@@ -58,6 +62,14 @@ module.exports = CoreView.extend({
     this.$el.toggleClass('is-selected', isSelected);
     this.$el.toggleClass('has-error', status === 'failed');
 
+    var tooltip = new TipsyTooltipView({
+      el: this.$el.find('.js-tooltip'),
+      title: function (e) {
+        return $(this).data('tooltip');
```
ivanmalagon: Since the text to show in the tooltip is not dynamic, we can get rid of this and simply return the same string we have above.
This way, we get rid of the `jquery` dependency, we don't need to pass `name` to the template and the template gets simplified also.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-item.tpl
```diff
@@ -1,5 +1,5 @@
 <% if (isDone) { %>
-  <div class="HorizontalBlockList-item-actionBlock CDB-Text CDB-Size-small u-upperCase">
+  <div class="HorizontalBlockList-item-actionBlock CDB-Text CDB-Size-small u-upperCase js-tooltip" data-tooltip="<%- name %>">
```
ivanmalagon: If we return the string (as stated in a comment above) we don't need to provide info via `data` attribute. Same thing for the next two `data-tooltip` attributes.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-view.js
```diff
@@ -53,7 +53,7 @@ module.exports = CoreView.extend({
       }
     });
 
-    this.$('.js-content').append(view.render().el);
+    this.$('.ie11-hotfix').append(view.render().el);
```
ivanmalagon: I'm confused here.

Previous to this change, something was appended to `.js-content`. Now, it appends to `.ie11-hotfix`. But I can't find any change in this PR with the `js-content` tag. Can you explain me this?

ivanmalagon: Explained by @rjimenezda 

The `.js-content` tag is still present at parent level.

ivanmalagon:
Requested some changes. Moving back to development.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-item-view.js
```diff
@@ -62,10 +60,11 @@ module.exports = CoreView.extend({
...
     this.$el.toggleClass('has-error', status === 'failed');
 
+    var self = this;
     var tooltip = new TipsyTooltipView({
       el: this.$el.find('.js-tooltip'),
-      title: function (e) {
-        return $(this).data('tooltip');
+      title: function () {
+        return Analyses.title(self._analysisNode.get('type') || '');
       }
```
xavijam: As you know, we can get rid of the `self` var adding .bind(this) to the function.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-item-view.js
```diff
@@ -62,10 +60,11 @@ module.exports = CoreView.extend({
     this.$el.toggleClass('is-selected', isSelected);
     this.$el.toggleClass('has-error', status === 'failed');
 
+    var self = this;
     var tooltip = new TipsyTooltipView({
       el: this.$el.find('.js-tooltip'),
```
xavijam: `this.$el.find('.js-tooltip')` == `this.$('.js-tooltip')`

xavijam:
It has to get back to development due to the tooltip is not right placed when it is loading.

screen shot 2017-05-25 at 12 44 49

nobuti:
Looks good to me, we missed the chance to fix to set selected the first analysis when one of them is deleted, it will be tackled in a future issue. 🚀  once 🍏 .

nobuti:
Frontend tests passing.

PR #12187, opened 2017-05-24
remove db:create and db:migrate from create_dev_user rake task

matallo:
no prior issue created

after discussion, the `create_dev_user` rake task shouldn't have `db:create` nor `db:migrate` responsibility, which appear as independent tasks in the installation README, as well as in the provision task in the VM, furthermore this made the provision tasks because the db already existed

this was surfaced after a change in the sequel gem, which now exit with a more aggressive code than before, which failed silently and exited successfully

/cc @javitonino @juanignaciosl


javitonino:
🍏 (before NEWS update). Old school CR request -> @juanignaciosl. If 👍 feel free to merge (as acceptance was already done by @matallo)

Issue #10850, opened 2016-11-28
Several issues with Join columns from 2nd layer analysis


saleiva:
### Context
Although the analysis work correctly, the UX of this form is not smooth. This issue summarizes a set of improvements to be done on it. This analysis is in the top3 of most used analysis.

### Steps to Reproduce

1. Using an account with many datasets imported (more than 200) apply the Join columns from 2nd Layer analysis.
2. Once it is applied, go to the layer list and go to edit the applied analysis.

### Current Result
- [x] Input 2 selector is loading although the layer was already selected before (this might be the root cause of some other issues described below).
- [x] While loading the UI seems broken (image attached)
- [x] After loading the node names is wrong, and the label for the inputs from the 2nd layer is wrong as well. Seems to be adding the node id instead of the dataset/layer name
 

### Expected result
The UI shouldn't seem broken and it shouldn't reload the 2nd layer.

You can import this .CARTO in order to test this faster.
[Where we work in Madrid (on 2016-11-28 at 13.53.18).carto.zip](https://github.com/CartoDB/cartodb/files/616486/Where.we.work.in.Madrid.on.2016-11-28.at.13.53.18.carto.zip)

cc/ @xavijam

xavijam:
Deployed!

PR #12194, opened 2017-05-24
Join analysis UI revamp

nobuti:
This PR implements #10850.

merge

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -120,6 +120,7 @@ Development
 * Updated copies for export image & download map (#12114)
 
 ### Bug fixes
+* Fixed UI when editing merge analysis (#10850)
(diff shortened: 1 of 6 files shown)


xavijam:
I found something in the acceptance, the label of the right-source is not being rendered.

screen shot 2017-05-25 at 14 57 43

xavijam:
Hey @saleiva, what would you do it in this case (using different nodes from same layer)?

screen shot 2017-05-25 at 15 09 02

Would you like something like that?

option 1
screen shot 2017-05-25 at 15 12 26
---
option 2
screen shot 2017-05-25 at 15 14 02

saleiva:
2nd!

xavijam:
_Oido cocina, gracias_

Issue #8081, opened 2016-06-22
review loading stylesheets


piensaenpixel:
![image](https://cloud.githubusercontent.com/assets/1214201/16276851/7f71631c-38b0-11e6-8f32-46540d49ff8e.png)

xavijam:
DEPLOYED!

PR #12203, opened 2017-05-24
Using last cartoassets changes

xavijam:
Related ticket: https://github.com/CartoDB/cartodb/issues/8081.
But, in fact, it is a task from https://github.com/CartoDB/cartodb/milestone/139 milestone.

Basically making use of the new changes added in CartoAssets https://github.com/CartoDB/CartoAssets/pull/162.
But, before merging this, we need to merge those changes in CartoAssets, then in CARTO.js, after that in Deep-insights, and finally here. 💃 

cc @CartoDB/frontend


xavijam:
No need to change NEWs, I feel.

xavijam:
Deploying it in ded04 servers.

xavijam:
Tested it in ded04 server and looks ok, from private builder to public embed, but could anyone ( @CartoDB/frontend ) take a look please? 💋

nobuti:
It works smoothly, I couldn't find any flaws. Also, some savings in the final bundle:

- 5kb en editor3css y 2 kb en deep_insights.css, uncompressed.
- gzips remain in almost the same value, totally understandable.

But the main goal of this ticket was not the final weight, so kudos :)

xavijam:
So... starting the party with CartoAssets ---> CARTO.js ---> Deep-Insights.js ---> Builder

xavijam:
Again, deployed in ded04, just in case someone wants to check it (again). If nobody complains about it, we will deploy it tomorrow morning 💃 .

Issue #9502, opened 2016-08-22
UI to check all fields on JOIN analysis


jsanz:
Maybe this could be generic and applied to any case where using checkboxes to select fields list but it's a very common scenario to have a table with a normalized field so you want **all** the fields from the origin table and just a few from the join table. 

The UI to select fields is quite small, using the keyboard to select fields is not easy and the scroll can easily jump more than one visual page so it's hard to quickly select all fields.

xavijam:
Let's summon @saleiva here.

saleiva:
a button for selecting ALL / NONE was originally designed on the first version of this component by it never got implemented. All yours @xavijam 

![screen shot 2016-08-22 at 16 46 12](https://cloud.githubusercontent.com/assets/132149/17858906/e0159786-6887-11e6-9654-614089776491.png)

zingbot:
+1 on this. Just struggled with selecting 30 checkboxes:)

jsanz:
The predictive analysis really needs this **select all/none** UI, friends :sweat_smile: 

![2016-10-06_15 11 39-selection](https://cloud.githubusercontent.com/assets/188264/19153526/3721f1b0-8bd7-11e6-8aaa-b7a116c8983d.png)

saleiva:
adding next, since this is the most popular analysis by far

xavijam:
_plegarias escuchadas_

xavijam:
For new guys

rubenmoya:
Should the 'ALL' and 'NONE' buttons de/select everything in the list, or should it be smarter and only select the search items?

Example:

1. User selects absolutely everything (20 items)
2. User types 'vamos', there are 3 items matching
3. User hits 'ALL'

What should be selected then? Just the three matches? Still everything?

Same question but with 'NONE', should it only deselect the search results?

xavijam:
That's a good one, @saleiva what's your opinion?. IMHO I'd deselect all items, because when you filter by something, rest of items don't matter.

jsanz:
This could be confusing without changing the label to something like `ALL SELECTED` but yea, the functionality would be well received, I can tell you.

zingbot:
Agreed it would be a really well-received option. 'All Filtered' or 'All Visible' could make that clear, too

nobuti:
In production.

saleiva:
BOOM!

PR #12216, opened 2017-05-25
All / none button for multi select component

rubenmoya:
This PR implements #9502

allnone

### Acceptance
- [x] ALL selects all the items
- [x] NONE deselects all the items
- [x] ALL and NONE are gone if query search is applied


xavijam:
Already mentioned off-line:

- Buttons should be smaller.
- Paddings are not correct.
- Search should be bigger, we need to gain more space.

rubenmoya:
Is this better? I'm using `u-lSpace--xl`, `u-lSpace` is too small imho, and there is no `u-lSpace--m`

screen shot 2017-05-25 at 17 49 07

xavijam:
About space and padding and so on, perfect, but buttons are still big.

rubenmoya:
That's true, this looks better :)

screen shot 2017-05-25 at 17 56 47

xavijam:
NOW WE ARE TALKING MY FRIEND! 👍

lib/assets/core/javascripts/cartodb3/components/custom-list/custom-list-search-view.js
```diff
@@ -36,7 +36,12 @@ module.exports = CoreView.extend({
 
   _checkButtons: function () {
     var query = this.model.get('query');
-    this.$('.js-clear').toggle(query.length > 0);
+
+    if (query.length > 0) {
```
xavijam: What about?

```js
this.$('.js-clear').css('opacity', query.length ? 0 : 1);
```

lib/assets/core/javascripts/cartodb3/components/custom-list/custom-list-search.tpl
```diff
@@ -1,6 +1,6 @@
-<div class="CDB-Box-modalHeaderItem">
+<div class="u-flex">
   <input type="text" name="text" autocomplete="off" placeholder="<%- searchPlaceholder %>" class="CDB-InputTextPlain CDB-Text js-search">
-  <button type="button" style="display:none" class="u-lSpace--xl js-clear">
+  <button type="button" style="opacity: 0;" class="u-lSpace--xl js-clear">
```
xavijam: Could we create a class and put it wherever we consider? We should not add inline styles. Let's do it! 💃

app/assets/stylesheets/editor-3/custom-list.css.scss
```diff
@@ -119,3 +119,7 @@
 .CustomList-list.ps-container > .ps-scrollbar-x-rail {
   margin-bottom: 2px;
 }
+
+.CustomList-search {
```
piensaenpixel: IMO this should a modifier :p and like we have `.CustomList--small` etc... what do you think if we use something like `.CustomList--full`
BTW we have `.CustomList-itemLink` and maybe we could consolidate this one.
My 2 ctmos.

lib/assets/core/javascripts/cartodb3/components/custom-list/custom-view.js
```diff
@@ -84,17 +95,29 @@ module.exports = CoreView.extend({
...
+    this.$('.js-header').prepend(this._searchView.render().el);
     this.addView(this._searchView);
   },
 
+  _renderActions: function () {
+    _.each(this.options.actions, function (a) {
+      var view = new CustomListAction(a);
+      this.$('.js-header').append(view.render().el);
+      this.addView(view);
+    }.bind(this));
```
xavijam: Good thing about _.each in underscore is you can pass the context as an argument (http://underscorejs.org/#each). No need to bind the scope to the function.

lib/assets/core/javascripts/cartodb3/components/form-components/editors/select/multi-select-view.js
```diff
@@ -96,11 +104,32 @@ Backbone.Form.editors.MultiSelect = Backbone.Form.editors.Base.extend({
       this._popupManager.untrack();
     }.bind(this);
 
-    this.collection.bind('change:selected', this._onItemSelected, this);
+    this.collection.bind('change:selected', _.debounce(this._onItemSelected, 50), this);
```
xavijam: I encourage you to bind the this scope in this case.

lib/assets/core/test/spec/cartodb3/components/form-components/editors/select/multi-select-view.spec.js
```diff
@@ -23,11 +24,80 @@ describe('components/form-components/editors/select/multi-select-view', function
     });
   });
 
-  it('should trigger change event on select item', function () {
+  it('should trigger change event on select item', function (done) {
     var onChange = jasmine.createSpy('onChange');
     this.view.on('change', onChange);
     this.view.collection.at(1).set({selected: true});
-    expect(onChange).toHaveBeenCalled();
+    setTimeout(function () {
```
xavijam: Better than use a timeout, https://jasmine.github.io/2.0/introduction.html#section-Mocking_the_JavaScript_Timeout_Functions, clocks!

lib/assets/core/test/spec/cartodb3/components/form-components/editors/select/multi-select-view.spec.js
```diff
@@ -23,11 +24,80 @@ describe('components/form-components/editors/select/multi-select-view', function
...
+      done();
+    }, 100);
+  });
+
+  describe('all/none buttons', function () {
+    it('should trigger only one change event (debounce)', function (done) {
+      var onChange = jasmine.createSpy('onChange');
+      this.view.on('change', onChange);
+      this.view._selectAll();
+      setTimeout(function () {
```
xavijam: Same as above ^^^  🍄

lib/assets/core/javascripts/cartodb3/components/custom-list/custom-view.js
rjimenezda: Good to know 🆒

lib/assets/core/javascripts/cartodb3/components/custom-list/custom-list-search.tpl
rjimenezda: Sure, how about 'u-transparent' 'u-opaque'? @rubenmoya

lib/assets/core/javascripts/cartodb3/components/form-components/editors/select/multi-select-view.js
xavijam: Also, after talking with @rjimenezda, we could change the way we trigger that change, what about to make the change of the attribute silently and then trigger a custom event like 'allSelected' or 'noneSelected' at the end of the loop?

saleiva:
I don't like the way the "cross" button appears... @urbanphes @piensaenpixel  any idea?

saleiva:
a possible solution is to remove those ALL/NONE buttons when having a search query applied

xavijam:
I like that @saleiva, but in any case, wouldn't you display the cross for "removing" the searched query?

saleiva:
yes, but not at the same time than the ALL / NONE buttons

xavijam:
I feel it, we are on the same page 💃 . @rubenmoya and @rjimenezda clear?

lib/assets/core/test/spec/cartodb3/components/form-components/editors/select/multi-select-view.spec.js
rubenmoya: It seems clocks don't affect setTimeout functions deep inside the code (like _.debounce), so we went with a different approach, mocking _.debounce.

xavijam:
Moving to development, but we are close!

lib/assets/core/javascripts/cartodb3/components/custom-list/custom-view.js
```diff
@@ -113,11 +115,15 @@ module.exports = CoreView.extend({
   _renderActions: function () {
     _.each(this.options.actions, function (action) {
       var view = new CustomListAction(action);
-      this.$('.js-header').append(view.render().el);
+      this.$('.js-actions').append(view.render().el);
       this.addView(view);
     }, this);
   },
 
+  _setActionsVisibility: function () {
```
nobuti: When we are listening to one key's changes `change:key`, the handler has the signature `function (model, value)`. This way we can save the `this.model.get('query')`:

```
_setActionsVisibility: function (model, query) {
   this.$('.js-actions').toggleClass('u-hide', query !== '');
}
```

nobuti:
👍

nobuti:
Frontend tests passing 🚀

PR #12223, opened 2017-05-26
DO measurement dropdown revamp.

nobuti:
This PR revamps the measurements dropdown.

do-dropdown


saleiva:
small detail, can we add a placeholder to the search saying... "Search by name..." instead of those "..."?

nobuti:
I've changed 100 as the max items to show, let me know if you see it ok from a performance point of view. This is in ruis.

nobuti:
Due this solution has been proved more performance, I'm merging it back to the original branch and keep working on there 🤖

PR #12231, opened 2017-05-29
Overlay issues in legends view.

### LINKED ISSUE #11825: Legend color bar gets small
### Steps to Reproduce

1. Get a map with a category widget. 
2. Select a layer and create a legend in it. A custom one, for example.
3. The legend form shows a color bar with a proper size.
4. Go to the category widget and select one of the categories.
5. The color bar gets very small and it keeps this size even when exiting the layer edition and coming back.

### Current Result

This is what happens.

peque-barra

### Expected result

The color bar remains equal when selecting a category. Its size should not be affected.

nobuti:
This PR fixes #11825. I've fixed in DO branch but it makes sense to pull it out. Also, once fixed, I've discovered other issues related it, so in order to test it:

- [x] Check the original issue is fixed.
- [x] The overlay is placed properly in the dom tree, for instance, the infoboxes are below it.
- [x] Make the layer not visible. Get the legend panel and make the layer visible again. The overlay should hide.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -121,6 +121,7 @@ Development
 * New style for add analysis button (#11902)
 
 ### Bug fixes
+* Fixed overlay for legends view (#11825)
 * Fixed UI when editing merge analysis (#10850)
 * Fixed uninitialized constant in Carto::Visualization when a viewer shares a visualization (#12129).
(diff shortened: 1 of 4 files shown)


xavijam:
Working like a charm! Moving it to deploy!

Issue #12227, opened 2017-05-29
Small UI issues after CARTOAssets update


xavijam:
I feel we have open the door of UI issues after fixing the problem we had with the CartoAssets bundling (https://github.com/CartoDB/cartodb/pull/12203). We have found those so far:

- [ ] In dataset view, when you open the column dropdown.
![screen shot 2017-05-29 at 11 43 45](https://cloud.githubusercontent.com/assets/132146/26545208/15b90af8-4465-11e7-9134-ee3fd5b36d58.png)

- [ ] In the dashboard dropdown.
![screen shot 2017-05-29 at 11 43 29](https://cloud.githubusercontent.com/assets/132146/26545224/22492e1a-4465-11e7-8718-27faec7878aa.png)

- [ ] In the "create/upload dataset" modal in Builder page.
![](https://cloud.githubusercontent.com/assets/132146/26361191/857eef7c-3fda-11e7-9788-1c032ead3d01.png)

xavijam:
cc @nobuti. Good for new hires.

PR #12234, opened 2017-05-29
fix small ui issues

rubenmoya:
Issue: #12227

- [x] Dataset view, when you open the column dropdown.

screen shot 2017-05-29 at 16 00 32

- [x]  Dashboard dropdown.

screen shot 2017-05-29 at 16 00 44

- [x] Color by value

screen shot 2017-05-31 at 15 30 35

- [x] "create/upload dataset" modal in Builder page.

dataset_loading

Depends on: https://github.com/CartoDB/CartoAssets/pull/163


rubenmoya:
retest this please

xavijam:
Going back to development because the modal we need to review is in Builder page, not in the dashboard.

xavijam:
@nobuti mentioned an issue related to fill -> style -> icon. Is it fixed too'

rubenmoya:
Seems like it is fixed too 👏 

screen shot 2017-05-30 at 15 23 18

xavijam:
Tested in staging and working like a charm. Let's merge everything tomorrow!

rubenmoya:
I've been talking with @piensaenpixel and we think the best way to fix it is by adding some more markup, for example, in this case moving the flex classes to other div:

screen shot 2017-05-31 at 12 01 27

I'll have to check the rest of the cases, if it works we won't need too add logic to the css, which is nice. What do you think @xavijam?

rubenmoya:
Last changes fix it with markdown instead of changing cartoassets, the only one we can't fix this way is the loading icon, which has to be fixed using this PR: https://github.com/CartoDB/CartoAssets/pull/164

xavijam:
Fantastic. I'm a bit worried about if we find more places where there are UI issues, but let's face them when appear.

xavijam:
Hey, there is another place where we haven't fixed it 😬 

screen shot 2017-06-01 at 11 11 08

Back to development, come on!

xavijam:
:+1:

Issue #11632, opened 2017-02-24
Color mismatch when map has null category


ivanmalagon:
### Context
Styling a map by category with a lot of categories (more than 10), one of the valid categories has a value of `null`. In the category style we can see that it's valid category with its own color. The `others` category has the right color, a grey one.

In the legend, the `others` category get mistaken for the `null` one, showing a wrong color, not the gray one.

I haven't done any research so I don't know if this is a bug in `cartodb` or `cartodb.js`. Please move the issue to the right repo once that gets clarified. 

### Steps to Reproduce

1. Import the attached .carto file.

1. Style by value choosing the `line` column.

Here's a recording of the problem.

![null category](https://cloud.githubusercontent.com/assets/1078228/23296750/cb0374fe-fa76-11e6-88c2-1ba4fccd6486.gif)

### Files

[Leer solo (on 2017-02-24 at 08.52.14).carto.zip](https://github.com/CartoDB/cartodb/files/798739/Leer.solo.on.2017-02-24.at.08.52.14.carto.zip)

nobuti:
Possibly related to https://github.com/CartoDB/cartodb/issues/11421

javierarce:
Acceptance: 👍

nobuti:
This is still happens after #11421.

xavijam:
Deployed

PR #12240, opened 2017-05-30
Label empty/null values in the same way as fill component.

nobuti:
This PR implements #11632.

My first approach was to make a distinction between `null` and empty values. However turbocarto converts values to String:

```
function processFilterValue (rawValue) {
  if (rawValue === 'false') {
    return false;
  } else if (rawValue === 'true') {
    return true;
  } else {
    return Number.isFinite(rawValue) ? rawValue : '"' + rawValue + '"';
  }
}
```

so, right now it's imposible to do it.

### DIFF
--- a/lib/assets/core/test/spec/cartodb3/editor/style/style-converter-fixtures.spec.js
+++ b/lib/assets/core/test/spec/cartodb3/editor/style/style-converter-fixtures.spec.js
@@ -559,13 +559,13 @@ module.exports = [
     },
     result: {
       point: {
-        cartocss: '#layer {\nmarker-width: ramp([cartodb_id], range(1, 20), category(6));\nmarker-fill: ramp([name], (#fabada, black, red, white), (wutang clan, form-components.editors.fill.input-categories.null, public enemy), "=", category);\nmarker-fill-opacity: 0.5;\nmarker-file: ramp([name], (url(\'image1\'), url(\'image2\'), url(\'image3\'), url(\'otherimage\')), (wutang clan, \'\', public enemy), "=");\nmarker-allow-overlap: true;\n}'
+        cartocss: '#layer {\nmarker-width: ramp([cartodb_id], range(1, 20), category(6));\nmarker-fill: ramp([name], (#fabada, black, red, white), (wutang clan, _t('form-components.editors.fill.input-categories.null'), public enemy), "=", category);\nmarker-fill-opacity: 0.5;\nmarker-file: ramp([name], (url(\'image1\'), url(\'image2\'), url(\'image3\'), url(\'otherimage\')), (wutang clan, \'\', public enemy), "=");\nmarker-allow-overlap: true;\n}'
       },
       line: {
         cartocss: '#layer {\n}'
       },
       polygon: {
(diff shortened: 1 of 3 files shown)


lib/assets/core/test/spec/cartodb3/editor/style/style-converter-fixtures.spec.js
```diff
@@ -559,13 +559,13 @@ module.exports = [
     },
     result: {
       point: {
-        cartocss: '#layer {\nmarker-width: ramp([cartodb_id], range(1, 20), category(6));\nmarker-fill: ramp([name], (#fabada, black, red, white), (wutang clan, , public enemy), "=", category);\nmarker-fill-opacity: 0.5;\nmarker-file: ramp([name], (url(\'image1\'), url(\'image2\'), url(\'image3\'), url(\'otherimage\')), (wutang clan, \'\', public enemy), "=");\nmarker-allow-overlap: true;\n}'
+        cartocss: '#layer {\nmarker-width: ramp([cartodb_id], range(1, 20), category(6));\nmarker-fill: ramp([name], (#fabada, black, red, white), (wutang clan, form-components.editors.fill.input-categories.null, public enemy), "=", category);\nmarker-fill-opacity: 0.5;\nmarker-file: ramp([name], (url(\'image1\'), url(\'image2\'), url(\'image3\'), url(\'otherimage\')), (wutang clan, \'\', public enemy), "=");\nmarker-allow-overlap: true;\n}'
```
xavijam: What about adding it like `_t('form-components.editors.fill.input-categories.null')`. It will return the same output, but for me this is a bit weird when you see the first time :D.

xavijam:
We need to merge master with this PR in order to make backend tests pass.

lib/assets/core/test/spec/cartodb3/editor/style/style-converter-fixtures.spec.js
```diff
@@ -559,13 +559,13 @@ module.exports = [
     },
     result: {
       point: {
-        cartocss: '#layer {\nmarker-width: ramp([cartodb_id], range(1, 20), category(6));\nmarker-fill: ramp([name], (#fabada, black, red, white), (wutang clan, , public enemy), "=", category);\nmarker-fill-opacity: 0.5;\nmarker-file: ramp([name], (url(\'image1\'), url(\'image2\'), url(\'image3\'), url(\'otherimage\')), (wutang clan, \'\', public enemy), "=");\nmarker-allow-overlap: true;\n}'
+        cartocss: '#layer {\nmarker-width: ramp([cartodb_id], range(1, 20), category(6));\nmarker-fill: ramp([name], (#fabada, black, red, white), (wutang clan, _t('form-components.editors.fill.input-categories.null'), public enemy), "=", category);\nmarker-fill-opacity: 0.5;\nmarker-file: ramp([name], (url(\'image1\'), url(\'image2\'), url(\'image3\'), url(\'otherimage\')), (wutang clan, \'\', public enemy), "=");\nmarker-allow-overlap: true;\n}'
```
xavijam: Hey, but that's a string, you should concatenate it right?

rubenmoya:
Looks good! 👏 

Btw, if someone wants to test it, keep in mind you have to edit the style so the CartoCSS gets updated and the null value appears in the legend.

Issue #12177, opened 2017-05-23
Time-series UI - Add context menu


ivanmalagon:
Split from https://github.com/CartoDB/cartodb/issues/11588

**CONTEXT MENU**

Add context menu to time-series widget. The options that we should show are:
- Show totals: in the same way that it does in the histogram widget, it should show the total values in gray bars and normalizes the height of the yellow ones.
- Edit
- Delete

kevin-reilly:
Please let me know when these enhancements are available on staging and how I might access them.  Thanks.

ivanmalagon:
Just sent you a mail with the steps to enter staging, @kevin-reilly

nobuti:
Deployed.

PR #12246, opened 2017-05-31
12177 timeseries context menu

ivanmalagon:
Multi PR.
- cartodb: https://github.com/CartoDB/cartodb/pull/12246
- cartodb.js: https://github.com/CartoDB/cartodb.js/pull/1665 (already merged)
- deep-insights.js: https://github.com/CartoDB/deep-insights.js/pull/547

Implements https://github.com/CartoDB/cartodb/issues/12177

show totals

This PR adds the context menu that is present in the others widgets to the time-series one. Available options are: `show totals`, `edit` and `delete`.
It's been added to both histogram and torque subtypes.

**Bonus tracks**

While developing this issue we had to fix some other issues:
- Number input in forms now has a debounce time, avoiding unnecessary server trips.
- Widget loader is wrapped and now the scroll bar while loading widget has disappeared.
- Fixed a race condition in histogram data model that triggered infinite loop of widget instantiation calls.

### Things to check:

**Histogram subtype**

Add a time series widget in a non-animated layer.
Select a layer that has some irregular shape in the histogram. In the attached dataset about bird tracking, you can select altitude. Anyway, don't select cartodb_id since it's a bad column for this test.

Check:
- [x] Menu button is present.
- [x] Available options are: show totals, edit, delete.
- [x] Show totals switch between normalized and non-normalized data in the histogram. Move the map bounding box to see only a part of the data for this check to make sense.
- [x] Edit the widget clicking on the menu button.
- [x] Change number of bins in both normalized and non-normalized modes and check that everything is behaving correctly.
- [x] Selecting a range shows the proper text about range boundaries.
- [x] Delete the widget

**Torque subtype**

Change the style to animated to transform the widget to a torque one

Check:
- [x] Menu button is present.
- [x] Available options are: show totals, edit, delete.
- [x] Show totals switch between normalized and non-normalized data in the histogram. Move the map bounding box to see only a part of the data for this check to make sense.
- [x] Edit the widget clicking on the menu button.
- [x] Change number of bins in both normalized and non-normalized modes and check that everything is behaving correctly.
- [x] Selecting a range shows the proper text about range boundaries.
- [x] The Play controls hide when a filter is present.
- [x] Delete the widget

**Menu position**
- [x] The menu always shows above the pointer since there's not enough room to show it in the standard position
- Add another widget of any class.
- Resize the screen to a narrower size to provoke the lateral widgets to be at the bottom of the time-series one.
- [x] The menu appears in the standard position now that it's enough space.

**Misc**

If you find any strange behaviour in the widget, please compare to the existing behaviour in production. This widget has several existing issues so it's likely that it's not a problem provoked by this PR. Anyway, we are on the way to improve it drastically so write down the problem to check if we have an open issue already or we should add a new one.

**Attached dataset: the famous bird tracking**
[bird_tracking.zip](https://github.com/CartoDB/cartodb/files/1058651/bird_tracking.zip)


lib/assets/core/javascripts/cartodb3/components/form-components/editors/number/number.js
```diff
@@ -55,6 +56,8 @@ Backbone.Form.editors.Number = Backbone.Form.editors.Base.extend({
       this.options.step = this._validators[1].step;
     }
 
+    this._debouncedTriggerChange = _.debounce(this._triggerChange, 333).bind(this);
```
nobuti: why 333?

ivanmalagon: It was a bait. ;)

No. I searched about the best time for debouncing inputs and there were no a set answer, but people choose between 300 and 400 ms. So I put something in the middle.

nobuti:
I've found an issue. The context menu is shown behind the map in some breakpoints:

widget-context-menu

Also, as you can see in the gifcast above, the play head gets lost when resizing, but I guess it's not related to this issue.

ivanmalagon:
Fixed the responsive issue.

After wrapping the loader to avoid the `overflow` property, I forgot to clean it in the small breakpoint.

It's already deployed in ded01 if you want to take a look, @nobuti

rjimenezda:
Found something strange, not sure if will happen to everyone.

Open Context Menu
Hit edit
Select 'No' in Dynamic
Open Context Menu again
Hit edit
Dynamic is neither 'yes' nor 'no'

If you refresh this is persisted, although it seems that it behaves as not dynami.

rubenmoya:
It happens to me too

rubenmoya:
I've opened a new issue for the radio buttons things @rjimenezda since it's happening in all widgets, and it's not related to the context menu. So, I think this is ready to deploy!

PR #12248, opened 2017-05-31
Remove icon model from userAssetsCollection if upload fails

### LINKED ISSUE #11980: Icon is saved even when the upload fails

rubenmoya:
Fixes #11980

image_upload

### Acceptance
- [ ] When the upload fails, the icon is not in "Your uploads" tab.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -123,6 +123,7 @@ Development
 * Show infowindow when user reaches max layer limit (#12167)
 
 ### Bug fixes
+* Fixed problem when icon upload fails (#11980)
 * Boolean fields are visible in the filter by column value analysis (#11546)
 * Fixed legend's color mismatch with empty values (#11632)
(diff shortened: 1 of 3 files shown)


xavijam:
Why the loading appears at the bottom of the screen?

rubenmoya:
Because I only recorded half screen 😂

xavijam:
Working like a charm"ander".

PR #12249, opened 2017-05-31
Select next analysis node when deleting one

### LINKED ISSUE #11899: Incorrect behaviour in Analysis tab
### Context

If you already have an analysis added and you add a new one but you cancel it before applying it, the interface doesn't move to the previous analysis UI. 
 
### Steps to Reproduce

1. Apply an analysis to a map
2. Add a second analysis (don't apply it)
3. Cancel the analysis 

### Current Result

Notice how the steps below stay the same when you cancel the analysis, and the previous analysis doesn't get selected automatically. 

analysis
 
### Expected result

The previous analysis should be selected automatically after cancelling the other analysis. 

### Browser and version

Chrome Version 57.0.2987.110 (64-bit)

rjimenezda:
Fixes #11899 

incorrect-analysis

This PR makes it so when you delete an analysis, the next one is selected. It also reduces the number of layer-content-analyses-view render calls.

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-list-view.js
+++ b/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-list-view.js
@@ -6,7 +6,6 @@ var checkAndBuildOpts = require('../../../../helpers/required-opts');
 var REQUIRED_OPTS = [
   'analysisDefinitionNodesCollection',
   'analysisFormsCollection',
-  'model',
   'layerId'
 ];
 
@@ -17,6 +16,12 @@ module.exports = CoreView.extend({
 
   initialize: function (opts) {
     checkAndBuildOpts(opts, REQUIRED_OPTS, this);
+
+    if (!opts.model) {
+      throw new Error('model is required');
+    }
+
+    this.model = opts.model;
   },
 
   render: function () {
@@ -33,7 +38,7 @@ module.exports = CoreView.extend({
     var view = new AnalysesWorkflowItemView({
       analysisNode: this._analysisDefinitionNodesCollection.get(formModel.id),
       formModel: formModel,
-      viewModel: this._model,
+      viewModel: this.model,
       layerId: this._layerId
     });
     this.$el.append(view.render().el);
--- a/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-list-view.js
+++ b/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-list-view.js
@@ -1,6 +1,14 @@
 var CoreView = require('backbone/core-view');
 var AnalysesWorkflowItemView = require('./analyses-workflow-item-view');
 var template = require('./analyses-workflow-list.tpl');
+var checkAndBuildOpts = require('../../../../helpers/required-opts');
+
+var REQUIRED_OPTS = [
+  'analysisDefinitionNodesCollection',
+  'analysisFormsCollection',
+  'model',
+  'layerId'
+];
 
 module.exports = CoreView.extend({
 
@@ -8,15 +16,7 @@ module.exports = CoreView.extend({
   tagName: 'ul',
 
   initialize: function (opts) {
-    if (!opts.analysisDefinitionNodesCollection) throw new Error('analysisDefinitionNodesCollection is required');
-    if (!opts.analysisFormsCollection) throw new Error('analysisFormsCollection is required');
-    if (!opts.model) throw new Error('model is required');
-    if (!opts.layerId) throw new Error('layerId is required');
-
-    this._analysisDefinitionNodesCollection = opts.analysisDefinitionNodesCollection;
-    this.analysis = opts.analysis;
-    this.analysisFormsCollection = opts.analysisFormsCollection;
-    this._layerId = opts.layerId;
+    checkAndBuildOpts(opts, REQUIRED_OPTS, this);
   },
 
   render: function () {
@@ -25,7 +25,7 @@ module.exports = CoreView.extend({
       layerId: this._layerId
     }));
 
-    this.analysisFormsCollection.each(this._renderWorkflowItem, this);
+    this._analysisFormsCollection.each(this._renderWorkflowItem, this);
     return this;
   },
 
@@ -33,7 +33,7 @@ module.exports = CoreView.extend({
     var view = new AnalysesWorkflowItemView({
       analysisNode: this._analysisDefinitionNodesCollection.get(formModel.id),
       formModel: formModel,
-      viewModel: this.model,
+      viewModel: this._model,
       layerId: this._layerId
     });
     this.$el.append(view.render().el);
--- a/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-view.js
+++ b/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-view.js
@@ -20,7 +20,10 @@ module.exports = CoreView.extend({
   initialize: function (opts) {
     checkAndBuildOpts(opts, REQUIRED_OPTS, this);
 
-    this.listenTo(this._analysisDefinitionNodesCollection, 'add', this.render);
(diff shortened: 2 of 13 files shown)


lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-controls-view.js
```diff
@@ -101,16 +101,21 @@ module.exports = CoreView.extend({
...
-          this.render();
-        }
-      });
+      this._bindAnalysisNodeStatusChanges();
     }
   },
 
+  _bindAnalysisNodeStatusChanges: function () {
+    this.listenTo(this._analysisNode, 'change:status', function () {
+      console.log('change:status', 'isDone', this._isAnalysisDone());
```
rubenmoya: console.log!

xavijam:
This is an important one, could you take a look too @ivanmalagon @nobuti ? And then in the acceptance we will need several pair of eyes.

xavijam:
We need to merge master branch also @rjimenezda 💋

rjimenezda:
I brought master (b60160e). Should I do it again? Poor travis

xavijam:
Do we have last master in this PR?

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-forms-collection.js
```diff
@@ -57,6 +57,22 @@ module.exports = Backbone.Collection.extend({
     }
   },
 
+  removeUselessModels: function () {
```
nobuti: I'm intrigued, why do we need this?

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/layer-content-analyses-view.js
```diff
@@ -116,15 +104,9 @@ module.exports = CoreView.extend({
...
-
-    this._layerDefinitionModel.on('change:source', this._onLayerDefSourceChanged, this);
-    this.add_related_model(this._layerDefinitionModel);
-
-    this._analysisFormsCollection.on('add', this._onFormModelAdded, this);
-    this._analysisFormsCollection.on('remove', this._onFormModelRemoved, this);
-    this.add_related_model(this._analysisFormsCollection);
+    this.listenTo(this._viewModel, 'change:selectedNodeId', this.render);
+    this.listenTo(this._layerDefinitionModel, 'change:source', this._onLayerDefSourceChanged);
+    this.listenTo(this._analysisFormsCollection, 'destroyedModels remove', this._renderWithDefaultNodeSelected);
```
nobuti: Why are we listening the `add` event in one view but `remove` in other?

rjimenezda: The add event is not necessary because we only add to analysisFormsCollection when coming back from the add-analysis modal (layer-content-view.js:127), and after that, the view is recreated. We can leave it, but there's no point.

I had to add the 'add' handler on the workflow because I removed a render call on analyses-view. The consequence is that neither the workflow list nor the apply button would update correctly after applying a newly added analysis. So I added more specific event handling instead of just re-rendering the whole view.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analysis-forms-collection.js
rjimenezda: When the user removes an analysis node (already stored), instead of just removing it from the collection we have to go through user actions, remove the node from everywhere except analysisFormCollection, and then update layerDefinition (set source).

Whenever layerDefinition is updated, we end up calling reset on the collection, which is done by mapping & augmenting the values from layerDefinition. The problem is, we don't get notified that a node is removed from the collection in this case, we just get a reset, which is called in other places as well.

So instead of doing a reset which is very opaque,  we silently remove the ones that are no longer necessary (hence the useless) and then fire a custom event, so we can react on it.

rjimenezda:
The number on the workflow view is not being updated either, because 'add' happens before 'change:source' on layer definition model, so it renders before source is updated, and it reports an 'incorrect' size. Got a solution though, hope you like it

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-view.js
```diff
@@ -20,7 +20,10 @@ module.exports = CoreView.extend({
   initialize: function (opts) {
     checkAndBuildOpts(opts, REQUIRED_OPTS, this);
 
-    this.listenTo(this._analysisDefinitionNodesCollection, 'add', this.render);
+    this.listenTo(this._analysisDefinitionNodesCollection, 'add', function () {
```
xavijam: Fine by me, but I'd add a test checking this behaviour 😬

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-list-view.js
```diff
@@ -25,15 +25,15 @@ module.exports = CoreView.extend({
...
+    this._analysisFormsCollection.each(this._renderWorkflowItem, this);
     return this;
   },
 
   _renderWorkflowItem: function (formModel) {
     var view = new AnalysesWorkflowItemView({
       analysisNode: this._analysisDefinitionNodesCollection.get(formModel.id),
       formModel: formModel,
-      viewModel: this.model,
+      viewModel: this._model,
```
xavijam: In any case, `this.model` is defined (something by default in backbone views), so please, let's keep using that way, because there are helper classes and methods using that reference.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-view.js
rjimenezda: You got it!

xavijam:
There is something I've found in acceptance and it was managed by the QuerySanity stuff (https://github.com/CartoDB/cartodb/pull/12249/files#diff-df1d90c928d7c866a602f579b8116b67L66). If you apply a wrong query, we should not let add any analysis, because it will fail, so this state appears:

screen shot 2017-06-01 at 10 53 25

But, with our changes, removing that, we don't have that state anymore. We should keep checking if there is an error in the "source" node query and display that. Right now, this is what we are displaying:

screen shot 2017-06-01 at 10 59 17

Back to development, but closer!

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-list-view.js
rjimenezda: I have a couple of ideas to improve required-opts which would improve this, but I'll fix it manually for now

lib/assets/core/javascripts/cartodb3/editor/layers/query-sanity-check.js
```diff
@@ -39,6 +39,10 @@ module.exports = {
   },
 
   _hasErrors: function () {
+    return this._checkErrors(this._querySchemaModel);
+  },
+
+  _checkErrors: function (querySchemaModel) {
     var errors = this.querySchemaModel.get('query_errors');
```
nobuti: this.querySchemaModel is undefined, right?

rjimenezda: Now I wonder how did that work locally 🤔

rjimenezda: Fixed

nobuti:
I couldn't find anything else, it works pretty well ✨

xavijam:
Nice!

nobuti:
And this definitely makes DO faster at the first rendering 🙌

nobuti:
🚀

rjimenezda:
🚢 :shipit:

xavijam:
> And this definitely makes DO faster at the first rendering 🙌 

Vamos!

PR #12258, opened 2017-06-01
formula widget uses operators editor

### LINKED ISSUE #12242: Formula widget form should make use of operation component
In the formula widget, we are requesting a column and an operation, but that's not necessary because we have already a component which makes that.

- This is the current thing in formula widget edit form:

screen shot 2017-05-30 at 15 18 48

- And this is the component we want to use in formula widget form, but being used in the category one.

screen shot 2017-05-30 at 15 18 41

Take into account that COUNT operation needs the `cartodb_id` column or any other numeric column in order to work properly (under the hood). I encourage you to test it in production and check how it works right now in order to understand the previous phrase.

rubenmoya:
Fixes #12242

widget_formula

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/editor/widgets/widgets-form/schema/widgets-form-formula-schema-model.js
+++ b/lib/assets/core/javascripts/cartodb3/editor/widgets/widgets-form/schema/widgets-form-formula-schema-model.js
@@ -4,42 +4,60 @@ var WidgetsFormBaseSchema = require('./widgets-form-base-schema-model');
 module.exports = WidgetsFormBaseSchema.extend({
 
   defaults: {
-    schema: {}
+    schema: {},
+    aggregate: {
+      attribute: '',
+      operator: 'count'
+    }
   },
 
   initialize: function (attrs, opts) {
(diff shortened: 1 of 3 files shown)


lib/assets/core/javascripts/cartodb3/editor/widgets/widgets-form/schema/widgets-form-formula-schema-model.js
```diff
@@ -4,43 +4,61 @@ var WidgetsFormBaseSchema = require('./widgets-form-base-schema-model');
...
     return {
-      data: 'column,operation,prefix,suffix,description',
+      data: 'aggregate,prefix,suffix,description',
       style: 'sync_on_bbox_change'
     };
   },
 
+  _updateAggregate: function () {
+    var aggregate = this.get('aggregate');
+    if (aggregate === undefined) {
```
nobuti: aggregate is defined in the `defaults` block, maybe I'm wrong but I think it could not be undefined ever. The else part will do the trick.

nobuti:
I'll take care @rubenmoya 💋

nobuti:
🚀

PR #12259, opened 2017-06-01
Show title & error on analysis workflow tooltip

### LINKED ISSUE #12250: Problem with analysis tooltip and error
@xavijam commented on [Wed May 31 2017](https://github.com/CartoDB/design/issues/798)

Checking some behaviors with analyses, I've found this problem:

screen shot 2017-05-31 at 15 05 32

Should we remove the analysis-name tooltip when there is an error in the analysis node?

Assigning @arianaescobar because she opened the ticket in CartoDB repository 😬 
cc @rjimenezda @rubenmoya 

---

@arianaescobar commented on [Wed May 31 2017](https://github.com/CartoDB/design/issues/798#issuecomment-305192214)

I'd say that's the fastest solution, yeah. 

But, just wondering, is it possible to "merge" both tooltips in a single one in those cases? Like "Centroid analysis failed"? 

---

@xavijam commented on [Wed May 31 2017](https://github.com/CartoDB/design/issues/798#issuecomment-305193780)

Sure!

---

@xavijam commented on [Wed May 31 2017](https://github.com/CartoDB/design/issues/798#issuecomment-305194033)

But take into account that the error could change and provide more info. It could be like:

`Centroid (A2): column whatever is not longer available`

What do you think?

---

@arianaescobar commented on [Wed May 31 2017](https://github.com/CartoDB/design/issues/798#issuecomment-305196064)

Love it.

rubenmoya:
Fix #12250 

If there's an error, it shows the tooltip in red and combines the Analysis name & the error message.

tooltip_2

### DIFF
--- a/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-item-view.js
+++ b/lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-item-view.js
@@ -5,6 +5,7 @@ var layerColors = require('../../../../data/layer-colors');
 var AnalysisTooltip = require('../../analysis-views/analyses-tooltip-error');
 var TipsyTooltipView = require('../../../../components/tipsy-tooltip-view');
 var Analyses = require('../../../../data/analyses');
+var NotificationErrorMessageHandler = require('../../notification-error-message-handler');
 
 module.exports = CoreView.extend({
 
@@ -25,14 +26,6 @@ module.exports = CoreView.extend({
 
     this._analysisNode.on('change:status', this.render, this);
     this.add_related_model(this._analysisNode);
-
-    this._analysisTooltip = new AnalysisTooltip({
-      analysisNode: this._analysisNode,
-      element: this.$el,
-      triggerSelector: '.Editor-ListAnalysis-itemError'
-    });
-
-    this.addView(this._analysisTooltip);
   },
 
   render: function () {
@@ -41,6 +34,14 @@ module.exports = CoreView.extend({
     var isSelected = this._analysisNode.id === this._viewModel.get('selectedNodeId');
     var letter = nodeIds.letter(this._analysisNode.id);
     var bgColor = layerColors.getColorForLetter(letter);
+    var analysisTitle = Analyses.title(this._analysisNode.get('type') || '');
+    var tooltipTitle = analysisTitle;
+    var hasFailed = status === 'failed';
+
+    if (hasFailed) {
+      var message = NotificationErrorMessageHandler.extractErrorFromAnalysisNode(this.analysisNode);
+      tooltipTitle = analysisTitle + ': ' + message;
+    }
 
     this.$el.html(
       template({
@@ -62,9 +63,10 @@ module.exports = CoreView.extend({
 
     var tooltip = new TipsyTooltipView({
       el: this.$el,
+      className: hasFailed ? 'is-error' : '',
       title: function () {
-        return Analyses.title(this._analysisNode.get('type') || '');
-      }.bind(this)
+        return tooltipTitle;
+      }
     });
     this.addView(tooltip);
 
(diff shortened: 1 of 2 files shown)


lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-item-view.js
```diff
@@ -41,6 +33,14 @@ module.exports = CoreView.extend({
     var isSelected = this._analysisNode.id === this._viewModel.get('selectedNodeId');
     var letter = nodeIds.letter(this._analysisNode.id);
     var bgColor = layerColors.getColorForLetter(letter);
+    var analysisTitle = Analyses.title(this._analysisNode.get('type') || '');
+    var tooltipTitle = analysisTitle;
+    var hasFailed = status === 'failed';
+
+    if (hasFailed) {
+      var message = NotificationErrorMessageHandler.extractErrorFromAnalysisNode(this.analysisNode);
```
nobuti: This is my monthly quota of "I don't like to define vars inside if block".  Also, as far I can see, `extractErrorFromAnalysisNode` already returns a string with `:` as part of the message key, so be careful with that.

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-item-view.js
```diff
@@ -41,6 +33,16 @@ module.exports = CoreView.extend({
     var isSelected = this._analysisNode.id === this._viewModel.get('selectedNodeId');
     var letter = nodeIds.letter(this._analysisNode.id);
     var bgColor = layerColors.getColorForLetter(letter);
+    var analysisTitle = Analyses.title(this._analysisNode.get('type') || '');
+    var tooltipTitle = analysisTitle;
+    var hasFailed = status === 'failed';
+
+    if (hasFailed) {
+      var message = NotificationErrorMessageHandler.extractErrorFromAnalysisNode(this.analysisNode).message;
```
nobuti: `extractErrorFromAnalysisNode` method uses the locale `notifications.analysis.failed`, that is `Analysis %{nodeId} failed`. Then it concatenates this with the error message. I think we are only interested in the message, so I guess we should use the `extractError` method instead. Also, according to the issue's description, the message should be something like `Centroid (A2): column whatever is not longer available`. I miss the id here.

rubenmoya: We talked about the id, and decided not to put it in the message. About the error message you're right. Do you know any way to produce an error in an analysis to test this?

rubenmoya:
Improved error message format, see updated gif in OP

lib/assets/core/javascripts/cartodb3/editor/layers/layer-content-views/analyses/analyses-workflow-item-view.js
```diff
@@ -36,14 +36,14 @@ module.exports = CoreView.extend({
     var analysisTitle = Analyses.title(this._analysisNode.get('type') || '');
     var tooltipTitle = analysisTitle;
     var hasFailed = status === 'failed';
-
-    if (hasFailed) {
-      var message = NotificationErrorMessageHandler.extractErrorFromAnalysisNode(this.analysisNode).message;
-      if (message) {
-        tooltipTitle = analysisTitle + ': ' + message;
-      }
+    var error = this._analysisNode.get('error');
```
nobuti: We only need `errorMessage` if `hasFailed`, so I'd avoid the `extractError` call check by default:
```
var error;
var errorMessage;

if (hasFailed) {
    error = this._analysisNode.get('error');
    errorMessage = error && error.message && NotificationErrorMessageHandler.extractError(error.message);
    tooltipTitle = analysisTitle + ': ' + errorMessage.message;
}
```
what do you think?

rubenmoya: You're right, but what happens if `errorMessage.message` is an empty string or undefined? we'll end up with something like: `Georeference: ` or `Georeference: undefined`

nobuti: Right, we should check that case too.

nobuti:
Picking this up to make the changes.

nobuti:
What do you think about last changes @rubenmoya?

rubenmoya:
I like that approach, pretty interesting!

rubenmoya:
It works like a charm! ✨

PR #12262, opened 2017-06-02
Number editor to respects min/max validations when using arrow keys

### LINKED ISSUE #12212: Increase and Decrease with arrows doesn't respect the minimum and maximum size
### Context

size

### Steps to Reproduce
1.  For example, try to change the size in the hexbis aggregation method
2. Increase and decrease with arrows
3. If I go to less than 10 (the minimum size) it still decreases the number

rubenmoya:
Fixes #12212

number_editor

### Acceptance
With the arrow keys:
- [ ] The user can't exceed the max value of the input.
- [ ]  The user can't go below the min value of the input.
- [ ] If the user types an amount bigger than the max permitted and use the arrow keys, the input is set to the maximum value.
- [ ] If the user types an amount smaller than the min permitted and use the arrow keys, the input is set to the min value.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,7 @@ Development
 * New loading button styles (#12132)
 
 ### Bug fixes / enhancements
+* Fixed arrow keys exceeding min/max values in number editor (#12212)
 * Fixed aligment problems after cartoassets update (#12234)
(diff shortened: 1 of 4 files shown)


nobuti:
🚀

Issue #11581, opened 2017-02-20
Break down deep-insights-integration class


xavijam:
Although `deep-insights-integration` JS class is a marvellous idea to connect Builder and CARTO.js interactions, we have realized it is growing and growing without control (almost ~1000 lines of code), and we should split it in several parts in order to make it manageable (it includes tests).

xavijam:
Work started in branch `deep-insights-integration-split` for Friday's leapfrog.

nobuti:
We need to pass smokes before to merge.

PR #12265, opened 2017-06-02
Deep insights integration split

xavijam:
Related ticket: https://github.com/CartoDB/cartodb/issues/11581.

As any @CartoDB/frontend knows, our proxy with Deep-Insights/CARTO.js is too dense. And that's not the only problem, I feel the main problem is writing tests is not as easy as we want. Let's try to fix it! In short, the steps we should follow are:

- [x] Create a Deep-insights-helpers class, the only way "for talking" with CARTO.js and DI (https://github.com/CartoDB/cartodb/pull/12265/files#diff-b455cbf417985bb3ea779060b81be078R1).
- [x] Split the code by different contexts, like layer, map, widgets, analyses, etc.
- [x] Decide how to communicate different operations between integrations (1).
- [x] Split the tests and create spec-helpers for easy debugging.

[(1)](https://github.com/CartoDB/cartodb/pull/12265/files#diff-3b0801115002fb01644a46bf23ce01d7R93) As you can check in the changes, I've decided to use a trigger-bind thing instead of passing the integration to any other. Just for simplicity, and because there is only one integration where it needs to know from another one.

## Acceptance

As you have seen, there are lot of changes and moves, so we should make a deep check involving all CARTO.js operations. So I've decided to create a smoke tests document:

https://goo.gl/FWuyfm


xavijam:
This week episode: 
- Helpers are created (\o/).
- Map-integration class has its own tests, removed from the original deep-insights-integration specs.
- Moved, but not integrated, rest of the tests to the proper integration class, that's the pending task + test test test test.

xavijam:
Close to the final steps. Just failing a few tests, but main tasks are already done. It would be great if you take a look when you have some time 🙏 .

matallo:
on my radar 👀

lib/assets/core/javascripts/cartodb3/editor.js
```diff
@@ -320,7 +320,6 @@ var mapCreation = function () {
       legendDefinitionsCollection: legendDefinitionsCollection,
       widgetDefinitionsCollection: widgetDefinitionsCollection,
       overlayDefinitionsCollection: overlaysCollection,
-      overlaysCollection: overlaysCollection,
```
ivanmalagon: Was this unnecessary?

lib/assets/core/javascripts/cartodb3/deep-insights-integrations.js
```diff
@@ -1,1156 +1,100 @@
...
+  LayersIntegration.bind('onLayerChanged', function (layerDefinitionModel) {
+    WidgetsIntegration.manageTimeSeriesForTorque(layerDefinitionModel);
+  });
 
-F.prototype._onBoundsSet = function (bounds) {
-  this._diDashboard._dashboard.vis.map.setBounds(bounds);
+  LayersIntegration.bind('onBaseLayerChanged', function () {
+    MapIntegration.setMapConverters();
+  });
 };
```
ivanmalagon: Super duper clean!! Way more useful and understable.

I think the integration between integration via event it's also a good approach instead of passing integrations to integrations. The connections between the different integrations are very explicit this way.

Kudos!

xavijam: 👏 great to see you like it! 💃

lib/assets/core/javascripts/cartodb3/editor.js
xavijam: We were passing the same variable with two different names.

rjimenezda:
What an undertaking, good job 👌

xavijam:
Updated with the acceptance process 💃 .

matallo:
👏 incredible work, loved those `deep-insights-helpers.js`

xavijam:
Hey, wait, it could break everything very badly, we need to run the smokes before doing anything. I've planned to do it next week.

nobuti:
Adding blocked tag then.

xavijam:
Tests passed in local, deploying to ded02 and passing them there :dancer:.

xavijam:
Ladies and gentlemen, after checking it in staging (with @rubenmoya in the last instance), we haven't found anything weird after passing the smoke tests shared above, so... merging tomorrow! (and keeping and eye in our JS error tracker 💃 ).

matallo:
BOOM

alonsogarciapablo:
Thanks for your hard work on thin @xavijam!

